PR #1451, opened 2015-02-07
Handle button release events

acrisci:
This enables the --release switch on mouse button bindings.

### DIFF
--- a/src/click.c
+++ b/src/click.c
@@ -200,6 +200,11 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
         }
     }
 
+    /* There is no default behavior for button release events so we are done. */
+    if (event->response_type == XCB_BUTTON_RELEASE) {
+        goto done;
+    }
+
     /* Any click in a workspace should focus that workspace. If the
      * workspace is on another output we need to do a workspace_show in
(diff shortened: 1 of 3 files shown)


src/click.c
```diff
@@ -348,7 +354,7 @@ int handle_button_press(xcb_button_press_event_t *event) {
     if ((con = con_by_window_id(event->event)))
         return route_click(con, event, mod_pressed, CLICK_INSIDE);
 
-    if (!(con = con_by_frame_id(event->event))) {
+    if (!(con = con_by_frame_id(event->event)) && event->response_type == XCB_BUTTON_PRESS) {
```
stapelberg: This check is subtly wrong: the previous `if (!(con = con_by_frame_id(event->event))) {` was intended to catch the case where `con` is `NULL` and abort the function. With your additional condition, the function is not aborted for `XCB_BUTTON_PRESS` events anymore, and in line 384, you can now have a null pointer dereference.

I think you should move the condition into the if statement on line 354 (`if (event->event == root)`)

stapelberg:
Looks good except for the one comment.

acrisci:
Ok I fixed it.

PR #1452, opened 2015-02-07
i3-msg: pretty print json response

acrisci:


stapelberg:
I see you’ve put some effort into this, but I’m not sure we should merge this in. i3-msg is supposed to be a **simple** example of how to communicate with i3. It intentionally is just a single file without too many dependencies. I realize that it depends on libi3, and I consider that somewhat unfortunate, but the past has shown that the details of using sockets in C are subtle, so having one place to fix is a good thing.

Your pull request’s effect can also be achieved with e.g. `i3-msg -t get_tree | json_xs` (from the JSON::XS perl module) or `i3-msg -t get_tree | jq ''`, so people who often use i3-msg for debugging can just define `func i3msg() { i3-msg $* | jq ''; }` in their shell.

Is there anything I’m missing here?

acrisci:
The reformatting was put in a separate file to separate the presentation of the result from the business logic ([separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns)).

The code to format the result looks complicated, but this is just an unfortunate consequence of the verbosity of the yajl library which only gives us a streaming parser. It just a simple matter of parsing the json and then reformatting it in a pretty-print form. This is about 5 lines of code in most frameworks.

This patch was adapted from [an official yajl example](https://lloyd.github.io/yajl/yajl-2.0.1/reformatter_2json_reformat_8c-example.html).

i3-msg is an interactive tool that is meant to be used on the command line. The results are meant primarily to be read by humans in the terminal. Pretty printed json is considerably more easy to read. Therefore, this change enhances the primary purpose of the utility.

For a query like `get_tree`, the result is impossible to meaningfully interpret in the minimized form. A user would be required to pipe it into another program to use the utility. This change makes i3-msg more user friendly by eliminating the need for that extra step which reduces the overhead for use.

stapelberg:
> i3-msg is an interactive tool that is meant to be used on the command line.

I don’t think i3-msg counts as interactive, based on the definition at http://en.wikipedia.org/wiki/Interactivity#Computing_science :).

> The results are meant primarily to be read by humans in the terminal. Pretty printed json is considerably more easy to read. Therefore, this change enhances the primary purpose of the utility.

The primary purpose of i3-msg is to serve as an example of how to use the IPC interface from a very low-level language such as C. The fact that it can be used to send messages to i3 is a nice side-effect and certainly used by a large number of people, but not the primary purpose. Note that I’m not saying that looking at the results is something a large number of people do — I think the opposite is the case. As another example to underline i3-msg’s limited scope, consider that i3-msg doesn’t have support for subscribing to events and displaying them.

I’m therefore still very hesitant to merge this pull request, even though I agree that the output is more useful in a pretty-printed form.

> For a query like get_tree, the result is impossible to meaningfully interpret in the minimized form. A user would be required to pipe it into another program to use the utility. This change makes i3-msg more user friendly by eliminating the need for that extra step which reduces the overhead for use.

Agreed, but the output of get_tree is hard to interpret even if pretty-printed, so in practice you either save the pretty-printed representation or directly use e.g. jq to filter/post-process it any further.

I still don’t see why the ways I suggested (using json_xs or jq or any other JSON processing tool) aren’t good enough — with regards to overhead you surely can’t mean resources, and I think the mental overhead of piping a program to another program should be sufficiently abstracted away by the shell function I provided.

My arguments for not merging this pull request are:

 1. Avoiding code complexity, especially in parts of the code which are intended to be read by other people to learn from.
 2. Avoiding changes in behavior that break existing use-cases, and unfortunately we see code from time to time which directly uses i3-msg output without properly parsing it as JSON (e.g. in shell scripts), even though we always recommend against that.

May I suggest adding the shell function definition from my last post as a suggested way to consume i3-msg’s output to i3-msg’s manpage instead of adding pretty-printing to i3-msg itself?

acrisci:
> Avoiding code complexity, especially in parts of the code which are intended to be read by other people to learn from.

This doesn't make sense to me. All of the ipc code is encapsulated in libi3 which doesn't have a public interface. As someone who learned the wire protocol, I didn't find i3-msg helpful. The functions in libi3 should be considered the reference implementation, not i3-msg.

As for complexity, while the code may be verbose, I don't think it's that complex. It doesn't affect any part of the logic except the presentation to the user.

> Avoiding changes in behavior that break existing use-cases, and unfortunately we see code from time to time which directly uses i3-msg output without properly parsing it as JSON (e.g. in shell scripts), even though we always recommend against that.

I don't see how this could possibly break anything. Pretty printing only adds new lines. It doesn't change the content of the message at all. Anyone who is awking or grepping json in an automated script without taking into account that there could be new lines is not following the json spec and deserves to have their code broken.

But anyway, this change is not critical for me. I already have all the aliases that I need set up in my shell rc. You can close this if you want.

stapelberg:
> As someone who learned the wire protocol, I didn't find i3-msg helpful.

That’s a fair data point, and if lots of people agree, we should change direction.

> I don't see how this could possibly break anything. Pretty printing only adds new lines. It doesn't change the content of the message at all. Anyone who is awking or grepping json in an automated script without taking into account that there could be new lines is not following the json spec and deserves to have their code broken.

I also think they are doing it wrong, but I don’t think anyone deserves to have their code broken, unless it’s unavoidable. In this case it’s clearly avoidable, so I’d rather err on the side of not breaking people’s code, no matter how they are abusing the interface we provide.

Closing this as suggested.

PR #1453, opened 2015-02-08
Update bar font config on reload

acrisci:


i3bar/src/ipc.c
```diff
@@ -169,6 +172,8 @@ void got_bar_config_update(char *event) {
         reconfig_windows(true);
     }
 
+    /* update fonts and colors */
+    init_xcb_late(config.fontname);
```
stapelberg: `init_xcb_late` is not yet adapted for freeing the current font before loading a new font, i.e. you’ll have an X11 resource leak unless you call `free_font`.

i3bar/src/ipc.c
```diff
@@ -159,6 +159,9 @@ void got_bar_config_update(char *event) {
     if (found_id == NULL)
         return;
 
+    /* reconfigure the bar based on the current outputs */
+    i3_send_msg(I3_IPC_MESSAGE_TYPE_GET_OUTPUTS, NULL);
```
stapelberg: Please mention this change in the commit message.

acrisci:
I added a better commit message and fixed the memory leak issue.

Issue #1294, opened 2014-06-24
i3 ignores changes in _NET_WM_STRUT_PARTIAL


i3bot:
[Originally reported by a.pronobis@…]
(I'm successfully using KDE's panel with i3. There is however one annoying problem. My panel appears on the top of the screen even if it is supposed to show on the bottom. The location of the panel seems to be detected based on the 
_NET_WM_STRUT_PARTIAL and xprop actually shows correct values
_NET_WM_STRUT_PARTIAL(CARDINAL) = 0, 0, 0, 24, 0, 0, 0, 0, 0, 0, 0, 1599
(the 4th non-zero value corresponds to the bottom of the screen) as described here:
http://standards.freedesktop.org/wm-spec/1.3/ar01s05.html

Further investigation revealed that the property is not set correctly when the window is created (STRUT_PARTIAL contains only zeros then) and soon changed to the value shown below. However, that change is not picked up by i3. Could we have handlers for changes in STRUT_PARTIAL as well just like there are for e.g. window title?

PR #1454, opened 2015-02-08
reconfigure dock clients on strut partial change

acrisci:
Strut partial tells i3 where to reserve space for i3bar. It is affected
by the `position` bar config directive.

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -1525,6 +1525,50 @@ void realloc_sl_buffer(void) {
     }
 }
 
+/* Strut partial tells i3 where to reserve space for i3bar. This is determined
+ * by the `position` bar config directive. */
+xcb_void_cookie_t config_strut_partial(i3_output *output) {
+    /* A local struct to save the strut_partial property */
+    struct {
+        uint32_t left;
+        uint32_t right;
+        uint32_t top;
+        uint32_t bottom;
+        uint32_t left_start_y;
+        uint32_t left_end_y;
+        uint32_t right_start_y;
+        uint32_t right_end_y;
+        uint32_t top_start_x;
+        uint32_t top_end_x;
+        uint32_t bottom_start_x;
+        uint32_t bottom_end_x;
+    } __attribute__((__packed__)) strut_partial;
+    memset(&strut_partial, 0, sizeof(strut_partial));
+
+    switch (config.position) {
+        case POS_NONE:
+            break;
+        case POS_TOP:
+            strut_partial.top = bar_height;
+            strut_partial.top_start_x = output->rect.x;
+            strut_partial.top_end_x = output->rect.x + output->rect.w;
+            break;
+        case POS_BOT:
+            strut_partial.bottom = bar_height;
+            strut_partial.bottom_start_x = output->rect.x;
+            strut_partial.bottom_end_x = output->rect.x + output->rect.w;
+            break;
+    }
+    return xcb_change_property(xcb_connection,
+                               XCB_PROP_MODE_REPLACE,
+                               output->bar,
+                               atoms[_NET_WM_STRUT_PARTIAL],
+                               XCB_ATOM_CARDINAL,
+                               32,
+                               12,
+                               &strut_partial);
+}
+
 /*
  * Reconfigure all bars and create new bars for recently activated outputs
  *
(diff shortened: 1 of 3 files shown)


acrisci:
I'm trying to fix #1294

acrisci:
This actually requires #1453 to work.

acrisci:
Now the bar position will change when we `reload`.

stapelberg:
I’ll review this once #1453 is in.

i3bar/src/xcb.c
```diff
@@ -1462,6 +1462,51 @@ void realloc_sl_buffer(void) {
...
+    }
+    return xcb_change_property(xcb_connection,
+            XCB_PROP_MODE_REPLACE,
+            output->bar,
+            atoms[_NET_WM_STRUT_PARTIAL],
+            XCB_ATOM_CARDINAL,
+            32,
+            12,
+            &strut_partial);
+
```
stapelberg: nit: stray newline

src/handlers.c
```diff
@@ -1134,6 +1134,76 @@ static bool handle_class_change(void *data, xcb_connection_t *conn, uint8_t stat
     return true;
 }
 
+/*
+ * Handles the _NET_WM_STRUT_PARTIAL property for allocating space for dock clients.
+ *
+ */
+static bool handle_strut_partial_change(void *data, xcb_connection_t *conn, uint8_t state, xcb_window_t window,
```
stapelberg: Can you add a testcase for this please?

acrisci:
I added a test, but it is failing. However, it seems to work fine under normal conditions with i3bar.

The issue is that the request for the strut partial property fails with error code 3 (which I think is BadWindow). However, the con is found in the tree and the window id seems to be valid.

Any ideas on what is going on with the test?

src/handlers.c
```diff
@@ -1174,6 +1174,85 @@ static bool handle_class_change(void *data, xcb_connection_t *conn, uint8_t stat
...
+
+    window_update_strut_partial(con->window, prop);
+
+    /* we only handle this change for dock clients */
+    if (con->parent == NULL || con->parent->type != CT_DOCKAREA) {
+        return true;
+    }
+
+    Con *search_at = croot;
+    Output *output = get_output_containing(con->geometry.x, con->geometry.y);
```
stapelberg: Using `con_get_output()` seems more straight-forward than `get_output_containing()` here.

src/handlers.c
```diff
@@ -1174,6 +1174,85 @@ static bool handle_class_change(void *data, xcb_connection_t *conn, uint8_t stat
...
+            con->window->dock = W_DOCK_TOP;
+        } else {
+            DLOG("geom->y = %d >= rect.height / 2 = %d, it is a bottom dock client\n",
+                    con->geometry.y, (search_at->rect.height / 2));
+            con->window->dock = W_DOCK_BOTTOM;
+        }
+    }
+
+    /* find the dockarea */
+    Con *dockarea = con_for_window(search_at, con->window, NULL);
```
stapelberg: `con_for_window` could return NULL. If you’re sure that it cannot under these circumstances, at least add an `assert`.

testcases/t/102-dock.t
```diff
@@ -143,6 +143,21 @@ wait_for_map $window;
...
+    PROP_MODE_REPLACE,
+    $window->id,
+    $atomname->id,
+    $atomtype->id,
+    32,         # 32 bit integer
+    12,
+    pack('L12', 0, 0, 0, 16, 0, 0, 0, 0, 0, 0, 1280, 0)
+);
+
+@docked = get_dock_clients('bottom');
```
stapelberg: You’re missing a `sync_with_i3;` before this line.

stapelberg:
Even with the `sync_with_i3`, the testcase makes i3 crash sometimes. Valgrind reports invalid writes. I think your new code (or perhaps old code, but that’s rather unlikely) is corrupting memory somewhere.

acrisci:
It works just fine outside of the test environment.

stapelberg:
So? Memory corruption tends to have that feature (“works just fine on my machine”).

acrisci:
Thanks for the hint about the sync in the testcases. That led me to the memory error and I fixed it.

I messed up a rebase. Do I need to close this and submit another pull request?

src/handlers.c
acrisci: I think it should never be NULL because there is always a dock area below the output (or root) for the dock client.

stapelberg:
Not sure what you mean with the messed up rebase? Can you not use interactive rebase and then just `git push -f`?

acrisci:
It was showing a lot of recreated commits in the pull request for some reason. Maybe it was a github bug. I think that issue is resolved though and this should be ready.

Issue #1433, opened 2015-01-19
KDE plasma panel windows change locations


i3bot:
[Originally reported by a.pronobis@…]
(I'm running i3 together with KDE 5.0. The problem I'm observing is that every second time I open a kicker window (the one shown in the image here: http://regmedia.co.uk/2014/07/28/kde_plasma_5_kickoff_menu_resized.jpg), it appears in the middle of the screen instead of it's normal location.

I'm opening and closing the window using the (K) button in the left bottom corner. Basically, every 2 clicks on the (K) button, the window appears in the center of the screen and then next time in it's correct position and so on.

However, I have noticed that if the window is closed by my i3 shortcut, the problem does not occur and the window is re-created in it's correct location every time.

Moreover, even if the window is not shown, restarting i3, will leave an "empty" window hanging in the middle of the screen. It seems that KDE does not really close the window, but hides it somehow.

This problem does not occur for current master branch, but only for next. I bisected it to the following commit:
43b447855df095c56e860117d53acd0dd09ac42b
When this commit is reverted, the problem does not occur even in the next branch.

acrisci:
Hey @pronobis, how is this patch working out for you? Can you make a pull request for this on the new github repo here?

pronobis:
It's working great for my Plasma 5.2 + i3 setup. Here's the PR https://github.com/i3/i3/pull/1456

Btw. glad to see i3 on github!

pronobis:
I do have some other KDE-related improvements in my fork, you can check it out.

pronobis:
Here you go. I did open/close a few of the panels before generating the log.
http://logs.i3wm.org/logs/5655869022797824.bz2

Airblader:
The PR has been adopted in #1712 due to inactivity of the original author of the PR.

Airblader:
Due to lack of activity we are closing this issue. If someone feels up to testing the existing PR, please let us know and we can consider merging the fix.

PR #1456, opened 2015-02-08
Use WM_SIZE_HINTS when available to set the geometry of floating windows

pronobis:
This helps a lot when using i3 with KDE/Plasma 5.2 since the intended location of the panel windows is stored in WM_SIZE_HINTS. Without it, the panel windows end up centered in the screen.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -428,6 +428,17 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     if (cwindow->dock)
         want_floating = false;
 
+    /* Plasma windows set their geometry in WM_SIZE_HINTS. */
+    if ((wm_size_hints.width > 0) && (wm_size_hints.height > 0))
+    {
+      DLOG("We are setting geometry according to wm_size_hints x=%d y=%d w=%d h=%d\n",
+           wm_size_hints.x, wm_size_hints.y, wm_size_hints.width, wm_size_hints.height);
+      geom->x = wm_size_hints.x;
+      geom->y = wm_size_hints.y;
+      geom->width = wm_size_hints.width;
+      geom->height = wm_size_hints.height;
+    }
+
     /* Store the requested geometry. The width/height gets raised to at least
      * 75x50 when entering floating mode, which is the minimum size for a
      * window to be useful (smaller windows are usually overlays/toolbars/…


src/manage.c
```diff
@@ -428,6 +428,17 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     if (cwindow->dock)
         want_floating = false;
 
+    /* Plasma windows set their geometry in WM_SIZE_HINTS. */
+    if ((wm_size_hints.width > 0) && (wm_size_hints.height > 0))
```
stapelberg: http://tronche.com/gui/x/xlib/ICC/client-to-window-manager/wm-normal-hints.html says:

> The x, y, width, and height members are now obsolete and are left solely for compatibility reasons.

Could we instead change plasma to just set that geometry on the window before letting the window manager manage it? That way, i3 would pick it up as well, without us having to add support for obsolete properties to i3.

src/manage.c
```diff
@@ -428,6 +428,17 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
...
+    {
```
stapelberg: nit: please use clang-format to make your code adhere to i3’s coding style.

src/manage.c
pronobis: We could try:
https://bugs.kde.org/show_bug.cgi?id=344018

But, also i3 could simply be a bit more tolerant to "older" specs, especially if they do not collide with new specifications.

stapelberg: > But, also i3 could simply be a bit more tolerant to "older" specs, especially if they do not collide with new specifications.

I don’t think it’s clear that this never collides. With your pull request, we’d prefer WM_SIZE_HINTS over the actual geometry, so if an application sets both, but to different values, the application will now exhibit different behavior in i3.

This is why I’m hesitant to merge this, especially given that it’s deprecated.

pronobis: I understand. The thing is that I am definitely not an X11 expert and my patch simply fixed things for me. I am sure, you  are much more qualified to re-write the code to rely on WM_SIZE_HINTS only as a fallback. Btw. how can I check the geometry properties set for the window easily? I use xprop to check the window properties and so far all of the windows that I know set concrete location use WM_SIZE_HINTS.

pronobis: Please also see the response of KDE developers here:
https://bugs.kde.org/show_bug.cgi?id=344018

You are welcome to add your opinion to that discussion.

stapelberg:
Can you upload a logfile? I’d like to see what geometry these windows set. See http://i3wm.org/docs/debugging.html

stapelberg:
From the logfile you provided in #1433:

```
03/10/2015 02:49:09 PM - manage.c:manage_window:208 - startup workspace = (null)
03/10/2015 02:49:09 PM - manage.c:manage_window:259 - Initial geometry: (0, 0, 632, 824)
[…]
03/10/2015 02:49:09 PM - manage.c:manage_window:449 - We are setting geometry according to wm_size_hints x=0 y=582 w=632 h=824
```

src/manage.c
stapelberg: Instead of checking whether the members are > 0, you need to check if flags contains the following bits:

`XCB_ICCCM_SIZE_HINT_US_POSITION` or `XCB_ICCCM_SIZE_HINT_P_POSITION` → x and y are set
`XCB_ICCCM_SIZE_HINT_US_SIZE` or `XCB_ICCCM_SIZE_HINT_P_SIZE` → width and height are set

stapelberg:
Please rebase the patch and address my comments and we can give this a try.

Airblader:
@pronobis It's been 2.5 months. Are you going to keep working on this?

steverweber:
+1

acrisci:
If you want to adopt this PR, @Airblader, go ahead.

Airblader:
@acrisci I could do the changes, but I can't update the log file @stapelberg wants.

Airblader:
@steverweber I have adopted this PR, but since I am not running Plasma, it would be helpful (necessary) for someone to verify it. Since you +1-ed this, maybe you can do that?

Airblader:
I'll actually close this PR given that it has been superseded. It'd be cool for someone to test it so we know whether it works. I'd do it in a VM but setting up Plasma 5 + i3 seems to be a pain (or I'm doing something wrong).

@stapelberg I hope that's okay with you. If not, we can reopen it.

steverweber:
sorry about long delay... confirmed its fixed. thanks!

Issue #628, opened 2012-01-26
Bells for scratchpad


i3bot:
[Originally reported by anonymous]
(Normally, when a window rings a bell, the appropriate titlebar and workspace light up red to notify you, for example, about an instant message you have received. A window in scratchpad has no workspace and bells are thus not displayed.

You can reproduce this problem this way:
-Focus an IRC client
-$ i3-msg move scratchpad
-Have someone highlight you
-Notice you aren't notified of this

My suggestion is that, upon receiving a bell from a window in scratchpad, a workspace is created and has the process' class as name and red as background color. In fact, the workspace could be visible even when there is no bell (but with a regular background), for the sake of accessibility.

Just to be clear: according to this suggestion, every window in scratchpad would have its own workspace.

Airblader:
@stapelberg "We have to add an indicator" is a little vague. Do you have a specific idea of what it should look like (where to put it, how to display the number of scratchpad windows, …)?

stapelberg:
I think it’d be best if the indicator would look just like the mode indicator we have in i3bar (e.g. like an urgent workspace). I think using “scratchpad: %d urgent” as a hard-coded string to signal how many windows are urgent on the scratchpad would work.

Airblader:
Ah, I see. I was thinking about a constant display if there are scratchpad windows, not just if there's an urgent one. Okay. I will try and see if I can get anywhere with this.

acrisci:
Instead of a separate indicator, you could just add a block to the statusline generator.

acrisci:
I implemented this feature in the [scratchpad generator](https://github.com/acrisci/i3-dstatus/blob/master/i3dstatus/generators/scratchpad) for my statusline project, i3-dstatus.

sauyon:
Is there any progress on this?

Airblader:
Nope, feel free to start if you are looking to contribute.

Airblader:
I'm going to close this issue because it can be addressed outside of i3.

PR #1457, opened 2015-02-08
Change iconify request to move window to scratchpad.

doctorjames:
By default the X11 version of Emacs maps key press Ctrl-z (suspend-frame) to iconify the window. In i3 it can be rather destructive pressing this by accident. Sending the window to the scratchpad on the other hand is extremely useful and consistent with the intent. In this case it can be scripted within Emacs lisp too. Presumably this functionality may be equally useful in other applications.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -799,8 +799,8 @@ static void handle_client_message(xcb_client_message_event_t *event) {
             /* this request is so we can play some animiation showing the
              * window physically moving to the tray before we close it (I
              * think) */
-            DLOG("Client has requested iconic state. Closing this con. (con = %p)\n", con);
-            tree_close(con, DONT_KILL_WINDOW, false, false);
+            DLOG("Client has requested iconic state. Moving this con to scratchpad. (con = %p)\n", con);
+            scratchpad_move(con);
             tree_render();
         } else {
             DLOG("Not handling WM_CHANGE_STATE request. (window = %d, state = %d)\n", event->window, event->data.data32[0]);


stapelberg:
I agree that the default behavior is not a good idea, i.e. having Emacs windows be closed on C-z is not great :).

A problem with just sending it to the scratchpad is that users will not know what happened, i.e. the scratchpad has no indicator (yet). I think if we had #628 implemented and would mark the window urgent after moving it to the scratchpad, that’d be okay.

One more thing that warrants a closer look: until now we’ve always closed these windows, and I think that was necessary for some apps, at least at some point in time, to prevent ghost windows, i.e. windows with no content that no app feels responsible for.

doctorjames:
Yes, I agree that an indicator with a count of scratchpad windows/urgency flag would generally be handy in addition to this case. 

A quick peruse of: http://tronche.com/gui/x/icccm/sec-4.html#s-4.2.5 suggests that perhaps it needs a UnmapNotify/MapNotify event sending in case the application is tracking the window state in order to save unnecessary rendering etc. This low level of X11 is not something I have any experience of though.

>The client can elect to be notified of these state changes by selecting for StructureNotify events on the top-level window. It will receive a UnmapNotify event when it goes Iconic and a MapNotify event when it goes Normal.

acrisci:
The spec is wide open in terms of how we can interpret this request. How we interpret it now is bad and I think it needs to change. The term "iconic state" I think is now what we would call "minimization" and the scratchpad is the closest thing we have to that, so I think this is a good idea.

stapelberg:
> One more thing that warrants a closer look: until now we’ve always closed these windows, and I think that was necessary for some apps, at least at some point in time, to prevent ghost windows, i.e. windows with no content that no app feels responsible for.

I think I recall that we just needed to unmap the window, as clients expect that — they typically recover just fine once they get a MapNotify because the window was mapped again.

So, yeah, moving to the scratchpad should work.

The only remaining blocker to merging this is the missing indicator. Also, I think I want to merge this after the next release so that we have some time to see whether it breaks anything (i.e. it will stay in `next` for longer).

stapelberg:
For more context, I just realized https://github.com/i3/i3/commit/136b3e345bfde637d3270eafaa2d8aa3555cc3d5 introduced the behavior in question, so it’s not something that we’ve had for years.

licaon-kter:
This could help this case too: https://github.com/virtual-programming/witcher2-linux/issues/100

stapelberg:
@licaon-kter Thanks for the additional input. Moving the window to scratchpad wouldn’t be perfect in the case you’re describing, but for sure better than closing.

acrisci:
It's really unfortunate that SDL has that behavior.

You have to put this in your shell rc file:

    export SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS=0

stapelberg:
This pull request is obsolete due to d551618cf03695b765b4f6d7f30df8e01847b79f where we reverted the handler for WM_CHANGE_STATE entirely (to fix #1516).

PR #1460, opened 2015-02-10
Correctly calculate clicks on i3bar status blocks

### LINKED ISSUE #1458: Click detection on i3bar is wrong
The current click detection does not take separators into account. When separators are used, this causes the click detection to be more off the more blocks there are and the further right the user clicks, eventually even leading to a click being detected for the wrong block.

The faulty code is `i3bar/src/xcb.c`. In `refresh_statusline`, the `x` coordinate takes `sep_block_width` into account:

````
x += block->width + block->sep_block_width + block->x_offset + block->x_append;
```

But in `handle_button`, the `sep_block_width` is ignored, causing it to slowly diverge:

````
TAILQ_FOREACH(block, &statusline_head, blocks) {
    last_block_x = block_x;
    block_x += block->width + block->x_offset + block->x_append;

    if (x <= block_x && x >= last_block_x) {
        send_block_clicked(event->detail, block->name, block->instance, event->root_x, event->root_y);
        return;
    }
}  
```

Log file: not necessary
Affected: All versions (i3 version 4.8 to make the bot happy, hopefully)

Airblader:
fixes #1458

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -117,6 +117,12 @@ int _xcb_request_failed(xcb_void_cookie_t cookie, char *err_msg, int line) {
     return 0;
 }
 
+uint32_t get_sep_offset(struct status_block *block) {
+    if (!block->no_separator && block->sep_block_width > 0)
+        return block->sep_block_width / 2 + block->sep_block_width % 2;
+    return 0;
+}
+
 /*
  * Redraws the statusline to the buffer
  *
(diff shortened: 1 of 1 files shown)


i3bar/src/xcb.c
```diff
@@ -148,8 +154,8 @@ void refresh_statusline(void) {
                     block->x_offset = padding_width;
                     break;
                 case ALIGN_CENTER:
-                    block->x_offset = padding_width / logical_px(2);
-                    block->x_append = padding_width / logical_px(2) + padding_width % logical_px(2);
+                    block->x_offset = padding_width / 2;
```
Airblader: I calculate half of a value by dividing by anything but 2, no matter what kind of display I have. :)

i3bar/src/xcb.c
```diff
@@ -362,22 +368,31 @@ void handle_button(xcb_button_press_event_t *event) {
                 continue;
             tray_width += (font.height + logical_px(2));
         }
+        if (tray_width > 0)
+            tray_width += logical_px(2);
 
         int block_x = 0, last_block_x;
-        int offset = (walk->rect.w - (statusline_width + tray_width)) - logical_px(10);
```
Airblader: I don't know where the `10` came from, but the code copying the statusline to the buffer uses an offset of `4`.

PR #1465, opened 2015-02-11
Make click events on status blocks work if 'workspace_buttons no' is set...

### LINKED ISSUE #1430: i3bar does not send click events when workspace buttons are hidden
[Originally reported by i3-new-ticket]
(When workspace buttons are hidden 

```
bar {
	...
	workspace_buttons no
}
```

i3bar does not send click events even if the appropriate header block is sent

```
{"version": 1, "click_events": true}
```

click events can be received if `workspace_buttons yes` is set in `bar` block.

Airblader:
....

1. Always subscribe to click events for i3bar.
2. Exit the click event handler if no current workspace was found only after clicks on status blocks have been handled.

fixes #1430

### DIFF
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -520,33 +520,35 @@ void child_click_events_key(const char *key) {
  *
  */
 void send_block_clicked(int button, const char *name, const char *instance, int x, int y) {
-    if (child.click_events) {
-        child_click_events_initialize();
-
-        yajl_gen_map_open(gen);
-
-        if (name) {
-            child_click_events_key("name");
-            yajl_gen_string(gen, (const unsigned char *)name, strlen(name));
-        }
-
-        if (instance) {
-            child_click_events_key("instance");
-            yajl_gen_string(gen, (const unsigned char *)instance, strlen(instance));
-        }
-
-        child_click_events_key("button");
-        yajl_gen_integer(gen, button);
-
-        child_click_events_key("x");
-        yajl_gen_integer(gen, x);
-
-        child_click_events_key("y");
-        yajl_gen_integer(gen, y);
-
-        yajl_gen_map_close(gen);
-        child_write_output();
+    if (!child.click_events) {
+        return;
     }
+
+    child_click_events_initialize();
+
+    yajl_gen_map_open(gen);
+
+    if (name) {
+        child_click_events_key("name");
+        yajl_gen_string(gen, (const unsigned char *)name, strlen(name));
+    }
+
+    if (instance) {
+        child_click_events_key("instance");
+        yajl_gen_string(gen, (const unsigned char *)instance, strlen(instance));
+    }
+
+    child_click_events_key("button");
+    yajl_gen_integer(gen, button);
+
+    child_click_events_key("x");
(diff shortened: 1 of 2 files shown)


i3bar/src/child.c
```diff
@@ -520,33 +520,35 @@ void child_click_events_key(const char *key) {
  *
  */
 void send_block_clicked(int button, const char *name, const char *instance, int x, int y) {
-    if (child.click_events) {
```
Airblader: `git diff` is bugging out here. This is just an early return to get rid of one level of intendation.

i3bar/src/xcb.c
```diff
@@ -364,6 +352,18 @@ void handle_button(xcb_button_press_event_t *event) {
         x = original_x;
     }
 
+    /* TODO: Move this to extern get_ws_for_output() */
```
Airblader: The actual fix. Since i3bar doesn't subscribe to the workspaces event if no workspace are to be drawn, it will not find a `cur_ws` and return. I figured moving this down to after the check for status block clicks is more reasonable than subscribing to the event.

i3bar/src/xcb.c
```diff
@@ -1460,10 +1460,8 @@ void reconfig_windows(bool redraw_bars) {
              * BUTTON_PRESS, to handle clicks on the workspace buttons
              * */
             values[2] = XCB_EVENT_MASK_EXPOSURE |
-                        XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT;
-            if (!config.disable_ws) {
-                values[2] |= XCB_EVENT_MASK_BUTTON_PRESS;
-            }
+                        XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT |
```
Airblader: Also part of the fix; at this point we don't know yet whether the child will request click events, so we need to subscribe to them in any case. This will cause a merge conflict to `i3/next`, but a trivial one.

i3bar/src/child.c
stapelberg: Can you move this to a separate commit, or leave it out of this pull request entirely? That’d make it easier to review the pull request itself.

stapelberg:
Looks good, aside from the one comment about making the commit diff easier to understand.

Airblader:
@stapelberg Sure. Since I can't change commits postum, you'll have to reject this PR and I'll submit a new one.

stapelberg:
@Airblader You can git push --force, can’t you?

Airblader:
@stapelberg I can, but I'm not sure if this will affect the PR. Let's find out!

Airblader:
Cool, it worked. :)

Airblader:
@stapelberg Don't forget to merge this up to `i3/next` :)

stapelberg:
Already did. Please send pull requests for next directly in the future.

Airblader:
@stapelberg This was a bug fix, which according to the guide should go to `i3/master`, no?

stapelberg:
Theoretically yes, but I’d prefer it if I did the backports myself when merging. From a developer point of view, you should only ever have to care about `next`.

Airblader:
Ah, okay. Willdo :)

PR #1478, opened 2015-02-14
Config: use default system monospace

acrisci:
Use the default monospace font for the system in the default config.
This should be a bit more portable for systems that do not have the
recommended font installed.

### DIFF
--- a/i3.config
+++ b/i3.config
@@ -11,9 +11,12 @@
 
 # Font for window titles. Will also be used by the bar unless a different font
 # is used in the bar {} block below.
+font pango:monospace 8
+
 # This font is widely installed, provides lots of unicode glyphs, right-to-left
 # text rendering and scalability on retina/hidpi displays (thanks to pango).
-font pango:DejaVu Sans Mono 8
+#font pango:DejaVu Sans Mono 8
+
 # Before i3 v4.8, we used to recommend this one as the default:
 # font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
 # The font above is very space-efficient, that is, it looks good, sharp and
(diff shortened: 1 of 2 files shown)


acrisci:
I haven't heard any complaints, but I was looking through the default config right now and I think it makes sense to do this. When the recommended font is not installed, it will use the default system font. With this in the config, we can make sure that is a monospace font.

stapelberg:
Can you explain how Pango choses which font gets used when just specifying “monospace”? Could we use a different string that would prefer DejaVu Sans Mono and fall back to the generic monospace, just like in CSS?

lkraav:
The ultimate font choice is most likely up to individual distros fontconfig setup. I'm overriding mine locally on gentoo:
```
    .config/fontconfig/fonts.conf

  <alias>
    <family>serif</family>
    <prefer>
      <family>DejaVu Serif</family>
    </prefer>
  </alias>

  <alias>
    <family>sans-serif</family>
    <prefer>
      <family>DejaVu Sans</family>
    </prefer>
  </alias>

  <alias>
    <family>monospace</family>
    <prefer>
      <family>DejaVu Sans Mono for Powerline</family>
    </prefer>
  </alias>
```

stapelberg:
What’s the default? And what do distributions such as Debian and Fedora use?

acrisci:
If I'm reading the config from [this package](https://packages.debian.org/jessie/fontconfig-config) correctly, it seems that Debian prefers Bitstream Vera, Deja Vu, Liberation, Inconsolata in that order for monospace fonts.

Bitstream Vera fonts fonts are almost identical to Deja Vu.

type `fc-match monospace` to see what your default font is.

stapelberg:
SGTM. Can you please also update i3.config.keycodes?

acrisci:
Ok, I made that change.

PR #1479, opened 2015-02-14
Add default media key support

### LINKED ISSUE #1467: Built in media key support
Media keys should do something by default.

For instance, play, pause, etc. can be implemented with the [mpris dbus specification](http://specifications.freedesktop.org/mpris-spec/latest/).

Volume up/down can be implemented by adjusting the volume of the default alsa card.

In some cases, this could be accomplished as simply as adding things to the default config.

acrisci:
Add bindings in the default config for XF86 multimedia key support for
controlling the audio device, screen brightness, and media players.

fixes #1467

### DIFF
--- a/i3.config
+++ b/i3.config
@@ -164,6 +164,27 @@ bar {
         status_command i3status
 }
 
+# Alsa audio controls
+bindsym XF86AudioRaiseVolume exec --no-startup-id amixer sset Master 5%+
+bindsym XF86AudioLowerVolume exec --no-startup-id amixer sset Master 5%-
+bindsym XF86AudioMute exec --no-startup-id amixer sset Master toggle
+
+# Pulse Audio controls
+#bindsym XF86AudioRaiseVolume exec --no-startup-id pactl set-sink-volume 0 +5%
+#bindsym XF86AudioLowerVolume exec --no-startup-id pactl set-sink-volume 0 -- -5%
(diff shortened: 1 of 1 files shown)


Airblader:
Isn't this just the temporary config used when first starting i3? I think the wizard will use `i3.config.keycodes` to write a new config.

i3.config
```diff
@@ -164,6 +164,27 @@ bar {
...
+bindsym XF86AudioRaiseVolume exec --no-startup-id amixer sset Master 5%+
+bindsym XF86AudioLowerVolume exec --no-startup-id amixer sset Master 5%-
+bindsym XF86AudioMute exec --no-startup-id amixer sset Master toggle
+
+# Pulse Audio controls
+#bindsym XF86AudioRaiseVolume exec --no-startup-id pactl set-sink-volume 0 +5%
+#bindsym XF86AudioLowerVolume exec --no-startup-id pactl set-sink-volume 0 -- -5%
+#bindsym XF86AudioMute exec --no-startup-id pactl set-sink-mute 0 toggle
+
+# Sreen brightness controls
```
stapelberg: typo: s/Sreen/Screen/

stapelberg:
What @Airblader said.

Also, please comment out all of these, as they use commands that may not be installed. The best case is that the commands are installed, but the worst case is that we bind a key and don’t do anything with it whereas another running program might be able to bind it and actually do something with it (e.g. some sort of desktop environment volume/brightness/media control program).

Issue #1480, opened 2015-02-17
i3bar flickers with many status blocks


Airblader:
A user reported a bug on my fork (https://github.com/Airblader/i3/issues/36) about i3bar flickering. After some debugging and poking around, it looks like i3bar has an issue.

If the JSON line becomes too long / has too many elements, parsing might become so slow, that `refresh_statusline` is called on an incompletely parsed line. This causes not all blocks to render for this iteration, causing the flicker.

Though it seems to first appear in my fork due to the increased number of JSON properties, this is definitely a bug in i3 as well and could happen when there are "too many" status blocks or too much content.

What's a reasonable way to address this issue? I had prepared a crude fix to verify my suspicion which has some sort of dirty flag (`is_processing`) that is set/unset in `i3bar/src/child.c` when the parser starts/finishes parsing a new statusline. In `refresh_statusline`, I simply return directly if the flag is currently set (see https://github.com/Airblader/i3/commit/d1a690017868812a1343e360db485039b54fb535)

The first indication is that this seems to work; however, I'm not so sure this is a really good fix. Opinions?

To make the bot shut up (at least about the version):
i3 version 4.8-314-g463a421 (2015-02-16, branch "feature-i3bar-use-pango-markup") © 2009-2014 Michael Stapelberg and contributors

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

Airblader:
One possible way to solve this more nicely would probably be to copy blocks during parsing to a temporary statusline and then copying over the entire statusline when the line is fully parsed.

Airblader:
I've attempted a fix using the latter idea in https://github.com/i3/i3/pull/1481 … see my comment there.

PR #1481, opened 2015-02-18
#1480 Introduce a buffer for the statusline

Airblader:
Some Notes:

* I'm actually really not that into C, so please pay attention to things like memory management. Comparing valgrind's output with and without the patch didn't show me much of a difference, at least nothing I could find in the many warnings that hinted to me introducing a memory problem here; but still…
* You may hate the way to fix this or the implementation of the idea. Again, my C-Fu is really bad and it took me way too long to make this work without segfaults and the like…
* Unlike my "crude fix", this has not yet been verified to actually solve the issue. I will ask the user who reported it to verify the fix once you agreed to the patch. I wouldn't merge it until then.

### DIFF
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -72,6 +72,7 @@ static void clear_statusline(struct statusline_head *head, bool free_resources) 
             I3STRING_FREE(first->full_text);
             FREE(first->color);
             FREE(first->name);
+            FREE(first->instance);
         }
 
         TAILQ_REMOVE(head, first, blocks);
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -54,22 +54,37 @@ typedef struct parser_ctx {
 
 parser_ctx parser_context;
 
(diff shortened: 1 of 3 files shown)


i3bar/src/child.c
```diff
@@ -54,26 +54,40 @@ typedef struct parser_ctx {
...
-    while (!TAILQ_EMPTY(&statusline_head)) {
-        first = TAILQ_FIRST(&statusline_head);
-        I3STRING_FREE(first->full_text);
-        TAILQ_REMOVE(&statusline_head, first, blocks);
+    while (!TAILQ_EMPTY(head)) {
+        first = TAILQ_FIRST(head);
+        if (free_resources) {
+            I3STRING_FREE(first->full_text);
+            FREE(first->color);
+            FREE(first->name);
```
Airblader: `instance` is missing here.

Airblader:
The user is using this patch now and is confirming that it works in terms of avoiding the flickering.

PR #1483, opened 2015-02-19
Docs: Update ipc library list

acrisci:
Remove unmaintained or deprecated ipc libraries from the ipc library
list and add a note about GObject introspection for i3ipc-glib.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -808,18 +808,10 @@ know):
 
 C::
 	* i3 includes a headerfile +i3/ipc.h+ which provides you all constants.
-	* https://github.com/acrisci/i3ipc-glib
+	* https://github.com/acrisci/i3ipc-glib - provides bindings to many languages with its https://wiki.gnome.org/Projects/GObjectIntrospection/Users[GOjbect Introspection] library.
 Go::
 	* https://github.com/proxypoke/i3ipc
-JavaScript::
-	* https://github.com/acrisci/i3ipc-gjs
-Lua::
-	* https://github.com/acrisci/i3ipc-lua
 Perl::
 	* https://metacpan.org/module/AnyEvent::I3
 Python::
 	* https://github.com/acrisci/i3ipc-python
-	* https://github.com/whitelynx/i3ipc (not maintained)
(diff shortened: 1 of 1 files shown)


stapelberg:
I don’t think we should entirely remove existing libraries just because they’re unmaintained. I’m not even sure we should annotate that fact, as it’s easier for users to just check the maintenance level by looking at the project and deciding whether they’re comfortable with that. That way, we don’t need to judge and update the maintenance status of each library regularly.

Therefore, can you revert the removals please?

acrisci:
I think that if a library is not being maintained, it should be removed from the list. It's helpful for people if there is a little quality control for what we list on this page. Delisting the unmaintained projects will help the google rank of the better projects so that more people will find the better projects more easily.

You should at least let me delist my own projects that I want to deprecate (the javascript and lua libraries). The ruby library has a note on the repository directs people to the i3ipc-glib ruby bindings, so I think that is also a clear case of something that should be removed.

i3ipc-python is now pure python and I will support it indefinitely, which makes it better than the other two. i3-py has not been updated in 3 years and lacks support for the window event, which is currently the most useful event. It also lacks python 3 support and has a weird interface.

stapelberg:
> I think that if a library is not being maintained, it should be removed from the list. It's helpful for people if there is a little quality control for what we list on this page. 

And we do that, but by telling people we think it’s unmaintained instead of unlisting entirely. So, please revert the two Python library removals.

> Delisting the unmaintained projects will help the google rank of the better projects so that more people will find the better projects more easily.

Make the link use rel=nofollow if ranking is a corncern, see also https://support.google.com/webmasters/answer/96569?hl=en

> You should at least let me delist my own projects that I want to deprecate (the javascript and lua libraries). The ruby library has a note on the repository directs people to the i3ipc-glib ruby bindings, so I think that is also a clear case of something that should be removed.

What bothered me about removing these is that readers will get the impression that there are no libraries for these languages when having a quick glance at the list.

Could you instead move i3ipc-glib to a new category called “C, JavaScript, Ruby, …” (put in whichever languages are known to work with the glib bindings)?

Issue #840, opened 2012-10-11
Allow cycling through windows when using '[title="foo"] focus'


i3bot:
[Originally reported by anonymous]
(At the moment when doing something like
 $ i3-msg '[title="foo"] focus'
the 'focus' command is issued to all of the matching windows. Presumably in order of their creation, as the focus rests on the window, which was created last.

While this behavior is quite useful when doing 
 $ i3-msg '[title="foo"] move workspace current
or
 $ i3-msg '[title="foo"] border none
I think it is not so useful with focus, as only one window can (or should?) actually have the focus. An additional side effect is that any urgent flags on matching windows get lost.

Would it be possible to add something like this
 [title="foo"] focus next
 [title="foo"] focus previous
or maybe like this using the 'window'-field I saw with get_tree
 [title="foo" window="next_higher"] focus
 [title="foo" window="next_lower"] focus
so that one can cycle through the windows one by one?

Thanks

aloispichler:
[alttab](https://github.com/sagb/alttab) does the job.

PR #1488, opened 2015-02-22
Focus next

wentasah:
Hi,

this is a RFC, not a full-featured pull request. I consider switching to i3 and I miss some features, namely something similar to awesome's run-or-raise. There is a solution at [i3 FAQ](https://faq.i3wm.org/question/2473/run-or-focus-in-i3/?answer=2792#post-id-2792) that might work (with some changes) for me but I find it unnecessarily complex.

With these patches, my desired run-or-raise functionality can be implemented easily as:

    if i3-msg '[class="XTerm" workspace=current] focus next'|jq -e '.[0].success == false'; then
        exec xterm
    fi

### DIFF
--- a/include/commands.h
+++ b/include/commands.h
@@ -187,6 +187,12 @@ void cmd_focus_level(I3_CMD, char *level);
 void cmd_focus(I3_CMD);
 
 /**
+ * Implementation of 'focus next'.
+ *
+ */
+void cmd_focus_next(I3_CMD);
+
+/**
  * Implementation of 'fullscreen [enable|disable|toggle] [global]'.
  *
  */
--- a/include/data.h
+++ b/include/data.h
@@ -408,6 +408,8 @@ struct Match {
     struct regex *instance;
     struct regex *mark;
     struct regex *window_role;
+    struct regex *workspace;
+    bool current_workspace;
     enum {
         U_DONTCHECK = -1,
         U_LATEST = 0,
--- a/include/match.h
+++ b/include/match.h
@@ -41,6 +41,13 @@ void match_copy(Match *dest, Match *src);
 bool match_matches_window(Match *match, i3Window *window);
 
 /**
+ * Check if a match data structure matches the workspace of the given
+ * container.
+ *
+ */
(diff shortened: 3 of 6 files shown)


stapelberg:
You’re not describing what “focus next” does. See https://github.com/i3/i3/issues/840#comment:1 as for why we’re not going to implement such a command (if it does what I think it does).

> I consider switching to i3 and I miss some features

Typically I advise that you do the switch and get accustomed to the way things work in i3 before trying to make i3 more like the thing you’ve used before. Often, you’ll realize that what you asked for is not necessary in i3. This is more a general comment than really applicable in this situation, though :).

> There is a solution at i3 FAQ that might work (with some changes) for me but I find it unnecessarily complex.

The code changes you’re trying to do in i3 are more complex than the script you’re referring to, judging by lines of code.

I’m going to close this in the assumption that I’m right about what “focus next” does, but feel free to reopen in case I got it all wrong.

wentasah:
> You’re not describing what “focus next” does. 

I do it in the commit message.

> See https://github.com/i3/i3/issues/840#comment:1 as for why we’re not
> going to implement such a command (if it does what I think it does).

I expected this answer (as mentioned in the commit message too), but
thanks for the link. I don't fully agree with the arguments there, but I
understand your position. The 'focus next' command is not crucial for me
now.

What about the [other patch (workspace
match)](https://github.com/wentasah/i3/commit/6d92e21a16731a87743efc48cc44c96af43caadc)?
Do you think it might be useful for others?

Thanks.
-Michal

stapelberg:
>> You’re not describing what “focus next” does.

> I do it in the commit message.

Ah, I missed that. The reply still stands :).

> What about the [other patch (workspace
> match)](https://github.com/wentasah/i3/commit/6d92e21a16731a87743efc48cc44c96af43caadc)?
> Do you think it might be useful for others?

Yeah, I think adding a workspace criterion is fine. As for the syntax for current, one way to avoid the conflict would be to add a “workspace” and a “current_workspace” criterion (just like we have an “urgent” criterion). Can you submit that in a new pull request please?

Issue #838, opened 2012-10-06
back and forth between windows


i3bot:
[Originally reported by simao.m@…]
(Feature Request:

Currently, there's a command that allows the user to go back and forth between workspaces.

The same feature for windows would be very good, one common use case in a developer workflow is to go back and forth between two windows (editor/terminal for example), so this would increase productivity.

Thanks!

mikedmcfarland:
FYI. I've also written a script that also gives this feature
http://thrownforaloop.com/posts/i3-finder/
uses i3-msg, not the ipc though.

acrisci:
This should be closed.

stapelberg:
Closing as requested.

gvalkov:
Hello,

I'm terribly used to awesomewm's `$mod+Tab` behavior (i.e. cycle between the two most recently focused windows). It would have been nice if this was in core (as implemented in #1492), but here's my take on an ipc solution.

**event-listener.py:**

```python
#!/usr/bin/env python3

import os
import asyncio
import i3ipc

socket = '{XDG_RUNTIME_DIR}/i3/event-listener.sock'.format_map(os.environ)
events = asyncio.Queue()
window_stack = []

def enqueue(*args):
    events.put_nowait(args)

async def i3_event_listener():
    while True:
        conn, event = await events.get()
        if event.change == 'focus':
            window_stack.insert(0, event.container.props.id)
            if len(window_stack) > 2:
                del window_stack[2:]

async def on_unix_connect(reader, writer):
    while True:
        cmd = await reader.readline()
        cmd = cmd.strip()
        if not cmd:
            break

        if cmd == b'swap_focus' and window_stack:
            cmd = '[con_id=%s] focus' % window_stack.pop()
            i3.command(cmd)

async def main():
    if os.path.exists(socket):
        os.unlink(socket)

    server = await asyncio.start_unix_server(on_unix_connect, socket)
    asyncio.ensure_future(i3_event_listener())
    await server.wait_closed()

i3 = i3ipc.Connection()
i3.on('window::focus', enqueue)
i3.event_socket_setup()

loop = asyncio.get_event_loop()
loop.add_reader(i3.sub_socket, i3.event_socket_poll)
loop.run_until_complete(main())
```

**i3/config**

```
bindsym $mod+Tab exec echo "swap_focus" | nc --send-only -U "$XDG_RUNTIME_DIR/i3/event-listener.sock"
exec --no-startup-id event-listener.py
```

[...]

PR #1492, opened 2015-02-26
feature: focus previous

choopm:
allows quick switching to the previous focused window on the current workspace

see also: https://github.com/i3/i3/issues/838


stapelberg:
See https://github.com/i3/i3/issues/838#issuecomment-72825126, this should be implemented using the IPC interface.

PR #1503, opened 2015-02-28
Glob filepath when calling append_layout

### LINKED ISSUE #1500: enhancement: append_layout should glob paths/replace ~
See also #1499 for context.

Airblader:
fixes #1500

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -900,11 +900,15 @@ void cmd_nop(I3_CMD, char *comment) {
 void cmd_append_layout(I3_CMD, char *path) {
     LOG("Appending layout \"%s\"\n", path);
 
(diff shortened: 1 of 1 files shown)


src/commands.c
```diff
@@ -900,6 +900,9 @@ void cmd_nop(I3_CMD, char *comment) {
 void cmd_append_layout(I3_CMD, char *path) {
     LOG("Appending layout \"%s\"\n", path);
 
+    /* Make sure we allow paths like '~/.i3/layout.json' */
+    path = resolve_tilde(path);
```
stapelberg: This is a memory leak. You’ll need to `free(path);`, since `resolve_tilde` allocates memory.

Airblader: @stapelberg Done

PR #1512, opened 2015-03-03
Fix scrolling in window decoration with hidden cursor.

### LINKED ISSUE #1511: Scrolling in window decoration with hidden cursor doesn't cycle stack.
If the mouse cursor is hidden (i.e. with unclutter), then scrolling in the window decoration doesn't cycle the stack. Instead, it just highlights all the window decorations simultaneously without changing the window. It is easily reproducible by...

1. unclutter -idle 1
2. place cursor on window decoration
3. wait for the cursor to disappear
4. scroll through stack

The bug was introduced in commit 491274d8 about a year ago to fix issue #1176.

I've fixed this and also confirmed it causes no regression on issue #1176. I'll submit a pull request soon.

mdtipton:
If the mouse cursor is hidden (by unclutter, for example), then scrolling
in the window decoration creates an event with a child
(i.e. event->child != XCB_NONE). This causes route_click() to be called
with dest=CLICK_INSIDE, which prevents scrolling through a stacked layout.

To fix this, check if a click is in the window decoration _before_
checking if the event has a child.

Fixes #1511

### DIFF
--- a/src/click.c
+++ b/src/click.c
@@ -380,11 +380,6 @@ int handle_button_press(xcb_button_press_event_t *event) {
         return 0;
     }
 
-    if (event->child != XCB_NONE) {
-        DLOG("event->child not XCB_NONE, so this is an event which originated from a click into the application, but the application did not handle it.\n");
-        return route_click(con, event, mod_pressed, CLICK_INSIDE);
-    }
-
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for the detailed description!

PR #1513, opened 2015-03-04
Move workspace to assigned output when renaming it

### LINKED ISSUE #1473: enhancement: rename workspace and assignments
When a workspace is renamed to the name of a workspace is assigned to a particular monitor, the workspace should be moved to that monitor.

Airblader:
fixes #1473


stapelberg:
Nice!

PR #1520, opened 2015-03-04
Properly error out when the layout file cannot be read.

### LINKED ISSUE #1499: Empty error message for append_layout errors.
@siassaj reports that with i3 4.9-2-gcc55ee4 (2015-02-28, branch "next") starting, using `i3-msg "workspace 2; append_layout ~/.i3/workspace_2.json"` results in:

```text
ERROR: Your command: (null)
ERROR:               (null)
ERROR: (null)
[{"success":true},{"success":false}]
```

Based on the logfile http://logs.i3wm.org/logs/5675267779461120.bz2, I think the error is empty when append_layout cannot open the provided path.

Airblader:
This will result in an actual error message for the user.
fixes #1499


stapelberg:
I merged this, but github doesn’t realize (because I cherry-picked it into master).

PR #1521, opened 2015-03-05
Use the already existing function to get rid of some code duplication

Airblader:


stapelberg:
Thanks for the cleanup :).

Issue #1527, opened 2015-03-06
Workspaces in startup sequences not renamed on "rename workspace"


mh21:
When renaming workspaces with "rename workspace", workspaces that are mentioned in startup sequences are not renamed.

This is problematic for e.g. firefox, where the transient keyring-passphrase window appearing several seconds after startup also uses the startup workspace. When the workspace was renamed in the meanwhile, this dialog ends up on a new workspace out of sight.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

mh21:
This was on `next`: i3 version 4.9-27-gb083471 (2015-03-07, branch "next") © 2009-2014 Michael Stapelberg and contributors

mh21:
Logs: http://logs.i3wm.org/logs/5685265389584384.bz2

Airblader:
Why would i3 rename workspaces in commands you haven't executed yet? I'm not sure I understand the problem correctly.

mh21:
Hi,

the following steps show the problem:

  -> start firefox

  -> a startup notification context is created and i3 sets up a
startup sequence. The launch workspace (WS1) is saved in the startup
sequence so the application window shows up on the correct workspace.

  -> the firefox window appears and is mapped on the correct workspace
(WS1), but the startup sequence stays around for a while longer.
Quoting src/startup.c:
/*
 * Some applications (such as Firefox) mark a startup sequence as completed
 * *before* they even map a window. Therefore, we cannot entirely delete the
 * startup sequence once it’s marked as complete. Instead, we’ll mark it for
 * deletion in 30 seconds and use that chance to delete old sequences.
 */

  -> somebody renames the workspace (with firefox on it) to something
else (WS1->WS2). The startup sequence still contains the old workspace
name (WS1), which is now not the same as the workspace with the
firefox window on it (WS2).

  -> firefox shows the keyring password dialog. Because the startup
sequence is still around, this window also gets mapped to the same
workspace *name* (WS1) as before. Because the workspace has been
renamed in the meanwhile (to WS2), a new workspace is created (for
WS1).

  -> Now there are two workspaces: one with the firefox window (WS2),
and one with the keyring password dialog (WS1), I think the intention
[...]

PR #1528, opened 2015-03-06
Rename workspaces in startup sequences

mh21:
When renaming workspaces with `rename workspace`, workspaces that are mentioned in startup sequences are not renamed (issue #1527).

First commit is a test case that fails in `next`.

I'm not really sure where `startup_sequence_rename_workspace(old_name, new_name)` should go exactly in `cmd_rename_workspace`.


mh21:
An updated version with the correct branch can be found in PR #1539.

PR #1539, opened 2015-03-10
Rename workspaces in startup sequences

mh21:
When renaming workspaces with `rename workspace`, workspaces that are mentioned in startup sequences are not renamed (issue #1527).

First commit is a test case that fails in `next`.

I'm not really sure where `startup_sequence_rename_workspace(old_name, new_name)` should go exactly in `cmd_rename_workspace`.

This is an updated version of PR #1528 with the correct branch attached.

### DIFF
--- a/include/startup.h
+++ b/include/startup.h
@@ -45,6 +45,12 @@ void startup_sequence_delete(struct Startup_Sequence *sequence);
 void startup_monitor_event(SnMonitorEvent *event, void *userdata);
 
 /**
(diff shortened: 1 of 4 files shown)


stapelberg:
This looks good to me, but you’re changing the file permission of the testcase. Can you please revert that? They shouldn’t be executable, as they should be executed by the `complete-run.pl` script only.

mh21:
I reverted the mode change in the first commit.

Issue #1538, opened 2015-03-09
Warnings when compiling with a higher optimisation level


ashinkarov:
If I compile i3 with higher level of optimisations:
```diff
diff --git a/common.mk b/common.mk
index b9e15a2..a4a8a00 100644
--- a/common.mk
+++ b/common.mk
@@ -1,5 +1,5 @@
 UNAME=$(shell uname)
-DEBUG=1
+DEBUG=0
 COVERAGE=0
 INSTALL=install
 LN=ln
@@ -36,7 +36,7 @@ ifeq ($(DEBUG),1)
 # Extended debugging flags, macros shall be available in gcc
 CFLAGS ?= -pipe -gdwarf-2 -g3
 else
-CFLAGS ?= -pipe -O2 -freorder-blocks-and-partition
+CFLAGS ?= -pipe -O3 -march=native -mtune=native -freorder-blocks-and-partition
 endif
 
 # Default LDFLAGS that users should be able to override

```
I get the following list of previously unseen warnings:
```
../i3/src/load_layout.c: In function ‘json_end_map’:
../i3/src/load_layout.c:108:17: warning: ignoring return value of ‘asprintf’, declared with attribute warn_unused_result [-Wunused-result]
                 asprintf(&(json_node->name), "%s_%d", base, cnt++);
                 ^
../i3/src/floating.c: In function ‘drag_pointer’:
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     ev_check_init(&loop.check, xcb_drag_check_cb);
     ^
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
../i3/src/click.c: In function ‘tiling_resize_for_border’:
../i3/src/click.c:51:10: warning: ‘search_direction’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     bool res = resize_find_tiling_participants(&first, &second, search_direction);
          ^
../i3/src/sighandler.c: In function ‘backtrace’:
../i3/src/sighandler.c:73:9: warning: ignoring return value of ‘pipe’, declared with attribute warn_unused_result [-Wunused-result]
         pipe(stdin_pipe);
         ^
[...]

PR #1540, opened 2015-03-10
Fix warnings

ashinkarov:
This fixes all the warnings from #1538 except this one:
```
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     ev_check_init(&loop.check, xcb_drag_check_cb);
     ^
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
../i3/src/click.c: In function ‘tiling_resize_for_border’:
```
I don't understand `ev` well enough to fix this.  Maybe someone else can have a look?

### DIFF
--- a/libi3/safewrappers.c
+++ b/libi3/safewrappers.c
@@ -81,3 +81,11 @@ int spipe(int pipefd[2]) {
         err(EXIT_FAILURE, "pipe((int[]){%d,%d})", pipefd[0], pipefd[1]);
     return result;
 }
+
+int ssymlink(const char *target, const char *linkpath) {
+    ssize_t result = symlink(target, linkpath);
+
+    if (-1 == result)
+        err(EXIT_FAILURE, "symlink(\"%s\", \"%s\")", target, linkpath);
+    return result;
+}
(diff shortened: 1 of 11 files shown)


libi3/safewrappers.c
```diff
@@ -56,3 +65,27 @@ int sasprintf(char **strp, const char *fmt, ...) {
     va_end(args);
     return result;
 }
+
+ssize_t swrite(int fd, const void *buf, size_t count) {
+    ssize_t result = write (fd, buf, count);
```
stapelberg: Please use `clang-format` to format your source code (affects this line and others).

While you’re at it, please revert the yoda conditions as well (e.g. `if (-1 == result)` should be `if (result == -1)` for consistency with the rest of the code base).

stapelberg:
See http://lists.schmorp.de/pipermail/libev/2010q1/000917.html for more infos on the libev-related warnings.

Airblader:
@ashinkarov Will you keep working on this?

ashinkarov:
No, I won't.  Most of the stuff is fixed by the #1575.

Airblader:
OK, then there's no point in keeping this open. If someone wants to do the rest, they can take over. I'll close this PR. 

Thanks for the feedback!

PR #1545, opened 2015-03-12
degendered terms

garbados:
changed gendered terms like "he" and "his" to "they" and "their"

### DIFF
--- a/docs/hacking-howto
+++ b/docs/hacking-howto
@@ -1012,7 +1012,7 @@ gets started in any way) and the window(s) which appear.
 
 Imagine for example using dmenu: The user starts dmenu by pressing Mod+d, dmenu
 gets started with PID 3390. The user then decides to launch Firefox, which
-takes a long time. So he enters firefox into dmenu and presses enter. Firefox
+takes a long time. So they enter firefox into dmenu and press enter. Firefox
 gets started with PID 4001. When it finally finishes loading, it creates an X11
 window and uses MapWindow to make it visible. This is the first time i3
 actually gets in touch with Firefox. It decides to map the window, but it has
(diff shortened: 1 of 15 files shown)


garbados:
for [this line](https://github.com/i3/i3/blob/next/src/randr.c#L659) i wasn't sure if "his" referred to a male user or some lowercase acronym, so i left it alone pending confirmation.

garbados:
hmm, but "he or she" excludes people who aren't he or she, ex: nonbinary folk. they doesn't.

garbados:
also, um

https://en.wikipedia.org/wiki/Singular_they#Usage_guidance_in_English_grammars

> According to A Comprehensive Grammar of the English Language (1985):
>
> > The pronoun they is commonly used as a 3rd person singular pronoun that is neutral between masculine and feminine. . . . At one time restricted to informal usage. it is now increasingly accepted in formal usage [...]

acrisci:
I really don't think "nonbinary folk" will feel not included by everyday normal English.

I can list several style guides that advise against it from the same article. The fact that it is controversial at all is a reason to avoid it when you can.

There is a difference between someone saying it is "increasingly accepted in formal usage" and putting a ban on gendered pronouns and mandating the exclusive usage of singular they like you want to do. Can you find any reputable source that says gendered pronouns are bad style?

garbados:
scare quotes around nonbinary? i'm sorry, i don't want to make this thread unprofessional. i'll message you on irc.

Airblader:
I'm strongly in favor of the singular they. As far as I know it's not academically accepted, but other than that fine. This isn't an academic document. 

It's much nicer to read than the terrible "he or she"  every single time a pronoun is used.

acrisci:
I thought our documentation language was already gender inclusive, but @garbados has told me that they do not feel included so I retract my strong objection to this.

stapelberg:
Thanks. I find it interesting how there are both the generic masculine and generic feminine form sprinkled across comments and docs.

With regards to the line you weren’t sure about: replacing “his” by “output” would be my suggestion.

garbados:
cool! i'll make another pr for the `his -> output` change :)

PR #1553, opened 2015-03-16
Remove the A_TO_OUTPUT option

Airblader:


Airblader:
@stapelberg I believe that the tests should be successful and that this is an unrelated timing issue. The test is also successful locally for me.

acrisci:
:+1:

stapelberg:
I think c35b3160f6e0c07cfbaa3a09baed6ebde48c8417 addresses the test failure. As you can see in the details, one run succeeded and one didn’t.

stapelberg:
Thanks for the cleanup!

PR #1555, opened 2015-03-18
Make the --workspace optional by defaulting to the focused workspace.

Airblader:
If no option is given, i3-save-tree should default to the currently focused workspace.
Specifying both --workspace and --output will still yield an error.


Airblader:
I guess I should modify the usage/help text, too.

PR #1556, opened 2015-03-18
Allow referencing the workspace by number.

Airblader:
This allows "--workspace 3" for numbered workspaces, e.g., if the name is
actually "3:foo". This introduces the same functionality the IPC already
offers in many places.

### DIFF
--- a/i3-save-tree
+++ b/i3-save-tree
@@ -222,7 +222,7 @@ my $tree = $i3->get_tree->recv;
 my $dump;
 if (defined($workspace)) {
     $dump = filter_containers($tree, sub {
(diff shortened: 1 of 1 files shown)


Airblader:
Just to add some reasoning: obviously there's no use-case to script calling `i3-save-tree`, but I, for example, name workspaces with an icon from an iconic font. If I want to call `i3-save-tree`, I have to go into my config and copy the name since I can't just type out the icon. Hence this introduces a nice convenience factor (for me :smile: )

Airblader:
I guess I should modify the usage/help text, too.

stapelberg:
This change looks good to me, but can you rebase it please?

Airblader:
@stapelberg Done

PR #1557, opened 2015-03-19
Bugfix: validation segfault

acrisci:
When `new_window` is given in the config, config validation with `i3 -C`
would segfault.

Add a NULL check in logical_px() to check for the case when the config
is being validated without an X server to prevent this.

### DIFF
--- a/libi3/dpi.c
+++ b/libi3/dpi.c
@@ -17,6 +17,12 @@ extern xcb_screen_t *root_screen;
  *
  */
 int logical_px(const int logical) {
+    if (root_screen == NULL) {
+        /* Dpi info may not be available when parsing a config without an X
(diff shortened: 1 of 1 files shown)


libi3/dpi.c
```diff
@@ -17,6 +17,13 @@ extern xcb_screen_t *root_screen;
  *
  */
 int logical_px(const int logical) {
+    if (root_screen == NULL) {
+        /* Dpi info may not be available when parsing a config without an X
+         * server, such as for validation.
```
stapelberg: nit: s/validation/config file validation/, also please end the comment on that same line (no need for the newline).

acrisci:
Ok, I made that change.

PR #1561, opened 2015-03-21
Added a troubleshooting section to the restoring layout docs

Airblader:


docs/layout-saving
```diff
@@ -231,3 +231,31 @@ layouts, you can either use a JSON parser that supports these deviations (for
 example libyajl), transform the layout file to a JSON-conforming file, or
 link:http://cr.i3wm.org/[submit a patch] to make +i3-save-tree(1)+ optionally
 output standard-conforming JSON.
+
+== Troubleshooting
+
+=== Restoring a vertically split container
```
stapelberg: suggestion: s/container/workspace, because it’s not actually a problem for arbitrary containers, only for workspaces.

Airblader:
Updated. :smile:

PR #1564, opened 2015-03-22
i3bar: set markup per block

acrisci:
Add `markup` to the i3bar protocol as a block member.

This is a string that determines how the block should be parsed as
markup. "pango" indicates the block should be parsed as Pango markup.
"none" indicates the block should not be parsed as markup.

### DIFF
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -38,11 +38,18 @@ struct status_block {
     i3String *short_text;
 
     char *color;
+
+    /* min_width can be specified either as a numeric value (in pixels) or as a
+     * string. For strings, we set min_width to the measured text width of
+     * min_width_str. */
     uint32_t min_width;
+    char *min_width_str;
+
     blockalign_t align;
 
     bool urgent;
     bool no_separator;
+    bool is_markup;
 
     /* The amount of pixels necessary to render a separater after the block. */
     uint32_t sep_block_width;
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -74,6 +74,7 @@ static void clear_statusline(struct statusline_head *head, bool free_resources) 
             FREE(first->color);
             FREE(first->name);
             FREE(first->instance);
+            FREE(first->min_width_str);
         }
 
         TAILQ_REMOVE(head, first, blocks);
@@ -163,6 +164,9 @@ static int stdin_start_map(void *context) {
     /* Default width of the separator block. */
     ctx->block.sep_block_width = logical_px(9);
 
+    /* Use markup by default */
+    ctx->block.is_markup = true;
+
     return 1;
 }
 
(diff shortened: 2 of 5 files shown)


acrisci:
Me and @vivien came up with this as a pretty good solution for the markup parsing problem some people have had with their i3blocks scripts.

acrisci:
There may be conflicts with #1562 but they will be easy to fix.

Airblader:
I don't mind this being merged first. #1562 is waiting on another merge anyway, so I can do the rebasing.

i3bar/src/child.c
```diff
@@ -191,6 +195,13 @@ static int stdin_string(void *context, const unsigned char *val, size_t len) {
     if (strcasecmp(ctx->last_map_key, "color") == 0) {
         sasprintf(&(ctx->block.color), "%.*s", len, val);
     }
+    if (strcasecmp(ctx->last_map_key, "markup") == 0) {
+        if (len == strlen("pango") && !strncmp((const char *)val, "pango", strlen("pango"))) {
```
stapelberg: Simpler and more robust (because case-insensitive and treating everything else but pango as false):

```c
ctx->block.is_markup = (len == strlen("pango") && strncasecmp((const char *)val, "pango", strlen("pango")));
```

i3bar/src/child.c
```diff
@@ -162,6 +163,9 @@ static int stdin_start_map(void *context) {
     /* Default width of the separator block. */
     ctx->block.sep_block_width = logical_px(9);
 
+    /* Use markup by default */
+    ctx->block.is_markup = true;
```
stapelberg: Hm, but in #1565 you suggested the opposite of this…?

acrisci: Yes, I'd like to decouple the two issues. Removing these two lines should be enough to fix #1565.

i3bar/include/common.h
```diff
@@ -42,6 +42,11 @@ struct status_block {
 
     bool urgent;
     bool no_separator;
+    bool is_markup;
+
+    /* When min_width is given as an str, the width of this string becomes
```
stapelberg: nit: s/an/a/, and I find that comment confusingly worded. I suggest moving the `min_width_str` member directly below the `min_width` member and adding this comment above `min_width`:

min_width can be specified either as a numeric value (in pixels) or as a string. For strings, we set min_width to the measured text width of min_width_str.

acrisci:
Ok, made that change.

include/libi3.h
```diff
@@ -211,6 +211,11 @@ size_t i3string_get_num_bytes(i3String *str);
  */
 bool i3string_is_markup(i3String *str);
 
+/*
```
stapelberg: nit: you’re missing the second asterisk here. Comments in .h files need to start with /** so that doxygen considers them documentation.

stapelberg:
Looks good. Will merge once you fix the final comment.

Airblader:
You probably have to rebase this now, too.

acrisci:
Ok, rebased and made the change.

PR #1568, opened 2015-03-23
testcases: dont run valgrind verbose

acrisci:


acrisci:
If you use the verbose info, we can leave it in.

I only use valgrind to find segfaults and memory leaks and verbose output has not ever helped me with that.

PR #1574, opened 2015-03-24
i3bar: check block text for NULL

acrisci:
Add a null check to `full_text` and `short_text` for a block before
setting the markup property to avoid a segfault.

### DIFF
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -286,7 +286,7 @@ static int stdin_end_array(void *context) {
     struct status_block *current;
     TAILQ_FOREACH(current, &statusline_head, blocks) {
         DLOG("full_text = %s\n", i3string_as_utf8(current->full_text));
-        DLOG("short_text = %s\n", i3string_as_utf8(current->short_text));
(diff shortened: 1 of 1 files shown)


i3bar/src/child.c
```diff
@@ -264,8 +264,11 @@ static int stdin_end_map(void *context) {
         i3string_free(text);
     }
 
-    i3string_set_markup(new_block->full_text, new_block->is_markup);
-    i3string_set_markup(new_block->short_text, new_block->is_markup);
+    if (new_block->full_text != NULL)
+        i3string_set_markup(new_block->full_text, new_block->is_markup);
```
acrisci: The check on `full_text` is not strictly required because it's caught as a spec violation earlier, but it's safer.

acrisci:
@Airblader, the segfault happens when you try to set the short text to markup.

Airblader:
Mh, odd. I also tried short text with markup because I wanted to make sure markup works there. Oh well, either way it'll be fixed now. :)

acrisci:
There are still some problems that need to be worked out.

Airblader:
Which are?

acrisci:
I think that should fix it.

PR #1575, opened 2015-03-24
Fix warnings

hwangcc23:
This fixes warnings from the issue #1538 except:
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     ev_check_init(&loop.check, xcb_drag_check_cb);
     ^
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
../i3/src/floating.c:743:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]

Need to study ev for this warning. Wondering it is caused by ev.

### DIFF
--- a/i3-nagbar/main.c
+++ b/i3-nagbar/main.c
@@ -164,7 +164,9 @@ static void handle_button_release(xcb_connection_t *conn, xcb_button_release_eve
     char *link_path;
     char *exe_path = get_exe_path(argv0);
     sasprintf(&link_path, "%s.nagbar_cmd", script_path);
-    symlink(exe_path, link_path);
+    if (symlink(exe_path, link_path) == -1) {
+        err(EXIT_FAILURE, "Failed to symlink %s to %s", link_path, exe_path);
+    }
 
     char *terminal_cmd;
     sasprintf(&terminal_cmd, "i3-sensible-terminal -e %s", link_path);
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -103,7 +103,7 @@ __attribute__((format(printf, 1, 2))) static void set_statusline_error(const cha
     char *message;
     va_list args;
     va_start(args, format);
-    vasprintf(&message, format, args);
+    (void)vasprintf(&message, format, args);
 
     struct status_block *err_block = scalloc(sizeof(struct status_block));
     err_block->full_text = i3string_from_utf8("Error: ");
@@ -455,11 +455,22 @@ void child_write_output(void) {
     if (child.click_events) {
         const unsigned char *output;
         size_t size;
+        ssize_t n;
 
         yajl_gen_get_buf(gen, &output, &size);
-        write(child_stdin, output, size);
-        write(child_stdin, "\n", 1);
+
+        n = writeall(child_stdin, output, size);
+        if (n != -1)
+            n = writeall(child_stdin, "\n", 1);
+
         yajl_gen_clear(gen);
+
+        if (n == -1) {
+            child.click_events = false;
+            kill_child();
+            set_statusline_error("child_write_output failed");
+            draw_bars(false);
+        }
     }
 }
 
--- a/i3bar/src/ipc.c
+++ b/i3bar/src/ipc.c
@@ -296,18 +296,7 @@ int i3_send_msg(uint32_t type, const char *payload) {
     if (payload != NULL)
         strncpy(walk, payload, len);
 
-    uint32_t written = 0;
-
-    while (to_write > 0) {
-        int n = write(i3_connection->fd, buffer + written, to_write);
-        if (n == -1) {
-            ELOG("write() failed: %s\n", strerror(errno));
-            exit(EXIT_FAILURE);
-        }
-
-        to_write -= n;
-        written += n;
-    }
+    swrite(i3_connection->fd, buffer, to_write);
 
     FREE(buffer);
 
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -135,6 +135,20 @@ char *sstrdup(const char *str);
 int sasprintf(char **strp, const char *fmt, ...);
 
 /**
+ * Wrapper around correct write which returns -1 (meaning that
+ * write failed) or count (meaning that all bytes were written)
+ *
+ */
+ssize_t writeall(int fd, const void *buf, size_t count);
+
+/**
+ * Safe-wrapper around writeall which exits if it returns -1 (meaning that
+ * write failed)
+ *
+ */
+ssize_t swrite(int fd, const void *buf, size_t count);
+
+/**
  * Build an i3String from an UTF-8 encoded string.
  * Returns the newly-allocated i3String.
  *
--- a/libi3/ipc_send_message.c
+++ b/libi3/ipc_send_message.c
@@ -32,33 +32,11 @@ int ipc_send_message(int sockfd, const uint32_t message_size,
         .size = message_size,
         .type = message_type};
 
-    size_t sent_bytes = 0;
-    int n = 0;
+    if (writeall(sockfd, ((void *)&header), sizeof(i3_ipc_header_t)) == -1)
+        return -1;
 
-    /* This first loop is basically unnecessary. No operating system has
-     * buffers which cannot fit 14 bytes into them, so the write() will only be
-     * called once. */
-    while (sent_bytes < sizeof(i3_ipc_header_t)) {
-        if ((n = write(sockfd, ((void *)&header) + sent_bytes, sizeof(i3_ipc_header_t) - sent_bytes)) == -1) {
-            if (errno == EAGAIN)
-                continue;
-            return -1;
-        }
-
-        sent_bytes += n;
-    }
-
-    sent_bytes = 0;
-
-    while (sent_bytes < message_size) {
-        if ((n = write(sockfd, payload + sent_bytes, message_size - sent_bytes)) == -1) {
-            if (errno == EAGAIN)
-                continue;
-            return -1;
-        }
-
-        sent_bytes += n;
-    }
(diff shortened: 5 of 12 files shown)


stapelberg:
See https://github.com/i3/i3/pull/1540 for discussion.

i3-nagbar/main.c
```diff
@@ -164,7 +164,9 @@ static void handle_button_release(xcb_connection_t *conn, xcb_button_release_eve
     char *link_path;
     char *exe_path = get_exe_path(argv0);
     sasprintf(&link_path, "%s.nagbar_cmd", script_path);
-    symlink(exe_path, link_path);
+    if (symlink(exe_path, link_path) == -1) {
+        printf("Failed to symlink %s, %s\n", exe_path, link_path);
```
stapelberg: Use `err()` instead.

i3bar/src/child.c
```diff
@@ -103,7 +103,11 @@ __attribute__((format(printf, 1, 2))) static void set_statusline_error(const cha
     char *message;
     va_list args;
     va_start(args, format);
-    vasprintf(&message, format, args);
+    if (vasprintf(&message, format, args) == -1) {
```
stapelberg: I doubt running `DLOG` in this situation will work. Note that we’re already talking about a function that is an error handler. It might be better to just use `(void)` to mark the return code unused.

i3bar/src/child.c
```diff
@@ -457,8 +461,12 @@ void child_write_output(void) {
         size_t size;
 
         yajl_gen_get_buf(gen, &output, &size);
-        write(child_stdin, output, size);
-        write(child_stdin, "\n", 1);
+        if (write(child_stdin, output, size) == -1) {
+            DLOG("Failed to write child_stdin\n");
```
stapelberg: Use `ELOG` instead

hwangcc23:
Thanks for the review and comments. 
Updated the patch. 
Please review it again.

i3-nagbar/main.c
```diff
@@ -164,7 +164,9 @@ static void handle_button_release(xcb_connection_t *conn, xcb_button_release_eve
     char *link_path;
     char *exe_path = get_exe_path(argv0);
     sasprintf(&link_path, "%s.nagbar_cmd", script_path);
-    symlink(exe_path, link_path);
+    if (symlink(exe_path, link_path) == -1) {
+        err(EXIT_FAILURE, "Failed to symlink %s, %s\n", exe_path, link_path);
```
stapelberg: Remove the `\n`, `err()` already adds that for you.

Also, either make it “Failed to symlink %s to %s” or “Failed to symlink(%s, %s)”

src/sighandler.c
```diff
@@ -70,8 +70,14 @@ static int backtrace(void) {
         int stdin_pipe[2],
             stdout_pipe[2];
 
-        pipe(stdin_pipe);
-        pipe(stdout_pipe);
+        if (pipe(stdin_pipe) == -1) {
+            DLOG("Failed to init stdin_pipe\n");
```
stapelberg: Here and below, use ELOG

i3bar/src/child.c
```diff
@@ -457,8 +457,12 @@ void child_write_output(void) {
         size_t size;
 
         yajl_gen_get_buf(gen, &output, &size);
-        write(child_stdin, output, size);
-        write(child_stdin, "\n", 1);
+        if (write(child_stdin, output, size) == -1) {
```
stapelberg: I’m thinking if `write` fails with return code `-1`, we should set `child.click_events = false;`, `kill(child.pid)` and display an error in the same way that `child_sig_cb` does, so that the user knows what’s going on.

Also, while you’re at it and making our `write()` calls better: http://michael.stapelberg.de/Artikel/c_write_pattern/ documents how the code should look to be correct :).

hwangcc23:
Fixed.
Please review the patch again.
Thanks.

i3-nagbar/main.c
```diff
@@ -164,7 +164,9 @@ static void handle_button_release(xcb_connection_t *conn, xcb_button_release_eve
     char *link_path;
     char *exe_path = get_exe_path(argv0);
     sasprintf(&link_path, "%s.nagbar_cmd", script_path);
-    symlink(exe_path, link_path);
+    if (symlink(exe_path, link_path) == -1) {
+        err(EXIT_FAILURE, "Failed to symlink %s to %s", exe_path, link_path);
```
stapelberg: See the man page of symlink, it’s actually the other way around. `link_path` will be created, pointing to `exe_path`.

i3bar/src/child.c
```diff
@@ -455,11 +455,43 @@ void child_write_output(void) {
     if (child.click_events) {
         const unsigned char *output;
         size_t size;
+        int n, written, ret;
 
         yajl_gen_get_buf(gen, &output, &size);
-        write(child_stdin, output, size);
-        write(child_stdin, "\n", 1);
+
+        written = 0;
```
stapelberg: I think it might be best to put this into a function, possibly in libi3. I think we might also want to use this code in other places in i3, and then you can get rid of the `goto`.

hwangcc23:
Updated. Please review again. Thanks.

i3bar/src/ipc.c
```diff
@@ -296,17 +296,10 @@ int i3_send_msg(uint32_t type, const char *payload) {
...
-        int n = write(i3_connection->fd, buffer + written, to_write);
-        if (n == -1) {
-            ELOG("write() failed: %s\n", strerror(errno));
-            exit(EXIT_FAILURE);
-        }
-
-        to_write -= n;
-        written += n;
+    ssize_t n;
+    n = swrite(i3_connection->fd, buffer, to_write);
```
stapelberg: Since you’re not using `n`, why declare it at all? I suggest:

```c
if (swrite(i3_connection->fd, buffer, to_write) == -1)
    err(EXIT_FAILURE, "Cannot send IPC message to i3");
```

stapelberg: Actually, now that I’ve read the definition of `swrite`, it’s impossible for swrite to return -1. Remove this check entirely.

i3bar/src/child.c
```diff
@@ -454,12 +454,22 @@ void child_sig_cb(struct ev_loop *loop, ev_child *watcher, int revents) {
...
-        write(child_stdin, output, size);
-        write(child_stdin, "\n", 1);
+
+        n = swrite(child_stdin, output, size);
+        if (n != -1)
+            n = swrite(child_stdin, "\n", 1);
+
         yajl_gen_clear(gen);
+
+        if (n == -1) {
```
stapelberg: This code path is never hit because `swrite` exits when it gets a -1. I suggest you introduce a function `writeall` and make `swrite` wrap that function, so that we can use either `writeall` or `swrite`, depending on whether we want to exit on errors or not.

stapelberg:
Please also squash your commits (google for git interactive rebase if you don’t know how).

hwangcc23:
Fixed. And also squash commits.
Please review again.
Thanks.

src/util.c
```diff
@@ -265,25 +265,14 @@ char *store_restart_layout(void) {
...
-            DLOG("write == 0?\n");
-            free(filename);
-            close(fd);
-            return NULL;
-        }
-        written += n;
-        DLOG("written: %zd of %zd\n", written, length);
+    ssize_t n;
+    n = writeall(fd, payload, length);
+    if (n == -1) {
```
stapelberg: add:
```c
ELOG("Could not write restart layout to \"%s\", layout will be lost: %s\n", filename, strerror(errno));
```

libi3/safewrappers.c
```diff
@@ -56,3 +58,30 @@ int sasprintf(char **strp, const char *fmt, ...) {
...
+ssize_t writeall(int fd, const void *buf, size_t count) {
+    int written = 0;
+    ssize_t n = 0;
+
+    while (written < count) {
+        n = write(fd, buf + written, count - written);
+        if (n == -1) {
+            if (errno == EINTR || errno == EAGAIN)
+                continue;
+            break;
```
stapelberg: replace the `break` with `return n;`

libi3/safewrappers.c
```diff
@@ -56,3 +58,30 @@ int sasprintf(char **strp, const char *fmt, ...) {
...
+        }
+        written += n;
+    }
+
+    return n;
```
stapelberg: replace `return n` with `return written`, as `n` is just a temporary return code

libi3/ipc_send_message.c
```diff
@@ -31,34 +31,15 @@ int ipc_send_message(int sockfd, const uint32_t message_size,
         .magic = {'i', '3', '-', 'i', 'p', 'c'},
         .size = message_size,
         .type = message_type};
+    ssize_t n;
 
-    size_t sent_bytes = 0;
-    int n = 0;
+    n = writeall(sockfd, ((void *)&header), sizeof(i3_ipc_header_t));
```
stapelberg: Here and below, you don’t use `n`, so change it to:
```c
if (writeall(…) == -1)
    return -1;
```

include/libi3.h
```diff
@@ -135,6 +135,20 @@ char *sstrdup(const char *str);
 int sasprintf(char **strp, const char *fmt, ...);
 
 /**
+ * Wrapper around correct write which returns -1 (meaning that
+ * write failed) or count (meaning that all are written)
```
stapelberg: nit: s/all are written/all bytes were written/

hwangcc23:
Fixed the bug in writeall and revised code based on your comments.
Please review again.
Thanks.

PR #1578, opened 2015-03-24
Add support for a custom separator symbol

### LINKED ISSUE #1472: configure statusline separator
Right now, the statusline is drawn as a line with the X drawing api. It should be a character (such as "|") which the user can configure. Then people can use little circles or diamonds or something.

Airblader:
This introduces a "separator_symbol" property for the i3bar configuration.
If set, the specified string will be used as a separator instead of a vertical line. Since it is an optional configuration, complete backwards compatibility is given.

fixes #1472

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1249,6 +1249,24 @@ bar {
 }
 --------------------------------------------------------------
 
+=== Custom separator symbol
+
+Specifies a custom symbol to be used for the separator as opposed to the vertical,
+one pixel thick separator. Note that you may have to adjust the +sep_block_width+ 
+property.
+
+*Syntax*:
+-------------------------
+separator_symbol <symbol>
+-------------------------
+
+*Example*:
+------------------------
+bar {
+    separator_symbol ":|:"
+}
+------------------------
+
 === Workspace buttons
 
 Specifies whether workspace buttons should be shown or not. This is useful if
--- a/i3bar/src/config.c
+++ b/i3bar/src/config.c
@@ -144,6 +144,13 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
+    if (!strcmp(cur_key, "separator_symbol")) {
+        DLOG("separator = %.*s\n", len, val);
+        I3STRING_FREE(config.separator_symbol);
+        config.separator_symbol = i3string_from_utf8_with_length((const char *)val, len);
+        return 1;
+    }
+
     if (!strcmp(cur_key, "outputs")) {
         DLOG("+output %.*s\n", len, val);
         int new_num_outputs = config.num_outputs + 1;
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -63,6 +63,9 @@ static i3Font font;
 /* Overall height of the bar (based on font size) */
 int bar_height;
 
+/* Cached width of the custom separator if one was set */
+int separator_symbol_width;
+
 /* These are only relevant for XKB, which we only need for grabbing modifiers */
 int xkb_base;
 int mod_pressed = 0;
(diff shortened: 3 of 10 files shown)


i3bar/src/xcb.c
```diff
@@ -156,6 +156,37 @@ int get_tray_width(struct tc_head *trayclients) {
...
+        uint32_t mask = XCB_GC_FOREGROUND | XCB_GC_BACKGROUND | XCB_GC_LINE_WIDTH;
+        uint32_t values[] = {colors.sep_fg, colors.bar_bg, logical_px(1)};
+        xcb_change_gc(xcb_connection, statusline_ctx, mask, values);
+        xcb_poly_line(xcb_connection, XCB_COORD_MODE_ORIGIN, statusline_pm, statusline_ctx, 2,
+                      (xcb_point_t[]){{center_x, logical_px(sep_voff_px)},
+                                      {center_x, bar_height - logical_px(sep_voff_px)}});
+    } else {
+        /* We have the font now, so we can cache the width here. */
+        if (!config.separator_width)
+            config.separator_width = predict_text_width(config.separator_symbol);
```
Airblader: I can't write this value when reading the config because the font won't necessarily be loaded by then yet.

i3bar/src/config.c
```diff
@@ -144,6 +144,16 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
+    if (!strcmp(cur_key, "separator_symbol")) {
+        DLOG("separator = %.*s\n", len, val);
+        FREE(config.separator_symbol);
+        char *text;
+        sasprintf(&text, "%.*s", len, val);
+        config.separator_symbol = i3string_from_utf8(text);
```
Airblader: Opinions on whether this should allow markup? This would allow typesetting the separator bold or whatever, but would potentially require escaping (since whether or not markup should be used can't be configured for this). I tend to think that if you want to go *that* fancy, just add separate blocks for it. Colors can be controled via the separator color anyway.

docs/userguide
```diff
@@ -1249,6 +1249,25 @@ bar {
...
+*Syntax*:
+-------------------------
+separator_symbol <symbol>
+-------------------------
+
+*Example*:
+------------------------
+bar {
+    separator_symbol |
+    separator_symbol ":|:"
```
stapelberg: nit: why two example? I’d go for the more robust example, i.e. the quoted one. Listing two examples doesn’t semantically make sense, as the latter option will override the former.

i3bar/src/config.c
stapelberg: Agreed, I don’t think we need markup here.

stapelberg: I think you can use `i3string_from_utf8_with_length` instead of calling `sasprintf`.

stapelberg:
Looks good overall.

i3bar/src/xcb.c
acrisci: The separator does not change size when you change the bar font and `reload`.

Airblader: @acrisci Should be fixed now. Found a better place to cache the value.

i3bar/src/config.c
Airblader: Done!

docs/userguide
Airblader: Yes, you're right. I reduced it now.

i3bar/src/config.c
```diff
@@ -144,6 +144,13 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
+    if (!strcmp(cur_key, "separator_symbol")) {
+        DLOG("separator = %.*s\n", len, val);
+        FREE(config.separator_symbol);
```
stapelberg: You need to use `i3string_free()` instead of `FREE` here, otherwise you’re leaking memory.

Airblader: done

PR #1580, opened 2015-03-25
Allow nop command without argument

### LINKED ISSUE #1318: Nop command requires a string as parameter
[Originally reported by javiervalencia80@…]
(Don't know if intended or not, but it asks for a string.

Also, it's undocumented in the User Guide, and might be useful for some of us.

Airblader:
fixes #1318

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2011,6 +2011,27 @@ bindsym $mod+minus scratchpad show
 bindsym mod4+s [title="^Sup ::"] scratchpad show
 ------------------------------------------------
 
+=== Nop
+
+There is a no operation command +nop+ which allows you to override default
+behavior. This can be useful for, e.g., disabling a focus change on clicks with
+the middle mouse button.
+
+The optional +comment+ argument is ignored, but will be printed to the log file
+for debugging purposes.
+
+*Syntax*:
+---------------
+nop [<comment>]
+---------------
+
+*Example*:
+----------------------------------------------
+# Disable focus change for clicks on titlebars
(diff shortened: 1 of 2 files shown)


Airblader:
Do we want to add this to the docs as suggested in #1318?

acrisci:
Yeah, this is useful for overriding mouse clicks.

stapelberg:
I agree with @acrisci.

Airblader:
@acrisci Sorry, I'm not following. Would you mind giving me an example on how to use it? I think it'd be useful to add such an example to the docs.

acrisci:
@Airblader If I want to make middle click over titlebars not focus the window, I can do `bindsym button2 nop`.

Airblader:
Ah, okay. Thanks. I've updated the PR.

docs/userguide
```diff
@@ -2011,6 +2011,26 @@ bindsym $mod+minus scratchpad show
 bindsym mod4+s [title="^Sup ::"] scratchpad show
 ------------------------------------------------
 
+=== Nop
+
+There is a no operation command +nop+ which allows you to override default
+behavior. This can be useful for, e.g., disabling a focus change on clicks with
+the middle mouse button.
+
+The optional +command+ argument is ignored.
```
stapelberg: Here and below, s/command/comment/

Please also mention that the comment is mentioned prominently in the i3 logfile (link to /docs/debugging), since that is useful for debugging.

Airblader: Done!

PR #1588, opened 2015-03-26
Respect XDG config directories in i3-config-wizard

### LINKED ISSUE #1566: i3-config-wizard should be aware of ~/.config/i3
If you have your i3 configuration under ~/.config, i3 wizard says i3 is not yet configured. It should respect it.

Deiz:
Should fix #1566

I've explicitly excluded two of i3's config detection mechanisms, SYSCONFDIR (usually /etc) and /etc/xdg (Normally default when $XDG_CONFIG_DIRS isn't set) because they're system- rather than user-level, and the wizard is for setting up a user-editable config.

The text changes are necessary to (hopefully) prevent overflow with long $XDG_CONFIG_HOME paths. It'd have been considerably uglier to dynamically widen the window and pixmap dependent on the longest line of text.

### DIFF
--- a/i3-config-wizard/main.c
+++ b/i3-config-wizard/main.c
@@ -482,17 +482,23 @@ static int handle_expose() {
         set_font_colors(pixmap_gc, get_colorpixel("#FFFFFF"), get_colorpixel("#000000"));
 
         txt(logical_px(10), 2, "You have not configured i3 yet.");
-        txt(logical_px(10), 3, "Do you want me to generate ~/.i3/config?");
-        txt(logical_px(85), 5, "Yes, generate ~/.i3/config");
-        txt(logical_px(85), 7, "No, I will use the defaults");
+        txt(logical_px(10), 3, "Do you want me to generate a config at");
+
+        char *msg;
+        sasprintf(&msg, "%s?", config_path);
+        txt(logical_px(10), 4, msg);
+        free(msg);
+
+        txt(logical_px(85), 6, "Yes, generate the config");
+        txt(logical_px(85), 8, "No, I will use the defaults");
 
         /* green */
         set_font_colors(pixmap_gc, get_colorpixel("#00FF00"), get_colorpixel("#000000"));
-        txt(logical_px(25), 5, "<Enter>");
+        txt(logical_px(25), 6, "<Enter>");
 
         /* red */
         set_font_colors(pixmap_gc, get_colorpixel("#FF0000"), get_colorpixel("#000000"));
-        txt(logical_px(31), 7, "<ESC>");
+        txt(logical_px(31), 8, "<ESC>");
     }
 
     if (current_step == STEP_GENERATE) {
@@ -502,7 +508,7 @@ static int handle_expose() {
         txt(logical_px(85), 4, "Win as default modifier");
         txt(logical_px(85), 5, "Alt as default modifier");
         txt(logical_px(10), 7, "Afterwards, press");
-        txt(logical_px(85), 9, "to write ~/.i3/config");
+        txt(logical_px(85), 9, "to write the config");
         txt(logical_px(85), 10, "to abort");
 
         /* the not-selected modifier */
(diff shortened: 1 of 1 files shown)


i3-config-wizard/main.c
```diff
@@ -792,20 +805,48 @@ int main(int argc, char *argv[]) {
         }
     }
 
-    /* Check if the destination config file does not exist but the path is
-     * writable. If not, exit now, this program is not useful in that case. */
+#define EXIT_IF_EXISTS(path)                                               \
```
stapelberg: Please use the `do { } while (0)` idiom in macros.

i3-config-wizard/main.c
```diff
@@ -792,20 +805,48 @@ int main(int argc, char *argv[]) {
...
+        printf("The config file \"%s\" already exists. Exiting.\n", path); \
+        exit(0);                                                           \
+    }
+
     struct stat stbuf;
-    if (stat(config_path, &stbuf) == 0) {
-        printf("The config file \"%s\" already exists. Exiting.\n", config_path);
-        return 0;
+
+    /* 1: check the traditional path under the home directory */
```
stapelberg: Instead of duplicating this logic, can we move it to libi3 and re-use the same function (i.e. `get_config_path()` from src/config.c?)

stapelberg:
Can you rebase this on top of current `next` please?

i3-config-wizard/main.c
Deiz: I don't think so, since the wizard only checks a subset of the paths that i3 itself does. However, they could use identical logic if the wizard did additional checks, e.g. only exiting if the first found config is owned by the current user.

I'm not sure that's more elegant than duplication, but if you'd prefer it I can do it that way.

Airblader: Why would we not want the wizard to look for the same paths as i3? That sounds like the desired behavior to me.

Deiz: The wizard shouldn't check system-level paths because it's providing a user-level choice, and a config under /etc will almost always be present, whether the user has installed i3 via their package manager or make install. So, effectively, checking the /etc fallback path would prevent the wizard from running in almost all circumstances.

Airblader: Yes, you're right. My mistake :)

stapelberg: We could just add a flag to that function to make it not consider system-level paths. That way we’d still have the config file path related logic in just one place.

Deiz: That seems viable, though I'd like to fix up the $XDG_CONFIG_DIRS logic I've implemented.

Looking at the XDG base directory spec, it seems that $XDG_CONFIG_HOME will always trump anything in $XDG_CONFIG_DIRS, so perhaps $XDG_CONFIG_DIRS should always be treated as system-level, regardless of whether it's defined or the /etc/xdg fallback is used.

Deiz:
With #1600 merged, I've removed the duplicate config path logic and rebased onto next.

i3-config-wizard/main.c
```diff
@@ -774,20 +780,29 @@ int main(int argc, char *argv[]) {
...
+
+    xdg_config_home = resolve_tilde(xdg_config_home);
+    sasprintf(&config_path, "%s/i3/config", xdg_config_home);
+
+    /* Create $XDG_CONFIG_HOME/i3 if it does not yet exist */
+    char *config_dir;
+    struct stat stbuf;
+    sasprintf(&config_dir, "%s/i3", xdg_config_home);
     if (stat(config_dir, &stbuf) != 0)
         if (mkdir(config_dir, 0755) == -1)
```
stapelberg: This won’t work if `~/.config` doesn’t exist yet, which can happen when a user installs a system and tries to set up i3 first. You could pull the `mkdirp` function of src/ipc.c into libi3 (in a separate pull request please) to implement this.

Deiz:
Updated to use mkdirp.

PR #1589, opened 2015-03-26
Make center coordinates relative to current workspace

### LINKED ISSUE #1211: "move position center" often chooses the wrong monitor
[Originally reported by jeremy.john.reeder@…]
(The command "move position center" should move the window to the center of the workspace that it's already on. But in a dual-monitor setup with one screen above the other, this command always moves windows from the lower monitor to the center of the upper monitor.

Deiz:
Should fix #1211. The bug is actually describing the symptom, as by failing to factor in the focused workspace's base coordinates, the centering mostly ended up on the upper left monitor.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1761,8 +1761,8 @@ void cmd_move_window_to_center(I3_CMD, char *method) {
         Rect newrect = focused->parent->rect;
 
         DLOG("moving to center\n");
-        newrect.x = wsrect->width / 2 - newrect.width / 2;
-        newrect.y = wsrect->height / 2 - newrect.height / 2;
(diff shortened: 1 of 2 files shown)


stapelberg:
Can you add a testcase for this please?

Deiz:
I've made some further changes locally, centralizing all the centering logic into a floating_center function that takes a con and a rect to center within. To avoid needing to rebase it, I'll submit it once you've reviewed and merged this.

stapelberg:
The test case executes the code path, but it doesn’t check the positions, which is what you’re changing, right?

Deiz:
Yes and no. The test is for the bug itself, where centering results in the window ending up on the wrong output. I've got more thorough, position-specific checks in another test I'll be submitting in a separate pull request.

PR #1590, opened 2015-03-26
i3bar: don't parse markup on statusline by default

### LINKED ISSUE #1565: Don't parse bar blocks as markup by default
When I suggested to use markup, I didn't think there would be any breaking changes to statusline generators. Now I know that there are some characters that may cause the string to not display properly.

Pango markup must be escaped like XML. For instance, the characters `<` and `&` must be escaped as `&lt;` and `&amp;` respectively.

This will cause problems with any block that displays a string from an outside source. Mediaplayer, clipboard, and window title blocks must escape their content.

There have been several complaints of breakages so far.

We can still keep the markup with a block member such as is implemented in #1564.

Sorry for the breakage. I don't think we have enough people using both the dev version and a custom statusline generator with scripts of this kind to identify this as a problem before the release.

acrisci:
fixes #1565

### DIFF
--- a/docs/i3bar-protocol
+++ b/docs/i3bar-protocol
@@ -117,10 +117,8 @@ click_events::
 === Blocks in detail
 
 full_text::
-	The most simple block you can think of is one which just includes the
-	only required key, the +full_text+ key. i3bar will display the string
-	value parsed as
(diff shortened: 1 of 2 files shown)


Airblader:
The docs have to be changed since they currently state that pango is the default.

acrisci:
Updates the docs.

PR #1591, opened 2015-03-26
Return in child parsing as soon as the match was made.

Airblader:
I didn't change the return value for cases where the type can't be handled for the current key as it could break someone's config.


stapelberg:
Why are you making this change? Please explain that in the commit message.

Airblader:
Done. Better? :)

PR #1593, opened 2015-03-27
Display i3-nagbar on primary output

Airblader:
I mainly gave this a shot to do some XCB-related stuff. This "only" makes i3-nagbar appear on the primary screen (if set and active), not on all screens. I'd be open to try the latter as well, but I'd need some hints on how to make a window appear on multiple outputs (?)

I basically followed `src/randr.c` to see what has to be done. Seems like a lot of work, so I'm not sure if there isn't an easier way to do this.

### DIFF
--- a/i3-nagbar/i3-nagbar.mk
+++ b/i3-nagbar/i3-nagbar.mk
@@ -4,7 +4,7 @@ CLEAN_TARGETS += clean-i3-nagbar
 
 i3_nagbar_SOURCES := $(wildcard i3-nagbar/*.c)
 i3_nagbar_HEADERS := $(wildcard i3-nagbar/*.h)
-i3_nagbar_CFLAGS   = $(XCB_CFLAGS) $(XCB_WB_CFLAGS) $(PANGO_CFLAGS)
+i3_nagbar_CFLAGS   = $(XCB_CFLAGS) $(XCB_WM_CFLAGS) $(PANGO_CFLAGS)
 i3_nagbar_LIBS     = $(XCB_LIBS) $(XCB_WM_LIBS) $(PANGO_LIBS)
 
 i3_nagbar_OBJECTS := $(i3_nagbar_SOURCES:.c=.o)
--- a/i3-nagbar/i3-nagbar.mk
+++ b/i3-nagbar/i3-nagbar.mk
@@ -4,8 +4,8 @@ CLEAN_TARGETS += clean-i3-nagbar
 
 i3_nagbar_SOURCES := $(wildcard i3-nagbar/*.c)
 i3_nagbar_HEADERS := $(wildcard i3-nagbar/*.h)
-i3_nagbar_CFLAGS   = $(XCB_CFLAGS) $(PANGO_CFLAGS)
-i3_nagbar_LIBS     = $(XCB_LIBS) $(PANGO_LIBS)
+i3_nagbar_CFLAGS   = $(XCB_CFLAGS) $(XCB_WB_CFLAGS) $(PANGO_CFLAGS)
+i3_nagbar_LIBS     = $(XCB_LIBS) $(XCB_WM_LIBS) $(PANGO_LIBS)
 
 i3_nagbar_OBJECTS := $(i3_nagbar_SOURCES:.c=.o)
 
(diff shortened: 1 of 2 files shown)


i3-nagbar/main.c
```diff
@@ -286,6 +287,77 @@ static int handle_expose(xcb_connection_t *conn, xcb_expose_event_t *event) {
...
+    }
+
+    int len = xcb_randr_get_screen_resources_current_outputs_length(res);
+    randr_outputs = xcb_randr_get_screen_resources_current_outputs(res);
+    for (int i = 0; i < len; i++) {
+        if (randr_outputs[i] != primary->output)
+            continue;
+
+        xcb_randr_get_output_info_cookie_t ocookie;
+        ocookie = xcb_randr_get_output_info(conn, randr_outputs[i], res->config_timestamp);
```
stapelberg: Instead of looping through all outputs, you could just use `xcb_randr_get_output_info(conn, primary->output, res->config_timestamp);`, right?

stapelberg:
You cannot display a single window on multiple outputs. You’d need to make i3-nagbar open multiple windows.

i3-nagbar/main.c
Airblader: Yes, you're right, of course. I've updated it.

Airblader:
If showing it on all outputs is the behavior we want to go for I'd still consider this an improvement and would suggest opening a separate ticket for the "sticky" solution.

i3-nagbar/main.c
```diff
@@ -286,6 +287,69 @@ static int handle_expose(xcb_connection_t *conn, xcb_expose_event_t *event) {
...
+    if ((primary = xcb_randr_get_output_primary_reply(conn, pcookie, NULL)) == NULL) {
+        DLOG("Could not determine the primary output.\n");
+        goto free_resources;
+    }
+
+    if ((res = xcb_randr_get_screen_resources_current_reply(conn, rcookie, NULL)) == NULL) {
+        goto free_resources;
+    }
+
+    xcb_randr_get_output_info_cookie_t ocookie;
```
stapelberg: Here and below, the calls are synchronous, so you might as well use this format to save variables and lines:
```c
xcb_randr_get_output_info_reply_t *output =
    xcb_randr_get_output_info_reply(conn, xcb_randr_get_output_info(
        conn, primary->output, res->config_timestamp), NULL));
```

i3-nagbar/main.c
```diff
@@ -286,6 +287,69 @@ static int handle_expose(xcb_connection_t *conn, xcb_expose_event_t *event) {
     return 1;
 }
 
+/**
+ * Return the position and size the i3-nagbar window should use.
+ * This will be the primary output or a fallback if it cannot be determined.
+ */
+static xcb_rectangle_t get_window_position() {
```
stapelberg: s/()/(void)/

i3-nagbar/i3-nagbar.mk
```diff
@@ -4,8 +4,8 @@ CLEAN_TARGETS += clean-i3-nagbar
 
 i3_nagbar_SOURCES := $(wildcard i3-nagbar/*.c)
 i3_nagbar_HEADERS := $(wildcard i3-nagbar/*.h)
-i3_nagbar_CFLAGS   = $(XCB_CFLAGS) $(PANGO_CFLAGS)
-i3_nagbar_LIBS     = $(XCB_LIBS) $(PANGO_LIBS)
+i3_nagbar_CFLAGS   = $(XCB_CFLAGS) $(XCB_WB_CFLAGS) $(PANGO_CFLAGS)
```
stapelberg: s/WB/WM/, I suppose? :)

stapelberg:
Looks good overall.

i3-nagbar/i3-nagbar.mk
Airblader: Yes :) Although this raises the question whether I even need this (?)

i3-nagbar/main.c
Airblader: Done. I don't have a multi monitor setup at home so I couldn't verify it still works (even though I of course only did refactoring)

Airblader:
@davidbaumann I'll try to make a quick video showing that it works. But feel free to just verify it yourself.

Airblader:
@davidbaumann See here: http://media.giphy.com/media/xTiTnncVxAHY0EBRyo/giphy.gif

PR #1597, opened 2015-03-28
Move urgency hint when moving container to another workspace

Airblader:


Airblader:
Good catch. I'll fix it. I had tested it manually anyway (with different commands as well), so it hopefully really should just be a fix of the test.

Airblader:
The test has been fixed. It broke without the fix for me now and runs through with the fix.

PR #1600, opened 2015-03-28
Move resolve_tilde and get_config_path into libi3

Deiz:
This is groundwork for a reworked version of #1588.

I pulled the "Unable to find the configuration file" die() call out of get_config_path because it seems bad to have that function return NULL on error in one instance, yet call die() in another.

### DIFF
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -434,3 +434,21 @@ char *get_exe_path(const char *argv0);
  *
  */
 int logical_px(const int logical);
+
+/**
+ * This function resolves ~ in pathnames.
+ * It may resolve wildcards in the first part of the path, but if no match
+ * or multiple matches are found, it just returns a copy of path as given.
+ *
+ */
+char *resolve_tilde(const char *path);
+
+/**
+ * Get the path of the first configuration file found. If override_configpath
+ * is specified, that path is returned and saved for further calls. Otherwise,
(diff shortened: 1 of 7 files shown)


libi3/resolve_tilde.c
```diff
@@ -0,0 +1,45 @@
...
+    tail = strchr(path, '/');
+    head = strndup(path, tail ? (size_t)(tail - path) : strlen(path));
+
+    int res = glob(head, GLOB_TILDE, NULL, &globbuf);
+    free(head);
+    /* no match, or many wildcard matches are bad */
+    if (res == GLOB_NOMATCH || globbuf.gl_pathc != 1)
+        result = sstrdup(path);
+    else if (res != 0) {
+        err(1, "glob() failed");
```
stapelberg: s/1/EXIT_FAILURE/

include/libi3.h
```diff
@@ -420,3 +420,21 @@ char *get_exe_path(const char *argv0);
  *
  */
 int logical_px(const int logical);
+
+/*
```
stapelberg: here and below, use `/**` in headers to make doxygen realize this is documentation.

stapelberg:
Looks good overall.

Deiz:
Fixed and rebased on next.

PR #1599, opened 2015-03-28
Added a --toggle switch to the mark command

Airblader:
Please don't merge yet, I want to add a bunch of testcases for different scenarios. I've tried them manually, but I'd sleep better having tests to rely on.

PR #1601, opened 2015-03-28
Show marks in the titlebar

### LINKED ISSUE #1464: Show marks on the titlebar
The titlebar should show the marks that a window has on the right side of the window. This would make marks much easier to work with.

Airblader:
fixes #1464

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1006,6 +1006,24 @@ focus::
 none::
     The window will neither be focused, nor be marked urgent.
 
+=== Drawing marks on window decoration
+
+If activated, marks on windows are drawn in their window decoration. However,
+any mark starting with an underscore in its name (+_+) will not be drawn even if
+this option is activated.
+
+The default for this option is +no+.
+
+*Syntax*:
+-------------------
+show_marks [yes|no]
+-------------------
+
+*Example*:
+--------------
+show_marks yes
+--------------
+
 == Configuring i3bar
 
 The bar at the bottom of your monitor is drawn by a separate process called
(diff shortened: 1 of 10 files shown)


Airblader:
This will conflict with #1599. Please merge #1599 first (unless, of course, you don't want to merge it at all). I'll rebase this and refactor the mark commands into proper functions.

stapelberg:
Can you please also provide a screenshot of how this looks?

I’m a bit wary of the hard-coded `[%s]` (and no, I wouldn’t want it to be configurable, I think). What about just printing the mark as-is, without the parens?

What happens when a window title is longer than the window? Will it be ellipsized at the correct place, will the mark still be shown?

Airblader:
I can remove the `[]` if you want, but I just felt like it makes it a bit more obvious that it's a mark. 

If the window title is too long, it will be elliptisized and the mark is still shown. 


I'll provide screenshots of both scenarios tomorrow.

Airblader:
@stapelberg Here you go:
* Normal mark: http://i.imgur.com/cDDrax5.png
* Mark if window title is too long: http://i.imgur.com/RLBPXn1.png

One thing I noticed is that tabbed layout displays the mark correctly, but in stacked layout it's not visible. I'll look into this.

Airblader:
@stapelberg Never mind. I must've forgotten the `show_marks` directive when I tested stacked. It works fine. This leaves us with the question of whether we want to keep the `[]`. I vote in favor of it because I think it emphasizes its nature of a mark. Even more so if the window title has to be shortened.

stapelberg:
Thanks for the screenshots. I’m still a bit unsure about the square brackets. I suppose we can merge it and wait for complaints :).

I’m also wondering if we should make this option default to “yes”, given that most people probably don’t use marks.

Airblader:
@stapelberg I also vote for making it the default. I don't see why you wouldn't want to see marks if you do use them and even if you don't, it's an easy fix. On the other hand, people would first have to learn about being able to display them.

stapelberg:
Let’s make it the default then. Can you make the change and rebase?

Airblader:
@stapelberg I've changed the default and rebased. As for my aforementioned refactoring, I will do this in a separate PR because I'll open a bug anyway.

Airblader:
Rebased again.

PR #1604, opened 2015-03-29
Cope with non-null-terminated x class properties.

mh21:


src/window.c
```diff
@@ -26,7 +26,11 @@ void window_update_class(i3Window *win, xcb_get_property_reply_t *prop, bool bef
     /* We cannot use asprintf here since this property contains two
      * null-terminated strings (for compatibility reasons). Instead, we
      * use strdup() on both strings */
-    char *new_class = xcb_get_property_value(prop);
+    void *prop_value = xcb_get_property_value(prop);
+    size_t prop_length = xcb_get_property_value_length(prop);
+
+    char *new_class = scalloc(prop_length + 1);
+    memcpy(new_class, prop_value, prop_length);
```
stapelberg: Use `memcpy(new_class, xcb_get_property_value(prop), prop_length)` instead to get rid of one variable.

src/window.c
```diff
@@ -26,7 +26,11 @@ void window_update_class(i3Window *win, xcb_get_property_reply_t *prop, bool bef
     /* We cannot use asprintf here since this property contains two
      * null-terminated strings (for compatibility reasons). Instead, we
      * use strdup() on both strings */
-    char *new_class = xcb_get_property_value(prop);
+    void *prop_value = xcb_get_property_value(prop);
+    size_t prop_length = xcb_get_property_value_length(prop);
```
stapelberg: Use `const size_t` so that the compiler knows this cannot change.

stapelberg:
Please add “fixes #1605” to the commit message and rebase this onto current next.

mh21:
Updated to include 
- const size_t
- smalloc + null at the end instead of scalloc
- reuse of prop_length in class check
- fixes line in commit message

PR #1608, opened 2015-03-30
Update old_name when renaming current workspace

### LINKED ISSUE #1607: 4.10.1 crashes upon renaming a workspace
i3 crashes upon renaming any workspace. This issue was not present in 4.9. Box popped up asking if I wanted to save a backtrace and/or restart in place. Rebuilt i3 with debug symbols enabled and attached said backtrace. Reproducible 100% for me. The log did not appear to contain anything about the crash, as i3 reloads upon crashing, apparently wiping the log. I'm open to suggestions about doing it better.

Log is here: http://logs.i3wm.org/logs/5681034041491456.bz2
Backtrace is here: http://pastebin.com/3Jn98b71

mh21:
should fix #1607

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1851,6 +1851,7 @@ void cmd_rename_workspace(I3_CMD, char *old_name, char *new_name) {
                    !strcasecmp(child->name, old_name));
     } else {
         workspace = con_get_workspace(focused);
+        old_name = workspace->name;
     }
 
     if (!workspace) {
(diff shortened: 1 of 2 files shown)


src/commands.c
```diff
@@ -1851,6 +1851,7 @@ void cmd_rename_workspace(I3_CMD, char *old_name, char *new_name) {
                    !strcasecmp(child->name, old_name));
     } else {
         workspace = con_get_workspace(focused);
```
mh21: Can workspace be NULL here?

stapelberg: No.

PR #1609, opened 2015-03-30
Handle floating centering in one function and test for consistency

Deiz:


src/floating.c
```diff
@@ -420,6 +417,11 @@ bool floating_maybe_reassign_ws(Con *con) {
     return true;
 }
 
+void floating_center(Con *con, Rect rect) {
```
stapelberg: nit: copy the comment from the header file (with just a single asterisk).

stapelberg:
Looks good overall.

Deiz:
Added the comment.

PR #1613, opened 2015-03-30
Added focus_on_window_activation directive

Airblader:
Before this gets merged, I want to add testcases for it. I just submitted the PR already because I'm not entirely sure about the naming. Opinions?

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -981,6 +981,31 @@ force_display_urgency_hint <timeout> ms
 force_display_urgency_hint 500 ms
 ---------------------------------
 
+=== Focus on window activation
+
+If a window is activated, e.g., via +google-chrome www.google.com+, it may request
+to take focus. Since this may not preferable, different reactions can be configured.
+
+Note that this does not apply to any window that is opened, i.e., typically opening
+an application will focus that window independent of this configuration.
+
+*Syntax*:
+----------------------------------------------------
+focus_on_window_activation <smart|urgent|focus|none>
+----------------------------------------------------
+
+The different modes will act as follows:
+
+smart::
+    This is the default behavior. If the window requesting focus is on an active
+    workspace, it will receive the focus. Otherwise, the urgency hint will be set.
+urgent::
+    The window will always be marked urgent, but the focus will not be stolen.
+focus::
+    The window will always be focused and not be marked urgent.
+none::
+    The window will neither be focused, nor be marked urgent.
+
 == Configuring i3bar
 
 The bar at the bottom of your monitor is drawn by a separate process called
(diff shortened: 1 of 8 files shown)


Airblader:
@stapelberg To make sure, we don't want to change the behavior for such requests coming from a pager, right?

docs/userguide
```diff
@@ -981,6 +981,31 @@ force_display_urgency_hint <timeout> ms
...
+
+If a window is activated, e.g., via +google-chrome www.google.com+, it may request
+to take focus. Since this may not preferable, different reactions can be configured.
+
+Note that this does not apply to any window that is opened, i.e., typically opening
+an application will focus that window independent of this configuration.
+
+*Syntax*:
+--------------------------------------------------
+focus_on_window_activation smart|urgent|focus|none
```
Deiz: Small nitpick, the parameter lists (here and in config.spec) should probably be enclosed in &lt;&gt;, e.g. &lt;smart|urgent|focus|none&gt;

acrisci:
This looks kind of ugly to me.

Airblader:
@acrisci I'll be happy to change it if you can be a bite more specific.

docs/userguide
Airblader: Yep, thanks. I've changed it.

include/config.h
```diff
@@ -167,6 +167,18 @@ struct Config {
      * flag can be delayed using an urgency timer. */
     float workspace_urgency_timer;
 
+    /** Behavior when a window sends a NET_ACTIVE_WINDOW message. */
+    enum {
+        /* Focus if the target workspace is visible, set urgency hint otherwise. */
+        FOWN_SMART,
```
stapelberg: These should be FOWA, not FOWN, right?

docs/userguide
```diff
@@ -981,6 +981,31 @@ force_display_urgency_hint <timeout> ms
 force_display_urgency_hint 500 ms
 ---------------------------------
 
+=== Behavior when a window is activated
```
stapelberg: I’d rename this to “Focus on window activation” to make it clearer what it’s talking about and consistent with how the option itself is called.

stapelberg:
Looks good overall.

Airblader:
@stapelberg I've added testcases, incorporated your comments and rebased it.

testcases/t/240-focus-on-window-activation.t
```diff
@@ -0,0 +1,206 @@
...
+}
+
+sub get_urgency_for_window_on_workspace {
+    my ($ws, $con) = @_;
+
+    my $current = first { $_->{window} == $con->{id} } @{get_ws_content($ws)};
+    return $current->{urgent};
+}
+
+#####################################################################
```
Airblader: I'm not testing "smart" here since it's the default and extensively tested in another test. I had to create a new test class for these since they need to run with their own config.

stapelberg:
Nice!

PR #1618, opened 2015-03-31
Allow single-child non-default layout cons to be moved between outputs

### LINKED ISSUE #1603: `move [direction]` does not move across outputs with default workspace layout
The `move [direction]` command does not move containers across outputs when a default workspace layout is given.

To reproduce, use this config directive:

    workspace_layout stacked

then open a window on a left monitor and give the command `move right`.

Deiz:
Includes a test case to verify the behaviour of 'move &lt;direction&gt;'.
Fixes #1603.

There's one caveat here: The non-default layout con surrounding the con that's being moved is destroyed. For example, if you have a single-child stacked container and move it onto a workspace with a tabbed layout, it will be enclosed by a tabbed container, or no container if the destination workspace has a default layout.

It'd be possible to avoid that by moving the container's parent instead of the container itself, but I prefer my current approach because most people hitting this code path are likely to be working with implicitly-created containers.

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -249,6 +249,14 @@ void tree_move(Con *con, int direction) {
                         ? AFTER
                         : BEFORE);
         insert_con_into(con, target, position);
+    } else if (con->parent->parent->type == CT_WORKSPACE &&
+               con->parent->layout != L_DEFAULT &&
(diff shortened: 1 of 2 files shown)


acrisci:
Works for me.

Issue #1614, opened 2015-03-30
Marks can be put onto multiple windows


Airblader:
How to reproduce:

* Open two instances of urxvt (or any other application)
* Execute `i3-msg '[class="(?i)urxvt"] mark uhoh'`

The reason is that `cmd_mark` loops over all matched windows. We should probably log an error and only work on the first match.

i3 version 4.10.1-33-g537fec1 (2015-03-30, branch "feature-titlebar-marks") © 2009-2014 Michael Stapelberg and contributors

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

PR #1620, opened 2015-03-31
Only allow unique marks (#1614)

Airblader:


Airblader:
By the way, is there a specific reason we don't allow marks to be used on multiple windows? Sure, `[con_mark="foo"] focus` becomes a weird thing to do then, but that's up to the user. I could also do something like `[con_mark="foo"] move container to …` or any other command.

testcases/lib/i3test.pm
```diff
@@ -614,7 +614,7 @@ Sends the specified command to i3.
 
 =cut
 sub cmd {
-    i3(get_socket_path())->command(@_)->recv
+    return i3(get_socket_path())->command(@_)->recv
```
stapelberg: This is unnecessary. The default return value is the last expression in Perl.

stapelberg:
> By the way, is there a specific reason we don't allow marks to be used on multiple windows? Sure, [con_mark="foo"] focus becomes a weird thing to do then, but that's up to the user. I could also do something like [con_mark="foo"] move container to … or any other command.

Since marks were envisioned to be used for specifically such commands (that operate on a single window), it was a good thing to avoid weird use-cases :).

testcases/lib/i3test.pm
Airblader: Unnecessary, but explicit. But okay, I'll remove it. I'll leave the description change of it, though. :)

Airblader:
@stapelberg Alright. If we want to stick to the uniqueness then, I've incorporated your comment and rebased it.

PR #1621, opened 2015-03-31
Added no_focus directive

Airblader:


Airblader:
Forgot the userguide. Doing it now.

Airblader:
@stapelberg Thinking about this I wonder whether we actually want to use the same config (and possibilities) here as for `_NET_ACTIVE_WINDOW`?

Airblader:
@stapelberg This makes me think whether (from a user's perspective) this is all too different from `_NET_ACTIVE_WINDOW`. Is it worth thinking about somehow combining these two things into one directive that allows specifying the behavior and criteria?

stapelberg:
They’re related for sure, but I don’t think they are so close to each other that we should combine them.

Instead, make the corresponding sections in the documentation refer to each other.

Airblader:
@stapelberg Done!

docs/userguide
```diff
@@ -589,6 +589,26 @@ for_window [title="x200: ~/work"] floating enable
 
 The valid criteria are the same as those for commands, see <<command_criteria>>.
 
+=== Don't focus window upon opening
+
+[[no_focus]]
+
+Generally, if an application is opened, its window will take the focus. The +no_focus+
```
stapelberg: Replace this line with:

When a new window appears, it will be focused. The +no_focus+…

docs/userguide
```diff
@@ -589,6 +589,26 @@ for_window [title="x200: ~/work"] floating enable
...
+directive allows preventing this from happening and can be used in combination with
+<<command_criteria>>.
+
+Note that this does not apply to all cases. Also refer to <<focus_on_window_activation>>.
```
stapelberg: Can you list the cases where this doesn’t apply?

Airblader:
@stapelberg  Done and rebased

src/manage.c
```diff
@@ -512,8 +512,11 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     /* Defer setting focus after the 'new' event has been sent to ensure the
      * proper window event sequence. */
     if (set_focus && !nc->window->doesnt_accept_focus && nc->mapped) {
-        DLOG("Now setting focus.\n");
-        con_focus(nc);
+        Assignment *assignment = assignment_for(cwindow, A_NO_FOCUS);
+        if (assignment == NULL) {
```
stapelberg: Use `if (assignment_for(cwindow, A_NO_FOCUS) == NULL) {` to avoid the unused variable.

docs/userguide
```diff
@@ -983,12 +1004,16 @@ force_display_urgency_hint 500 ms
...
 
+[[focus_on_window_activation]]
+
 If a window is activated, e.g., via +google-chrome www.google.com+, it may request
 to take focus. Since this may not preferable, different reactions can be configured.
 
 Note that this does not apply to any window that is opened, i.e., typically opening
 an application will focus that window independent of this configuration.
 
+A similar concept exists for opening new windows. See <<no_focus>>.
```
stapelberg: Replace this with:

To prevent new windows from being focused, see <<no_focus>>.

Airblader: I've actually rephrased the paragraph before a bit as well since it was misleading. Also incoroprated your other comment.

PR #1622, opened 2015-03-31
Make resizing on border click configurable

Deiz:
This is quite minor, but most of my border clicks are accidental (when I want to resize, I tend to use the keyboard or the modifier key) and I figured the behaviour may as well be configurable.

### DIFF
--- a/parser-specs/config.spec
+++ b/parser-specs/config.spec
@@ -290,6 +290,8 @@ state FONT:
 state BINDING:
   release = '--release'
       ->
+  border = '--border'
+      ->
   whole_window = '--whole-window'
       ->
   modifiers = 'Mod1', 'Mod2', 'Mod3', 'Mod4', 'Mod5', 'Shift', 'Control', 'Ctrl', 'Mode_switch', '$mod'
@@ -302,10 +304,12 @@ state BINDING:
 state BINDCOMMAND:
   release = '--release'
       ->
+  border = '--border'
+      ->
   whole_window = '--whole-window'
       ->
   command = string
-      -> call cfg_binding($bindtype, $modifiers, $key, $release, $whole_window, $command)
+      -> call cfg_binding($bindtype, $modifiers, $key, $release, $border, $whole_window, $command)
 
 ################################################################################
 # Mode configuration
@@ -339,6 +343,8 @@ state MODE_IGNORE_LINE:
 state MODE_BINDING:
   release = '--release'
       ->
+  border = '--border'
+      ->
   modifiers = 'Mod1', 'Mod2', 'Mod3', 'Mod4', 'Mod5', 'Shift', 'Control', 'Ctrl', 'Mode_switch', '$mod'
       ->
   '+'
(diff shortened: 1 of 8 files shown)


acrisci:
What's the use case for this?

acrisci:
Couldn't you just set the border to none?

Deiz:
In some cases yes, but not always. I like to keep a 1px border on my terminals, and mostly run into unwanted resizes when quickly selecting a block of text at one of the edges.

Whether anybody else is interested in disabling the behaviour, I'm unsure.

acrisci:
I'm going to oppose this change because it's a mouse-centric config directive that has a pretty good workaround.

stapelberg:
Yeah, a new config directive is a bit much for this. If we could make this behavior achievable using mouse bindings, I’d possibly be more open to merging it.

Deiz:
Ah, I'd completely forgotten about mouse bindings. How about adding a `--border` flag to mouse bindings? That would allow e.g. `bindsym --border button3 nop` to do precisely what I want.

Deiz:
I noticed that some floating window actions will accept any mouse button, such as clicks on decorations (anything except for right triggers drag) and borders (all buttons trigger resize).

Presumably those should be restricted slightly to mirror their tiling counterparts (left on decorations triggers move, left/right on borders starts resize). Any opposition to that?

acrisci:
@Deiz I had some trouble understanding that. Can you rephrase that question?

Deiz:
Tiling windows only trigger resize on decoration and border left/right click, whereas floating windows allow any button to trigger resize. I'm proposing that floating windows ignore non-left/right clicks, like tiling windows already do.

acrisci:
Yeah, I agree that should be made consistent.

stapelberg:
No, we want to preserve the current behavior. I expect floating windows to be resizable and draggable with the left mouse button, and plenty of notebook users with unfortunately bad touchpads will be thankful.

Deiz:
I'm only proposing to stop middle mouse, etc. as well as back/forward buttons from triggering floating resize.

Floating move is already restricted to the left mouse button, and I wouldn't be doing anything to it. Floating resize would continue to be triggered by left or right click on borders, or right click on decorations.

stapelberg:
Ah. Filtering out any mouse buttons but left/right sounds reasonable.

Deiz:
Alright, since Github doesn't support changing the target branch of a pull request, I've clobbered this PR's branch with an implementation of `--border`. It's functional, tests pass, etc., but it still needs to be documented in the user guide.

If adding `--border` is desirable, I'll update the userguide and resubmit this as a new pull request.

stapelberg:
@acrisci what do you think? Personally I’m fine with having --border in the mouse bindings.

acrisci:
Yeah, that seems to be a much cleaner approach to me. I'm ok with that.

parser-specs/config.spec
```diff
@@ -339,6 +343,8 @@ state MODE_IGNORE_LINE:
 state MODE_BINDING:
   release = '--release'
       ->
+  border = '--border'
```
Airblader: This also needs to go to the state MODE_BINDCOMMAND

Deiz: Yeah, I corrected that locally after noticing the changes you made in #1631.

Deiz:
Closing this as it's superseded by #1632.

PR #1624, opened 2015-04-01
Position i3-input at window with input focus

### LINKED ISSUE #1058: i3-input malfunctions if primary monitor is not at 0x0
[Originally reported by ryzion@…]
(Using other positions than 0x0 for the primary monitor induces i3-input to malfunction. i3-input will still accept inputs but the input bar is not visible.

Tested command: bindsym $mod+n exec i3-input -F 'rename workspace to "%s"' -P 'New name: '

Airblader:
This positions the i3-input window at the window holding the input focus' position, plus a small offset.

fixes #1058

### DIFF
--- a/i3-input/main.c
+++ b/i3-input/main.c
@@ -314,6 +314,50 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
     return 1;
 }
 
+static xcb_rectangle_t get_window_position(void) {
+    xcb_rectangle_t result = (xcb_rectangle_t){logical_px(50), logical_px(50), logical_px(500), font.height + logical_px(8)};
+
+    xcb_get_input_focus_reply_t *input_focus = NULL;
+    xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    /* In rare cases, the window holding the input focus might disappear while we are figuring out its
+     * position. To avoid this, we grab the server in the meantime. */
+    xcb_grab_server(conn);
+
+    input_focus = xcb_get_input_focus_reply(conn, xcb_get_input_focus(conn), NULL);
+    if (input_focus == NULL || input_focus->focus == XCB_NONE) {
+        DLOG("Failed to receive the current input focus or no window has the input focus right now.\n");
+        goto free_resources;
+    }
+
+    geometry = xcb_get_geometry_reply(conn, xcb_get_geometry(conn, input_focus->focus), NULL);
+    if (geometry == NULL) {
+        DLOG("Failed to received window geometry.\n");
+        goto free_resources;
+    }
+
+    coordinates = xcb_translate_coordinates_reply(
+        conn, xcb_translate_coordinates(conn, input_focus->focus, root, geometry->x, geometry->y), NULL);
+    if (coordinates == NULL) {
+        DLOG("Failed to translate coordinates.\n");
+        goto free_resources;
+    }
+
+    DLOG("Determined coordinates of window with input focus at x = %i / y = %i", coordinates->dst_x, coordinates->dst_y);
+    result.x += coordinates->dst_x;
+    result.y += coordinates->dst_y;
+
+free_resources:
+    xcb_ungrab_server(conn);
+    xcb_flush(conn);
+
+    FREE(input_focus);
+    FREE(geometry);
+    FREE(coordinates);
+    return result;
+}
+
 int main(int argc, char *argv[]) {
     format = strdup("%s");
     socket_path = getenv("I3SOCK");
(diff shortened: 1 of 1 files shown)


i3-input/main.c
```diff
@@ -314,6 +314,45 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
...
+    xcb_get_input_focus_reply_t *input_focus = NULL;
+    xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    input_focus = xcb_get_input_focus_reply(conn, xcb_get_input_focus(conn), NULL);
+    if (input_focus == NULL) {
+        DLOG("Failed to receive the current input focus.\n");
+        goto free_resources;
+    }
+
```
stapelberg: Please also check whether `input_focus->focus != XCB_NONE` :)

i3-input/main.c
```diff
@@ -314,6 +314,45 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
...
+    geometry = xcb_get_geometry_reply(conn, xcb_get_geometry(conn, input_focus->focus), NULL);
+    if (geometry == NULL) {
+        DLOG("Failed to received window geometry.\n");
+        goto free_resources;
+    }
+
+    coordinates = xcb_translate_coordinates_reply(conn, xcb_translate_coordinates(conn,
```
stapelberg: I think adding a line break after the opening paren (i.e. …reply(\nconn…) will lead to clang-format making this more readable.

i3-input/main.c
```diff
@@ -314,6 +314,45 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
...
 }
 
+static xcb_rectangle_t get_window_position(void) {
+    xcb_rectangle_t result = (xcb_rectangle_t){logical_px(50), logical_px(50), logical_px(500), font.height + logical_px(8)};
+
+    xcb_get_input_focus_reply_t *input_focus = NULL;
+    xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    input_focus = xcb_get_input_focus_reply(conn, xcb_get_input_focus(conn), NULL);
```
stapelberg: I think to make this code race-condition free you also need to grab the X11 server before getting the input focus (and pay attention to ungrabbing it!).

i3-input/main.c
```diff
@@ -402,15 +441,17 @@ int main(int argc, char *argv[]) {
     if (prompt != NULL)
         prompt_offset = predict_text_width(prompt);
 
+    xcb_rectangle_t win_pos = get_window_position();
```
stapelberg: nit: make this const

i3-input/main.c
Airblader: The race condition being that the input focus changes while this is going on? How long should the server be grabbed then? If I only grab it for this function, the same could pretty much still happen, no? But grabbing it for the entire lifecycle of i3-input seems excessive.

stapelberg: The race condition is that the window on which the input focus is disappears. This means your get_input_focus call will succeed, but get_geometry and/or translate_coordinates will fail.

Airblader: Ah, yes, that makes more sense. I'll add the grab.

Airblader:
@stapelberg Done.

i3-input/main.c
```diff
@@ -314,6 +314,49 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
...
+        DLOG("Failed to translate coordinates.\n");
+        goto free_resources;
+    }
+
+    DLOG("Determined coordinates of window with input focus at x = %i / y = %i", coordinates->dst_x, coordinates->dst_y);
+    result.x += coordinates->dst_x;
+    result.y += coordinates->dst_y;
+
+free_resources:
+    xcb_ungrab_server(conn);
```
stapelberg: Please add an `xcb_flush(conn);` right after the `xcb_ungrab_server(conn);` so that the server is grabbed as briefly as possible.

i3-input/main.c
```diff
@@ -314,6 +314,49 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
...
+    xcb_get_input_focus_reply_t *input_focus = NULL;
+    xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    /* In rare cases, the window holding the input focus might disappear while we are figuring out its
+     * position. To avoid this, we grab the server in the meantime. */
+    xcb_grab_server(conn);
+
+    input_focus = xcb_get_input_focus_reply(conn, xcb_get_input_focus(conn), NULL);
+    if (input_focus == NULL || input_focus == XCB_NONE) {
```
stapelberg: s/input_focus/input_focus->focus/ in the second condition.

Airblader: Done (and the other one)

PR #1626, opened 2015-04-01
Remove unused `tests` dir

acrisci:


acrisci:
This directory seems to be unused. Is it being used for anything?

stapelberg:
It used to be for experimental code. I think I have a small number of uncommitted tests as well.

These days we have enough test coverage that the files are not adding any value, I think.

PR #1628, opened 2015-04-01
complete-run: add coverage report generation

acrisci:
When `complete-run.pl` is given `--coverage-testing`, try to generate an
html coverage testing report for the run. This requires i3 to be
compiled with coverage testing support.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1,6 +1,5 @@
 UNAME=$(shell uname)
 DEBUG=1
-COVERAGE=0
 INSTALL=install
 LN=ln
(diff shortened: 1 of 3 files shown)


Airblader:
> This requires i3 to be compiled with coverage testing support.

And how do I do that? Could this maybe be documented?

acrisci:
Do `CFLAGS=-coverage LDFLAGS=-coverage make`.

There is a make variable in one of the files, but it's hardcoded in there right now.

That should be changed and coverage should be documented.

PR #1630, opened 2015-04-01
Make floating window mouse handling consistent with tiled windows

Deiz:
It seems that this was the intended behaviour all along, according to
the comments.

Per the discussion on #1622, floating window border clicks only trigger resize on left/right click, and decoration clicks only start moving the window if they're left clicks.

### DIFF
--- a/src/click.c
+++ b/src/click.c
@@ -180,6 +180,9 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
     if (con->parent->type == CT_DOCKAREA)
         goto done;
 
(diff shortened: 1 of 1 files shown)


src/click.c
```diff
@@ -166,6 +166,9 @@ static bool tiling_resize(Con *con, xcb_button_press_event_t *event, const click
     return false;
 }
 
+#define is_left_or_right_click(event)         \
```
stapelberg: Instead of using a macro, please just declare a const bool variable in route_click.

Deiz:
Made the suggested change.

PR #1634, opened 2015-04-03
Use a reasonable default sep_block_width if a separator_symbol is given

Airblader:
There's no reason not to provide a reasonable default; sorry for not doing this right away. I don't think it's a problem to add this now because it only means the user doesn't have to do something they would otherwise have to do. Also, it caused some confusion, see https://faq.i3wm.org/question/5727/how-to-set-separator-in-i3bar/

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1319,8 +1319,7 @@ bar {
 === Custom separator symbol
 
 Specifies a custom symbol to be used for the separator as opposed to the vertical,
-one pixel thick separator. Note that you may have to adjust the +sep_block_width+ 
-property.
+one pixel thick separator.
 
 *Syntax*:
 -------------------------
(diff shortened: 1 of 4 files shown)


Airblader:
Also, I'd vote for merging this into the 4.10.2 release that I suppose will happen rather soon (?).

docs/userguide
```diff
@@ -1319,8 +1319,7 @@ bar {
 === Custom separator symbol
 
 Specifies a custom symbol to be used for the separator as opposed to the vertical,
-one pixel thick separator. Note that you may have to adjust the +sep_block_width+
```
Airblader: Note that this was actually wrong because the property is named `separator_block_width`. `sep_block_width` is just the internal name.

PR #1636, opened 2015-04-04
Update copyright notices and get rid of ranges

Deiz:
Takes about two seconds to run on the whole repo on my machine. Should make it considerably easier to keep the copyright notices current.

### DIFF
--- a/docs/asciidoc-git.conf
+++ b/docs/asciidoc-git.conf
@@ -647,7 +647,7 @@ endif::doctype-manpage[]
 </div>
 {disable-javascript%<div id="footnotes"><hr /></div>}
 <div id="footer" lang="de">
-© 2009-2014 Michael Stapelberg, <a href="http://i3wm.org/impress.html">Impressum</a>
+© 2009 Michael Stapelberg, <a href="http://i3wm.org/impress.html">Impressum</a>
 </div>
 </body>
 </html>
--- a/generate-command-parser.pl
+++ b/generate-command-parser.pl
@@ -2,7 +2,7 @@
 # vim:ts=4:sw=4:expandtab
 #
 # i3 - an improved dynamic tiling window manager
-# © 2009-2012 Michael Stapelberg and contributors (see also: LICENSE)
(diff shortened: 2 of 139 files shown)


Deiz:
I was just looking at the output of a local run, and it appears that quite a few files include copyright notices for individual authors rather than the more typical "© $start-$end $name and contributors (see also: LICENSE)".

Presumably those shouldn't be automatically updated, so it might be a good idea to only have the script modify lines containing 'and contributors'. That said, most of those single-author notices are out of date at this point, as i3 has enough contributors that it's rare to find a file that's only been worked on by one person.

Airblader:
I'd argue that it makes more sense to not have such notices in any file other than the LICENSE file. It isn't necessarily legally anyway. As for a single file being taken and passed around, wouldn't if suffice to instead include in the header to which project the file belongs and referencing the LICENSE file?

stapelberg:
After reading http://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated, I think the way we should go is to specify the year when the file in question was created, and not bother updating it.

@Deiz, could you please change your pull request to do that? You could use the script for that :).

Deiz:
Yeah, I can do that. Do you want `i3 --version` to be an exception to that (i.e. keep the © 2009-2014 style), or should it simply be © 2009?

stapelberg:
@Deiz i3 --version should also just be © 2009

Deiz:
Updated and rebased. Since getting the commit that created a file entails traversing the whole commit history, it's a fair bit slower now (takes around 15s, instead of the original implementation's 2-3s) but not unbearably slow.

stapelberg:
@Deiz Sorry, I should have been more clear: I’m not actually interested in the script, just the results, since this is a one-time cleanup.

Deiz:
Alright, updated again. i3status and i3lock could also be updated if you'd like.

stapelberg:
Thanks! Yes, please also update i3status and i3lock.

Issue #1489, opened 2015-02-23
i3 exits when no output available (docking station scenario) reopen #926


scoddy:
As described in #926 i3 correctly dies when no output is available. Beside the unlikely scenario that you disable all outputs via xrandr, there's another scenario that is a pretty inconvenient.

Using a docking station that does not announce the undock (e.g. ACPI Undock event triggered through manually pushing a button _before_ actually ejecting the notebook). During the removal of the notebook from the docking station, all outputs disappear momentarily. As the ACPI Undock event is triggered at the time the notebook is removed, there's no way to send a SIGSTOP to i3 before the graphic outputs disappear, thus i3 dies.

The only workaround is to manually call an 'undock' script before ejecting the notebook from the dock and sending a SIGSTOP to i3 as proposed in #926 . 
This is a major inconvenience and it would be great if a solution for #926 could be considered that actually allows all outputs to disappear (at least for a configurable period, before letting i3 die).

i3 version 4.8 (2014-06-15, branch "4.8") © 2009-2014 Michael Stapelberg and contributors

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
I’d argue that having the outputs disappear is a bug in your driver that should be fixed.

That said, I agree that it’s an inconvenience for which there doesn’t seem to be a good workaround. I’ll accept a pull request which adds a timeout called `delay_exit_on_zero_displays` (see `force_display_urgency_hint` for another implemented timeout).

hwangcc23:
Hi, 
I want to help this. 
I have implemented the new timeout delay_exit_on_zero_displays. But I have problems on testing: I have only one desktop PC and one monitor. Don't know how to do the test (to disable an output and enable it again).
Is there any suggestion?
Thanks.

Airblader:
@hwangcc23 Use `xrandr` to disable and enable the output. If you don't want to mess with your machine, I'd recommend setting up a virtual machine via VirtualBox and installing it in there.

hwangcc23:
@Airblader 
Thanks.  :)
I used "xrandr --output Virtual1 --off; xrandr --output Virtual1 --auto" to verify my patch.
Have created the pull request for review.

acrisci:
[...]

PR #1638, opened 2015-04-05
Add a timeout: delay_exit_on_zero_displays

hwangcc23:
This fixes the issue #1489.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1003,6 +1003,31 @@ force_display_urgency_hint <timeout> ms
 force_display_urgency_hint 500 ms
 ---------------------------------
 
+=== Delaying exiting on zero displays
+
+Outputs may disappear momentarily and come back later. For example,
+using a docking station that does not announce the undock (e.g. ACPI Undock 
+event triggered through manually pushing a button before actually ejecting 
+the notebook). During the removal of the notebook from the docking station,
+all outputs disappear momentarily.
+
+To prevent i3 from exiting when no output is available momentarily, you can 
+tell i3 to delay a certain time first and check available outputs again using 
+the +delay_exit_on_zero_displays+ directive. Setting the value to 0 disables 
+this feature.
+
+The default is 500ms.
+
+*Syntax*:
+----------------------------------------
+delay_exit_on_zero_displays <timeout> ms
+----------------------------------------
+
+*Example*:
+----------------------------------
+delay_exit_on_zero_displays 500 ms
+----------------------------------
+
 === Focus on window activation
 
 [[focus_on_window_activation]]
--- a/include/config.h
+++ b/include/config.h
@@ -167,6 +167,10 @@ struct Config {
      * flag can be delayed using an urgency timer. */
     float workspace_urgency_timer;
 
+    /** Use a timer to delay exiting when no output is available.
+      * This can prevent i3 from exiting when all outputs disappear momentarily. */
+    float zero_disp_exit_timer_ms;
+
     /** Behavior when a window sends a NET_ACTIVE_WINDOW message. */
     enum {
         /* Focus if the target workspace is visible, set urgency hint otherwise. */
--- a/src/randr.c
+++ b/src/randr.c
@@ -69,7 +69,7 @@ Output *get_first_output(void) {
     if (output->active)
         return output;
 
-    die("No usable outputs available.\n");
+    return NULL;
 }
 
 /*
@@ -564,6 +564,8 @@ static void handle_output(xcb_connection_t *conn, xcb_randr_output_t id,
     if (!new->active) {
         DLOG("width/height 0/0, disabling output\n");
         return;
+    } else {
+        new->to_be_disabled = false;
     }
 
     DLOG("mode: %dx%d+%d+%d\n", new->rect.width, new->rect.height,
@@ -585,11 +587,7 @@ static void handle_output(xcb_connection_t *conn, xcb_randr_output_t id,
     new->changed = true;
 }
 
-/*
- * (Re-)queries the outputs via RandR and stores them in the list of outputs.
- *
- */
-void randr_query_outputs(void) {
+static bool __randr_query_outputs(void) {
     Output *output, *other, *first;
     xcb_randr_get_output_primary_cookie_t pcookie;
     xcb_randr_get_screen_resources_current_cookie_t rcookie;
@@ -603,7 +601,7 @@ void randr_query_outputs(void) {
     xcb_randr_output_t *randr_outputs;
 
     if (randr_disabled)
-        return;
+        return true;
 
     /* Get screen resources (primary output, crtcs, outputs, modes) */
     rcookie = xcb_randr_get_screen_resources_current(conn, root);
@@ -615,7 +613,7 @@ void randr_query_outputs(void) {
         DLOG("primary output is %08x\n", primary->output);
     if ((res = xcb_randr_get_screen_resources_current_reply(conn, rcookie, NULL)) == NULL) {
         disable_randr(conn);
-        return;
+        return true;
     }
     cts = res->config_timestamp;
 
@@ -697,6 +695,11 @@ void randr_query_outputs(void) {
             DLOG("Output %s disabled, re-assigning workspaces/docks\n", output->name);
 
             first = get_first_output();
+            if (!first) {
+                FREE(res);
+                FREE(primary);
+                return false;
+            }
 
             /* TODO: refactor the following code into a nice function. maybe
              * use an on_destroy callback which is implement differently for
@@ -812,6 +815,32 @@ void randr_query_outputs(void) {
 
     FREE(res);
     FREE(primary);
+
+    return true;
+}
+
+/*
+ * (Re-)queries the outputs via RandR and stores them in the list of outputs.
+ *
+ */
+void randr_query_outputs(void) {
+    static bool first_query = true;
+
+    if (first_query) {
+        /* find monitors at least once via RandR */
+        if (!__randr_query_outputs())
+            die("No usable outputs available.\n");
+        first_query = false;
+    } else {
+        /* requery */
+        if (!__randr_query_outputs()) {
+            DLOG("sleep %f ms due to zero displays\n", config.zero_disp_exit_timer_ms);
+            usleep(config.zero_disp_exit_timer_ms * 1000);
+
+            if (!__randr_query_outputs())
+                die("No usable outputs available.\n");
+        }
+    }
 }
 
 /*
--- a/include/config_directives.h
+++ b/include/config_directives.h
@@ -51,6 +51,7 @@ CFGFUN(force_focus_wrapping, const char *value);
 CFGFUN(force_xinerama, const char *value);
 CFGFUN(fake_outputs, const char *outputs);
 CFGFUN(force_display_urgency_hint, const long duration_ms);
+CFGFUN(delay_exit_on_zero_displays, const long duration_ms);
 CFGFUN(focus_on_window_activation, const char *mode);
 CFGFUN(show_marks, const char *value);
 CFGFUN(hide_edge_borders, const char *borders);
--- a/parser-specs/config.spec
+++ b/parser-specs/config.spec
@@ -39,6 +39,7 @@ state INITIAL:
   'workspace_auto_back_and_forth'          -> WORKSPACE_BACK_AND_FORTH
   'fake_outputs', 'fake-outputs'           -> FAKE_OUTPUTS
   'force_display_urgency_hint'             -> FORCE_DISPLAY_URGENCY_HINT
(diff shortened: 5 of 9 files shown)


docs/userguide
```diff
@@ -981,6 +981,30 @@ force_display_urgency_hint <timeout> ms
 force_display_urgency_hint 500 ms
 ---------------------------------
 
+=== Delaying exit on zero displays
+
+i3 dies immediately when no output is available. But this may be inconvenient for
+some use cases. For example, Using a docking station that does not announce the undock
```
Airblader: nit: /Using/using/

docs/userguide
```diff
@@ -981,6 +981,30 @@ force_display_urgency_hint <timeout> ms
 force_display_urgency_hint 500 ms
 ---------------------------------
 
+=== Delaying exit on zero displays
+
+i3 dies immediately when no output is available. But this may be inconvenient for
```
Airblader: Since the default is 500ms, this isn't true now anymore.

include/config.h
```diff
@@ -167,6 +167,11 @@ struct Config {
      * flag can be delayed using an urgency timer. */
     float workspace_urgency_timer;
 
+    /** By default, i3 exits immediately when no output is avaiable. It is
```
Airblader: Same as above: because of this very setting and its default, this isn't true.

acrisci:
This is testable.

src/randr.c
```diff
@@ -589,7 +591,7 @@ static void handle_output(xcb_connection_t *conn, xcb_randr_output_t id,
  * (Re-)queries the outputs via RandR and stores them in the list of outputs.
  *
  */
-void randr_query_outputs(void) {
+int __randr_query_outputs(void) {
```
stapelberg: Please move the comment. Please mark the function `static` as it should only be used within this file.

stapelberg:
To be more clear about what @acrisci is trying to say: please add a testcase for this.

stapelberg:
Also travis indicates that this change makes the test suite fail. Please address that.

Airblader:
@acrisci @stapelberg Since I could use this for tests on another ticket – is there some way of disabling the fake outputs during runtime? Is there a test already that does something like this?

hwangcc23:
Fixed comments and the test suite first.
Working on the test case. (Not family with Perl. Studying the document for creating one.)

docs/userguide
```diff
@@ -1003,6 +1003,31 @@ force_display_urgency_hint <timeout> ms
...
 
+=== Delaying exit on zero displays
+
+Outputs may disappear momentarily and come back later. For example,
+using a docking station that does not announce the undock (e.g. ACPI Undock 
+event triggered through manually pushing a button before actually ejecting 
+the notebook). During the removal of the notebook from the docking station,
+all outputs disappear momentarily.
+
+To prevent i3 from exit when no output is available momentarily, you can tell
```
Airblader: Still s/exit/exiting/

docs/userguide
```diff
@@ -1003,6 +1003,31 @@ force_display_urgency_hint <timeout> ms
...
+
+Outputs may disappear momentarily and come back later. For example,
+using a docking station that does not announce the undock (e.g. ACPI Undock 
+event triggered through manually pushing a button before actually ejecting 
+the notebook). During the removal of the notebook from the docking station,
+all outputs disappear momentarily.
+
+To prevent i3 from exit when no output is available momentarily, you can tell 
+i3 to delay a certain time first and check available outputs again using the 
++delay_exit_on_zero_displays+ directive.  Setting the value to 0 disables
```
Airblader: There's a double space here now.

hwangcc23: Fixed the typo. Thanks.  :)

src/randr.c
```diff
@@ -585,11 +587,7 @@ static void handle_output(xcb_connection_t *conn, xcb_randr_output_t id,
     new->changed = true;
 }
 
-/*
- * (Re-)queries the outputs via RandR and stores them in the list of outputs.
- *
- */
-void randr_query_outputs(void) {
+static int __randr_query_outputs(void) {
```
stapelberg: It seems like this function either succeeds or it doesn’t. Can you change the return type to bool and use true/false to signal success/failure please?

src/randr.c
```diff
@@ -812,6 +815,23 @@ void randr_query_outputs(void) {
...
+}
+
+/*
+ * (Re-)queries the outputs via RandR and stores them in the list of outputs.
+ *
+ */
+void randr_query_outputs(void) {
+    if (__randr_query_outputs() < 0) {
+        DLOG("sleep %f ms due to zero displays\n", config.zero_disp_exit_timer_ms);
+        usleep(config.zero_disp_exit_timer_ms * 1000);
```
stapelberg: The unconditional sleep introduces a delay when i3 falls back to using the X root window on systems without RandR, I think. Could you make the conditional so that it only sleeps when i3 found monitors via RandR at least once?

src/randr.c
```diff
@@ -69,7 +69,7 @@ Output *get_first_output(void) {
     if (output->active)
         return output;
 
-    die("No usable outputs available.\n");
+    return NULL;
```
stapelberg: src/main.c also calls get_first_output and will need to be adapted, now that you introduced the possibility of returning NULL.

src/randr.c
```diff
@@ -564,6 +564,8 @@ static void handle_output(xcb_connection_t *conn, xcb_randr_output_t id,
     if (!new->active) {
         DLOG("width/height 0/0, disabling output\n");
         return;
+    } else {
+        new->to_be_disabled = false;
```
stapelberg: Can you explain why you needed to add this line?

hwangcc23: If the output is disabled, __randr_query_outputs() doesn't handle it immediately and returns false first.
After timeout, if the output is active again, its to_be_disabled flag should be clear.
If the to_be_disabled flag is still set, the output will be set as disabled (in the line 694).

692     TAILQ_FOREACH(output, &outputs, outputs) {
693         if (output->to_be_disabled) {
694             output->active = false;
695             DLOG("Output %s disabled, re-assigning workspaces/docks\n", output->name);

hwangcc23:
Hi,
I have problems on implementing the new test case for this patch.
I originally used xrandr to disable the output for testing.
But I found that it is not working in Xephyr. ("xrandr --output default --off")
Is there any suggestion?
Thanks.

stapelberg:
@acrisci When you said that this is testable, was that a guess or can you explain how it can be tested specifically? As @hwangcc23 noted, Xephyr’s RANDR support merely allows for changing the resolution, but you cannot turn displays off.

acrisci:
This should be testable by

```
xrandr --output default --off
sleep 0.4
xrandr --output default --auto
# does i3 live?
```

within the test env.

hwangcc23:
@acrisci,
It is not working in my environment.
For debugging the problem, I tried to open an xterm and run xrandr with --off in Xephyr directly. It didn't work neither.
screenshot2

P.S. My testcase 243-delay-exit-on-zero-displays.t:
    use i3test i3_autostart => 0;
    use Time::HiRes qw(sleep);
    my $config = <<EOT;
    # i3 config file (v4)
    font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
    delay_exit_on_zero_displays 1000 ms
    EOT
    my $pid = launch_with_config($config);
    cmd qq|exec "xrandr --output default --off"|;
    sleep(1);
    cmd qq|exec "xrandr --output default --auto"|;
    does_i3_live;
    exit_gracefully($pid);
    done_testing;

acrisci:
It might be a little tricky to make this testable after all.

I can get the xrandr commands to work in Xephyr, but only if I initialize it with startx like this

    startx -- /usr/bin/Xephyr :2

otherwise it won't work.

hwangcc23:
@acrisci,
Thanks. I tried your method. But it still didn't work in my environment.
screenshot3

hwangcc23:
Hi all,
Is there any suggestion or another method for the test?
(Cannot get the patch merged until the testcase is added successfully.) 
Thanks.

Airblader:
I think if it can't be tested, it can still be merged (manual testing fallback). @stapelberg just needs comments in a PR to get a notification to look into it again. :)

However, it looks like there are still comments on the changes itself.

hwangcc23:
@Airblader, thanks.  :)
@stapelberg, I have updated the commit (to revise code according to your suggestions) and replied your question. Could you review it again? Thanks.

acrisci:
Xephyr claims support of randr so the test as I described should work. This leads me to believe we have found a bug in Xephyr. I will contact someone on the project to get some clarification on this issue.

Since the issue with the test could take some extra effort to fix, I wouldn't have a problem merging this and adding an issue to the bug tracker to write a test later if everything else is ok with the pr.

acrisci:
I looked into it a bit more and I found that `xrandr --output default --off` is indeed disabling the output. i3 then checks the outputs and finds none and dies. It just looks as though i3 is still running because it doesn't unmap any of the windows, but the i3 interface is unresponsive.

However, the i3 process does not seem to actually terminate as the code would suggest here:

https://github.com/i3/i3/blob/next/src/randr.c#L72

I think this is a bug in i3. I believe if i3 would cleanly shut down in this case, the test would work. I'll do some more testing on this theory.

stapelberg:
> Xephyr claims support of randr so the test as I described should work.

@acrisci, when starting i3 with `DISPLAY` pointing to a Xephyr instance, I see the following log lines:

```
2015-04-26 23:45:24 - main.c:main:607 - Checking for XRandR...
2015-04-26 23:45:24 - randr.c:randr_query_outputs:615 - primary output is 00000000
2015-04-26 23:45:24 - ERROR: No outputs found via RandR, disabling
```

Which suggests that Xephyr doesn’t support RandR.

> I looked into it a bit more and I found that xrandr --output default --off is indeed disabling the output. i3 then checks the outputs and finds none and dies.

I cannot reproduce this. When I run `DISPLAY=:1 xrandr --output default --off`, I see the following lines in the i3 log:

```
2015-04-26 23:45:27 - handlers.c:handle_event:1234 - event type 88, xkb_base 85
2015-04-26 23:45:27 - handlers.c:handle_screen_change:430 - RandR screen change
2015-04-26 23:45:27 - handlers.c:handle_screen_change:440 - root geometry reply: (0, 0) 640 x 480
2015-04-26 23:45:27 - scratchpad.c:scratchpad_fix_resolution:256 - Current resolution: (0, 0) 640 x 480
2015-04-26 23:45:27 - scratchpad.c:scratchpad_fix_resolution:265 - output xroot-0's resolution: (0, 0) 640 x 480
2015-04-26 23:45:27 - scratchpad.c:scratchpad_fix_resolution:278 - new width = 640, new height = 480
2015-04-26 23:45:27 - scratchpad.c:scratchpad_fix_resolution:285 - Scratchpad size unchanged.
[…]
```

But i3 continues to run just fine. Further, xrandr still gives precisely the same output as before trying to turn the default output off:

```
midna ~ $ DISPLAY=:1 xrandr --output default --off
xrandr: Failed to get size of gamma for output default
midna ~ $ DISPLAY=:1 xrandr                       
xrandr: Failed to get size of gamma for output default
Screen 0: minimum 160 x 160, current 640 x 480, maximum 1600 x 1200
default connected 640x480+0+0 (normal left inverted right x axis y axis) 0mm x 0mm
   1600x1200      0.00  
   1400x1050      0.00  
   1280x960       0.00  
   1280x1024      0.00  
   1152x864       0.00  
   1024x768       0.00  
   832x624        0.00  
   800x600        0.00  
   720x400        0.00  
   480x640        0.00  
   640x480        0.00* 
   640x400        0.00  
   320x240        0.00  
   240x320        0.00  
   160x160        0.00  
```

I’m curious why our experiences are so different. Could it be that RandR support in Xephyr depends on the environment in which it’s running in?

acrisci:
It seems to not be enabling the output when it starts. If you manually enable the output before starting i3 with `xrandr --output default --auto`, RandR will be enabled. The output will also be enabled automatically when you start Xephyr with startx. This complicates testing because we would need to make sure i3 starts with the output enabled.

Also, I can't determine exactly how to tell whether an output is off. The output of `xrandr` seems to me to be exactly the same whether the output is on or off.

I think we might need to contact someone from Xephyr regarding these issues.

stapelberg:
> It seems to not be enabling the output when it starts. If you manually enable the output before starting i3 with xrandr --output default --auto, RandR will be enabled.

I can’t reproduce this here:

```
$ Xephyr :1 -screen 1024x768 -dpi 96 &
$ DISPLAY=:1 xrandr --output default --auto
xrandr: Failed to get size of gamma for output default
$ DISPLAY=:1 i3 -V -d all
[…]
2015-05-05 09:35:58 - ERROR: No outputs found via RandR, disabling
```

I think we’ve tried finding a way to test this automatically for long enough. I’ll manually test this patch and then move forward with it.

PR #1641, opened 2015-04-06
Store Git commit identifier in its own object

### LINKED ISSUE #1640: --version/--moreversion cannot be trusted after compilation without “make clean”
**TL;DR:** Just running `make` and omitting `make clean` aparently may result in mix-match of binaries that (apart from other potential problems) may report the older version.

Happened after checking out commit eb04a64 and re-building, with left-over binaries from tag 4.10.1.  Tree clean in both cases; Fedora 21 w/ git-2.1.0-4.fc21.x86_64 and gcc-4.9.2-6.fc21.x86_64.

---

I came to my machine (with i3 built from 4.10.1 tag running) with intent to quickly verify a bug fixed few hours ago.

    # cd'd to the i3 git repo
    git checkout master
    make
    sudo make install

Then restarted w/ $mod+R (actually twice, as the verification required).  I could immediately see the effect of the changed behavior so that i3 was definitely reloaded.

But, out of curiosity I ran `i3 --moreversion` and to my surprise, both reported versions were 4.10.1, just as before restarting!

    Binary i3 version:  4.10.1 (2015-03-29, branch "4.10.1") © 2009-2014 Michael Stapelberg and contributors
    Running i3 version: 4.10.1 (2015-03-29, branch "4.10.1") (pid 1552)

So I got back to the git repo, and this time ran `make clean` before `make`, and I could immediately see that many more binaries were getting built. After installing and reloading, versions were right:

    Binary i3 version:  4.10.1-6-geb04a64 (2015-04-06, branch "master") © 2009-2014 Michael Stapelberg and contributors
    Running i3 version: 4.10.1-6-geb04a64 (2015-04-06, branch "master") (pid 1552)

I guess this could lead to pretty strange situations with misleading data, if anybody uses the output for bug reporting.

Deiz:
Fixes #1640. I believe I caught all instances (within i3 itself) of the version string being compiled in as part of a static string.

Note that I've only updated i3 itself, the standalone utilities are mostly not susceptible to this by virtue of mostly being single files, though i3bar is a notable exception.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,12 @@ include i3-dump-log/i3-dump-log.mk
 include docs/docs.mk
 include man/man.mk
 
+# Update $(TOPDIR)/LAST_VERSION if it differs from $I3_VERSION
+CACHED_VERSION := '$(shell [ -f $(TOPDIR)/LAST_VERSION ] && cat $(TOPDIR)/LAST_VERSION)'
+ifneq ($(CACHED_VERSION),$(I3_VERSION))
+$(shell echo -n ${I3_VERSION} > $(TOPDIR)/LAST_VERSION)
+endif
+
 real-all: $(ALL_TARGETS)
 
 install: $(INSTALL_TARGETS)
--- a/src/i3.mk
+++ b/src/i3.mk
@@ -32,6 +32,10 @@ include/all.h.pch: $(i3_HEADERS)
 	echo "[i3] PCH all.h"
 	$(CC) $(I3_CPPFLAGS) $(XCB_CPPFLAGS) $(CPPFLAGS) $(i3_CFLAGS) $(I3_CFLAGS) $(CFLAGS) -x c-header include/all.h -o include/all.h.pch
 
+src/version.o: src/version.c LAST_VERSION $(i3_HEADERS_DEP)
+	echo "[i3] CC $<"
+	$(CC) $(I3_CPPFLAGS) $(XCB_CPPFLAGS) $(CPPFLAGS) $(i3_CFLAGS) $(I3_CFLAGS) $(CFLAGS) $(PCH_FLAGS) -c -o $@ ${canonical_path}/$<
+
 src/%.o: src/%.c $(i3_HEADERS_DEP)
 	echo "[i3] CC $<"
 	$(CC) $(I3_CPPFLAGS) $(XCB_CPPFLAGS) $(CPPFLAGS) $(i3_CFLAGS) $(I3_CFLAGS) $(CFLAGS) $(PCH_FLAGS) -c -o $@ ${canonical_path}/$<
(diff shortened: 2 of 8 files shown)


stapelberg:
How does this fix #1640 specifically? The Makefile still doesn’t have a dependency on the version itself, does it?

Deiz:
Since make only rebuilds objects when the source is newer than the object, there should be no difference between running `make` on commits A, B, C and running `make clean && make` on commit C.

The only problem I see in #1640 is that the version strings were separately embedded in numerous places, and so `i3 --version`, etc. would always be incorrect unless switching branches/merging commits happened to touch main.c

stapelberg:
My point is that the Makefile doesn’t see the git revision as an input to version.c, but it must, otherwise it won’t rebuild version.c.

Deiz:
Ah, I forgot about that. I just pushed a commit that makes version.o get rebuilt when .git/HEAD or .git/index change, which is a bit too aggressive (since staging changes won't actually change the output of `git describe`) but I think that's acceptable. The only other option would be to cache the output of `git describe` in a file and compare against it, and I think that's overkill.

stapelberg:
> The only other option would be to cache the output of git describe in a file and compare against it, and I think that's overkill.

I actually think that’s not overkill, but the best way to avoid rebuilds when unnecessary and still rebuild whenever necessary. Can you do that please?

Deiz:
Updated it again, though I just realized I3_VERSION (the file) is now somewhat overloaded, because it's the canonical version for source tarballs, but not respected when within a Git repo (which of course includes cases where someone's unpacked a source tarball within a Git repo).

I think it'd be better to cache the version string as $(TOPDIR)/.last_version or similar.

Deiz:
Gave it a bit more thought: My proposed behaviour is to respect I3_VERSION (the file, again) if it exists, and otherwise generate/compare against LAST_VERSION, which will be the cached copy of the version string.

src/i3.mk
```diff
@@ -92,4 +96,6 @@ install-i3: i3
 
 clean-i3:
 	echo "[i3] Clean"
-	rm -f $(i3_OBJECTS) $(i3_SOURCES_GENERATED) $(i3_HEADERS_CMDPARSER) include/loglevels.h loglevels.tmp include/all.h.pch i3-command-parser.stamp i3-config-parser.stamp i3 test.config_parser test.commands_parser src/*.gcno src/cfgparse.* src/cmdparse.*
+	rm -f $(i3_OBJECTS) $(i3_SOURCES_GENERATED) $(i3_HEADERS_CMDPARSER) include/loglevels.h loglevels.tmp include/all.h.pch i3-command-parser.stamp i3-config-parser.stamp i3 test.config_parser test.commands_parser src/*.gcno src/cfgparse.* src/cmdparse.* LAST_VERSION
+
+.INTERMEDIATE: LAST_VERSION
```
stapelberg: Why are you marking this file intermediate? I’ve read https://www.gnu.org/software/make/manual/html_node/Chained-Rules.html#Chained-Rules and don’t understand why that’s necessary.

Makefile
```diff
@@ -22,6 +22,16 @@ include i3-dump-log/i3-dump-log.mk
 include docs/docs.mk
 include man/man.mk
 
+# Check and update $(TOPDIR)/LAST_VERSION if I3_VERSION doesn't exist
```
stapelberg: Can you expand on why you do that? Wouldn’t the code also work just as well if you unconditionally created LAST_VERSION?

src/i3.mk
Deiz: I've marked it as intermediate because it's not generated when building from a source tarball, which would cause dependency failures since there isn't actually a rule for creating it.

If you'd prefer that LAST_VERSION always be created/updated (including when building from a source tarball) then it doesn't need to be intermediate.

stapelberg: See my other comment. I think (without having tested it) that unconditionally creating that file would make this idea simpler.

Deiz:
Updated to always generate LAST_VERSION, and verified that it works properly with source tarballs as well as Git.

PR #1644, opened 2015-04-08
Respect EXEC_PREFIX and a users' choice of PKG_CONFIG.

tanderson92:
The Makefiles should put binaries in $(EXEC_PREFIX) and
architecture-independent files in $(PREFIX). Also a user may have a
prefixed- pkg-config, as in the case of cross compiling on Exherbo
Linux, so respect the well-accepted $(PKG_CONFIG) variable for this
purpose.

If accepted, these commits should be squashed -- I forgot about what the common case should be.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -6,7 +6,7 @@ ifndef PREFIX
   PREFIX=/usr
 endif
 ifndef EXEC_PREFIX
-  EXEC_PREFIX=/usr
+  EXEC_PREFIX=$(PREFIX)
 endif
 ifndef SYSCONFDIR
   ifeq ($(PREFIX),/usr)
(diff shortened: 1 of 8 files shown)


common.mk
```diff
@@ -59,8 +62,12 @@ I3_CPPFLAGS += -DI3__FILE__=__FILE__
 
 ## Libraries flags
 
+ifndef PKG_CONFIG
 ifeq ($(shell which pkg-config 2>/dev/null 1>/dev/null || echo 1),1)
 $(error "pkg-config was not found")
+else
+PKG_CONFIG = `which pkg-config`
```
stapelberg: Instead of doing this, just define `PKG_CONFIG=pkg-config` above (see line 1), that’s more consistent.

stapelberg:
Can you squash your commits please? I don’t want a commit titled “oops” to end up in our history.

tanderson92:
I was under the impression github could squash commits together during a pull request, is this not correct?

stapelberg:
@tanderson92 You didn’t have to create a new pull request, you could have just pushed to your branch after squashing the commits with `git push --force`.

PR #1650, opened 2015-04-10
_NET_WM_STATE_HIDDEN on stacked/tabbed containers

Airblader:
I'm not entirely sure this is the right / only place, though from playing around it seems to work fine.

I will also try to add a testcase for it.


Airblader:
This causes the entire X server to crash if i3 is restarted inplace with a fullscreened container in a stacking layout (a test for this exists and fails). I'll check on it.

src/con.c
```diff
@@ -186,6 +186,31 @@ void con_detach(Con *con) {
...
+
+    unsigned int num = 0;
+    uint32_t values[1];
+    if (hidden) {
+        DLOG("setting _NET_WM_STATE_HIDDEN for con = %p\n", con);
+        values[num++] = A__NET_WM_STATE_HIDDEN;
+    } else
+        DLOG("removing _NET_WM_STATE_HIDDEN for con = %p\n", con);
+
+    xcb_change_property(conn, XCB_PROP_MODE_REPLACE, con->window->id, A__NET_WM_STATE, XCB_ATOM_ATOM, 32, num, values);
```
Airblader: I'm wondering whether I should use APPEND or PREPEND here instead so that other states (fullscreen, urgent) aren't lost (and does it matter whether I append or prepend)? If so, is there any way to remove *only* the hidden atom? Do I have to fetch the properties, remove the atom from the result and then change it?

src/con.c
```diff
@@ -202,6 +227,14 @@ void con_focus(Con *con) {
         con_focus(con->parent);
 
     focused = con;
+
+    if (con->parent->layout == L_STACKED || con->parent->layout == L_TABBED) {
+        Con *child;
+        TAILQ_FOREACH(child, &(con->parent->nodes_head), nodes) {
```
Airblader: This is too spammy. I'll change it to only make the calls on the previously focused and now focused container.

Airblader:
This is definitely not enough. When a container is moved from the tabbed container somewhere else, the tabbed container is left with all windows set to hidden.

Airblader:
So as much as I would love to have this for multiple reasons – I think this might be too hard. We have to keep this updated when containers are moved around, even using criteria matches and the like, so the moved container might be hidden or it might not be. It might be moved within the tabbed container or it might be moved to somewhere else.

Any opinions or suggestions on whether there is a good place for this? Otherwise I think it's really easy to overlook cases here.

stapelberg:
I’d put the logic into src/render.c, because that’s where every container that is affected by any operation (moves etc.) gets evaluated anyway. That way, it’d be impossible to miss cases.

Airblader:
@stapelberg I had thought about it but was afraid it wouldn't be a clean solution. If you're okay with it, I will try to move the logic there. I haven't checked whether all relevant containers go through there (because if I move the container out of the tabbed container, I also need to update the other children of the tabbed container), but I would guess so (since at least the decoration has to be redrawn).

acrisci:
I don't think this belongs in `render_con()`

>
/*
 * "Renders" the given container (and its children), meaning that all rects are
 * updated correctly. Note that this function does not call any xcb_*
 * functions, so the changes are completely done in memory only (and
 * side-effect free). As soon as you call x_push_changes(), the changes will be
 * updated in X11.
 *
 */
>

I think it belongs in `x_push_changes()`. Right?

Airblader:
@acrisci Yes, you're right.

I also completely disregarded nested containers. Since the approach has changed fundamentally, I will close this PR and make a new one as I will be working on a new branch.

PR #1651, opened 2015-04-10
Respect EXEC_PREFIX and a users' choice of PKG_CONFIG.

tanderson92:
The Makefiles should put binaries in $(EXEC_PREFIX) and
architecture-independent files in $(PREFIX). Also a user may have a
prefixed- pkg-config, as in the case of cross compiling on Exherbo
Linux, so respect the well-accepted $(PKG_CONFIG) variable for this
purpose.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -2,8 +2,12 @@ UNAME=$(shell uname)
 DEBUG=1
 INSTALL=install
 LN=ln
+PKG_CONFIG=pkg-config
 ifndef PREFIX
   PREFIX=/usr
+endif
+ifndef EXEC_PREFIX
+  EXEC_PREFIX=$(PREFIX)
 endif
 ifndef SYSCONFDIR
   ifeq ($(PREFIX),/usr)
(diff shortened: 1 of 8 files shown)


common.mk
```diff
@@ -59,8 +62,12 @@ I3_CPPFLAGS += -DI3__FILE__=__FILE__
 
 ## Libraries flags
 
+ifndef PKG_CONFIG
 ifeq ($(shell which pkg-config 2>/dev/null 1>/dev/null || echo 1),1)
 $(error "pkg-config was not found")
+else
+PKG_CONFIG=pkg-config
```
stapelberg: Please move the `PKG_CONFIG=pkg-config` definition up to the very top of the file (i.e. into line 5) and remove the ifndef/else/endif/

tanderson92:
What's the hold up on this? I had hoped it would go in before the next release (too late now)

Airblader:
@tanderson92 4.10.2 is a bugfix release. Since this isn't a bug, why put it into that release?

stapelberg:
You didn’t reply with a comment, so I didn’t get an email notification.

Regardless of that, we wouldn’t have cherry-picked it for the 4.10.2 release unless you had specifically asked for it and provided a good reason.

tanderson92:
Ah, my fault then. Apologies.

It is a bug in the sense that it goes against what most packages on linux do when using autotools or Makefiles in general. The reason i submitted the patch is because it really does break cross-compiling on a linux distribution.

PR #1654, opened 2015-04-11
Allow complete-run.pl to be run from any directory

Deiz:
More or less a Perl equivalent to `cd "$(dirname $0)"`.

The BEGIN block is needed to ensure $dirname is defined before `use lib $dirname . '/lib';` is evaluated.

### DIFF
--- a/testcases/complete-run.pl
+++ b/testcases/complete-run.pl
@@ -8,7 +8,6 @@ use v5.10;
 use utf8;
 # the following are modules which ship with Perl (>= 5.10):
 use Pod::Usage;
-use Cwd qw(abs_path);
 use File::Temp qw(tempfile tempdir);
 use Getopt::Long;
 use POSIX ();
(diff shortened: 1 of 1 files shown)


testcases/complete-run.pl
```diff
@@ -17,8 +16,18 @@ package complete_run;
...
 use Time::HiRes qw(time);
 use IO::Handle;
+
+my $dirname;
+
+BEGIN {
+   use File::Basename;
+   use Cwd 'abs_path';
+
+   $dirname = (fileparse(abs_path($0)))[1];
```
stapelberg: Why not just `$dirname = dirname(abs_path($0));`?

testcases/complete-run.pl
```diff
@@ -17,8 +16,18 @@ package complete_run;
 use TAP::Parser::Aggregator;
 use Time::HiRes qw(time);
 use IO::Handle;
+
+my $dirname;
+
+BEGIN {
+   use File::Basename;
+   use Cwd 'abs_path';
```
stapelberg: We use `use Cwd qw(abs_path);` everywhere else, can you please keep the style in `use` directives consistent?

testcases/complete-run.pl
Deiz: `dirname` emulates the utility of the same name, which means that for `/foo/bar/` it would return `/foo` even though bar is a directory as well. It'd be safe enough to use here since `$0` will always end with a file, but I prefer to avoid functions that deliberately emulate quirks.

stapelberg: Fair enough. Please add a comment stating that fileparse will always return the directory portion of a path and refer to File::Basename(3p) for more details.

Deiz:
Made the changes and also fixed my indenting (had used 3-space tabs instead of 4).

PR #1657, opened 2015-04-14
i3bar: fix flickering shortened status bar on other output(s)

Georgiy-Tugai:
When one `i3bar` instance is used on more than one output, and an output other than the last one (in internal linked list) uses `short_text` rendering, the subsequent outputs (which would normally be showing a full-size status) also show a shortened status for a second or so.

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -1792,6 +1792,8 @@ void reconfig_windows(bool redraw_bars) {
 void draw_bars(bool unhide) {
     DLOG("Drawing bars...\n");
     int workspace_width = 0;
+    /* Is the currently-rendered statusline using short_text items? */
+    bool rendered_statusline_is_short = false;
 
     refresh_statusline(false);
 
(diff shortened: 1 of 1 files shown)


Airblader:
Please remove that apostrophe in "short_text's"  (can't line comment on mobile, apparently). :)

i3bar/src/xcb.c
```diff
@@ -1792,6 +1792,7 @@ void reconfig_windows(bool redraw_bars) {
 void draw_bars(bool unhide) {
     DLOG("Drawing bars...\n");
     int workspace_width = 0;
+    bool rendered_statusline_is_short = false; /* Is the currently-rendered statusline using short_text items? */
 
     refresh_statusline(false);
```
stapelberg: I personally find your suggestion a bit hard to follow (I get what it does now, but it took a bit). How about just moving the `refresh_statusline(false);` call a couple of lines down, i.e. before line 1804? That should also fix the bug, right?

stapelberg:
Also, please add a comment whenever you push new commits, because only comments trigger email notifications.

And please squash your commits (then use push -f) so that the commit history stays clean.

Georgiy-Tugai:
I have squashed the commits, as per your request.

As for the reasoning behind the suggestion; I saw three different ways to solve the problem.

1. Re-render for every output (as you suggested)
2. Re-render only when the status line must be rendered differently (`short_text` vs `full_text`).
3. Introduce another pixmap for `short_text` statusbar.

The solution chosen (2) is a trade-off between complexity and scalability; Solution 1 will result in excess rendering for anyone using more than one output (even if the outputs are the same width), while solution 3 is complicated and results in excess memory usage for anyone using only one output, or who is not using `short_text`; on the other hand, this solution scales well to numbers of outputs greater than 2, which cause excess rendering calls when using solution 2.

stapelberg:
With regards to solution (3), you could introduce the short_text pixmap only when it’s actually necessary, but that’s even more complex and probably not worth thinking a lot about, given that the vast majority of users use either 1 or 2 outputs.

I’m fine with the approach you’re taking after hearing the rationale.

Can you please move the comments that are on the same line as code to their own line to be more consistent with the rest of our codebase?

Georgiy-Tugai:
Done.

P.S: Is more "intelligent" handling of `short_text` on the roadmap somewhere? Admittedly, the "right" way to choose which entries to shorten is a can of worms I don't want to open yet.

stapelberg:
@Georgiy-Tugai There’s no official roadmap. i3 is driven mostly by contributions, there is no full-time development. I personally don’t care about `short_text`, so if you want to see it being handled better, the best way to achieve that is by coordinating with us and sending pull requests :).

PR #1662, opened 2015-04-15
Allow moving windows to marks

### LINKED ISSUE #1643: [enhancement] Support moving windows to marked containers
For now, i3 supports moving windows to workspaces. Using the for_window directive I can make the window of an X11 application always open on another workspace so it doesn't interfere with the layout of my development workspace. Unfortunately, this only works for multi-monitor setups.

By adding a command to move a container into a marked container window placing rules can be implemented by using the for_window command.
To illustrate use cases of this command:
This is how my dev workspace looks like
```
+-------------------+
|                   |
|Editor, inside     |
|tabbed container   |
|with Glade, other  |
|Terminals, etc     |
+-------------------+
+--------+ +--------+
|Term1   | |Term2   |
+--------+ +--------+
```` 
When starting my application from Term1, i3 places its window i the split container of Term1 and Term2, which obviously isn't where I want my application to pop up. Instead, it'd be more useful if I can tell i3 to put the application window in the big tabbed container above the terminals.

Another use case: You click on a link in your browser, a PDF download starts, your browser is configured to automatically launch a pdf viewer when the download has finished. Since the download takes some time, you switch to another workspace an do some other stuff. When the pdf viewer finally opens, you move it to the workspace where it's supposed to be, only to find out that the focused container on that workspace was the wrong one, so you end up shuffling the pdf viewer around to place where you want it. This situation also can be simplified by telling i3 to move all pdf viewer windows to the container containing the browser and pdf documents that opened immediately.

The syntax for this command could look like this:

```move [window|container] [to] container <mark>``` where mark is the one previously set using ```mark``` In case a leaf container has been marked, the parent container should be used.

Airblader:
This is split into a few commits because it should make reviewing this a whole lot easier since the changes might otherwise be too complicated to oversee. I had to do some refactoring to include a number of edge cases and I hope I haven't overlooked anything.

If there are more ideas for test cases, please let me know. I've added a test for everything that gave me trouble on the way as well as the things that seemed worth testing.

fixes #1643

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1837,6 +1837,26 @@ bindsym $mod+x move workspace to output right
 bindsym $mod+x move container to output VGA1
 --------------------------------------------------------
 
+=== Moving containers/workspaces to marks
+
+To move a container to another container with a specific mark (see <<vim_like_marks>>),
+you can use the following command.
+
+The window will be moved right after the marked container in the tree, i.e., it ends up
+in the same position as if you had opened a new window when the marked container was
+focused. If the mark is on a split container, the window will appear as a new child
+after the currently focused child within that container.
+
+*Syntax*:
+------------------------------------
+move window|container to mark <mark>
+------------------------------------
+
+*Example*:
+--------------------------------------------------------
+for_window [instance="tabme"] move window to mark target
+--------------------------------------------------------
+
 [[resizingconfig]]
 
 === Resizing containers/windows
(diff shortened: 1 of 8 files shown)


stapelberg:
Just a couple of nits, otherwise this looks good. Nicely done!

Airblader:
@stapelberg All taken care of and rebased :)

Airblader:
@stapelberg Done.

Airblader:
@stapelberg And done.

PR #1664, opened 2015-04-17
Set _NET_WM_STATE_HIDDEN in tabbed/stacked containers

### LINKED ISSUE #1648: [enhancement] Set _NET_WM_STATE_HIDDEN in tabbed / stacked containers
http://standards.freedesktop.org/wm-spec/1.3/ar01s05.html: 
> _NET_WM_STATE_HIDDEN should be set by the Window Manager to indicate that a window would not be visible on the screen if its desktop/viewport were active and its coordinates were within the screen bounds. The canonical example is that minimized windows should be in the _NET_WM_STATE_HIDDEN state. Pagers and similar applications should use _NET_WM_STATE_HIDDEN instead of WM_STATE to decide whether to display a window in miniature representations of the windows on a desktop.

Shouldn't this apply to unfocused containers in stacked / tabbed layouts?

Airblader:
Again, I hope the testcases cover everything, but please do ask me to write more tests if you can think of anything. This can be crucial since some applications will "freeze" while this atom is set (e.g., cmp. the visibility API in Chrome)

fixes #1648


src/x.c
```diff
@@ -612,6 +612,40 @@ void x_deco_recurse(Con *con) {
...
+        DLOG("setting _NET_WM_STATE_HIDDEN for con = %p\n", con);
+        values[num++] = A__NET_WM_STATE_HIDDEN;
+    } else {
+        if (!is_hidden)
+            goto after_set_hidden_state;
+
+        DLOG("removing _NET_WM_STATE_HIDDEN for con = %p\n", con);
+    }
+
+    xcb_change_property(conn, XCB_PROP_MODE_REPLACE, con->window->id, A__NET_WM_STATE, XCB_ATOM_ATOM, 32, num, values);
```
Airblader: Should I use `XCB_PROP_MODE_APPEND` here for the case where I add the atom? How can I use the `reply` above to only remove the hidden atom when removing it?

Airblader:
Mh. Not sure why that test is failing, but looking into it.

src/x.c
```diff
@@ -612,6 +612,40 @@ void x_deco_recurse(Con *con) {
...
+ *
+ */
+static void set_hidden_state(Con *con) {
+    if (con->window == NULL) {
+        return;
+    }
+
+    /* check whether the window is currently marked as hidden. */
+    xcb_get_property_reply_t *reply =
+        xcb_get_property_reply(conn, xcb_get_property(conn, false, con->window->id, A__NET_WM_STATE, XCB_GET_PROPERTY_TYPE_ANY, 0, UINT32_MAX), NULL);
```
Airblader: This might be fairly expensive since it's done on every container everytime `x_push_changes` is called. Would it be worth storing this as a property on the container?

include/atoms.xmacro
```diff
@@ -5,6 +5,7 @@ xmacro(_NET_WM_MOVERESIZE)
 xmacro(_NET_WM_STATE_FULLSCREEN)
 xmacro(_NET_WM_STATE_DEMANDS_ATTENTION)
 xmacro(_NET_WM_STATE_MODAL)
+xmacro(_NET_WM_STATE_HIDDEN)
```
Airblader: There's a spot in `src/ewmh.c` that I need to change due to this. I'll send an update to the PR.

include/con.h
```diff
@@ -48,6 +48,13 @@ bool con_has_managed_window(Con *con);
  */
 bool con_is_split(Con *con);
 
+/*
```
stapelberg: nit: use `/**` in .h files.

src/con.c
```diff
@@ -258,6 +258,29 @@ bool con_is_split(Con *con) {
...
+ * This will only return true for containers which have some parent with
+ * a tabbed / stacked parent of which they are not the currently focused child.
+ *
+ */
+bool con_is_hidden(Con *con) {
+    Con *current = con;
+
+    /* ascend to the workspace level and memorize the highest-up container
+       which is stacked or tabbed. */
+    do {
```
stapelberg: Why the `do { } while` loop? A regular `while` loop should work just fine here, shouldn’t it?

src/con.c
```diff
@@ -258,6 +258,29 @@ bool con_is_split(Con *con) {
...
 /*
+ * This will only return true for containers which have some parent with
+ * a tabbed / stacked parent of which they are not the currently focused child.
+ *
+ */
+bool con_is_hidden(Con *con) {
+    Con *current = con;
+
+    /* ascend to the workspace level and memorize the highest-up container
+       which is stacked or tabbed. */
```
stapelberg: nit: add `*` in front of “which”

src/x.c
stapelberg: Only updating it when it actually needs to change would definitely be better, yeah.

src/x.c
stapelberg: In `con_set_fullscreen_mode` we already unconditionally replace this atom, so you don’t need to use `XCB_PROP_MODE_APPEND`.

src/x.c
Airblader: @stapelberg Mh, maybe that's not so easy afterall. It will bring back the problem that we need to update the cached state when containers are moved around and the like.

Airblader:
@stapelberg I've incorporated your comments.

src/x.c
stapelberg: Why? You can still run the code to figure out in which state the window should be, just not update it when it doesn’t need updating.

testcases/t/231-ipc-window-close.t
```diff
@@ -44,6 +44,7 @@ $i3->subscribe({
 my $window = open_window;
 
 cmd 'kill';
+sync_with_i3;
```
stapelberg: This seems like an unrelated bugfix. Can you send that in a separate PR?

src/x.c
Airblader: @stapelberg That I already am doing.

testcases/t/231-ipc-window-close.t
Airblader: The test reliably failed after this PR due do this, so I can't put it into a separate PR and still pass the build.

Airblader:
@stapelberg Not sure line comments create notifications, but I've done everything I think I can do. See the line comments for answers.

testcases/t/231-ipc-window-close.t
stapelberg: You could send it in another PR and then rebase this PR after the other one is merged.

src/x.c
stapelberg: I don’t follow. Currently, you’re querying X11 for the window state, I’m suggesting to add another field to `struct con_state` that you update unconditionally, but based on which you conditionally update X11 (only if it changed).

stapelberg:
FWIW, line comments do trigger email notifications.

src/x.c
Airblader: But I'll have to figure out all the places where I need to update that field, which would end up to be just as complicated as just doing the entire logic in those places (which is what we originally discussed in the ticket as something we can avoid by putting it here). For example, if a container is moved from a tabbed container (anywhere upwards in the tree), I have to change it. The same for if a container is moved anywhere into a tabbed container. Or am I overlooking something?

stapelberg: Why can’t you keep iterating over all windows like you do currently, compare the desired state (`con_is_hidden(con)`) with the last state (the to-be-introduced field) and then use `xcb_change_property`?

Airblader: Mh. I tried this earlier and it broke tests. Maybe I was too quick to judge what happened. I will try again with `con_state`. I agree that I don't see why it shouldn't work.

testcases/t/231-ipc-window-close.t
Airblader: OK.

Airblader:
@stapelberg Done. I've deleted the last commit that now has its own PR, so the build here might fail until I rebase.

src/x.c
```diff
@@ -36,6 +36,7 @@ typedef struct con_state {
     bool mapped;
     bool unmap_now;
     bool child_mapped;
+    bool is_hidden;
```
Airblader: How does the state behave during an inplace restart?

Airblader:
@stapelberg As explained in #1675, that PR doesn't seem to be needed anymore. So I currently adressed all of your feedback.

src/x.c
stapelberg: AFAIR, con_state are not preserved during an in-place restart, i.e. they behave as if you just opened the window. I don’t think that’s a problem here, is it?

src/x.c
```diff
@@ -612,6 +613,36 @@ void x_deco_recurse(Con *con) {
...
+static void set_hidden_state(Con *con) {
+    if (con->window == NULL) {
+        return;
+    }
+
+    con_state *state = state_for_frame(con->frame);
+
+    unsigned int num = 0;
+    uint32_t values[1];
+    if (con_is_hidden(con)) {
```
stapelberg: nit: If you introduce a `const bool hidden = con_is_hidden(con);`, you can get rid of one of the two returns, i.e. just use `if (hidden == state->is_hidden) return;`. Further, you can set the correct value at the bottom of the function without having to use `is_hidden = !is_hidden`.

stapelberg:
Can you rebase this please to resolve the merge conflict?

src/x.c
Airblader: Just making sure. It shouldn't be a problem, it will just trigger an unnecessary request the first time the container goes through here. That's okay, though.

Airblader:
@stapelberg Changed it according to your feedback and rebased.

PR #1666, opened 2015-04-19
Fix ‘FALSE’ undeclared (first use in this function) when i3 is built w/o PANGO

Xarthisius:
When trying to build without PANGO support you get:
```
../i3/src/handlers.c: In function ‘handle_client_message’:
../i3/src/handlers.c:883:53: error: ‘FALSE’ undeclared (first use in this function)
                 floating_resize_window(con->parent, FALSE, &fake);
                                                     ^
../i3/src/handlers.c:883:53: note: each undeclared identifier is reported only once for each function it appears in
```

See downstream [bug](https://bugs.gentoo.org/show_bug.cgi?id=546444) for more info.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -880,7 +880,7 @@ static void handle_client_message(xcb_client_message_event_t *event) {
                 floating_drag_window(con->parent, &fake);
                 break;
             case _NET_WM_MOVERESIZE_SIZE_TOPLEFT... _NET_WM_MOVERESIZE_SIZE_LEFT:
-                floating_resize_window(con->parent, FALSE, &fake);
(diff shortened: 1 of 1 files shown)


Airblader:
Why do we even use `FALSE` at all there?

stapelberg:
What @Airblader said. Please use `false`.

Airblader:
@stapelberg I'll just ping you since @Xarthisius didn't. :)

stapelberg:
@Xarthisius Can you please squash your commits so that we have a clean history?

Also, please trigger an email notification once you’ve done that.

Xarthisius:
@stapelberg done

PR #1669, opened 2015-04-19
Documentation Improvement – Debug i3bar

Airblader:


docs/debugging
```diff
@@ -10,10 +10,10 @@ Thank you for being interested in debugging i3. It really means
 something to us to get your bug fixed. If you have any questions about the
 process and/or need further help, do not hesitate to contact us!
 
-== Verify you are using i3 ≥ 4.7
+== Verify you are using the latest version
```
stapelberg: I’d rather specify a version here, actually. “latest version” is not very precise. Can you revert this line and update release.sh to take care of it please?

Airblader: I'll revert this line entirely and open a separate PR for it since I need to look at that script first.

docs/debugging
```diff
@@ -121,3 +121,16 @@ When debugging with us in IRC, be prepared to use a so called nopaste service
 such as http://nopaste.info or http://pastebin.com because pasting large
 amounts of text in IRC sometimes leads to incomplete lines (servers have line
 length limitations) or flood kicks.
+
+== Debugging i3bar
+
+To debug i3bar problems, add +verbose yes+ to the +bar {}+ block in your i3 config
```
stapelberg: maybe s/the +bar {}+ block/all +bar {}+ blocks/?

Airblader:
@stapelberg Done

docs/debugging
```diff
@@ -121,3 +121,16 @@ When debugging with us in IRC, be prepared to use a so called nopaste service
...
 length limitations) or flood kicks.
+
+== Debugging i3bar
+
+To debug i3bar problems, add +verbose yes+ to all +bar {}+ blocks in your i3 config
+and then restart all i3bar instances like this:
+
+----------------------------------------------------------------------------
+$ killall i3bar
+$ i3bar --bar_id=$(i3-msg -t get_bar_config | cut -d\" -f 2) >i3bar.log 2>&1
```
stapelberg: Thinking about it, this will only work when there is precisely one bar block. Can you extend this to be a for loop over the i3-msg output, using separate log files?

Airblader: @stapelberg Parsing this in the commandline without node or anything isn't easy, but this should work.

Airblader:
@stapelberg Done.

docs/debugging
```diff
@@ -121,3 +121,18 @@ When debugging with us in IRC, be prepared to use a so called nopaste service
...
 length limitations) or flood kicks.
+
+== Debugging i3bar
+
+To debug i3bar problems, add +verbose yes+ to all +bar {}+ blocks in your i3 config
+and then restart all i3bar instances like this:
+
+---------------------------------------------------------------------------
+$ killall i3bar
+$ for c in $(i3-msg -t get_bar_config | grep -oP '(?<=")\w.*?(?=")'); do \
```
stapelberg: Hm, the -P parameter to grep is described as highly experimental, and e.g. FreeBSD doesn’t support it: https://www.freebsd.org/cgi/man.cgi?query=grep

How about using this python oneliner instead? Python is widely available and comes with a JSON module in its standard library (for Perl, that’s only the case for perl ≥ v5.13.9 AFAICT):

```bash
i3-msg -t get_bar_config | python -c 'import json,sys;print "\n".join(json.load(sys.stdin))'
```

Airblader: And here I spent half an hour coming up with this because I didn't think you'd want me to use Python :cry: I'll change it.

Airblader:
@stapelberg Updated.

docs/debugging
```diff
@@ -121,3 +121,18 @@ When debugging with us in IRC, be prepared to use a so called nopaste service
...
 length limitations) or flood kicks.
+
+== Debugging i3bar
+
+To debug i3bar problems, add +verbose yes+ to all +bar {}+ blocks in your i3 config
+and then restart all i3bar instances like this:
+
+-----------------------------------------------------------------------------------------------------------------
+$ killall i3bar
+$ for c in $(i3-msg -t get_bar_config | python -c 'import json,sys;print("\n".join(json.load(sys.stdin)))'); do \
```
stapelberg: Can you add another line break after the -c so that the command is easier to handle on average-sized displays?

Airblader:
@stapelberg Done (sorry, forgot the ping)

Issue #1661, opened 2015-04-15
Mouse bindings problem in FreeBSD 10.1


kdbill:
In FreeBSD 10.1-RELEASE-p9 the binding
    `bindsym --release button2 kill`
seems to have no effect. In Arch Linux this works flawlessly.

i3 version:
i3 version 4.10.1 (2015-03-29, branch "4.10.1") © 2009-2014 Michael Stapelberg and contributors

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
@acrisci Do you have an idea what could be going on here?

acrisci:
I'm not very familiar with FreeBSD and I don't have a machine set up to test, so I'm going to need to look at a log to see if I can determine if the issue is with the code or not.

If anyone else confirms the bug, I'll set up a machine and look into it further.

kdbill:
log from minimal X session:

startx
open terminal window
click twice with middle mouse button over terminal window titlebar
close X session

http://logs.i3wm.org/logs/5647648723828736.bz2

I could not get the instructions from http://i3wm.org/docs/debugging.html for obtaining the debug logfile to work. When executing i3 from `~/.xinitrc` with
`exec i3 --shmlog-size=26214400`
i got from `i3-dump-log` the error:
`i3-dump-log: Could not shm_open SHM segment for the i3 log (/tmp/i3-log-2393): No such file or directory`

I finally got the logfile by executing i3 from `~/.xinitrc` with
`exec i3 -V -d all > ~/.i3/logfile`

kdbill:
Sorry, by accident i temporarily closed this issue, this was not intentional. Just started using github :)

acrisci:
Based on this log, I can tell there is very likely a bug in i3. I'll set up a FreeBSD machine and look into it.

PR #1672, opened 2015-04-20
bugfix: serialize con_id with %p in run_binding()

### LINKED ISSUE #1668: Window title changes should trigger window swallowing (layout saving)
@teto reports that i3 doesn’t swallow urxvt with the following layout JSON file:

``` json
// vim:ts=4:sw=4:et
{
    "border": "pixel",
    "current_border_width": 1,
    "floating": "auto_off",
    "geometry": {
       "height": 388,
       "width": 734,
       "x": 0,
       "y": 0
    },
    "name": "/home/teto",
    "percent": 1,
    "swallows": [
       {
        "class": "^URxvt$",
       // "instance": "^rxvt\\-unicode$",
        "title": "^/home/teto/dce$"                                                                      
       // "transient_for": "^$"
       }
    ],
    "type": "con"
}
```

@teto starts urxvt like this:

``` bash
#!/bin/sh

i3-msg "workspace 9; append_layout $HOME/.i3/test.json"                                                  

urxvt -cd $HOME/dce
```

Having a look at the logfile reveals that urxvt first starts up with a different title and later changes its `_NET_WM_NAME` property to set the correct title.

i3 should, once it got the new title, check whether the window can be matched by any of the placeholders and swallow it if so.

This affects not only window titles, but all properties that can be used to swallow a window.

acrisci:
When a con is given to run_binding() to run the command in the context
of this con such as when running a mouse binding, serialize the con_id
with %p. This seems to be more consistent across Unixes.

fixes #1668

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -431,7 +431,7 @@ CommandResult *run_binding(Binding *bind, Con *con) {
     if (con == NULL)
         command = sstrdup(bind->command);
     else
-        sasprintf(&command, "[con_id=\"%d\"] %s", con, bind->command);
+        sasprintf(&command, "[con_id=\"%p\"] %s", con, bind->command);
 
     Binding *bind_cp = binding_copy(bind);
     CommandResult *result = parse_command(command, NULL);


stapelberg:
Can you elaborate on how that fixes #1668 please?

Airblader:
And if it does, we might want a testcase for it. :)

acrisci:
It actually fixes #1661 :squirrel:

stapelberg:
@acrisci Okay, but why? I.e., can you provide the output before/after on FreeBSD so that I can better understand what’s going on?

acrisci:
From the log on the issue, here is the click that is supposed to kill the window:

>
04/16/2015 23:22:42 - click.c:route_click:175 - --> click properties: mod = 0, destination = 1
04/16/2015 23:22:42 - click.c:route_click:176 - --> **OUTCOME = 0x80d119f80**
04/16/2015 23:22:42 - click.c:route_click:177 - type = 2, name = (null)
04/16/2015 23:22:42 - ../i3-4.10.1/src/bindings.c:get_binding_from_xcb_event:208 - (removed numlock, state = 512)
04/16/2015 23:22:42 - ../i3-4.10.1/src/bindings.c:get_binding_from_xcb_event:212 - (removed upper 8 bits, state = 0)
04/16/2015 23:22:42 - ../i3-4.10.1/src/bindings.c:get_binding_from_xcb_event:217 - (checked mode_switch, state 0)
04/16/2015 23:22:42 - commands_parser.c:parse_command:271 - **COMMAND: *[con_id="219258752"] kill***
>

Emphasis is on the con that should be killed and then the command that should kill the con.

Then a few lines later, we find that it does not match the con:

>
04/16/2015 23:22:42 - commands.c:cmd_criteria_match_windows:284 - checking if con 0x80d119f80 / (null) matches
04/16/2015 23:22:42 - commands.c:cmd_criteria_match_windows:290 - doesnt match
>

With the change, the command would be serialized like this:

    [con_id="0x80d119f80"] kill

which is the correct ID for the con in base 16.

The value will be parsed here: https://github.com/i3/i3/blob/next/src/commands.c#L338 with strtol.

stapelberg:
@acrisci Thanks, that makes it more clear. Note that the problem is not whether the ID is represented as base10 or base16, but rather the fact that %d seems to be equivalent to %ld on Linux when necessary, whereas it isn’t on FreeBSD. I.e., in your example, 219258752 is equivalent to 0xD119F80, which is the correct value, but with the most significant bits cut off.

https://www.freebsd.org/cgi/man.cgi?printf%283%29 says that %p is equivalent to %x or %lx, so that’s why it fixes the issue.

Can you please update your commit description to state this instead of the rather vague argument about consistency across Unices?

acrisci:
That is a good analysis.

I think I understand what is going on, but this is something I'm a little weak on. Can you please suggest a new commit message for me to review and resubmit?

An interesting question would be if this will work on processor architectures where pointers are not the same size as long integers.

stapelberg:
I suggest:

> Bugfix: serialize con_id with %p in run_binding()
>
> %p is equivalent to either %x or %lx, depending on the pointer size of the platform.
> Before this commit, we always used %d, which has the same behavior on Linux,
> but is not automatically expanded to %ld on e.g. FreeBSD.

acrisci:
Ok made that change.

PR #1675, opened 2015-04-20
Small test fix

Airblader:


testcases/t/231-ipc-window-close.t
```diff
@@ -44,6 +44,7 @@ $i3->subscribe({
 my $window = open_window;
 
 cmd 'kill';
+sync_with_i3;
```
stapelberg: I’m not convinced this is necessary here. I get that you were saying you need to change the test because otherwise it repeatedly fails with your other pull request, but why add `sync_with_i3`? The test in question doesn’t use X11 directly, so there should not be a need to synchronize the two connections (see http://i3wm.org/docs/testsuite.html#i3_sync for details).

Can you explain why `sync_with_i3` is necessary or dig a bit deeper for a different fix?

Airblader:
@stapelberg I think this might have been a problem with the first version of my pull request. Both Travis and a local run are green now, while before both Travis and the local run were reliably failing. Hence I'll close this PR as it doesn't seem to be needed anymore.

PR #1679, opened 2015-04-23
Focus containers by number

Airblader:
I'd like to get input first. As mentioned in the issue:
* Do we want both "concepts" (local, global)? The global concept can be emulated by focusing the workspace first, but the extra implementation effort is almost none and I think it's a valid usecase.
* I'm not happy with the current syntax. For the "local" mode, I could imagine using `focus container <n>`, but I'd a) like feedback on this and b) still need a good way for the "global" concept (if we want to keep it).

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1428,6 +1428,43 @@ void cmd_focus_level(I3_CMD, char *level) {
 }
 
 /*
+ * Implementation of 'focus local|global <n>'
+ *
+ */
+void cmd_focus_number(I3_CMD, char *context, char *_n) {
+    DLOG("Focusing %s-th container on %s level\n", _n, context);
+
+    /* parse into a number. */
+    char *endptr = NULL;
+    long n = strtol(_n, &endptr, 10);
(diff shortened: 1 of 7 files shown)


src/commands.c
```diff
@@ -1428,6 +1428,43 @@ void cmd_focus_level(I3_CMD, char *level) {
...
+    char *endptr = NULL;
+    long n = strtol(_n, &endptr, 10);
+    if (n == LONG_MIN || n == LONG_MAX || n <= 0 || endptr == _n) {
+        ELOG("failed to parse \"%s\" into a number", _n);
+        yerror("failed to parse \"%s\" into a number", _n);
+        return;
+    }
+
+    Con *con;
+    if (strcmp(context, "global") == 0 || focused->type == CT_WORKSPACE)
```
Airblader: The second check should probably actually be moved to the `else` by changing it to `else if (focused->type != CT_WORKSPACE)`

Airblader:
I'll fix the formatting next time I push.

Airblader:
I'll close this PR for now as it's very unlikely that it gets merged this way and we will have to start a discussion on this first (see discussion on the linked issue).

Issue #1673, opened 2015-04-20
add/remove monitor bug


unera:
If You add/remove monitor You are forced to restart some applications because they will not work properly.

for example:

Skype: will popup notifies at the center of new monitor against right bottom corner
Goldendict: will popup translate-window outside monitor

etc

Skype and Goldendict and the other applications works properly with the other WMs

PS:

PS: my xrandr output:

```txt

Screen 0: minimum 320 x 200, current 2560 x 1848, maximum 8192 x 8192
eDP1 connected 1366x768+597+1080 (normal left inverted right x axis y axis) 260mm x 140mm
   1366x768      60.00*+
   1360x768      59.80    59.96  
   1024x768      60.00  
   800x600       60.32    56.25  
   640x480       59.94  
DP1 disconnected (normal left inverted right x axis y axis)
HDMI1 connected primary 2560x1080+0+0 (normal left inverted right x axis y axis) 677mm x 290mm
   2560x1080     60.00*+
   1920x1080     60.00    50.00    59.94    30.00    29.97  
   1920x1080i    60.00    50.00    59.94  
   1680x1050     59.88  
   1600x900      59.98  
   1280x1024     75.02    60.02  
   1152x864      75.00  
   1280x720      60.00    50.00    59.94  
   1024x768      75.08    60.00  
   800x600       75.00    60.32  
   720x576       50.00  
   720x480       60.00    59.94  
   640x480       75.00    60.00    59.94  
   720x400       70.08  
DP2 disconnected (normal left inverted right x axis y axis)
HDMI2 disconnected (normal left inverted right x axis y axis)
```

If I start skype before plug-in HDMI1, skype will show popups at the center HDMI1 (after plug-in)

If I start Goldendict before plug-in HDMI1, it will show popups out of screen (after plug-in)

I used all these programs with the other WM (I used fluxbox before i3wm) and these programs worked properly before and after plug-in any monitors.

Sorry for my English.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

unera:
```
i3 --version
i3 version 4.8 (2014-06-15, branch "4.8") © 2009-2014 Michael Stapelberg and contributors
```

i3bot:
[...]

PR #1680, opened 2015-04-23
Check if output is disabled in handle_output()

acrisci:
Check if the `connection` of the randr output is
XCB_RANDR_CONNECTION_DISONNECTED and disable the output if it is.

This fixes an issue where the output would not be disabled if the output was
physically unplugged from the machine.


acrisci:
This might fix #1673, but there might be another problem with that issue.

PR #1685, opened 2015-04-26
Initialize workspace rect to the output's upon creation

Deiz:
The comment immediately following implied that this was the intended behaviour. Not doing so means that compound commands that both move a window to a new workspace as well as do something that depends on the workspace's geometry (e.g. 'move position center' or 'floating enable' on a tiled window) would use the workspace's calloc'd 0x0+0x0 geometry.

The issue can be reproduced by running the following on a tiled window, assuming workspace 11 is unused:
`i3-msg -t command 'mark foo; [con_mark="foo"] move workspace 11, floating enable'`

### DIFF
--- a/testcases/t/244-new-workspace-floating-enable-center.t
+++ b/testcases/t/244-new-workspace-floating-enable-center.t
@@ -1,0 +1,53 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • http://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • http://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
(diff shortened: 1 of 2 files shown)


testcases/t/244-new-workspace-floating-enable-center.t
```diff
@@ -0,0 +1,51 @@
...
+# Open a tiled window, and then simultaneously move it to another
+# workspace and float it, ensuring that it ends up centered.
+#####################################################################
+
+my $window = open_window;
+my $unused = get_unused_workspace();
+
+cmd "mark foo; [con_mark=\"foo\"] move workspace $unused, floating enable";
+
+my $pos = $window->rect;
```
stapelberg: I think you need a `sync_with_i3` call here since you’re using the X11 and the i3 connection. See http://i3wm.org/docs/testsuite.html#i3_sync

Deiz:
Made the suggested change.

PR #1689, opened 2015-04-26
Debugging docs – minimal setup, small log file, closed source software

Airblader:
* Motivate users to come up with clear and minimal instructions on how to reproduce a problem before submitting an issue.
* Encourage users to restart i3 before reproducing the problem so that the log file can stay small and noise-free.
* Mention the non-support of closed source software.

### DIFF
--- a/docs/debugging
+++ b/docs/debugging
@@ -67,6 +67,30 @@ fly:
 i3-msg 'debuglog on; shmlog on; reload'
 ---------------------------------------
 
+== Reproducing the problem
+
+Before submitting an issue, please make sure to close down on the problem as
+much as you can yourself. Here are some steps you should consider:
+
+* Find a deterministic, reliable way to reproduce the problem and provide it
+  with your bug report.
+* Try using the default i3 config to reproduce the problem. If the issue does
+  not appear with the default config, gradually adapt it to track down what 
+  change(s) to the config introduce the problem.
+* Reproduce the problem with a minimal setup, i.e., only use as few applications,
+  windows and steps as necessary.
+* In addition, try to stick to applications that are common and, even more
+  importantly, free / open source.
+* Before obtaining the log file, restart i3 in-place, execute the steps to
+  reproduce the problem and then save the logs. This keeps the log file as
+  small as possible and necessary.
+
+Please be aware that we cannot support compatibility issues with closed-source
+software, as digging into compatibility problems without having access to the
+source code is too time-consuming. Additionally, experience has shown that
+often, the software in question is responsible for the issue. Please raise an
+issue with the software in question, not i3.
+
 == Obtaining the debug logfile
 
 No matter whether i3 misbehaved in some way without crashing or whether it just


docs/debugging
```diff
@@ -67,6 +67,21 @@ fly:
...
+Before submitting an issue, try to close down on the problem as much as you
+can by yourself. Ideally, find a way to reliably reproduce the problem
+with a minimal setup. This means with as few running applications as possible,
+only as many steps as necessary and, if possible, using only common and open
+source software.
+
+Also, make sure to restart i3 right before going through the procedure to
+reproduce the problem to keep the log file as small as possible.
+
+Note that we do not support open source software unless it can be proven that
```
Deiz: Looks like that 'open' should be 'closed'.

Airblader:
@Deiz Right you are, I've updated it.

docs/debugging
```diff
@@ -67,6 +67,21 @@ fly:
 i3-msg 'debuglog on; shmlog on; reload'
 ---------------------------------------
 
+== Reproducing the problem
+
+Before submitting an issue, try to close down on the problem as much as you
+can by yourself. Ideally, find a way to reliably reproduce the problem
+with a minimal setup. This means with as few running applications as possible,
+only as many steps as necessary and, if possible, using only common and open
+source software.
```
stapelberg: Can you add “use the default config or track down what needs to be changed in the default config to reproduce the issue”?

Airblader: @stapelberg I've added it, but took the liberty to rephrase the entire thing to make it into a list. I think it looks more readable that way.

docs/debugging
```diff
@@ -67,6 +67,29 @@ fly:
...
 
+== Reproducing the problem
+
+Before submitting an issue, please make sure to close down on the problem as
+much as you can yourself. Here are some steps you should consider:
+
+* Find a deterministic, reliable way to reproduce the problem and provide it
+  with your bug report.
+* Try using the default i3 config to reproduce the problem. If the issue does
+  not appear in this case, gradually adapt it to track down what change(s) to
```
stapelberg: s/in this case/with the default config/

docs/debugging
```diff
@@ -67,6 +67,29 @@ fly:
...
+  the config introduce the problem.
+* Reproduce the problem with a setup as minimal as possible, e.g., only use
+  as few applications, windows and steps as necessary.
+* In addition, try to stick to applications that are common and, even more
+  importantly, free / open source.
+* Before obtaining the log file, restart i3 in-place, execute the steps to
+  reproduce the problem and then save the logs. This keeps the log file as
+  small as possible and necessary.
+
+Please note that we do not support closed-source software unless it can be
```
stapelberg: Can you just copy the paragraph out of CONTRIBUTING.md instead of adding another phrasing of the same thing? :)

Airblader:
@stapelberg Done

docs/debugging
```diff
@@ -67,6 +67,30 @@ fly:
...
 
+== Reproducing the problem
+
+Before submitting an issue, please make sure to close down on the problem as
+much as you can yourself. Here are some steps you should consider:
+
+* Find a deterministic, reliable way to reproduce the problem and provide it
+  with your bug report.
+* Try using the default i3 config to reproduce the problem. If the issue does
+  not appear with the defualt config, gradually adapt it to track down what
```
stapelberg: s/defualt/default/

docs/debugging
```diff
@@ -67,6 +67,30 @@ fly:
...
+
+Before submitting an issue, please make sure to close down on the problem as
+much as you can yourself. Here are some steps you should consider:
+
+* Find a deterministic, reliable way to reproduce the problem and provide it
+  with your bug report.
+* Try using the default i3 config to reproduce the problem. If the issue does
+  not appear with the defualt config, gradually adapt it to track down what 
+  change(s) to the config introduce the problem.
+* Reproduce the problem with a setup as minimal as possible, e.g., only use
```
stapelberg: I think “as minimal as possible” doesn’t make a lot of sense. Either a setup is minimal or it isn’t :). Maybe just use “Reproduce the problem with the minimal setup, i.e.” (also note the i.e. at the end instead of e.g.)

Airblader:
@stapelberg Done. The typo was, of course, just a test for you :)

PR #1693, opened 2015-04-30
Don't duplicate property value on class change.

mh21:
- replace smalloc/memcpy/strlen/sstrdup by strnlen/strndup

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -27,30 +27,27 @@ void window_update_class(i3Window *win, xcb_get_property_reply_t *prop, bool bef
      * null-terminated strings (for compatibility reasons). Instead, we
      * use strdup() on both strings */
     const size_t prop_length = xcb_get_property_value_length(prop);
-    char *new_class = smalloc(prop_length + 1);
-    memcpy(new_class, xcb_get_property_value(prop), prop_length);
-    new_class[prop_length] = '\0';
+    char *new_class = xcb_get_property_value(prop);
+    const size_t class_class_index = strnlen(new_class, prop_length) + 1;
 
     FREE(win->class_instance);
     FREE(win->class_class);
 
-    win->class_instance = sstrdup(new_class);
-    if ((strlen(new_class) + 1) < prop_length)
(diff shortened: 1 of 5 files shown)


src/window.c
```diff
@@ -27,30 +27,28 @@ void window_update_class(i3Window *win, xcb_get_property_reply_t *prop, bool bef
      * null-terminated strings (for compatibility reasons). Instead, we
      * use strdup() on both strings */
     const size_t prop_length = xcb_get_property_value_length(prop);
-    char *new_class = smalloc(prop_length + 1);
-    memcpy(new_class, xcb_get_property_value(prop), prop_length);
-    new_class[prop_length] = '\0';
+    char *new_class = xcb_get_property_value(prop);
+    /* length including the \0 at the end, might be longer than prop_length */
```
stapelberg: If `instance_size` is longer than `prop_length`, your call to `strndup` in line 39 will fail because you’re passing a negative size.

Also note that `strnlen(new_class, prop_length)` will always be less than or equal to `prop_length`. Maybe this comment is just worded in such a way that it’s easy to misunderstand?

src/window.c
```diff
@@ -27,30 +27,28 @@ void window_update_class(i3Window *win, xcb_get_property_reply_t *prop, bool bef
...
+    const size_t instance_size = strnlen(new_class, prop_length) + 1;
 
     FREE(win->class_instance);
     FREE(win->class_class);
 
-    win->class_instance = sstrdup(new_class);
-    if ((strlen(new_class) + 1) < prop_length)
-        win->class_class = sstrdup(new_class + strlen(new_class) + 1);
+    win->class_instance = strndup(new_class, prop_length);
```
stapelberg: `sstrdup` checks for `strdup` returning `NULL`, but your direct call to `strndup` doesn’t. Please introduce `sstrndup` and also switch the two other callsites in our code base to use `sstrndup`.

mh21:
- introduced a secure `strndup` wrapper
- renamed variable `instance_size` to `class_class_index`

PR #1697, opened 2015-05-02
Extend mouse commands on i3bar

### LINKED ISSUE #1695: [enhancement] Control i3bar click events
Currently, we already can change the behavior of scroll events via `wheel_up_cmd` and `wheel_down_cmd`. This should also work for click events, e.g., to allow disabling clicks on workspace buttons.

Airblader:
fixes #1695


acrisci:
I like the idea of this change. I would personally use it to open a window switcher (skippy-xd) when I right click on the bar. This approach is superior to having a separate command for each button for complicated mice with lots of buttons.

I think this change should also include the deprecation of the old `wheel_up_cmd` and `wheel_down_cmd` commands. By this I mean it should be removed from the documentation, and maybe issuing a warning during validation stating the command may be removed in a future release.

As for the command itself, what about using syntax like `bindsym button1 cmd` in the bar block?

Airblader:
> As for the command itself, what about using syntax like bindsym button1 cmd in the bar block?

I avoided that to have a semantic clarification that we don't support anything other than mouse clicks here. I also very much doubt that this will change in the future.

As for the deprecation – if it were up to me, I'd remove it altogether :) So maybe @stapelberg can just tell me how far we actually want to go with deprecation here and I'll do it. I already *pretty much* removed it from the docs, but left in a small bit for those who already use it and want to look it up.

I like the idea of issuing a warning. I'll add that.

Airblader:
@acrisci I've added the warning

teto:
Also like that. thanks @acrisci for the discovery of skippy-xd.

Airblader:
Note that a command is for "all" workspace buttons, so a user can't execute button-specific commands upon clicks. If someone has a need for this, we can always add that in a separate PR.

docs/userguide
```diff
@@ -1182,21 +1182,26 @@ Available modifiers are Mod1-Mod5, Shift, Control (see +xmodmap(1)+).
...
+or running scripts that implement custom behavior for these buttons.
+
+Please note that +wheel_up_cmd+ and +wheel_down_cmd+ are deprecated and will be
+removed in a future release. We strongly recommend using the more general +mouse_cmd+
+with +button4+ and +button5+ instead.
 
 *Syntax*:
----------------------
+---------------------------
+mouse_cmd <button> <command>
```
stapelberg: Why is this mouse_cmd and not bindsym, like for other mouse bindings?

Airblader: See the conversation between Tony and me on this PR. I agree that it's more consistent, but also suggests that it's the same command, which isn't true, as it will only wok with mouse buttons and nothing else. But I don't care for the keyword too much, so if you want me to change it to bindsym, let me know and I'll change it.

stapelberg: We also have `exec` which is both a command a configuration directive, so this wouldn’t be the first time :). With regards to non-mouse-button bindings, we should emit a warning.

I think having similar features have similar/the same names makes it easier for people to start using them because of familiarity. From my current POV, that overweighs the possible confusion.

Airblader: @stapelberg OK, I've renamed the command.

docs/userguide
```diff
@@ -1182,23 +1182,28 @@ Available modifiers are Mod1-Mod5, Shift, Control (see +xmodmap(1)+).
...
+or running scripts that implement custom behavior for these buttons.
+
+Please note that +wheel_up_cmd+ and +wheel_down_cmd+ are deprecated and will be
+removed in a future release. We strongly recommend using the more general +bindsym+
+with +button4+ and +button5+ instead.
 
 *Syntax*:
----------------------
+---------------------------
+bindsym <button> <command>
```
acrisci: Should read `bindsym button<n> <command>`.

docs/userguide
```diff
@@ -1182,23 +1182,28 @@ Available modifiers are Mod1-Mod5, Shift, Control (see +xmodmap(1)+).
...
 wheel_up_cmd <command>
 wheel_down_cmd <command>
```
acrisci: I think these two commands should be removed from the docs here.

docs/userguide
Airblader: Right, thanks. I've changed it. I also removed the two lines for the deprecated syntax.

src/ipc.c
```diff
@@ -469,6 +469,19 @@ void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     y(map_close);
 }
 
+static void dump_mouse_commands(yajl_gen gen, Barconfig *config) {
```
acrisci: There's going to be a problem extending this. How will modifiers and a flag such as `--release` be represented?

Airblader: Is the barconfig request really something we consider exposed API and that we need to keep compatible? If so, then yes, we need to serialize this in a more extensible way. But if not (which is what I assumed), then this can be changed when we actually need to extend it, but there wouldn't be any need to make it more complex than it has to be now.

acrisci: Yes, we do consider this exposed API that needs to be kept compatible. For instance, it is used in ipc libraries. If the library is implemented in a strongly-typed language (such as i3ipc-glib), this can cause problems when the api is eventually updated. And I do think that there is a strong possibility that the extended features could be requested and also that someone might start to rely on the api in a stand-alone i3bar.

Airblader: OK. I'll turn this into a more extensible format then. Maybe something like

    "bindings": {
        "button1": {
            "command": "..."
        }
    }

This would allow adding a `flags` section or similar later on. What do you think?

acrisci: I think a nice solution might be to try to stay close to the binding event

http://i3wm.org/docs/ipc.html#_binding_event

so that this type could potentially be reused either in the i3bar code or in an ipc library implementing the api.

Airblader: OK, that makes sense. So how about this?

    "bindings": [
        {
            "command": "nop",
            "input_code": 1
        }
    ]

This would be both syntactically and semantically compatible. I'd leave out all the keys we currently don't need for now, though.

Airblader:
@stapelberg @acrisci I've adapted the format and changed the directive to bindsym.

acrisci:
I think this is testable.

Airblader:
In theory, yes. I'd have to fake a button press event, though. I don't think X11::XCB can do this yet, can it? So I'd have to put together an event myself and send it, but that would also involve figuring out i3bar's window id and so on.

If X11::XCB can do this already, it'd be great if someone could tell me how. Otherwise I'd appreciate on writing such a test with custom assembly of the event.

acrisci:
Try using xdotool.

Airblader:
@acrisci Thanks. I didn't know we had tests using xdotool before. I've added testcases for all five buttons now.

Airblader:
I'll look into the failing test, of course.

Airblader:
I forgot the skip command. However, I think it makes more sense to install xdotool on Travis so that Travis actually runs *all* tools.

Airblader:
@acrisci @stapelberg OK, I rebased and fixed the skip command (and also used command instead of which).

src/config_directives.c
```diff
@@ -530,14 +530,44 @@ CFGFUN(bar_modifier, const char *modifier) {
         current_bar.modifier = M_SHIFT;
 }
 
+static void bar_configure_binding(const char *button, const char *command) {
+    if (strncasecmp(button, "button", sizeof("button") - 1) != 0) {
```
stapelberg: Here and below, instead of using `sizeof("button")-1`, why not use `strlen("button")` directly? Modern compilers turn that into a constant at compile-time.

i3bar/src/config.c
```diff
@@ -112,20 +144,6 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
-    if (!strcmp(cur_key, "wheel_up_cmd")) {
```
stapelberg: I think it’d be better to leave this code in (and possibly add a TODO to remove it in a future version, I think when also removing the `wheel_up_cmd` and `wheel_down_cmd` directives themselves). The reason is that if you move it out, this will break user configs for people who are restarting i3bar before restarting i3, which might happen for a variety of reasons. Given that it’s just 8 lines of code which can easily be changed to fill the new bindings data structure, I prefer having it more robust.

stapelberg:
Aside from my comments, this looks good.

i3bar/src/config.c
Airblader: Is there any good way to check if TAILQ_INIT has been called already? Because that's sort of a problem, I can' just call it when receiving wheel_[up|down]_cmd as that would mean only the latter works. Is this really so bad? It will just log an error and is fixed with a simple restart (which IMHO makes a lot of sense when updating a software).

Airblader:
@stapelberg I've changed according to your second comment. For the other one, see my reply.

i3bar/src/config.c
stapelberg: You can check if `(head)->tqh_last == &(head)->tqh_first`, which will be a precise check for memory that is allocated using `calloc`, and will work most of the time for arbitrarily initialized memory.

If that doesn’t help, would it be possible to call TAILQ_INIT in a different place?

Airblader: I'll just call the init in the function that starts the JSON parsing.

Airblader:
@stapelberg I've moved the TAILQ_INIT call now. Locally I did this:

* run i3 before this fix and old-syntax commands in the config
* merge the patch and compile, but only restart i3bar
* adapt one of the two config directives to use the new syntax and restart i3 + i3bar
* completely switch to the new syntax and restart i3 + i3bar

All of this worked fine for me.

docs/userguide
```diff
@@ -1209,23 +1209,27 @@ Available modifiers are Mod1-Mod5, Shift, Control (see +xmodmap(1)+).
...
 Specifies a command to run when a button was pressed on i3bar to override the
-default behavior. Currently only the mouse wheel buttons are supported. This is
-useful for disabling the scroll wheel action or running scripts that implement
-custom behavior for these buttons.
+default behavior. This is useful, e.g., for disabling the scroll wheel action
+or running scripts that implement custom behavior for these buttons.
+
+Please note that the old +wheel_up_cmd+ and +wheel_down_cmd+ commands are deprecated
+and will be removed in a future release. We strongly recommend using the more general
++bindsym+ with +button4+ and +button5+ instead.
```
stapelberg: Can you add an explanation of which <n> are available, i.e. which mouse button is button1, 2, 3, …

AFAIR, X11 doesn’t handle buttons > 5, but some mice come with a lot of buttons. Please clarify and add a note about that as well.

i3bar/src/config.c
```diff
@@ -112,17 +142,23 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
+    /* TODO Remove this after v5.0 */
```
stapelberg: Here and below, I’m not sure that makes sense. We don’t have a stated deprecation policy, even across major versions. Given that it’s really just a couple lines of code, maybe just remove the comment and live with the fact that there is code for backwards compatibility? :)

docs/userguide
Airblader: Using `xev`, I at least see ButtonPress / ButtonRelease events for buttons 8 and 9. I also just checked and a bindsym on button9 with the whole-window flag works both on the window and the window decoration, so I guess we support all.

I think that XCB is just missing constants for "only button 8", but the value we now use (any button = 0) is correctly interpreted by X11.

i3bar/src/config.c
Airblader: But i3 doesn't send these keys anymore now. We only need this for the "only restart i3bar first" scenario you described. If we remove this in any future release, the *only* usecase where something breaks is if a user updates from 4.10.2 (or before) directly to a version >= 5.1 *and* restarts i3bar but not i3.

I know we don't have a deprecation policy, but keeping dead code around for this usecase? Especially because the fix is a simple in-place restart of i3?

Airblader:
@stapelberg I've adapted the userguide. I've also adapted the comment to leave it as more of a general comment rather than a specific target. Is that OK (see my reply on the outdated diff)?

i3bar/src/config.c
```diff
@@ -112,17 +142,25 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
+    /* TODO This is for backwards compatibility with <= 4.10.2. It could be
+     * removed in a future version. */
```
stapelberg: > But i3 doesn't send these keys anymore now. We only need this for the "only restart i3bar first" scenario you described. If we remove this in any future release, the only usecase where something breaks is if a user updates from 4.10.2 (or before) directly to a version >= 5.1 and restarts i3bar but not i3.

Consider that people might be restarting i3bar as part of elaborate work flows with regards to multi-monitor / dock configuration. It’s not as unlikely as it might seem at first glance.

Also consider that there are users of Debian stable on their desktops, for whom an upgrade from ≤ 4.10.2 to a new major version would be a perfectly normal thing.

> I know we don't have a deprecation policy, but keeping dead code around for this usecase? Especially because the fix is a simple in-place restart of i3?

I don’t think it matters how hard the fix is, but rather how big the cost of maintaining this in the future is. For me, not breaking user setups is a great thing, and I take 5 lines of code that are unlikely to need changing over breakage at any time.

As a nit, your current wording “It could be removed in a future version” is too vague for my taste.

Regardless, I still think you should remove the part about removing the code entirely.

Airblader:
@stapelberg OK. I've adapted the wording to not mention removing it, but explain why it's there.

PR #1698, opened 2015-05-03
Added 'move position mouse'

### LINKED ISSUE #1696: [enhancement] Move window to mouse
I guess there's a good chance this will be rejected since it is about floating containers and mouse, both of which aren't core i3 features, but anyway:

We have a command to move a floating container to the center of the screen. However, many dialogs, which open floating, I'd rather automatically move to wherever my mouse is. This could be file dialogs, alerts or panels for volume control etc.

This could be done via `move position mouse` which should fit nicely into the `move [absolute] position center` syntax.

Airblader:
fixes #1696

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1637,15 +1637,17 @@ bindsym $mod+f fullscreen toggle
 bindsym $mod+t floating toggle
 --------------
 
-=== Focusing/Moving containers
+[[_focusing_moving_containers]]
 
-To change the focus, use the focus command: +focus left+, +focus right+, +focus
-down+ and +focus up+.
+=== Focusing containers
 
-There are a few special parameters you can use for the focus command:
+To change focus, you can use the +focus+ command. The following options are
+available:
 
+left|right|up|down::
+        Sets focus to the nearest container in the given direction.
 parent::
-	Sets focus to the +Parent Container+ of the current +Container+.
+	Sets focus to the parent container of the current container.
 child::
 	The opposite of +focus parent+, sets the focus to the last focused
 	child container.
@@ -1659,23 +1661,16 @@ output::
 	Followed by a direction or an output name, this will focus the
 	corresponding output.
 
-For moving, use +move left+, +move right+, +move down+ and +move up+.
-
 *Syntax*:
------------------------------------
-focus <left|right|down|up>
-focus <parent|child|floating|tiling|mode_toggle>
-focus output <<left|right|down|up>|output>
-move <left|right|down|up> [<px> px]
-move [absolute] position [[<px> px] [<px> px]|center]
------------------------------------
-
-Note that the amount of pixels you can specify for the +move+ command is only
-relevant for floating containers. The default amount is 10 pixels.
+----------------------------------------------
+focus left|right|down|up
+focus parent|child|floating|tiling|mode_toggle
+focus output left|right|up|down|<output>
+----------------------------------------------
 
 *Examples*:
-----------------------
-# Focus container on the left, bottom, top, right:
+-------------------------------------------------
+# Focus container on the left, bottom, top, right
 bindsym $mod+j focus left
 bindsym $mod+k focus down
 bindsym $mod+l focus up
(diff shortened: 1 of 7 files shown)


src/floating.c
```diff
@@ -426,6 +426,43 @@ void floating_center(Con *con, Rect rect) {
...
 
+/*
+ * Moves the given floating con to the current pointer position.
+ *
+ */
+void floating_move_to_pointer(Con *con) {
+    assert(con->type == CT_FLOATING_CON);
+
+    xcb_query_pointer_cookie_t pcookie = xcb_query_pointer(conn, root);
+    xcb_query_pointer_reply_t *reply = xcb_query_pointer_reply(conn, pcookie, NULL);
```
Airblader: Forgot a free call for this.

docs/userguide
```diff
@@ -1637,15 +1637,15 @@ bindsym $mod+f fullscreen toggle
 bindsym $mod+t floating toggle
 --------------
 
-=== Focusing/Moving containers
+=== Focusing containers
```
stapelberg: Please add `[[_focusing_moving_containers]]` above this line so that URLs to the userguide keep working.

docs/userguide
```diff
@@ -1659,23 +1659,16 @@ output::
...
-move <left|right|down|up> [<px> px]
-move [absolute] position [[<px> px] [<px> px]|center]
------------------------------------
-
-Note that the amount of pixels you can specify for the +move+ command is only
-relevant for floating containers. The default amount is 10 pixels.
+----------------------------------------------
+focus left|right|down|up
+focus parent|child|floating|tiling|mode_toggle
+focus output left|right|up|down|<output>
```
stapelberg: I agree that your suggestion makes more sense, but this makes it inconsistent with the rest of the file, such as the move command you describe below.

Could you do a refactoring of the entire userguide to follow this syntax in a separate PR please?

src/floating.c
```diff
@@ -426,6 +426,43 @@ void floating_center(Con *con, Rect rect) {
...
+    /* Correct target coordinates to be in-bounds. */
+    x = MAX(x, (int32_t)output->rect.x);
+    y = MAX(y, (int32_t)output->rect.y);
+    if (x + con->rect.width > output->rect.x + output->rect.width)
+        x = output->rect.x + output->rect.width - con->rect.width;
+    if (y + con->rect.height > output->rect.y + output->rect.height)
+        y = output->rect.y + output->rect.height - con->rect.height;
+
+    /* Update container's coordinates to position it correctly. */
+    Rect new_position = {
```
stapelberg: Why declare a variable for this at all? You can use `floating_reposition(con, (Rect){ x, y, con->rect.width, con->rect.height });`

stapelberg:
> I would've loved to test some more, especially for example having two outputs next to each other and making sure output boundaries are respected. It seems like fake-outputs doesn't play well with that, though.

What do you mean? I think it would be good to have a multi-monitor test for this feature (see the `-m` flag for testcases/new-test).

docs/userguide
Airblader: Sure, I can do that. I might not get around to it for a few days as I'm on vacation, but after that I will make it consistent.

Airblader:
@stapelberg I wrote such a testcase, but it seemed as if the fake outputs boundaries were simply not respected. I'm not sure why, but even putting in some sleeps and watching the test unfold proved that it was as if they didn't exist. When I test the feature on my machine with multiple outputs, it works fine, so it must be something about the test.

If you want, I can add another test in a separate PR and we can discuss there what isn't working and why not. I think it'd be okay to keep it as a separate PR, though.

stapelberg:
> If you want, I can add another test in a separate PR and we can discuss there what isn't working and why not. I think it'd be okay to keep it as a separate PR, though.

Alright, as long as you don’t forget it :).

Airblader:
I'll prepare it as soon as I tackled your other comments. :)

docs/userguide
Airblader: @stapelberg Had a few minutes and sent a separate PR for consistency of the docs.

Airblader:
@stapelberg I've incorporated your comments. I'll send the PR for the test (don't worry, I won't forget :smile: ). Maybe today or in the next few days, whenever I have half an hour.

Airblader:
@stapelberg I've found the problem with the test I was talking about. I didn't have the "i3 config file (v4)" hint in the test config, which causes the fake outputs to not be initialized correctly. *insert big sigh here*.

I've updated this PR with the new test as there's no need for a separate PR anymore.

PR #1700, opened 2015-05-07
Tray icon size cleanup.

mh21:
- icon_size instead of hardcoding font.height
- consistent spacing with tray_spacing_px instead of 2 or logical_px(2)
- the spacing is used on the top, bottom, right of the tray area and between icons

This change makes it possible to easily patch tray_spacing_px to zero so that 
- moving the mouse to the bottom or right of the screen and clicking activates the correct tray icons (i.e. obeying Fitt's law). This is already true for all other parts of the bar.
- icons are shown in their full 18x18px glory (at least on my machine)

If you think it useful, I could also add a config option for this in a separate PR.

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -60,6 +60,9 @@ xcb_connection_t *conn;
 /* The font we'll use */
 static i3Font font;
 
+/* Icon size (based on font size) */
+int icon_size;
+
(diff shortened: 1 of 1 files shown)


stapelberg:
I think having a config option for this would indeed be useful. Particularly Fitt’s law is a convincing argument :).

PR #1704, opened 2015-05-11
Detect base 16 in cmd criteria

acrisci:
Try to detect base 16 numbers given to `con_id` and `id` for command criteria.

This is necessary because mouse bindings now may be serialized as hex as of
5c32de4.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -335,7 +335,7 @@ void cmd_criteria_add(I3_CMD, char *ctype, char *cvalue) {
 
     if (strcmp(ctype, "con_id") == 0) {
         char *end;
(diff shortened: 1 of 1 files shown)


src/commands.c
```diff
@@ -335,7 +335,8 @@ void cmd_criteria_add(I3_CMD, char *ctype, char *cvalue) {
 
     if (strcmp(ctype, "con_id") == 0) {
         char *end;
-        long parsed = strtol(cvalue, &end, 10);
```
stapelberg: A simpler fix is to just replace `10` by `0` here, see `strtol(3)`. Can you do that please?

acrisci:
Ok made that change.

PR #1706, opened 2015-05-12
Made syntax of syntax descriptions consistent:

Airblader:
- <xyz> denotes that some string must be used which is not a fixed value (e.g., a command), but a variable string (text, a number, ...)
- [xyz] denotes that the parameter is optional
- abc|xyz denotes that either abc or xyz must be given


stapelberg:
Just one nit, otherwise looks good. Thanks a lot!

Airblader:
@stapelberg Updated!

PR #1709, opened 2015-05-17
Graph support in i3bar

wrzoski:
Hi,

I've spent a couple of days adding graph widget to i3bar. It works for my needs so i thought I share it with community. Certainly it needs some improvements and some design / architecture changes, so I kindly ask for review. I've changed i3bar protocol and i3 main configuration, but I did not break backward-compatibility.
Graphs in action
Graphs in action
# Short description of the change.
* i3 config
  * every bar can have any number of graph configurations defined like below:
    * graph_config - name of graph configuration, used to reference in i3bar protocol
    * min, max - minimal and maximal value presented on the graph
    * width - width in pixel of the graph
    * time_range - the 'amount' of time presented on the graph, aka virtual width of the graph.
    * colorTOP, colorMIDDLE, colorBOTTOM - three colors of the grandient used to display a graph
```
bar {
   ...
   graph {
     graph_config cpu
     min 0
     max 100
     width 128
     time_range 100000
     colorTOP #FF0000
     colorMIDDLE #FFFF00
     colorBOTTOM #00FF00
   }
   graph {
     graph_config mem
     min 0
     max 5000
     width 100
     time_range 100000
     colorTOP #00FF00
     colorMIDDLE #FF00FF
     colorBOTTOM #FFFFFF
   }
}
```
* i3bar protocol
  * extended the i3bar block with following attributes:
    * graph_config - this is a name of graph configuration used to display a graph
    * graph_value - value to be registered by the graph
    * graph_timestamp - moment in time of the event occurence
  * example blocks to feed a graph with  cpu cores usage and memory usage
```
{ { "full_text" : "Core-0:", "graph_config" : "cpu", "graph_value" : 33, "graph_timestamp" : 10000000 },
  { "full_text" : "Core-1:", "graph_config" : "cpu", "graph_value" : 16, "graph_timestamp" : 10000000 },
  { "full_text" : "Memory", "graph_config" : "mem", "graph_value" : 1000, "graph_timestamp" : 10000000 },
```

To make it more clear, here is what one needs to do in order to display a graph:
1. define a graph configuration in i3 main configuration for a given bar
2. include in i3status information a block that references a graph configuration
Sounds simple?

Some rules / restrictions:
* One block can display only one graph
* Graph configuration can be referenced any number of times, so it can be reused by many graphs
* graph configuration for a given graph cannot be changed (it can, but this will throw away all collected graph values/time-stamps)
* graph values/times-tamps are:
  * interpolated if the granularity of the graph is grater then time-stamps
  * overwritten  if the granularity of the graph is lower then time-stamp
* the graph value/time-stamp cannot be omitted (if done so, the graph will be destroyed internally)

Some future improvements:

* sending more value/time-stamp pairs in one block
* let omit value/time-stamp pairs in block so that it is easier to display a static graph
* more options to configure the look of the graph (line, dotted line, bar, etc)
* make the graph more general (i.e. display many data streams on one graph).

BR
Lukasz Wrzosek


Airblader:
There's quite a few things I'd want to point out but before getting into that I think we should clarify whether we want to have this in i3bar at all.

acrisci:
This is really interesting.

I doubt this will be accepted into the main project, but I'd love to see a fork of i3bar that accepts interesting features like this.

Airblader:
I would definitely consider merging (a modified version of) this into i3-gaps if it doesn't make it into i3 (assuming I am allowed to).

wrzoski:
I haven't seen i3-gaps until now. Wonderful idea. Feel free to merge it. I'm open for suggestions, criticism and ideas of improvements.

Airblader:
Thanks. I'll wait for now to see what happens here. Then I'll still need to think about it because I want to make sure my fork stays maintainable. So let's see what Michael says first.

stapelberg:
I don’t think we want to merge this, for the following reasons:

* Whereever possible, i3 prefers text/keyboard, hence adding a primarily graphical way of displaying information is inconsistent:
 * i3 doesn’t display window icons, but offers stacked mode, to be combined with meaningful window titles.
 * i3status doesn’t display wireless signal strength in e.g. “3 of 5 bars”, but as a text percentage such as 85%.
 * Displaying tray icons in i3bar is a notable exception because the tray icon protocol doesn’t offer us any flexibility on this — we have to display icons or we don’t have a tray at all.
* This seems like a feature which introduces a bunch of configuration options and is likely to require more in the future to satisfy personal tastes. Our configuration is already pretty complex, and I’d like to avoid adding new options.
* I realize that graphs are high-density displays of historical information, but i3status displays **current** information. This point is a bit hard to convey, but I don’t think graphs fit well into the i3status philosophy. See also http://i3wm.org/i3status/manpage.html#_what_about_memory_usage_or_cpu_frequency, especially the last 2 paragraphs.

In general, I think graphs about system behavior are an area where there already are plenty of solutions. As just one example, there is collectd with its various frontends. I don’t see how replicating this functionality into i3 makes it any better. i3 first and foremost is a window manager, not a system monitor.

I hope what I wrote makes sense in light of your use-case. Since you just described what your code does, I don’t really know what your use-case actually is :). Note that https://github.com/i3/i3/blob/next/CONTRIBUTING.md asks contributors to file a feature request before sending such contributions.

Airblader:
I think I agree that it's not necessary for i3bar to do any of this. It would be both more flexible and less complexity to allow i3bar to display images and render the image some other way (though as Michael explained that is still not really something for the main project). 

I think that's also what I'd prefer merging into i3-gaps. This PR just seems to diverge too much in an area that isn't covered by tests for me to feel comfortable merging it into my fork. Even now any change to i3bar like the recent refactorings are prone to introduce bugs.

wrzoski:
Your answers cover all the details I should have known before making a contribution, but in fact I wasn't going to contribute this in a first place. I've changed my mind later on, when I realized how much code I touched. I would say this is my use-case: lower the cost of  maintenance.

So now it seems reasonable to close this PR.

PS. As it comes to the consistency: support for mouse events in i3bar is against consistency too ;). And what is more, IMO consistency is not a real life use-case, isn't it?

coleifer:
> support for mouse events in i3bar is against consistency too

He's got you there bro.

mh21:
Just as a side note, something like this can be implemented with a program that shows the graphs in the tray area (shameless plug: e.g. [indicator-multiload](https://launchpad.net/indicator-multiload) with the --trayicon parameter) and support for non-square tray icons (a first ugly version can be found in https://github.com/mh21/i3/tree/non-square-icons).

This approach only requires minimal changes to i3 for the non-square icons. @stapelberg, would you consider merging something like this?

stapelberg:
@mh21 AFAIR, non-square tray icons violate the spec, don’t they?

Airblader:
@stapelberg Mh, I'm not familiar with the spec, but if [this](http://standards.freedesktop.org/systemtray-spec/systemtray-spec-latest.html) is the entire spec, then the only reference to size I can find is

> Tray icons may be assigned any size by the system tray, and should do their best to cope with any size effectively.

mh21:
No, as far as I know there is no requirement for square icons. Although that sadly does not prevent DE from enforcing them ;-) - an exception would be e.g. [stalonetray](http://stalonetray.sourceforge.net/manpage.html).

mh21:
Somewhat related, the Ubuntu application indicators (a.k.a. KDE status notifiers) which can be regarded as a dbus-based replacement for tray icons also do not enforce a square layout (at least as implemented in Unity).

stapelberg:
@mh21 Fair enough. I think it’s okay to merge support for non-square tray icons, then.

PR #1725, opened 2015-05-29
Allow window decoration to change focus

### LINKED ISSUE #1056: focus does not follow mouse over window titles
[Originally reported by modulistic@…]
(
**What did you do?**

```
+--------------------+-------------+
| xterm1         (A) | (B) xterm2  |
+--------------------+-------------+
| ...                | ...         |
| ...                | ...         |
| ...                | ...         |
| ...                | ...         |
| ...                +-------------+
| ...            (C) | (D) xterm3  |  
| ...                +-------------+
| ...                | ...         |
| ...                | ...         |
+--------------------+-------------+
```

* I have a horizontal container with a simple window on the left and a vertical container on the right; I have window titles enabled; see ascii-art above.

* Please note (A), (B)... etc are just points to explain myself, nothing visible on the screen; (A), (B) and (D) are points inside the window title. Point C is inside xterm1's workspace, left of xterm3's window title.

* When I move the mouse cursor across window borders I get what I feel weird focus-follows-mouse behavior.

**What did you see?**
 * CASE 1: having the focus on xterm2 and the mouse cursor on point (B), moving cursor from point (B) to point (A) in a straight line, focus **stays** on xterm2.
 * CASE 2: having the focus on xterm1 and the mouse cursor on point (C), moving the mouse in a horizontal straight line to point (D), the focus goes to the active window on the right vertical split, i.e. the focus moves to the vertical split but not to the window whose title bar in under the mouse cursor.

**What did you expect?**

* CASE 1: I would expect the focus to move to xterm1.
* CASE 2: I would expect the focus to move to xterm3, or to stay on xterm1 if the CASE 1 behavior is the intended behavior.

* I have to admit that perhaps strict focus-following-mouse on tabbed (or stacked!) containers when hovering over window titles might be a bit weird too. But I'm posting this anyway, "just in case".

Airblader:
~~Note: I still want to add more testcases that prove that tabbed and stacked containers still behave as before.~~

Additional tests have been added now.

fixes #1056

### DIFF
--- a/testcases/t/246-window-decoration-focus.t
+++ b/testcases/t/246-window-decoration-focus.t
@@ -1,0 +1,148 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • http://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • http://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • http://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Ensure that hovering over the window decoration of a window causes it to focus
+# correctly.
+# Ticket: #1056
+# Bug still in: 4.10.2-174-g8029ff0
+use i3test;
+
+my ($ws, $A, $B, $C, $target, $y);
+my @cons;
+
+# ==================================================================================
+# Given the following layout (= denotes the window decoration),
+# when moving the mouse from 1 to 2,
+# then the C should be focused.
+#
+# This should especially be the case if B is the focus head of its vertically split parent container.
+#
+# +===A===+===B===+
+# |       |       |
+# |     1 +=2=C===+
+# |       |       |
+# +-------+-------+
+#
+# ==================================================================================
+
+$ws = fresh_workspace;
+
+open_window;
+$B = open_window;
+cmd 'split v';
+open_window;
+$target = get_focused($ws);
+
+@cons = @{get_ws($ws)->{nodes}};
+$A = $cons[0];
+$C = $cons[1]->{nodes}[1];
+
+$y = $C->{rect}->{y} - 0.5 * $C->{deco_rect}->{height};
+
+# make sure that B is the focus head of its parent
+cmd '[id="' . $B->{id} . '"] focus';
+
+# move pointer to position 1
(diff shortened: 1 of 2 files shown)


testcases/t/246-window-title-focus.t
```diff
@@ -0,0 +1,64 @@
...
+open_window;
+$B = open_window;
+cmd 'split v';
+open_window;
+
+@cons = @{get_ws($ws)->{nodes}};
+$A = $cons[0];
+$C = $cons[1]->{nodes}[1];
+
+$y = $C->{rect}->{y} - 0.5 * $C->{deco_rect}->{height};
```
Airblader: @stapelberg This surprised me a little. Shouldn't "rect" include the deco_rect part of the container? Why does rect->y point to the y position of the actual client window (at least in this case of a v-split parent)? Isn't that what window_rect is for?

testcases/t/246-window-decoration-focus.t
```diff
@@ -0,0 +1,64 @@
...
+open_window;
+$B = open_window;
+cmd 'split v';
+open_window;
+
+@cons = @{get_ws($ws)->{nodes}};
+$A = $cons[0];
+$C = $cons[1]->{nodes}[1];
+
+$y = $C->{rect}->{y} - 0.5 * $C->{deco_rect}->{height};
```
Airblader: @stapelberg This surprised me a little. Shouldn't "rect" include the deco_rect part of the container? Why does rect->y point to the y position of the actual client window (at least in this case of a v-split parent)? Isn't that what window_rect is for?

Airblader:
@stapelberg If we have a layout `H[term V[term T[term term]]]` and move from position C to D as described in #1056, do we want this to focus the focus head of the tabbed container or the tabbed container itself?

Airblader:
I've also tested this with more complex layouts (I didn't consider all of these to be worthy testing since they are the same, I just wanted to see how different it feels). I think this is really a major improvement to the usability. This bug literally lurks in every corner and I'm surprised it hasn't been reported over and over again. I myself would've noticed for sure if I wasn't using i3 without window decorations. ;)

testcases/t/246-window-decoration-focus.t
stapelberg: The window title is always drawn on the parent’s pixmap, which is why `deco_rect` is separate. `window_rect` is a position within `rect`, inset by the number of pixels determined by the i3 border style and by the X11 border which the application determines.

stapelberg:
I’m guessing plenty of people (me included) use `workspace_layout` with stacked/tabbed, in which case they aren’t affected by this bug :).

testcases/t/246-window-decoration-focus.t
Airblader: OK, thanks for the explanation. I always assumed the rect contained the decoration (despite on which pixmap it is drawn, which I knew about). Good to know!

Issue #1727, opened 2015-05-30
[Feature request] size floating windows to specific size


rr-:
Right now we can only grow or shrink window sizes. It would make sense to resize a window exactly to a given size. This would prove useful especially in `for_window` statements.

I saw some people doing this by computing a difference between actual window size and desired window size, and then sending resize requests using `i3-msg`. But not only was this cumbersome, it wouldn't resize the window properly if the difference was too small. To work around this, someone resized the window to 10000x10000 and then downsized it to the desired size (*facepalm*). I've settled for using `xdotool windowsize`, but...

1. That's another dependency
2. Frankly, giving users option to resize window to a given size is more reasonable than offering only to shrink or enlarge the windows. The latter could be achieved using 3rd party tools, not the other way around IMO.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
I'd personally welcome this. I know it's about a better floating support, but I can see the usecase and the patch should be quite simple.

rr-:
I could come up with the patch tomorrow.

Airblader:
Please note that Michael makes the decision on what is accepted, so you should wait for his feedback.

rr-:
Sure (I'm okay with rejected pull requests though)

Airblader:
If you don't mind it perhaps being rejected, you are free to submit a PR anytime you want. :)

stapelberg:
I think it might make sense to have the suggested feature (mostly because it helps in working around broken applications and shouldn’t be too invasive on the code base), so please send a PR.

Airblader:
Closing since this was merged.

PR #1728, opened 2015-05-30
Add "move relative position" support

rr-:
Right now we can move floating windows in following fashion:

- to an absolute position
- to the center of the screen
- to the center of all screens
- in the specified direction

The last one might be sufficient for simple key bindings, but it's rather cumbersome to use in scripts. This proposal adds `move relative position X Y` command, that offsets current window position by X value in horizontal axis and Y value in vertical axis.

This makes it complementary with `resize grow X Y` and `resize shrink X Y` commands.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1728,10 +1728,11 @@ Use the +move+ command to move a container.
 # defaults to 10 pixels.
 move <left|right|down|up> [<px> px]
 
-# Moves the container either to a specific location
-# or to the center of the screen. If 'absolute' is
-# used, it is moved to the center of all outputs.
-move [absolute] position [[<px> px] [<px> px]|center]
+# Moves the container either to a specific location,
+# by a specific offset or to the center of the screen.
+# If 'absolute' is used, it is moved to the center
+# of all outputs.
+move [absolute|relative] position [[<px> px] [<px> px]|center]
 
 # Moves the container to the current position of the
 # mouse cursor. Only affects floating containers.
--- a/testcases/t/124-move.t
+++ b/testcases/t/124-move.t
@@ -235,6 +235,13 @@ cmd 'move position 5 px 15 px';
 is($floatcon[0]->{rect}->{x}, 5, 'moved to position 5 x');
 is($floatcon[0]->{rect}->{y}, 15, 'moved to position 15 y');
 
+cmd 'move relative position +3 px -12 px';
+
+@floatcon = @{get_ws($tmp)->{floating_nodes}};
+
+is($floatcon[0]->{rect}->{x}, 8, 'moved by offset 3 x');
+is($floatcon[0]->{rect}->{y}, 3, 'moved by offset -12 y');
+
 cmd 'move absolute position center';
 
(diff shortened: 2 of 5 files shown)


testcases/t/124-move.t
```diff
@@ -235,6 +235,13 @@ cmd 'move position 5 px 15 px';
 is($floatcon[0]->{rect}->{x}, 5, 'moved to position 5 x');
 is($floatcon[0]->{rect}->{y}, 15, 'moved to position 15 y');
 
+cmd 'move relative position +3 px +7 px';
```
Airblader: I would like to see a test that uses a negative value.

rr-:
Note that `i3-msg` would need to be invoked like this, `i3-msg -- move relative position -5 -7` so that it doesn't interpret `-3` as option `3`.

docs/userguide
```diff
@@ -1728,10 +1728,11 @@ Use the +move+ command to move a container.
 # defaults to 10 pixels.
 move <left|right|down|up> [<px> px]
 
-# Moves the container either to a specific location
-# or to the center of the screen. If 'absolute' is
-# used, it is moved to the center of all outputs.
-move [absolute] position [[<px> px] [<px> px]|center]
+# Moves the container either to a specific location,
+# by a specific offset or to the center of the screen.
+# If 'absolute' is # used, it is moved to the center
```
Airblader: nit: extra `#` in there

Airblader:
Please also squash your commits so the history stays clean (but you can wait for Michael to review it first). Of course we'll also have to wait on feedback if he even wants to accept such a patch at all. Assuming so, it looks good to me now.

Airblader:
@rr- Just so I understand you right, you know that we already have `move right 35` which will do exactly the same thing, you just want a different API, right?

I'm not sure we should merge this. This would only make sense if we removed the old API, but we generally don't break compatibility. Having two APIs do the same seems unnecessary and just adds complexity. If they then even expose slightly different behavior, we just end up with confusion.

rr-:
Well, if you look at it this way then yes, I want different API - just like I do in #1727...

Airblader:
I don't think this is the same as #1727. There you ask for a feature that is currently simply not possible in i3 itself. This PR is different, though, because `move relative position <x> <y>` is equivalent (almost, see above comment) to `move right <x>, move down <y>`. What's wrong with using that?

rr-:
There's nothing wrong with it, but it's the wording (which won't change due to BC) that makes me feel like this one is missing. That, and consistency: we have `resize grow` and `resize shrink` that takes both X and Y rather than `grow horizontal` or `grow vertical`.

Airblader:
> (if we have `move absolute position X Y`[…]

But we don't. At least not how you might expect. The only things `absolute` does if used with `position <x> <y>` (rather than `position center`, where it really makes a difference) is omit a check of the target position is contained in an output and it doesn't change the scratchpad state.

stapelberg:
I agree with @Airblader, I dislike having two different APIs for the same feature.

PR #1733, opened 2015-05-31
Added a separate color directive for the binding mode indicator

### LINKED ISSUE #1732: [enhancement] Separate binding mode color
The color currently used for binding mode indicators is the same as that for urgent workspaces. I know this is a detail, but I'd like to have a separate `binding_mode` directive in the colors that can be used.

The reason being that I want my urgent workspace to be really present since I really utilize urgency flags, but I feel the binding mode is too present due to that.

The idea would be to fall back to the urgent workspace setting if it is not present, providing backwards compatibility. The patch should be simple and not add any overhead.

Airblader:
fixes #1732

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -520,21 +520,23 @@ statusline::
 	Text color to be used for the statusline.
 separator::
 	Text color to be used for the separator.
-focused_workspace_text/focused_workspace_bg::
-	Text color/background color for a workspace button when the workspace
+focused_workspace_text/focused_workspace_bg/focused_workspace_border::
+	Text/background/border color for a workspace button when the workspace
 	has focus.
-active_workspace_text/active_workspace_bg::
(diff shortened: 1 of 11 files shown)


docs/ipc
```diff
@@ -520,21 +520,23 @@ statusline::
...
 	is active (visible) on some output, but the focus is on another one.
 	You can only tell this apart from the focused workspace when you are
 	using multiple monitors.
-inactive_workspace_text/inactive_workspace_bg::
-	Text color/background color for a workspace button when the workspace
+inactive_workspace_text/inactive_workspace_bg/inactive_workspace_border::
+	Text/background/border color for a workspace button when the workspace
 	does not have focus and is not active (visible) on any output. This
 	will be the case for most workspaces.
-urgent_workspace_text/urgent_workspace_bar::
```
Airblader: s/_bar/_bg (and below)

PR #1740, opened 2015-06-08
Added cmd_size

rr-:
Implements #1727.

- Doesn't move window, for this one should use `move direction` or `move position`.
- Size hints are untested, since I had no idea how to test them.
- Doesn't do anything for containers. IMO resizing containers to specific sizes is different enough to deserve a separate feature/pull request.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1818,6 +1818,23 @@ mode "resize" {
 bindsym $mod+r mode "resize"
 ----------------------------------------------------------------------
 
+=== Resizing containers/windows
+
+If you want to give a window a specific size, you can use the +resize+ command,
+which takes the desired size as its argument. This command will only work on
+floating containers.
+
+*Syntax*:
+---------------------------------------------------------
+resize <width> [px] <height> [px]
+---------------------------------------------------------
+
+*Examples*:
+------------------------------------------------
+# Set urxvt size to 640x480
+for_window [class="urxvt"] resize 640 480
+------------------------------------------------
+
 === Jumping to specific windows
 
 Often when in a multi-monitor environment, you want to quickly jump to a
--- a/include/commands.h
+++ b/include/commands.h
@@ -61,6 +61,12 @@ void cmd_move_con_to_workspace_name(I3_CMD, char *name);
 void cmd_move_con_to_workspace_number(I3_CMD, char *which);
 
 /**
+ * Implementation of 'resize <px> [px] <px> [px]'.
+ *
+ */
+void cmd_size(I3_CMD, char *cwidth, char *cheight);
+
+/**
  * Implementation of 'resize grow|shrink <direction> [<px> px] [or <ppt> ppt]'.
  *
  */
--- a/include/floating.h
+++ b/include/floating.h
@@ -177,6 +177,15 @@ drag_result_t drag_pointer(Con *con, const xcb_button_press_event_t *event,
 void floating_reposition(Con *con, Rect newrect);
 
 /**
+ * Sets size of the CT_FLOATING_CON to specified dimensions. Might limit the
+ * actual size with regard to size constraints taken from user settings.
+ * Additionally, the dimensions may be upscaled until they're divisible by the
+ * window's size hints.
+ *
+ */
+void floating_resize(Con *floating_con, int x, int y);
+
+/**
  * Fixes the coordinates of the floating window whenever the window gets
  * reassigned to a different output (or when the output’s rect changes).
  *
--- a/src/commands.c
+++ b/src/commands.c
@@ -791,6 +791,35 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
 }
 
 /*
+ * Implementation of 'resize <px> [px] <px> [px]'.
+ *
+ */
+void cmd_size(I3_CMD, char *cwidth, char *cheight) {
+    DLOG("resizing to %sx%s px\n", cwidth, cheight);
+    // TODO: We could either handle this in the parser itself as a separate token (and make the stack typed) or we need a better way to convert a string to a number with error checking
+    int x = atoi(cwidth);
+    int y = atoi(cheight);
+    if (x <= 0 || y <= 0) {
+        ELOG("Resize failed: dimensions cannot be negative (was %sx%s)\n", cwidth, cheight);
+        return;
+    }
+
+    HANDLE_EMPTY_MATCH;
+
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        Con *floating_con;
+        if ((floating_con = con_inside_floating(current->con))) {
+            floating_resize(floating_con, x, y);
+        }
+    }
+
+    cmd_output->needs_tree_render = true;
+    // XXX: default reply for now, make this a better reply
+    ysuccess(true);
+}
+
+/*
  * Implementation of 'border normal|none|1pixel|toggle|pixel'.
  *
  */
--- a/testcases/t/240-floating-size.t
+++ b/testcases/t/240-floating-size.t
@@ -1,0 +1,45 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • http://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • http://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • http://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Test behavior of "resize <width> <height>" command.
+# Ticket: #1727
+# Bug still in: 4.10.2-1-gc0dbc5d
+use i3test;
+
+################################################################################
+# Check that setting floating windows size works
+################################################################################
+
+my $tmp = fresh_workspace;
+
+open_floating_window;
+
+my @content = @{get_ws($tmp)->{floating_nodes}};
+is(@content, 1, 'one floating node on this ws');
+
+my $oldrect = $content[0]->{rect};
+
+cmd 'resize 100 px 250 px';
+
+@content = @{get_ws($tmp)->{floating_nodes}};
+cmp_ok($content[0]->{rect}->{x}, '==', $oldrect->{x}, 'x untouched');
+cmp_ok($content[0]->{rect}->{y}, '==', $oldrect->{y}, 'y untouched');
+cmp_ok($content[0]->{rect}->{width}, '!=', $oldrect->{width}, 'width changed');
+cmp_ok($content[0]->{rect}->{height}, '!=', $oldrect->{width}, 'height changed');
+cmp_ok($content[0]->{rect}->{width}, '==', 100, 'width changed to 100 px');
+cmp_ok($content[0]->{rect}->{height}, '==', 250, 'height changed to 250 px');
+
+done_testing;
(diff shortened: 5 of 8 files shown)


docs/userguide
```diff
@@ -1818,6 +1818,24 @@ mode "resize" {
 bindsym $mod+r mode "resize"
 ----------------------------------------------------------------------
 
+=== Resizing containers/windows
+
+If you want to set window size to specific size, you can use the +size+ command:
+
+*Syntax*:
+---------------------------------------------------------
+size <px> [px] <px> [px]
```
Airblader: Using width and height for the argument names is probably a bit clearer.

docs/userguide
```diff
@@ -1818,6 +1818,24 @@ mode "resize" {
 bindsym $mod+r mode "resize"
 ----------------------------------------------------------------------
 
+=== Resizing containers/windows
+
+If you want to set window size to specific size, you can use the +size+ command:
```
Airblader: This sentence sounds odd. How about "If you want to give a window a specific size,..."?

docs/userguide
```diff
@@ -1818,6 +1818,24 @@ mode "resize" {
...
+size <px> [px] <px> [px]
+---------------------------------------------------------
+
+The pixel argument specifies what dimensions pixels a *floating container*
+should be resized to.
+
+*Examples*:
+------------------------------------------------
+# Set urxvt size to 640x480
+for_window [class="urxvt"] floating enable size 640 480
```
Airblader: I don't think this would work? And the example should probably only contain the command. The fact that it only works on floating windows should be mentioned in the documentation above.

rr-: Even if this works only for floating windows? It works, I tested it.

docs/userguide
```diff
@@ -1818,6 +1818,24 @@ mode "resize" {
...
+=== Resizing containers/windows
+
+If you want to set window size to specific size, you can use the +size+ command:
+
+*Syntax*:
+---------------------------------------------------------
+size <px> [px] <px> [px]
+---------------------------------------------------------
+
+The pixel argument specifies what dimensions pixels a *floating container*
```
Airblader: I think the sentence should be above with the other one, but it also sounds off, so my suggestion: "It takes the desired size as its argument. This command will only work on floating containers."

src/commands.c
```diff
@@ -610,6 +610,24 @@ static void cmd_resize_floating(I3_CMD, char *way, char *direction, Con *floatin
         floating_con->scratchpad_state = SCRATCHPAD_CHANGED;
 }
 
+static void cmd_size_floating(I3_CMD, Con *floating_con, int x, int y) {
+    LOG("floating size\n");
```
Airblader: It'd be nice to log the arguments (container, x, y)

src/commands.c
```diff
@@ -610,6 +610,24 @@ static void cmd_resize_floating(I3_CMD, char *way, char *direction, Con *floatin
         floating_con->scratchpad_state = SCRATCHPAD_CHANGED;
 }
 
+static void cmd_size_floating(I3_CMD, Con *floating_con, int x, int y) {
```
Airblader: I'd like this to be more of an equivalent of `floating_reposition`, i.e., move it into a function in `floating.c` named `floating_resize` or something.

src/commands.c
```diff
@@ -791,6 +809,36 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
 }
 
 /*
+ * Implementation of 'size <px> [px] <px> [px]'.
+ *
+ */
+void cmd_size(I3_CMD, char *cx, char *cy) {
+    DLOG("resizing to %sx%s px\n", cx, cy);
+    int x = atoi(cx);
```
Airblader: You'll have to do some validation of this in case reading it as integers fails.

src/commands.c
```diff
@@ -791,6 +809,36 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
...
+    int y = atoi(cy);
+
+    HANDLE_EMPTY_MATCH;
+
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        /* Don't handle dock windows (issue #1201) */
```
Airblader: I'm guessing you copy-pasted this from somewhere. I doubt that this is relevant here, since docks aren't floating.

src/commands.c
rr-: Please take a look at line 773.

testcases/t/240-floating-size.t
```diff
@@ -0,0 +1,45 @@
...
+# • http://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • http://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# TODO: Description of this file.
```
Airblader: TODO :)

testcases/t/240-floating-size.t
```diff
@@ -0,0 +1,45 @@
...
+################################################################################
+# Check that setting floating windows size works
+################################################################################
+
+my $tmp = fresh_workspace;
+
+open_floating_window;
+
+my @content = @{get_ws($tmp)->{floating_nodes}};
+cmp_ok(@content, '==', 1, 'one floating node on this ws');
```
Airblader: Just use `is(@content, 1, …)`

docs/userguide
Airblader: Yeah, it seems to work (I didn't know the comma/semicolon was optional). Either way, the rest of the comment still stands.

src/commands.c
Airblader: Well, we can at least check for <= 0 since those values don't make sense, though technically the error behavior of atoi isn't specified. It might be better to use strtol here.

Airblader:
You're gonna have to rework the commits. For one, there should be some squashing, but also the commit messages need to be improved by quite a bit. They should explain what the commit does, the corresponding commit should contain a "fixes #xxxx" part and lastly commits shouldn't necessarily contain smilies of your emotional state. :wink:

rr-:
>You're gonna have to rework the commits

Usually I squash everything into one commit once it has no more issues left.

Airblader:
> Usually I squash everything into one commit once it has no more issues left.

That's absolutely fine.

src/commands.c
rr-: I see nice `strtol()` calls in line 356, but... they appear in 338 too. At the same time, there are more `atoi()` calls in this file, each one with `//TODO` comment that suggests to move the validation to the token parser. Perhaps instead of coming with my own solution, it'd be wiser to add similar `//TODO` and fix this properly in another commit?

Airblader: We can do that. I still think we should reject <= 0 since this makes no sense here and probably breaks the window.

rr-: Please note that `<=0` window sizes won't actually happen thanks to floating window size constraints enforced by the call to `floating_check_size()`.

rr-:
I fixed the mentioned issues.
By the way, what happens when I reference an issue in the commit message, and then I overwrite history? Does Github remove the references in the original issues to the dead commits, or does it leave them off?

Airblader:
I think references are kept, but closing hooks will go away (not entirely sure, though)

docs/userguide
```diff
@@ -1818,6 +1818,23 @@ mode "resize" {
...
 
+=== Resizing containers/windows
+
+If you want to give a window a specific size, you can use the +size+ command,
+which takes the desired size as its argument. This command will only work on
+floating containers.
+
+*Syntax*:
+---------------------------------------------------------
+size <width> [px] <height> [px]
```
stapelberg: I think it’d be better to make this feature a subcommand of the existing “resize” command.

include/commands.h
```diff
@@ -61,6 +61,12 @@ void cmd_move_con_to_workspace_name(I3_CMD, char *name);
 void cmd_move_con_to_workspace_number(I3_CMD, char *which);
 
 /**
+ * Implementation of 'size <px> [px] <px> [px]'.
+ *
+ */
+void cmd_size(I3_CMD, char *cx, char *cy);
```
stapelberg: Here and below, the parameters should be called “width” and “height” instead of “x” and “y”, since the latter are used for positions, not dimensions.

src/commands.c
```diff
@@ -791,6 +791,33 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
...
+ * Implementation of 'size <px> [px] <px> [px]'.
+ *
+ */
+void cmd_size(I3_CMD, char *cx, char *cy) {
+    DLOG("resizing to %sx%s px\n", cx, cy);
+    // TODO: We could either handle this in the parser itself as a separate token (and make the stack typed) or we need a better way to convert a string to a number with error checking
+    int x = atoi(cx);
+    int y = atoi(cy);
+    if (x <= 0 || y <= 0)
+        return;
```
stapelberg: Use `ELOG` to make the user aware of the invalid input rather than just not doing anything silently.

include/floating.h
```diff
@@ -176,6 +176,15 @@ drag_result_t drag_pointer(Con *con, const xcb_button_press_event_t *event,
  */
 void floating_reposition(Con *con, Rect newrect);
 
+/*
```
stapelberg: nit: Use `/**` instead of `/*` so that Doxygen picks up this comment.

rr-:
Updated

docs/userguide
```diff
@@ -1818,6 +1818,23 @@ mode "resize" {
...
 
+=== Resizing containers/windows
+
+If you want to give a window a specific size, you can use the +resize+ command,
+which takes the desired size as its argument. This command will only work on
+floating containers.
+
+*Syntax*:
+---------------------------------------------------------
+resize <width> [px] <height> [px]
```
Airblader: This is awkward now because resize is an existing command and already has documentation. One directive should only be described in one place.

docs/userguide
```diff
@@ -1818,6 +1818,23 @@ mode "resize" {
...
+---------------------------------------------------------
+
+*Examples*:
+------------------------------------------------
+# Set urxvt size to 640x480
+for_window [class="urxvt"] resize 640 480
```
Airblader: I think the syntax should be more explicit than this. How about "resize set 640 480"?

src/commands.c
```diff
@@ -791,6 +791,35 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
 }
 
 /*
+ * Implementation of 'size <px> [px] <px> [px]'.
```
Airblader: This is wrong now.

include/commands.h
```diff
@@ -61,6 +61,12 @@ void cmd_move_con_to_workspace_name(I3_CMD, char *name);
 void cmd_move_con_to_workspace_number(I3_CMD, char *which);
 
 /**
+ * Implementation of 'size <px> [px] <px> [px]'.
```
Airblader: This is wrong now.

Airblader:
@stapelberg Just a bumper to make sure you look at this PR since it's been a while. :)

stapelberg:
@Airblader I didn’t comment on this because your comments were not addressed so far (e.g. about the userguide being updated in the wrong place).

Looking at the PR itself, it seems like the original repository was deleted.

@rr- Are you still interested in working on this?

rr-:
Well...

>wait on what @stapelberg thinks, though

stapelberg:
The proposed “resize set” syntax sounds good to me, but we can hash out such details later. I didn’t realize that you wanted an answer to that before addressing the other comments.

rr-:
Continued in #1893 for reasons stated in the other pull request.

PR #1744, opened 2015-06-09
Documentation Improvements

Airblader:
Was the stack-limit thing a pre-v4 feature? Either way, it hasn't lived for a long time. If we do ever implement it (should we raise a feature ticket for it?), we should probably rethink it and just start from scratch with the documentation as well.

As for the "1pixel" directive, this seems to be a backwards compatibility thing, too, right? Maybe we should remove it from the docs altogether to not encourage using it anymore?


stapelberg:
Yeah, it was a v3 feature. Given that nobody ever asked for it after v4 was released, I think it doesn’t make sense to add it back.

docs/userguide
```diff
@@ -2051,49 +2045,26 @@ seperate bindings for a specific set of labels and then only use those labels.
...
-stack-limit cols 2
-
-# Not more than 5 rows in this stacking container
-stack-limit rows 5
--------------------
-
-image:stacklimit.png[Container limited to two columns]
-///////////////////////////////////////////////////////////////////////////////
+----------------------------------------------
+# use window decoration, but no border
```
acrisci: s/window decoration/title bar

Borders are also window decorations.

acrisci:
I think we should consider removing all mention of `border 1pixel` in the documentation since `border pixel [n]` does the same thing but is more general.

Airblader:
@acrisci Yeah, that's what I thought would be best, too. Updated.

stapelberg:
> I think we should consider removing all mention of border 1pixel in the documentation since border pixel [n] does the same thing but is more general.

I think we should retain at least one mention of “1pixel”, which could be something like “"1pixel" is a synonym for "pixel 1".”, just so that when users search the documentation for what 1pixel means, they can discover the current state of things.

Airblader:
@stapelberg I've added a mention of it back.

PR #1745, opened 2015-06-10
Configurable tray padding.

mh21:
Makes the i3bar tray padding configurable (as discussed in #1700):
- documentation a bit short
- introduces `cfg_bar_init` to keep default at `2px`, could go away if the default would be changed to `0px`
- no testcases for tray icons other than config validation
- this is not output specific, I have no idea whether the `logical_px(...)` is enough if somebody has both a  HiDPI and a LoDPI display

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1351,6 +1351,23 @@ Note that you might not have a primary output configured yet. To do so, run:
 xrandr --output <output> --primary
 -------------------------
 
+=== Tray padding
+
+The tray is shown on the right-hand side of the bar. By default, a padding of 2
+pixels is used for the upper, lower and right-hand side of the tray area and
+between the individual icons.
+
+*Syntax*:
+-------------------------
+tray_padding <px> [px]
+-------------------------
+
+*Example*:
+-------------------------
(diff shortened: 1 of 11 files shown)


Airblader:
This has a dependency with #1697 since I introduce `cfg_bar_start` there.

docs/userguide
```diff
@@ -1333,6 +1333,21 @@ Note that you might not have a primary output configured yet. To do so, run:
 xrandr --output <output> --primary
 -------------------------
 
+=== Tray padding
+
+By default, the tray icons have a padding of 2 pixels.
```
Airblader: I think there should be a sentence about what padding this refers to.

stapelberg:
Agreed with @Airblader’s comments. Using `logical_px` is enough for supporting hi-dpi displays.

mh21:
I will fix and rebase it once #1697 is merged.

Airblader:
@mh21 Merged :)

mh21:
Updated, now with longer documentation and Airbladers parsing functions.

PR #1747, opened 2015-06-10
Implement "title_format"

### LINKED ISSUE #1723: Support a user-defined title format
It would be useful if we could attach some text to be shown to the left of the application-defined title.  It's possible to configure some programs (e.g. vim, terminals via bash, etc) to do this, but most applications do not allow users to modify their titles in any way.   A small bit of application-specific text would allow a quicker scan of window titles, currently this requires reading and interpreting each title.  

Something like this in i3 config would be used to set up the text (longer strings should be possible, I just happen to like single characters) :

for_window [class="Gvim"] title_prefix "☯ "
for_window [class="Urxvt"] title_prefix "↵ "

I don't have this implemented, but I do have bash and vim configured to use the above characters in their titles.  The i3 implementation shouldn't look very different, except it would be possible for the other applications as well:

image

My i3 version is 4.10.2.

Airblader:
I verified that not using a pango font doesn't mistakenly escape the window title. I do strongly suggest we get at least someone else (@acrisci, @ambihelical being canonical candidates) to play around with this first.

One open question for me is if we need to handle other characters when escaping pango markup. The glib implementation seems to do something (see my comment in #1723 where I pasted their implementation), but I don't really understand what's happening there, so input is much appreciated.

Here are some quick screenshots:
screenshot-2015-06-10_19-10-57
screenshot-2015-06-10_19-12-37
screenshot-2015-06-10_19-15-20

fixes #1723

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -564,6 +564,8 @@ hide_edge_borders vertical
 
 === Arbitrary commands for specific windows (for_window)
 
+[[for_window]]
+
 With the +for_window+ command, you can let i3 execute any command when it
 encounters a specific window. This can be used to set windows to floating or to
 change their border style, for example.
@@ -2065,6 +2067,37 @@ bindsym $mod+g exec i3-input -p 'goto ' -l 1 -P 'Goto: '
 Alternatively, if you do not want to mess with +i3-input+, you could create
 seperate bindings for a specific set of labels and then only use those labels.
 ///////////////////////////////////////////////////////////////////
+
+=== Window title format
+
+By default, i3 will simply print the X11 window title. Using +title_format+,
+this can be customized by setting the format to the desired output. This
+directive supports
+https://developer.gnome.org/pango/stable/PangoMarkupFormat.html[Pango markup]
+and the following placeholders which will be replaced:
+
++%title+::
+    The X11 window title (_NET_WM_NAME or WM_NAME as fallback).
+
+Using the <<for_window>> directive, you can set the title format for any window
+based on <<command_criteria>>.
+
+*Syntax*:
+---------------------
+title_format <format>
+---------------------
+
+*Examples*:
+-------------------------------------------------------------------------------------
+# give the focused window a prefix
+bindsym $mod+p title_format "Important | %title"
+
+# print all window titles bold
+for_window [class=".*"] title_format "<b>%title</b>"
+
+# print window titles of firefox windows red
+for_window [class="(?i)firefox"] title_format "<span foreground='red'>%title</span>"
+-------------------------------------------------------------------------------------
 
 === Changing border style
 
--- a/src/commands.c
+++ b/src/commands.c
@@ -1900,6 +1900,17 @@ void cmd_scratchpad_show(I3_CMD) {
 }
 
 /*
+ * Implementation of 'title_format <format>'
+ *
+ */
+void cmd_title_format(I3_CMD, char *format) {
+    DLOG("setting title_format to %s\n", format);
+
+    cmd_output->needs_tree_render = true;
+    ysuccess(true);
+}
+
+/*
  * Implementation of 'rename workspace [<name>] to <name>'
  *
  */
--- a/src/commands.c
+++ b/src/commands.c
@@ -1904,7 +1904,25 @@ void cmd_scratchpad_show(I3_CMD) {
  *
  */
 void cmd_title_format(I3_CMD, char *format) {
-    DLOG("setting title_format to %s\n", format);
+    DLOG("setting title_format to \"%s\"\n", format);
+    HANDLE_EMPTY_MATCH;
+
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        if (current->con->window == NULL)
+            continue;
+
+        DLOG("setting title_format for %p / %s\n", current->con, current->con->name);
+        FREE(current->con->window->title_format);
+
+        /* If we only display the title without anything else, we can skip the parsing step,
+         * so we remove the title format altogether. */
+        if (strcasecmp(format, "%title") != 0)
+            current->con->window->title_format = sstrdup(format);
+
+        /* Make sure the window title is redrawn immediately. */
+        current->con->window->name_x_changed = true;
+    }
 
     cmd_output->needs_tree_render = true;
     ysuccess(true);
--- a/src/x.c
+++ b/src/x.c
@@ -302,6 +302,45 @@ void x_window_kill(xcb_window_t window, kill_window_t kill_window) {
     free(event);
 }
 
+static i3String *parse_title_format(char *format, i3String *_title) {
+    /* We need to ensure that we only escape the window title if pango
+     * is used by the current font. */
+    const bool is_markup = font_is_pango();
+
+    i3String *title = is_markup ? i3string_escape_markup(_title) : _title;
+    const char *escaped_title = i3string_as_utf8(title);
+
+    /* We have to first iterate over the string to see how much buffer space
+     * we need to allocate. */
+    int buffer_len = strlen(format) + 1;
+    for (char *walk = format; *walk != '\0'; walk++) {
+        if (STARTS_WITH(walk, "%title")) {
+            buffer_len = buffer_len - strlen("%title") + strlen(escaped_title);
+            walk += strlen("%title") - 1;
+        }
+    }
+
+    /* Now we can parse the format string. */
+    char buffer[buffer_len];
+    char *outwalk = buffer;
+    for (char *walk = format; *walk != '\0'; walk++) {
+        if (*walk != '%') {
+            *(outwalk++) = *walk;
+            continue;
+        }
+
+        if (STARTS_WITH(walk + 1, "title")) {
+            outwalk += sprintf(outwalk, "%s", escaped_title);
+            walk += strlen("title");
+        }
+    }
+    *outwalk = '\0';
+
+    i3String *formatted = i3string_from_utf8(buffer);
+    i3string_set_markup(formatted, is_markup);
+    return formatted;
+}
+
 /*
  * Draws the decoration of the given container onto its parent.
  *
@@ -549,10 +588,13 @@ void x_draw_decoration(Con *con) {
         I3STRING_FREE(mark);
     }
 
-    draw_text(win->name,
+    i3String *title = win->title_format == NULL ? win->name : parse_title_format(win->title_format, win->name);
+    draw_text(title,
               parent->pixmap, parent->pm_gc,
               con->deco_rect.x + logical_px(2) + indent_px, con->deco_rect.y + text_offset_y,
               con->deco_rect.width - logical_px(2) - indent_px - mark_width - logical_px(2));
+    if (win->title_format != NULL)
+        I3STRING_FREE(title);
 
 after_title:
     /* Since we don’t clip the text at all, it might in some cases be painted
--- a/include/commands.h
+++ b/include/commands.h
@@ -277,6 +277,12 @@ void cmd_move_scratchpad(I3_CMD);
 void cmd_scratchpad_show(I3_CMD);
 
 /**
+ * Implementation of 'title_format <format>'
+ *
+ */
+void cmd_title_format(I3_CMD, char *format);
+
+/**
  * Implementation of 'rename workspace <name> to <name>'
  *
  */
--- a/include/data.h
+++ b/include/data.h
@@ -363,6 +363,8 @@ struct Window {
 
     /** The name of the window. */
     i3String *name;
+    /** The format with which the window's name should be displayed. */
+    char *title_format;
 
     /** The WM_WINDOW_ROLE of this window (for example, the pidgin buddy window
      * sets "buddy list"). Useful to match specific windows in assignments or
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -244,6 +244,11 @@ bool i3string_is_markup(i3String *str);
 void i3string_set_markup(i3String *str, bool is_markup);
 
 /**
+ * Escape pango markup characters in the given string.
+ */
+i3String *i3string_escape_markup(i3String *str);
+
+/**
  * Returns the number of glyphs in an i3String.
  *
  */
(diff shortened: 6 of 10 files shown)


Airblader:
I completely forgot the documentation. Will add that, of course. :)

acrisci:
I think you should just use the glib implementation.

Airblader:
You mean just blindly use their logic? I'm not so much a fan of committing code I don't understand.

acrisci:
I think you just made the case against using any libraries at all.

The truth is that nobody understands how computers work.

Airblader:
But this isn't a library, this is own code that we need to maintain, so we should understand it. Or did you mean pulling in glib as a dependency?

acrisci:
We get glib for free with pango.

Airblader:
Oh  okay. Then it's definitely better than reimplementing this. I'll change it.

Airblader:
The documentation for `g_markup_escape_text` contains the explanation, by the way:

> Note also that this function will produce character references in
> the range of &#x1; ... &#x1f; for all control sequences
> except for tabstop, newline and carriage return.  The character
> references in this range are not valid XML 1.0, but they are
> valid XML 1.1 and will be accepted by the GMarkup parser.

Airblader:
OK, I've switched to the glib implementation and added the documentation.

ambihelical:
Nice!  I'll try to give it a go this weekend.   I still think having some kind of left/right/center region would be useful, and the size balancing not as hard to handle as you have implied.   Having the ability to change the appearance of the mark as shown on your window would be one benefit.

Airblader:
Yeah, I did come around and think alignment options are a superior solution, though we should still define clear behavior about the balancing part upfront. This PR is just the first version of this feature (iterative process), see #1750 for the followup.

teto:
First of all thanks for working on this and for all the rest of your work. I've tried the examples given in your doc and I succeed in adding a prefix in front of my title with your binding.
But when doing this:
`for_window [class="(?i)firefox"] title_format "<span foreground='red'>%title</span>" `,  "%title" is replaced but the <span> is not interpreted and is printed as is. So the firefox title becomes:
"<span foreground='red'>My tab title</span>"
I believe this is a fantastic feature and I am looking forward to it.

Airblader:
Thanks for the feedback!

> but the &lt;span&gt; is not interpreted and is printed as is.

Could you maybe take a screenshot? It works fine for me. Also, can you upload your config somewhere? Finally, please post your reply in the issue, it probably shouldn't be in the PR. Thanks!

ambihelical:
Feedback; it seems to be working pretty well for me.  I tried @teto's example as well and could not reproduce the issue he was having. 

One issue I found, with large fonts (size='xx-large') and stacked windows, the text is drawn partially outside of the window title area onto the title area of the window above.

Airblader:
Thanks for the feedback. I think the issue you mentioned is not something we can reasonably fix.

teto:
I was surprised by the 4.8 so I rechecked the PR and my git log on the pulled PR is:
```
634bdcf Added documentation for "title_format".
f177929 Parse the title_format and display the customized window title if a format was set. The format string set with "title_format" 
249af33 Implemented a helper function to escape a string for pango markup.
08d7f23 Added a helper function to decide whether the currently saved font is a pango font.
68cf77d Added command directive 'title_format'. This directive will be used to customize the window title.
1d4b586 Merge pull request #1737 from Airblader/feature-xdotool-on-travis
f262445 Install xdotool to run tests requiring it on Travis.
7b8745f Merge pull request #1735 from shdown/strdup
fe006f0 i3bar: fix freeing static strings
6b505d8 Merge pull request #1733 from Airblader/feature-1732
```
and
`i3 version 4.10.2-193-g634bdcf (2015-06-11, branch "feature-1723") © 2009 Michael Stapelberg and contributors`
Are those ok ? because I still have the same problem.

Airblader:
@teto I just checked your config. You're not using a pango font, but rather xft. Of course pango markup cannot work in this case. :)

teto:
Thanks !! I had no idea what a pango font was (even though you speak of it in yourt first post, shame on me). Now it works as intended ! Once again thanks for your many contributions on i3.

Airblader:
Great. Thanks for testing this and the feedback.
@stapelberg I think we have enough feedback now.

docs/userguide
```diff
@@ -2048,6 +2050,33 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
 seperate bindings for a specific set of labels and then only use those labels.
 ///////////////////////////////////////////////////////////////////
 
+=== Window title format
+
+By default, i3 will simply print the window's title. Using +title_format+, this can be customized
+by setting the format to the desired output. This directive supports pango markup and the following
```
acrisci: This should link to pango markup.

Airblader:
@acrisci Done

src/x.c
```diff
@@ -302,6 +302,45 @@ void x_window_kill(xcb_window_t window, kill_window_t kill_window) {
...
+
+    i3String *title = is_markup ? i3string_escape_markup(_title) : _title;
+    const char *escaped_title = i3string_as_utf8(title);
+
+    /* We have to first iterate over the string to see how much buffer space
+     * we need to allocate. */
+    int buffer_len = strlen(format) + 1;
+    for (char *walk = format; *walk != '\0'; walk++) {
+        if (STARTS_WITH(walk, "%title")) {
+            buffer_len += strlen(escaped_title);
```
stapelberg: I think this should be `buffer_len = buffer_len - strlen("%title") + strlen(escaped_title)`, shouldn’t it?

src/x.c
```diff
@@ -302,6 +302,45 @@ void x_window_kill(xcb_window_t window, kill_window_t kill_window) {
...
+        }
+
+        if (STARTS_WITH(walk + 1, "title")) {
+            outwalk += sprintf(outwalk, "%s", escaped_title);
+            walk += strlen("title");
+        }
+    }
+    *outwalk = '\0';
+
+    i3String *formatted = i3string_from_utf8(buffer);
```
stapelberg: Where is the string freed? Isn’t this a memory leak?

docs/userguide
```diff
@@ -2048,6 +2050,34 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
...
+    The window title.
+
+Using the <<for_window>> directive, you can set the title format for any window based on <<command_criteria>>.
+
+*Syntax*:
+---------------------
+title_format <format>
+---------------------
+
+*Examples*:
```
stapelberg: I think the first example should be one that motivated the feature in the first place, not one that shows what’s technically possible :).

stapelberg:
Can you please squash the two commits that deal with the escaping function? I.e., I don’t like to have the commit in the history which adds code that is never used and replaced a commit later :).

Airblader:
@stapelberg I've incorporated your comments. Is using `i3string_copy` like this ok for you?

docs/userguide
stapelberg: You’ve now changed this to:

```
# give the focused window a prefix
bindsym $mod+p title_format "[x] %title"
```

But is [x] really a worthwhile prefix? I can’t imagine why I would want to prefix window titles with an [x].

This might seem like nitpicking, but the userguide has plenty of strong examples, and I’d like to see a good one here as well please.

src/x.c
```diff
@@ -549,10 +588,12 @@ void x_draw_decoration(Con *con) {
         I3STRING_FREE(mark);
     }
 
-    draw_text(win->name,
+    i3String *title = win->title_format == NULL ? i3string_copy(win->name) : parse_title_format(win->title_format, win->name);
```
stapelberg: You can avoid the copy by using `if (win->title_format != NULL) I3STRING_FREE(title);` below. I think doing that is worthwhile, considering that this function is run for every window title drawing.

stapelberg:
AFAICT, my comment about the commit history is not yet addressed.

src/x.c
Airblader: Yeah, I just thought that looks somewhat ugly, but what you said was also my concern. I'll change it.

Airblader:
@stapelberg I had squashed the two individual commits for the two helper functions. Now also squashed with the usage of them. My apologies.

Airblader:
I forgot one comment…

Airblader:
@stapelberg Adressed the last comment now, too.

docs/userguide
```diff
@@ -2048,6 +2050,35 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
...
+-------------------------------------------------------------------------------------
+# give the focused window a prefix
+bindsym $mod+p title_format "Important | %title"
+
+# print all window titles bold
+for_window [class="^.*"] title_format "<b>%title</b>"
```
stapelberg: nit: the shortest way to match all windows is `for_window [class="."]`, but I can see how `.*` might be more familiar, i.e. people might more easily recognize that as a regular expression. The current form of `^.*` I don’t quite understand — why the anchoring in the beginning? It doesn’t seem to have any purpose, so please remove it.

docs/userguide
```diff
@@ -2048,6 +2050,35 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
 seperate bindings for a specific set of labels and then only use those labels.
 ///////////////////////////////////////////////////////////////////
 
+=== Window title format
+
+By default, i3 will simply print the window's title. Using +title_format+, this can be customized
```
stapelberg: I think it might be good to mention that the application defines the title. In the criteria section of the userguide, we use this terminology:

> title
> Compares the X11 window title (_NET_WM_NAME or WM_NAME as fallback).

So, I think changing “window's title” to “X11 window title” here would be good for consistency.

docs/userguide
```diff
@@ -2048,6 +2050,35 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
...
+by setting the format to the desired output. This directive supports 
+https://developer.gnome.org/pango/stable/PangoMarkupFormat.html[Pango markup] and the following placeholders
+which will be replaced:
+
++%title+::
+    The window title.
```
stapelberg: See above. Change to “The X11 window title (_NET_WM_NAME or WM_NAME as fallback)” please.

docs/userguide
```diff
@@ -2048,6 +2050,35 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
 seperate bindings for a specific set of labels and then only use those labels.
 ///////////////////////////////////////////////////////////////////
 
+=== Window title format
+
+By default, i3 will simply print the window's title. Using +title_format+, this can be customized
+by setting the format to the desired output. This directive supports 
+https://developer.gnome.org/pango/stable/PangoMarkupFormat.html[Pango markup] and the following placeholders
```
stapelberg: It seems like the paragraphs you’re adding are not line-wrapped at 80 chars. Use gqq if you’re using vim :).

src/commands.c
```diff
@@ -1900,6 +1900,35 @@ void cmd_scratchpad_show(I3_CMD) {
...
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        if (current->con->window == NULL)
+            continue;
+
+        DLOG("setting title_format for %p / %s\n", current->con, current->con->name);
+        FREE(current->con->window->title_format);
+
+        /* If we only display the title without anything else, we can skip the parsing step,
+         * so we remove the title format altogether. */
+        if (strcmp(format, "%title") != 0)
```
stapelberg: This check is different from what `BEGINS_WITH` does — the macro uses `strncasecmp`, so use `strcasecmp` here as well.

src/commands.c
```diff
@@ -1900,6 +1900,35 @@ void cmd_scratchpad_show(I3_CMD) {
...
+            current->con->window->title_format = sstrdup(format);
+
+        /* Make sure the window title is redrawn immediately. */
+        current->con->window->name_x_changed = true;
```
stapelberg: This line might as well go into the previous if condition, right? If the title format is just %title, no redraw is needed.

src/x.c
```diff
@@ -302,6 +302,45 @@ void x_window_kill(xcb_window_t window, kill_window_t kill_window) {
     free(event);
 }
 
+static i3String *parse_title_format(char *format, i3String *_title) {
+    /* We need to ensure that we only escape the window title if pango
+     * is used by the current font. */
+    bool is_markup = font_is_pango();
```
stapelberg: nit: use const bool

src/commands.c
```diff
@@ -1900,6 +1900,35 @@ void cmd_scratchpad_show(I3_CMD) {
 }
 
 /*
+ * Implementation of 'title_format <format>'
+ *
+ */
+void cmd_title_format(I3_CMD, char *format) {
+    DLOG("setting title_format to %s\n", format);
```
stapelberg: nit: please use `setting title_format to \"%s\"` so that readers can more easily identify where the title starts/ends.

src/commands.c
Airblader: I think it is, because the title_format might have been altered before and is now being reset to the default. In this case we need the redraw, too, right?

stapelberg: True. Nevermind then.

Airblader:
@stapelberg Done.

docs/userguide
```diff
@@ -2066,6 +2068,36 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
...
+=== Window title format
+
+By default, i3 will simply print the X11 window title. Using +title_format+,
+this can be customized by setting the format to the desired output. This
+directive supports
+https://developer.gnome.org/pango/stable/PangoMarkupFormat.html[Pango markup]
+and the following placeholders which will be replaced:
+
++%title+::
+    The X11 window title.
```
stapelberg: This should read “The X11 window title (_NET_WM_NAME or WM_NAME as fallback).”.

docs/userguide
```diff
@@ -2066,6 +2068,36 @@ Alternatively, if you do not want to mess with +i3-input+, you could create
...
+
+Using the <<for_window>> directive, you can set the title format for any window based on <<command_criteria>>.
```
stapelberg: This is still not line-wrapped.

Airblader: Mh, I tried your shortcut in vim. For some reason it only seems to work if I select the lines explicitly and press gq. I'll look into that, sorry. :)

Airblader:
@stapelberg But now, hopefully. :)

PR #1768, opened 2015-06-26
Fix TODO in src/bindings.c

hwangcc23:
I found a TODO in bindings.c when studying i3 source code.
Send a pull request to fix it.

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -63,10 +63,11 @@ Binding *configure_binding(const char *bindtype, const char *modifiers, const ch
 
         new_binding->symbol = sstrdup(input_code);
     } else {
-        // TODO: strtol with proper error handling
-        new_binding->keycode = atoi(input_code);
+        char *endptr;
+        long keycode = strtol(input_code, &endptr, 10);
+        new_binding->keycode = keycode;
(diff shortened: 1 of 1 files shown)


src/bindings.c
```diff
@@ -63,10 +63,10 @@ Binding *configure_binding(const char *bindtype, const char *modifiers, const ch
 
         new_binding->symbol = sstrdup(input_code);
     } else {
-        // TODO: strtol with proper error handling
-        new_binding->keycode = atoi(input_code);
+        char *endptr;
+        new_binding->keycode = strtol(input_code, &endptr, 10);
         new_binding->input_type = B_KEYBOARD;
-        if (new_binding->keycode == 0) {
+        if (errno != 0 || *endptr != '\0' || endptr == input_code) {
```
stapelberg: As per the manpage `strtol(3)` (section NOTES), if you use `errno != 0`, you must first set it to 0 yourself, as `strtol` might not modify it at all.

FWIW, in other places in i3 (e.g. src/util.c:80), we check the return value, not errno. Maybe use the same approach for consistency?

hwangcc23:
@stapelberg I have updated code again. Please review. Thanks.

PR #1771, opened 2015-06-29
Implement new criterion 'workspace'.

### LINKED ISSUE #1769: enhancement: Workspace selection criteria
Implement workspace selection criteria. This would select any window in a particular workspace.

Examples:

Kill all the windows in workspace 4

    [workspace="4"] kill

Kill all firefox windows on workspace 4

    [workspace="4" class="Firefox"] kill

Airblader:
If the match expression is a plain number (e.g., '99'), the number of a workspace will be compared strictly. Otherwise, the match expression is taken as a regular expression and compared against the workspace's name.
This allows all of the following:

````
for_window [workspace=5] ...
for_window [workspace="5:foo"] ...
for_window [workspace="foo"] ...
```

fixes #1769

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1597,14 +1597,16 @@ urgent::
 	Compares the urgent state of the window. Can be "latest" or "oldest".
 	Matches the latest or oldest urgent window, respectively.
 	(The following aliases are also available: newest, last, recent, first)
+workspace::
+	Compares the workspace name of the workspace the window belongs to.
 con_mark::
 	Compares the mark set for this container, see <<vim_like_marks>>.
 con_id::
 	Compares the i3-internal container ID, which you can get via the IPC
 	interface. Handy for scripting.
 
-The criteria +class+, +instance+, +role+, +title+ and +mark+ are actually
-regular expressions (PCRE). See +pcresyntax(3)+ or +perldoc perlre+ for
+The criteria +class+, +instance+, +role+, +title+, +workspace+ and +mark+ are
+actually regular expressions (PCRE). See +pcresyntax(3)+ or +perldoc perlre+ for
 information on how to use them.
 
 [[exec]]
--- a/include/data.h
+++ b/include/data.h
@@ -417,6 +417,7 @@ struct Match {
     struct regex *instance;
     struct regex *mark;
     struct regex *window_role;
+    struct regex *workspace;
     xcb_atom_t window_type;
     enum {
         U_DONTCHECK = -1,
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -51,6 +51,7 @@ state CRITERIA:
   ctype = 'con_mark'    -> CRITERION
   ctype = 'title'       -> CRITERION
   ctype = 'urgent'      -> CRITERION
+  ctype = 'workspace'   -> CRITERION
   ']' -> call cmd_criteria_match_windows(); INITIAL
 
 state CRITERION:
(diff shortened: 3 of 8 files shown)


docs/userguide
```diff
@@ -1597,14 +1597,18 @@ urgent::
 	Compares the urgent state of the window. Can be "latest" or "oldest".
 	Matches the latest or oldest urgent window, respectively.
 	(The following aliases are also available: newest, last, recent, first)
+workspace::
+        Matches the workspace a window belongs to. If the match expression is a
+        plain number, it will match against a workspace's number. Otherwise it
```
stapelberg: I’m not sure this is a good idea. I recall that @acrisci introduced a feature where e.g. “workspace 9” would be equivalent to “workspace number 9” and it broke people’s configs, hence we rolled it back. Maybe we should use a `workspace` and a `workspace_num` criterion?

Airblader: What scenario exactly would not work here? It certainly wouldn't break anyone's config since it's a new feature. We can add separate criteria for it, but I think we should only do that if we actually have a scenario that wouldn't work as expected.

stapelberg: Well, some people have multiple workspaces that start with a number, e.g. “9” and “9: www”. With your current implementation, “9” might be unreachable, depending on the internal order of workspaces in memory.

Airblader: If I have two workspaces "9" and "9:foo", both have `num = 9`, correct? In that case the criterion `workspace=9` would work for windows on either workspace, which seems correct and expected to me. If I want only the first one, I can use `workspace="^9$"` and if I only want the second one I can use `workspace="9:foo"` or similar. So I'm still not sure about what is breaking here.

The reason I chose this solution is because otherwise `workspace=9` will match any workspace that contains a 9 in its name, which feels unexpected to me.

stapelberg: I agree with everything you wrote, but given how `workspace=^9` is effectively doing what you are implementing, I don’t think we should try to outsmart the user and instead document (using examples) how the different use-cases can be achieved by using regular expressions. I.e., I’d prefer it if we just treated the `workspace` criterion as another standard regular expression, without the extra logic you’re proposing.

If `workspace=9` matching anything containing a 9 is unexpected, I’d argue it’s a good time to fix one’s mental model about how regular expression matching in i3 works :). Aside from that, in the default case (workspaces range from 1-9), the correct thing will happen anyway.

Airblader: Agreed. Do we still want a separate criterion then or do we just go with `workspace` (in the form of a simple regex) for now?

Airblader:
@stapelberg I've removed the special handling now. If you do want a separate `workspace_num` criterion, let me know. I don't think it's necessary, though, since for unnamed workspaces the name will be the number.

docs/userguide
```diff
@@ -1597,14 +1597,16 @@ urgent::
 	Compares the urgent state of the window. Can be "latest" or "oldest".
 	Matches the latest or oldest urgent window, respectively.
 	(The following aliases are also available: newest, last, recent, first)
+workspace::
+        Matches the workspace a window belongs to by its name.
```
stapelberg: There seems to be a white-space difference in this line compared to the ones above and below it.

Also, all other descriptions start with “Compares”. I suggest “Compares the workspace of the window”.

Airblader: @stapelberg I worded it slightly different, but fixed the indentation. I don't really use tab stops, ever. :)

Issue #1785, opened 2015-07-06
thinkpad multimedia buttons not recognized


matthiaskrgr:
i3 does not recognize my thinkpad media buttons

In short, I added this to my config
```
bindsym XF86AudioRaiseVolume exec xmessage 'it works A'
bindsym XF86AudioLowerVolume exec xmessage 'it works B'

bindcode 122  exec xmessage 'it works C'
bindcode 123  exec xmessage 'it works D'
```
bug nothing happens upon pressing the buttons.

xev output:

```
KeyPress event, serial 29, synthetic NO, window 0x2200001,
    root 0xc4, subw 0x0, time 17921607, (199,510), root:(1003,530),
    state 0x0, keycode 122 (keysym 0x1008ff11, XF86AudioLowerVolume), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 32, synthetic NO, window 0x2200001,
    root 0xc4, subw 0x0, time 17921734, (199,510), root:(1003,530),
    state 0x0, keycode 122 (keysym 0x1008ff11, XF86AudioLowerVolume), same_screen YES,
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyPress event, serial 32, synthetic NO, window 0x2200001,
    root 0xc4, subw 0x0, time 17922589, (199,510), root:(1003,530),
    state 0x0, keycode 123 (keysym 0x1008ff13, XF86AudioRaiseVolume), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 32, synthetic NO, window 0x2200001,
    root 0xc4, subw 0x0, time 17922685, (199,510), root:(1003,530),
    state 0x0, keycode 123 (keysym 0x1008ff13, XF86AudioRaiseVolume), same_screen YES,
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False
```

xmodmap -pke : http://lpaste.net/3916315304078082048
full i3 config: http://lpaste.net/4355387581660135424
i3 log:  http://logs.i3wm.org/logs/5706163895140352.bz2
i3 -v:  i3 version 4.10.2 (2015-04-16, branch "4.10.2") © 2009-2014 Michael Stapelberg and contributors

[...]

PR #1786, opened 2015-07-08
Display which config is used in --more_version

hwangcc23:
Display the path of the used config and its last modified time in the output of i3 --moreversion.
See the issue #1785 for the discussion and the usage.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -582,11 +582,14 @@ human_readable (string)::
 	build date and branch name. When you need to display the i3 version to
 	your users, use the human-readable version whenever possible (since
 	this is what +i3 --version+ displays, too).
+loaded_config_file_name (string)::
+	The current config path.
 
 *Example:*
 -------------------
 {
    "human_readable" : "4.2-169-gf80b877 (2012-08-05, branch \"next\")",
+   "loaded_config_file_name" : "/home/hwangcc23/.i3/config",
    "minor" : 2,
    "patch" : 0,
    "major" : 4
--- a/src/display_version.c
+++ b/src/display_version.c
@@ -16,20 +16,25 @@
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <fcntl.h>
+#include <time.h>
 #include "all.h"
 
-static bool human_readable_key;
-static char *human_readable_version;
+static bool human_readable_key, loaded_config_file_name_key;
+static char *human_readable_version, *loaded_config_file_name;
 
 static int version_string(void *ctx, const unsigned char *val, size_t len) {
     if (human_readable_key)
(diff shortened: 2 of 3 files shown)


docs/ipc
```diff
@@ -582,11 +582,14 @@ human_readable (string)::
 	build date and branch name. When you need to display the i3 version to
 	your users, use the human-readable version whenever possible (since
 	this is what +i3 --version+ displays, too).
+used_config (string)::
```
stapelberg: Can you rename this to the more descriptive `loaded_config_file_name` please?

src/display_version.c
```diff
@@ -104,6 +109,16 @@ void display_running_version(void) {
 
     printf("\rRunning i3 version: %s (pid %s)\n", human_readable_version, pid_from_atom);
 
+    struct stat sb;
+    char mtime[64];
+    mtime[0] = '\0';
+    if (stat(used_config, &sb) == -1) {
+        ELOG("Cannot stat config file %s\n", used_config);
```
stapelberg: Please use `\"%s\"` to clearly separate the path in the error message.

src/display_version.c
```diff
@@ -104,6 +109,16 @@ void display_running_version(void) {
...
+    } else {
+        strftime(mtime, sizeof(mtime), "(last modified: %Y-%m-%d %H:%M)", localtime(&(sb.st_mtime)));
```
stapelberg: Instead of hardcoding a date/time format, please use `%c` which prints the time according to the current locale.

Also, can you make the code also print how many seconds ago that was? That way, when people provide the output, outsiders in different time zones can easily spot whether a stale config file was used.

src/display_version.c
```diff
@@ -104,6 +109,16 @@ void display_running_version(void) {
...
+    }
+    printf("\rUsed i3 config: %s %s\n", used_config, mtime);
```
stapelberg: The `\r` in the beginning is not necessary, it’s only required in the earlier message because we need to clear the progress message that is printed in line 72.

hwangcc23:
@stapelberg Done. Please review the patch again. Thanks.

src/display_version.c
```diff
@@ -104,6 +109,19 @@ void display_running_version(void) {
 
     printf("\rRunning i3 version: %s (pid %s)\n", human_readable_version, pid_from_atom);
 
+    struct stat sb;
```
stapelberg: Please introduce an `if` statement that skips this code if loaded_config_file_name is not present in i3’s JSON reply. This is to cover the case where the i3 binary is newer than the i3 binary the user is running.

hwangcc23:
@stapelberg Done. Please review it. Thanks.

PR #1789, opened 2015-07-10
Use safe wrappers wherever possible

shdown:


stapelberg:
I applaud the effort to use safe-wrappers consistently throughout the code base. Two questions:

1. Why did you change the signature of scalloc? Is there any benefit? As-is, this makes this PR pretty tedious to review.

2. Can you think of a way to avoid people directly using the non-safe-wrapped versions in the future, so that we don’t need to keep fixing this?

shdown:
> Why did you change the signature of scalloc?

To not break the pattern that wrappers take the same arguments that thier non-safe conterparts do. I guess that’s confusing. Also, calloc handles overflow on multiplication for us.

> Can you think of a way to avoid people directly using the non-safe-wrapped versions in the future, so that we don’t need to keep fixing this?

Macros?

````c
#define malloc(size) smalloc(size)
// ...
void *smalloc(size_t size) {
    void *result = (malloc)(size); // ← real malloc() is called
    if (result == NULL)
        err(EXIT_FAILURE, "malloc(%zd)", size);
    return result;
}

````

stapelberg:
Instead of redefining the function to mean something else, could we make it so that plain malloc is marked as deprecated and smalloc isn’t?

shdown:
Yep.

````c
void *smalloc(size_t size) {
    void *result = malloc(size);
    if (result == NULL)
        err(EXIT_FAILURE, "malloc(%zd)", size);
    return result;

}
// until this, malloc was not marked deprecated
void *malloc(size_t size) __attribute__((deprecated)); // now it is
````

stapelberg:
Can you update your PR to include a commit marking all non-safe-wrapped functions as deprecated please?

shdown:
Should I deprecate `asprintf()` and `vasprintf()`? They are used in some places when there is no plan to exit if memory allocation failed.

stapelberg:
Well, we have sasprintf, so it’s safe to deprecate asprintf. For vasprintf, that’s only used in src/sd-daemon.c (with error checking) and the safe-wrapper itself and i3bar. I think it’s better to just fix the i3bar usage, as it’s the only usage.

shdown:
If I put those declarations into `include/libi3.h`, gcc warns about the use of deprecated functions in the wrappers. Is it OK to put `#pragma GCC diagnostic ignored "-Wdeprecated-declarations"` into `safewrappers.c`?

stapelberg:
Yeah, I think that’s fine, since the file only contains wrappers, so there is little chance other functions will be deprecated that we’d miss because of that.

shdown:
It fails to compile with `-D_FORTIFY_SOURCE=2`, which is default in Debian, because `strndup` becomes a macro.

stapelberg:
Hm. Maybe instead of trying to verify this at compile-time, we could move that responsibility to Travis and come up with a pre-submit-time check? Perhaps just grepping for all non-wrapped functions (excluding the safewrappers.c file itself) would work?

shdown:
Done. What’s left is to eliminate use of `asprintf`:
````
    if (asprintf(&startup_id, "%.*s", xcb_get_property_value_length(startup_id_reply),
^ ./src/startup.c calls a function that has a safe counterpart.
    if (asprintf(&new_role, "%.*s", xcb_get_property_value_length(prop),
^ ./src/window.c calls a function that has a safe counterpart.
        if (asprintf(&content, "%u", *((unsigned int *)xcb_get_property_value(prop_reply))) == -1) {
        if (asprintf(&content, "%.*s", xcb_get_property_value_length(prop_reply),
^ ./libi3/root_atom_contents.c calls a function that has a safe counterpart.
````
Is it OK to just replace `asprintf` with `sasprintf` in the cases above?

stapelberg:
Sorry for the late reply. For `src/{startup,window}.c`, using sasprintf is okay. For libi3/, I’d prefer it if we could keep asprintf+check, because libraries are not supposed to ever terminate execution.

shdown:
OK, I changed `asprintf`s to `sasprintf`s in `src/{startup,window}.c`.

> For libi3/, I’d prefer it if we could keep asprintf+check, because libraries are not supposed to ever terminate execution.

I’d disagree: libi3 is only used internally, and it already uses safe wrappers.

stapelberg:
> it already uses safe wrappers.

That’s a fair point. Go ahead and convert it, then.

shdown:
Done.

stapelberg:
Thanks!

PR #1796, opened 2015-07-14
i3bar/child: Tell shell to always exec child

sardemff7:
Fixes an orphan issue when:
- Shell does not exec status_command (like dash)
- status_command is not using SIGSTOP/SIGCONT (like j4status)

Signed-off-by: Quentin Glidic <sardemff7+git@sardemff7.net>

### DIFF
--- a/i3bar/include/child.h
+++ b/i3bar/include/child.h
@@ -46,7 +46,7 @@ typedef struct {
  * about arguments and such
  *
  */
-void start_child(char *command);
+void start_child(const char *command);
 
 /*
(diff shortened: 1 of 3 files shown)


stapelberg:
This might break user configs. If the user specified e.g. `(echo foo | myscript.sh)` as status_command, `exec (echo foo | myscript.sh)` is a syntax error.

I think it might be better to update the docs to suggest using exec.

sardemff7:
I was updating the doc and just saw that there is already a dash-specific hint on using `exec`. Closing.

Issue #1749, opened 2015-06-11
Config file line continuation


acrisci:
Config files should have line continuation, similar to bash, which is indicated by `\` before the new line character. This would allow users to break up longer commands into multiple lines for better organization of the config file and allow for longer `exec` lines to be used without needing to use a script.

Examples:

```
bindsym $mod+r exec --no-startup-id \
    dmenu_run -fn 'Fira Mono -20' -f -i
```

```
bindsym $mod+Tab \
    move container to output right; \
    focus output right;
```
```
bindsym Mod1+Shift+e exec \
    "test -f /tmp/i3-nagbar$DISPLAY.lock || \
    (touch /tmp/i3-nagbar$DISPLAY.lock && \
    i3-nagbar -t warning -m 'You pressed the exit shortcut. Do you really want to exit i3? This will end your X session.' \
    -b 'Yes, exit i3' 'i3-msg exit' && \
    rm /tmp/i3-nagbar$DISPLAY.lock)"
```

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
I like this. It'd also be really useful for pango markup.

stapelberg:
Sounds good.

parnmatt:
:+1:

acrisci:
Fixed in #1797

Issue #20, opened 2009-04-04
Using Mod1+Workspacenumber does not change focus when changing monitor


i3bot:
[Originally reported by Mirko Augsburger <mirko@…>]
(When changing to another monitor with Alt+5 e.g. the focus is not set correctly. So no window bar is highlighted but the old window is the one where the keystrokes go. The window on the new workspace should be the one with focus.

PR #1797, opened 2015-07-14
Support config file line continuation

hwangcc23:
Allow to use the line continuation, which is indicated by \ before the new line character, in config files.
See the issue #1749 for the discussion and the usage.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1078,6 +1078,19 @@ show_marks yes|no
 show_marks yes
 --------------
 
+[[line_continuation]]
+
+=== Line continuation
+
+Config files support line continuation, which is indicated by \ before the
+new line character.
+
+*Examples*:
+-------------------
+bindsym Mod1+f \
+fullscreen toggle
+-------------------
+
 == Configuring i3bar
 
 The bar at the bottom of your monitor is drawn by a separate process called
--- a/src/config_parser.c
+++ b/src/config_parser.c
@@ -837,11 +837,11 @@ static char *migrate_config(char *input, off_t size) {
  */
 bool parse_file(const char *f, bool use_nagbar) {
     SLIST_HEAD(variables_head, Variable) variables = SLIST_HEAD_INITIALIZER(&variables);
-    int fd, ret, read_bytes = 0;
+    int fd;
     struct stat stbuf;
     char *buf;
     FILE *fstr;
-    char buffer[1026], key[512], value[512];
+    char buffer[4096], key[512], value[512], *continuation = NULL;
 
     if ((fd = open(f, O_RDONLY)) == -1)
         die("Could not open configuration file: %s\n", strerror(errno));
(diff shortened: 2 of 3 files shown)


Airblader:
Cool! I think it'd be great to also test these:
* Line continuations within a string
* Making sure string escaping still works

hwangcc23:
@stapelberg I will send the PR again. Please don't review this one.
@Airblader Thanks.

hwangcc23:
@Airblader Done.
@stapelberg Please help review the patch. Thanks.

src/config_parser.c
```diff
@@ -837,11 +837,11 @@ static char *migrate_config(char *input, off_t size) {
  */
 bool parse_file(const char *f, bool use_nagbar) {
     SLIST_HEAD(variables_head, Variable) variables = SLIST_HEAD_INITIALIZER(&variables);
-    int fd, ret, read_bytes = 0;
+    int fd;
     struct stat stbuf;
     char *buf;
     FILE *fstr;
-    char buffer[1026], key[512], value[512];
+    char buffer[1026], key[512], value[512], *continuation = NULL;
```
stapelberg: Maybe we should increase the buffer size so that people won’t hit that length limit when doing complex things in their config file. Can you raise it to 4096 please?

Also, instead of using 1024 (which, curiously, mismatches the 1026 bytes above), please use `sizeof(buffer)`.

In order to gracefully fail when users use continuation with _many_ lines, can you please also add an error message for the case where `buffer` is filled up, i.e. we would pass `size=0` to `fgets`?

hwangcc23:
@stapelberg Done. Please help review the patch. Thanks.

src/config_parser.c
```diff
@@ -850,24 +850,36 @@ bool parse_file(const char *f, bool use_nagbar) {
...
+                buffer[511] = '\0';
+                ELOG("Variable's name/value overflow\n");
+            } else
+                continue;
+        }
+
+        strncpy(buf + strlen(buf), buffer, strlen(buffer) + 1);
 
         /* sscanf implicitly strips whitespace. Also, we skip comments and empty lines. */
         if (sscanf(buffer, "%s %[^\n]", key, value) < 1 ||
```
stapelberg: Let’s use `sscanf(buffer, "%511s %511[^\n]", key, value)` here instead of the if conditional you added above. The `if (strncasecmp(buffer, "set ", strlen("set ")) == 0) {` is not _exactly_ the same as how sscanf handles it, because of sscanf’s more liberal whitespace handling.

hwangcc23:
@stapelberg Done. Thanks.  :)

docs/userguide
```diff
@@ -2457,3 +2457,16 @@ and you are in multi-monitor mode (see <<multi_monitor>>).
 Because i3 is not a compositing window manager, there is no ability to
 display a window on two screens at the same time. Instead, your presentation
 software needs to do this job (that is, open a window on each screen).
+
+[[line_continuation]]
+
+=== Line continuation
```
stapelberg: I just realized you appended this section to the very end of the userguide, but I think it’d be better if it was in the “Configuring i3” section, i.e. if it became section 4.25.

hwangcc23:
@stapelberg Done. Move it to the end of the "Configuraing i3" section.

testcases/t/247-config-line-continuation.t
```diff
@@ -0,0 +1,201 @@
...
+0090-This is a very very very very very very very very very very very very very very very very very long cmd \
+0091-This is a very very very very very very very very very very very very very very very very very long cmd \
+0092-This is a very very very very very very very very very very very very very very very very very long cmd \
+0093-This is a very very very very very very very very very very very very very very very very very long cmd \
+0094-This is a very very very very very very very very very very very very very very very very very long cmd \
+0095-This is a very very very very very very very very very very very very very very very very very long cmd \
+0096-This is a very very very very very very very very very very very very very very very very very long cmd \
+0097-This is a very very very very very very very very very very very very very very very very very long cmd \
+0098-This is a very very very very very very very very very very very very very very very very very long cmd \
+0099-This is a very very very very very very very very very very very very very very very very very long cmd
```
acrisci: lol

acrisci:
I think this commit has caused a problem with i3-style [#20](https://github.com/acrisci/i3-style/issues/20).

Issue #1702, opened 2015-05-11
i3*.xession.desktop should allow a hook


RoUS:
Feature request: Have i3's Xsession `.desktop` files list a script as the `Exec` value, rather than the `i3`binary.

I'm using i3 on Fedora 21:

    $ rpm -q i3
    i3-4.9.1-1.fc21.x86_64
    $ i3 -v
    i3 version 4.9.1 (2015-03-07, branch "4.9.1") © 2009-2014 Michael Stapelberg and contributors

The `i3*.xsession.desktop` files contain `Exec=i3` and `TryExec=i3`, which means the binary will be directly `exec`ed by the Xsession sequence.

Unfortunately, that means the *only* hook for user intervention in this scenario (which is apparently where all of the Xsession stuff is now going) is the user's `$HOME/.profile` login script.  That gets invoked by `/etc/X11/xinit/xinitrc-common` and no other user script is in the sequence until the `.desktop` file's `Exec` is processed.

My feature request is that the `.desktop` files list a script as the {Try.}Exec key value, and that the script look for a user hook script to source before `exec`ing the i3 binary.

Airblader:
What do you need to execute before starting i3?

RoUS:
Things like setting the `PATH`, starting `gpg-agent`, `urxvtd`, `nm-applet` -- pretty much anything that someone might have put into `~/.xsession` is ye elder days.  Stuff that should be in `i3`'s environment when it actually starts.

Airblader:
Why not put it into `~/.xinitrc`? Shouldn't that be executed?

RoUS:
It isn't.  The gdm/Xsession model apparently goes straight to whatever was passed on its command line, which gets picked up from the .desktop files. :-(

Airblader:
What version of gdm is Fedora 21 using? gdm not respecting the proper startup files seems like a bug to me. According to [this](http://webcache.googleusercontent.com/search?q=cache:p9C-AeXU6DkJ:subtle.subforge.org/blogs/show/13+&cd=4&hl=en&ct=clnk&gl=us), gdm had this bug a few years back. According to [this](https://wiki.archlinux.org/index.php/Xprofile), at least newer versions pick up on `~/.xinitrc`.

RoUS:
[...]

PR #1801, opened 2015-07-18
Add setenv command

Merovius:
Enables users to change the environment of a running i3 session as well as changing the environment in the config file (so that users don't need to do that in their display manager and/or don't need a .xsession).

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2241,6 +2241,27 @@ bindsym $mod+minus scratchpad show
 bindsym mod4+s [title="^Sup ::"] scratchpad show
 ------------------------------------------------
 
+=== Setting environment variables
+
+There is a command to set environment variables for the i3 process. This is
+useful if you want some environment for everything in your i3 session but
+don't have the means to do so from your display manager (or what you might
+otherwise use to start i3). Environment variables in the value are expanded.
+
+*Syntax*:
+--------------------
+setenv <key> <value>
+--------------------
+
+*Example*:
+
+---------------------------
+# Set language
+setenv LANG=de_DE.utf8
+# Amend PATH by ~/bin
+setenv PATH=$PATH:$HOME/bin
+---------------------------
+
 === Nop
 
 There is a no operation command +nop+ which allows you to override default
--- a/src/commands.c
+++ b/src/commands.c
@@ -11,6 +11,7 @@
  */
 #include <float.h>
 #include <stdarg.h>
+#include <wordexp.h>
 
 #include "all.h"
 #include "shmlog.h"
@@ -2176,5 +2177,41 @@ void cmd_debuglog(I3_CMD, char *argument) {
         set_debug_logging(false);
     }
     // XXX: default reply for now, make this a better reply
+    ysuccess(true);
+}
+
+/*
+ * Implementation of 'setenv <key> <value>'
+ *
+ */
+void cmd_setenv(I3_CMD, char *key, char *value) {
+    wordexp_t p;
+
+    switch (wordexp(value, &p, WRDE_NOCMD)) {
+        case 0:
+            break;
+        case WRDE_BADCHAR:
+            yerror("Bad character in expansion of \"%s\" in setenv command\n", value);
+            return;
+        case WRDE_CMDSUB:
+            yerror("Command substitution not allowed in setenv\n");
+            return;
+        case WRDE_NOSPACE:
+            yerror("Out of memory when expanding \"%s\"\n", value);
+            return;
+        case WRDE_SYNTAX:
+            yerror("Syntax error when expanding \"%s\"\n", value);
+            return;
+        default:
+            yerror("Unknown error when expanding \"%s\"\n", value);
+            return;
+    }
+
+    if (p.we_wordc == 0) {
+        setenv(key, value, 1);
+    } else {
+        setenv(key, p.we_wordv[0], 1);
+    }
+    wordfree(&p);
     ysuccess(true);
 }
--- a/testcases/t/526-setenv.t
+++ b/testcases/t/526-setenv.t
@@ -1,0 +1,66 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • http://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • http://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • http://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Tests if a 'move <key> <value>' command will set the environment variable <key> to <value>.
+#
+use i3test;
+
+use POSIX qw(mkfifo);
+use File::Temp qw(:POSIX tempfile);
+
+#####################################################################
+# Try to set FOOBAR to baz
+#####################################################################
+
+cmd('setenv FOOBAR baz');
+
+my $fifo = tmpnam();
+mkfifo($fifo, 0600) or BAIL_OUT("Could not create FIFO in $fifo");
+
+cmd("exec echo \$FOOBAR > $fifo");
+
+open(my $fh, '<', $fifo);
+# Block on the FIFO, this will return exactly when the command is done.
+my $text = <$fh>;
+chomp($text);
+close($fh);
+unlink($fifo);
+
+is($text, "baz", "simple setenv command");
+
+#####################################################################
+# Try to append to FOOBAR, to test expansion
+#####################################################################
+
+cmd('setenv FOOBAR foo$FOOBAR');
+
+cmd("exec sh -c 'echo \$FOOBAR > $fifo'");
+
+$fifo = tmpnam();
+mkfifo($fifo, 0600) or BAIL_OUT("Could not create FIFO in $fifo");
+
+cmd("exec echo \$FOOBAR > $fifo");
+
+open($fh, '<', $fifo);
+# Block on the FIFO, this will return exactly when the command is done.
+$text = <$fh>;
+chomp($text);
+close($fh);
+unlink($fifo);
+
+is($text, "foobaz", "setenv expansion");
+
+done_testing;
--- a/include/commands.h
+++ b/include/commands.h
@@ -305,3 +305,9 @@ void cmd_shmlog(I3_CMD, char *argument);
  *
  */
 void cmd_debuglog(I3_CMD, char *argument);
+
+/*
+ * Implementation of 'setenv <key> <value>'
+ *
+ */
+void cmd_setenv(I3_CMD, char *key, char *value);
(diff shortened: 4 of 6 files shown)


docs/userguide
```diff
@@ -2241,6 +2241,27 @@ bindsym $mod+minus scratchpad show
 bindsym mod4+s [title="^Sup ::"] scratchpad show
 ------------------------------------------------
 
+=== Setting environment variables
+
+There is a command to set environment variables for the i3 process. This is
+usefull if you want some environment for everything in your i3 session but
```
Airblader: nit: s/usefull/useful/

docs/userguide
```diff
@@ -2241,6 +2241,27 @@ bindsym $mod+minus scratchpad show
...
+--------------------
+setenv <key> <value>
+--------------------
+
+*Example*:
+
+---------------------------
+# Set language
+setenv LANG=de_DE.utf8
+# Ammend PATH by ~/bin
```
Airblader: nit: s/Ammend/Amend/

Airblader:
Couldn't I just do this in my `~/.xinitrc`? That file will be used by `startx`, so I don't need a display manager for it.

Merovius:
Of course, if you are using startx to start your i3, you can add it to the .xinitrc. But not everyone is (and I actually think that using a display manager is probably one of the more common ones).

But even if we were to assume that anyone could find some other way to set this up: As this is a command, it also makes it possible to change the environment of a *running* i3 session. If you set your environment from your display manager, or your .xsession, or your .xinitrc, you would necessarily need to restart your i3 for the change to take affect (or [do weird hacks](http://blog.merovius.de/2013/10/11/inject-environment-variables-int.html)). And you can't even do an in-place restart, as that just does an exec, which uses the same process (and thus has the same environment as before).

So I do believe that there are valid usecases for this change :)

acrisci:
I like this feature.

I haven't looked too closely at the code yet though.

Merovius:
One note: The code only calls wordfree(), when wordexp() was successfull. The manpage doesn't specify if that's correct (i.e. if wordexp() might allocate even when failing), but [gem](http://sources.debian.net/src/gem/1:0.93.3-8/src/Gem/Files.cpp/?hl=105#L105) does the same, so I assume that's the correct behavior. I have also found some sources, where the result of wordexp isn't checked and wordfree is called unconditionally. I think that produces undefined behavior, though (as the manpage also doesn't define what p contains if wordexp fails). I love C.

src/commands.c
```diff
@@ -2178,3 +2179,38 @@ void cmd_debuglog(I3_CMD, char *argument) {
...
+            ELOG("Unknown error when expanding \"%s\"\n", value);
+            return;
+    }
+
+    if (p.we_wordc == 0) {
+        setenv(key, value, 1);
+    } else {
+        setenv(key, p.we_wordv[0], 1);
+    }
+    wordfree(&p);
```
acrisci: It needs to call ysuccess() to indicate whether it succeeded or failed.

testcases/t/526-setenv.t
```diff
@@ -0,0 +1,66 @@
...
+#####################################################################
+
+cmd('setenv FOOBAR foo$FOOBAR');
+
+cmd("exec sh -c 'echo \$FOOBAR > $fifo'");
+
+$fifo = tmpnam();
+mkfifo($fifo, 0600) or die "Could not create FIFO in $fifo";
+
+cmd("exec sh -c 'echo \$FOOBAR > $fifo'");
```
acrisci: The `sh -c` is redundant. It sould be tested without it.

testcases/t/526-setenv.t
```diff
@@ -0,0 +1,66 @@
...
+use File::Temp qw(:POSIX tempfile);
+
+#####################################################################
+# Try to set FOOBAR to baz
+#####################################################################
+
+cmd('setenv FOOBAR baz');
+
+my $fifo = tmpnam();
+mkfifo($fifo, 0600) or die "Could not create FIFO in $fifo";
```
acrisci: use `BAIL_OUT` instead of `die` or the testcases will not report the error clearly.

testcases/t/526-setenv.t
Merovius: Indeed. Remnant of some iterations of trying to get the test to pass.

src/commands.c
Merovius: like so? :)

acrisci: Yeah except without the newline :)

stapelberg:
Actually, I think two use-cases are conflated here:

1. Modifying the environment of the running i3 process, a one-off operation. This is what you implemented, because you introduced a command.

2. Configuring i3 to always set a specific environment variable, which would be best done by a configuration directive.

In the docs, you’re citing “you cannot do it from your DM” as the motivation, but that use-case would be better covered by a configuration directive. To cover that use-case with a command, you’d need to use `exec i3-msg setenv <key>=<value>` in your i3 config, and that would be prone to race-conditions, i.e. would make using other exec statements impossible.

In general, I’m not convinced that the window manager is the correct layer to do such configuration. Environment variables are not window manager specific, they are just passed through because of how the process hierarchy in X11 works. If you were to switch your window manager, you wouldn’t want to lose your environment variable adjustments.

For one-off cases where you really need to set/change an environment variable without restarting your session, I think it might bottom-line be preferable to just modify your i3 config to use `exec <key>=<value> i3-sensible-terminal`, i.e. modify your exec bindings to set that variable temporarily, until you get around to restarting your session.

What do you think? Do you have a convincing use-case? I think there’s some hack value in your pull request, but that doesn’t mean we should accept a rarely-used edge case feature into i3 :).

Merovius:
I'm uncertain about the precise differences between configuration directives and commands in regards to this feature. I thought it was possible to put commands inside your config-file (exec is a command right? I use exec to start stuff that definitely runs when I start i3). I considered it a command, because that's just what it is to me: It calls a library function, that changes something about the process, it *does* something :) (no, that's not a very strong defense, just an explanation for my initial choice) I am ultimately indifferent (though I agree that needing to do a exec i3-msg setenv… would be… sad), but I consider it important (if anything) that this can happen both at runtime *and* at configuration time. If need be, I would rather add a configuration directive *and* a command, than take the difference as a reason to only support one of the two usecases.

When I need to add an environment variable permanently to my session (notable examples where that was needed in the past for me personally are RXVT_SOCKET, AWT_TOOLKIT, GTK_IM_MODULE, PATH, GOPATH and localization-envs, so quite a bunch) I'd find it unpleasant to a) need to update every exec directive in my config file to reflect the added variable, b) add it to my .xsession and c) remember at *some* point in the distant feature, when I next reboot (which is when my the next non-inplace-restart usually happens) to again update my i3 config to remove all the now superfluous environment variables. Compare this with a) putting a setenv foo bar in my i3 config and b) doing a one-off i3-msg setenv foo bar (or just restart i3 in-place). I guess this ultimately comes down to frequency of use (see below for that).

> In general, I’m not convinced that the window manager is the correct layer to do such configuration. Environment variables are not window manager specific, they are just passed through because of how the process hierarchy in X11 works.

I agree. I also do think (and have thought so many times before), that the process-hierarchy approach of unix in general and environment variables in particular is very elegant, but completely breaks down in the face of X-sessions and the like. Do you have a good way to solve that? Because I haven't found one so far. In theory, a logind + systemd-user-session setup seems promising for this kind of stuff but I have no idea if it is in practice and I was unsuccessfull in setting it up so far.

In the abscense of a better solution, I'm leaning towards this (just like e.g. weird terminal execution hacks for i3-nagbar) maybe being a necessary evil to still put there.

> If you were to switch your window manager, you wouldn’t want to lose your environment variable adjustments.

True. But the same goes for my display manager (or whatever I use to start i3). I am unconvinced, that the frequency of one change overwhelms the frequency of the other (for example, I have not changed my wm in 5 years or so, whereas in the same timeframe I have changed my display manager two or three times on my laptop alone and used 5 or so on different machines in total).

Regarding the frequency with which this is needed: I have witnessed a bunch of new installations of i3 in the last couple of years and in a lot of them, environment-variables where the *only* reason, we had to write an .xssession. And while that's not a huge problem, it is kind of annoying.
I consider this feature to be pretty much equivalent to export in a shell (well, that's what it is) and I've been long of the opinion that everything long-running that launches processes, should have some kind of interface to influence the environment of the started processes. So, i3 as a window manager probably doesn't need it, but i3 as an application launcher does, I think.

Airblader:
For reference, we had a fairly similar discussion in #1702.

stapelberg:
> I'm uncertain about the precise differences between configuration directives and commands in regards to this feature. I thought it was possible to put commands inside your config-file (exec is a command right? I use exec to start stuff that definitely runs when I start i3).

`exec` is both a command and a configuration directive, see http://i3wm.org/docs/userguide.html#exec and http://i3wm.org/docs/userguide.html#_automatically_starting_applications_on_i3_startup

> I also do think (and have thought so many times before), that the process-hierarchy approach of unix in general and environment variables in particular is very elegant, but completely breaks down in the face of X-sessions and the like. Do you have a good way to solve that?

“good way” is subjective, but I presented the feasible solution of changing your exec command. It might be easier (depending on your taste) to use e.g. `exec $HOME/envwrap.sh i3-sensible-terminal` and modify `~/envwrap.sh` everytime you want to add an environment variable.

> But the same goes for my display manager (or whatever I use to start i3).

Yes, which is why you don’t customize your session in your DM either — the customization belongs into xsession, which is what all DMs can execute, and from where you can execute all window managers.

> So, i3 as a window manager probably doesn't need it, but i3 as an application launcher does, I think.

This is a very good point: I think another way of phrasing the debate we’re having is “how convenient shall we make the different aspects of launching processes from i3”.

The trivial case (“i3 should not be able to launch processes at all”) is clearly undesirable, as the user interface of i3 in and of itself is, in its default configuration, not something you can meaningfully interact with. Hence, i3 must have a way to launch processes, which is why the `exec` command exists. (As an aside, the `exec` configuration directive exists so that you have a way to run commands once the window manager is fully initialized, which might be necessary occasionally — personally, I do my entire session customization in xsession, because I have no such commands.)

The other extreme would be to compare i3 to e.g. an init system, whose most important job it is to launch (and supervise, etc.) processes. init systems comes with a bunch of configuration directives to influence environment variables, permissions, cgroups, etc. You could make a similar case for each of them, e.g. “we should have an i3 command to change the cgroups of future processes”.

With these two cases in mind, I interpret this pull request as “environment variables are an important-enough aspect of launching processes that methods for their manipulation should be made a first-class citizen in i3”.

I understand that your habits (very rarely restarting your session) make having this feature very compelling, but I’m not convinced this habit is representative of our user base at large. Specifically, I’m trying hard to keep the mental complexity of i3’s surface area with regards to commands and config directives as small as possible. I think having new users be able to understand i3 in its entirety is more valuable than making edge cases easier — a philosophy you might recognize from golang :).

Currently, the latter argument carries more weight for me. What do you think?

Merovius:
I think it's basically a judgement call and as I don't have any personal stakes in this (personally, I have an .xsession and gdb to make me happy), I probably shouldn't be the one to make it. Close this or merge it, I'm fine either way :)

stapelberg:
Fair enough, closing it for the reasons explained earlier.

PR #1803, opened 2015-07-18
Use BAIL_OUT instead of die in testcase 180

Merovius:


stapelberg:
Why?

acrisci:
I put `die` at the beginning of `t/222-regress-dock-resize.t`.

```
# 

#   Failed test 'testcase called exit_gracefully()'
#   at /home/tonyctl/projects/i3/testcases/lib/TestWorker.pm line 134.


Test Summary Report
-------------------
Use of uninitialized value in subtraction (-) at /usr/share/perl5/core_perl/TAP/Formatter/Base.pm line 386.
Negative repeat count does nothing at /usr/share/perl5/core_perl/TAP/Formatter/Base.pm line 386.
t/222-regress-dock-resize.t (Wstat: 0 Tests: 1 Failed: 1)
  Failed test:  1
  Parse errors: No plan found in TAP output
Use of uninitialized value in subtraction (-) at /usr/share/perl5/core_perl/TAP/Formatter/Base.pm line 386.
Negative repeat count does nothing at /usr/share/perl5/core_perl/TAP/Formatter/Base.pm line 386.
t/208-regress-floating-criteria.t (Wstat: 0 Tests: 2 Failed: 1)
  Failed test:  2
  Parse errors: Tests out of sequence.  Found (2) but expected (1)
                Tests out of sequence.  Found (1) but expected (2)
                Bad plan.  You planned 1 tests but ran 2.
Files=181, Tests=2335,  6 wallclock secs ( 0.77 usr +  0.01 sys =  0.78 CPU)
Result: FAIL
```

Compare this to `BAIL_OUT` at the top of the same test.

```
test t/222-regress-dock-resize.t bailed out: bye
```

The message is "bye". It is not reported in the first case. It is reported very clearly in the second.

stapelberg:
Can we make the error message better when tests die instead of never having tests die? :) The effect of using BAIL_OUT is that all testing is stopped immediately, which might hide valuable information about how the tests went. For example, the difference between “that one test is broken” and “I just broke 50 tests” is valuable in judging where the bug in your code changes might be.

acrisci:
`BAIL_OUT` I think should never be used when there is an i3 error for the reason you said. There is only one case where it is used that way in `t/117-workspace.t` and I think that is a mistake.

In this case, it is being used because of a system error that is unrelated to any changes you made in i3. It can't create the named pipe. I'm not exactly sure why it can't create the named pipe, but it actually seems to happen with some frequency.

stapelberg:
I think it’s somewhat hard to draw a clear line between “error that is caused by i3, in whichever indirect way” and “error that is caused by the system”.

With regards to the usage of `BAIL_OUT` in `t/117-workspace.t`, I’m not entirely sure — it’s an edge case where you can argue both ways. It certainly wasn’t a mistake, as the error message even clarifies the intention was to indeed stop all tests if you cannot create a workspace, as then virtually all other tests will fail :).

If creating the named pipe fails with some frequency, we should investigate the root cause of that. Regardless, I still think there’s value in improving the error message that gets printed when tests die.

Merovius:
I'm going to close this for now, it was a reaction to a comment on #1801. Once consensus is reached I'm sure you'll be able to recreate this PR :)

PR #1816, opened 2015-07-28
Changes for compiling i3 on Illumos

tcreech:
Very minor changes for compiling i3 on Illumos/Solaris.

The main issue was a collision between libi3's `mkdirp` and that of libgen on Illumos/Solaris. Here I have resolved this by renaming `mkdirp`. (I'm hoping that it is acceptable since libi3 is internal.)

### DIFF
--- a/i3-config-wizard/main.c
+++ b/i3-config-wizard/main.c
@@ -799,7 +799,7 @@ int main(int argc, char *argv[]) {
     struct stat stbuf;
     sasprintf(&config_dir, "%s/i3", xdg_config_home);
     if (stat(config_dir, &stbuf) != 0)
-        if (!mkdirp(config_dir))
+        if (mkdirp(config_dir, DEFAULT_DIR_MODE) != 0)
             err(EXIT_FAILURE, "mkdirp(%s) failed", config_dir);
     free(config_dir);
     free(xdg_config_home);
--- a/libi3/mkdirp.c
+++ b/libi3/mkdirp.c
@@ -8,24 +8,26 @@
  * Emulates mkdir -p (creates any missing folders)
  *
  */
-bool mkdirp(const char *path) {
-    if (mkdir(path, S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH) == 0)
-        return true;
+
+#if !defined(__sun)
+int mkdirp(const char *path, mode_t mode) {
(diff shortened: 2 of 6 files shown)


stapelberg:
I’d prefer it if we could avoid making the names of functions longer just to reduce the likelihood of colliding :).

According to http://illumos.org/man/3gen/mkdirp, mkdirp on Illumos does the same thing that ours is doing. So, I think using a conditional (like we already do with strndup, which is only necessary `#if defined(__APPLE__)`) in include/libi3.h and libi3/mkdirp.c is a better solution. You’ll also need to add the mode parameter to our implementation and make the function return an int.

tcreech:
Ok, thanks. Sounds good. I'll update the commits to conditionally build mkdirp when not on Illumos.

tcreech:
Ok, I've updated and squashed the changes.

libi3/mkdirp.c
```diff
@@ -38,13 +39,14 @@ bool mkdirp(const char *path) {
...
         }
-        return false;
+        return -1;
     }
     *sep = '\0';
-    bool result = false;
-    if (mkdirp(copy))
-        result = mkdirp(path);
+    int result = -1;
+    if (mkdirp(copy, mode))
```
stapelberg: This should read `if (mkdirp(copy, mode) == 0)`, right?

i3-config-wizard/main.c
```diff
@@ -799,7 +799,7 @@ int main(int argc, char *argv[]) {
     struct stat stbuf;
     sasprintf(&config_dir, "%s/i3", xdg_config_home);
     if (stat(config_dir, &stbuf) != 0)
-        if (!mkdirp(config_dir))
+        if (mkdirp(config_dir, DEFAULT_DIR_MODE))
```
stapelberg: Here and below, please make this `if (mkdirp(config_dir, DEFAULT_DIR_MODE) != 0)`

libi3/mkdirp.c
```diff
@@ -6,26 +6,27 @@
 
 /*
  * Emulates mkdir -p (creates any missing folders)
- *
```
stapelberg: Why did you remove this line from the comment?

tcreech:
@stapelberg: thanks for having a look. I think I've addressed your comments.

PR #1828, opened 2015-08-03
Ignoring desktop windows and moving them on the bottom of the stack.

pronobis:
This is important when i3 is used together with KDE Plasma Desktop.


pronobis:
That's the other important change I have that makes i3 usable with plasma. Without it, the desktop becomes a floating window.

stapelberg:
Can you squash your commits please?

Also, can one not disable the desktop windows entirely in plasma?

pronobis:
Squashed. Regarding the desktop window, I am not sure to be honest if it's possible to disable it. The thing is that having it enabled is actually not a bad thing as it also gives you a meaningful plasma-integrated desktop that is not conflicting with i3. I don't see any harm in making a window manager treat desktop windows as desktop windows in general. With that commit, the only task that needs to be done to enable i3 as WM in plasma is setting the env. var `KDEWM=i3`. That's all.

stapelberg:
> The thing is that having it enabled is actually not a bad thing as it also gives you a meaningful plasma-integrated desktop that is not conflicting with i3

I’m worried about whether this is always true, i.e. whether just placing the window as you do it in your PR is enough for making all the features and edge-cases work. Further, since this is a fairly unusual setup, supporting desktop windows officially means one more hard-to-test and easy-to-break code path in i3.

> Regarding the desktop window, I am not sure to be honest if it's possible to disable it.

Can you find out please? :)

pronobis:
The standard is clear about what that flag means, but doesn't really
specify what wm should do with such window. I would not suggest to add some
edge case to i3. My fork is good enough for that :) My thinking was that i3
is not aware of that flag at all right now and also that desktop windows
are usually, well, used as a desktop, meaning below other windows,
certainly not as floating windows. If you think that's not a standard use
of that flag, just ignore the PR.

Citing http://standards.freedesktop.org/wm-spec/1.3/ar01s05.html :

_NET_WM_WINDOW_TYPE_DESKTOP indicates a desktop feature. This can include a
single window containing desktop icons with the same dimensions as the
screen, allowing the desktop environment to have full control of the
desktop, without the need for proxying root window clicks.

stapelberg:
> If you think that's not a standard use of that flag, just ignore the PR.

Just to be clear, I do think that’s the standard use, but I’m not sure what other implications there are when we start supporting the _NET_WM_WINDOW_TYPE_DESKTOP type.

For now, I’d rather not merge this.

Thaodan:
> Just to be clear, I do think that’s the standard use, but I’m not sure what other implications there are when we start supporting the _NET_WM_WINDOW_TYPE_DESKTOP type

This PR is quite old by now but what other standard is there? All other tools I know use the same standard or set root xwindow ala feh or xsetroot.

I does sound like a natural extension to implement this part of the  EWMH since i3 already implements part of the spec.

The bug referenced by Github shows an example where its not just about "fixing Plasma":

https://github.com/jszczerbinsky/lwp/issues/57

orestisfl:
I don't really think not managing these windows (this PR) is standard-compliant while ignoring the desktop type technically is (the standard uses *SHOULD* wrt these windows, which means that the window manager is recommended to do a set of actions but not required to).

References to this type in the [spec](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html):

1.
>  _NET_WM_WINDOW_TYPE_DESKTOP indicates a desktop feature. This can include a single window containing desktop icons with the same dimensions as the screen, allowing the desktop environment to have full control of the desktop, without the need for proxying root window clicks. 

2.
> **Implementation notes**
> **File Manager desktop**
> 
> This spec suggests implementing the file manager desktop by mapping a desktop-sized window (no shape) to all desktops, with _NET_WM_WINDOW_TYPE_DESKTOP. This makes the desktop focusable and greatly simplifies implementation of the file manager. It is also faster than managing lots of small shaped windows. The file manager draws the background on this window. There should be a root property with a window handle for use in applications that want to draw the background (xearth). 

3.

> **Stacking order**
> 
> To obtain good interoperability between different Desktop Environments, the following layered stacking order is recommended, from the bottom:
> 
> - windows of type _NET_WM_TYPE_DESKTOP
> - windows having state _NET_WM_STATE_BELOW
> - windows not belonging in any other layer
> - windows of type _NET_WM_TYPE_DOCK (unless they have state _NET_WM_TYPE_BELOW) and windows having state _NET_WM_STATE_ABOVE
> - focused windows having state _NET_WM_STATE_FULLSCREEN

Thaodan:
How should the wm manage the windows besides putting them on the lowest
level of the desktop?

The window is moved to the lowest level by ignoring it, everything else
is layered above as intended.

PR #1830, opened 2015-08-04
Use bash instead of sh in i3-sensible scripts

jubalh:
In the script you use 'command -v' which AFAIK is a bash built-in and
thus not available on all shells. Although /bin/sh links to /bin/bash on
many systems it does not on every system. So it is possible that it does
not work.

### DIFF
--- a/i3-sensible-editor
+++ b/i3-sensible-editor
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 #
 # This code is released in public domain by Han Boetes <han@mijncomputer.nl>
 #
--- a/i3-sensible-pager
+++ b/i3-sensible-pager
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 #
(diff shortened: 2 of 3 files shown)


shdown:
I suggest using `#!/usr/bin/env bash` instead of `#!/bin/bash`, as on some systems (*BSD, NixOS) bash may not be in `/bin`. This is not true for `sh`, since FHS requires it to be there.

jubalh:
Alright, shall I update the PR?

Airblader:
Yes, please do :)

jubalh:
Done

stapelberg:
> In the script you use 'command -v' which AFAIK is a bash built-in and
> thus not available on all shells

`command -v` is not a bashism. See http://www.unix.com/man-page/posix/1posix/command/

With which shell did you run into trouble here?

jubalh:
Actually it was a friend who said he had trouble with this, and after looking I thought this is the reason. However I can not reach him right now..
Feel free to close this if it's not bashism after all.

stapelberg:
Closing for now until we have evidence of this being an actual problem.

Issue #1834, opened 2015-08-06
Mention xinitrc in Docu


Kaligule:
since many people abuse i3 as an initlauncher, wich is not what i3 is for. Its a job for xinitrc.
For example, instead setting a background image in i3config, one could add a line in xinitrc:
```
feh --bg-fill --no-fehbg ~/Documents/Bilder/Hintergrundbilder/Saturn.jpg
```

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
Can you send a pull request for this please? That way, it’ll be easier to discuss where exactly you think the change should be made.

Airblader:
It's true that applications can be started in many ways, but starting it from i3 also has its perks — for example that the window manager is actually running when exec calls are executed by i3.

I don't mind a friendly hint about alternatives, but I think we should keep it short and rather recommend the user to inform themselves using other sources. We want to document i3, not Linux.

acrisci:
I don't believe starting applications via i3 config on startup is abuse. Where did you get that idea from?

Kaligule:
acrisci: It just is not the job of an window manager to do so, I think. A setup shouldn't break only because the window manager is exchanged. See [Ravioli Code](https://en.wikipedia.org/wiki/Spaghetti_code#Ravioli_code).

Kaligule:
I created a pull request here: https://github.com/i3/i3/pull/1844

Airblader:
The PR has been withdrawn. Closing the issue.

PR #1844, opened 2015-08-12
Update userguide

Kaligule:
Mention ```.xinitrc``` and ```.xbindkeysrc``` as alternative for ```exec``` in userguide.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1667,6 +1667,9 @@ launching. So, if an application is not startup-notification aware (most GTK
 and Qt using applications seem to be, though), you will end up with a watch
 cursor for 60 seconds.
 
+You might want to consider using +.xinitrc+ (for launching programs at startup) and
++.xbindkeysrc+ (for keybindings) instead.
+
 === Splitting containers
 
 The split command makes the current window a split container. Split containers


Kaligule:
This should resolve the issiue https://github.com/i3/i3/issues/1834

docs/userguide
```diff
@@ -1667,6 +1667,9 @@ launching. So, if an application is not startup-notification aware (most GTK
 and Qt using applications seem to be, though), you will end up with a watch
 cursor for 60 seconds.
 
+You might want to consider using +.xinitrc+ (for launching programs at startup) and
++.xbindkeysrc+ (for keybindings) instead.
```
stapelberg: Your addition is at the wrong place. You’re adding to the exec command, you should be adding to the exec configuration directive.

acrisci:
`.xinitrc` is specific to `xinit` which is a display manager. This is only one of many display managers that someone can be using.

I believe if this is to be added, you must include at least some discussion about display managers. For anyone who uses xinit, this is pretty useless information, because they already know this. For anyone who doesn't use xinit, this is also pretty useless because it doesn't apply to them, and it's not immediate why it is better than the alternative of the i3 `exec` statement that they just read about.

acrisci:
In #1834, you said

> A setup shouldn't break only because the window manager is exchanged

and called this result "ravioli code", but I could also argue that

> A setup shouldn't break only because the display manager is exchanged

by the same logic to indicate that this suggestion to use `.xinitrc` rather than i3 `exec` is also ravioli code.

Kaligule:
acrisi: I didn't know ```xinit``` was a display manager and just installed ```lightdm``` to prove you wrong. I expected everything to work perfectly fine since ```xinitrc``` is always called when X is started.

Turns out I was wrong and there I stood with no terminal to disable lightdm and no dmenu to start a terminal :D
It took me a while to figure out I can get a loginshell with ```Ctrl-Alt-F8``` when in lightdm.
Seems your arguments are perfectly right and I just didn't understand enough of Linux and X.

My fault, thanks for your time.

Airblader:
I do believe that some other display managers use the xinitrc as well, but just for compatibility's sake.

Issue #1775, opened 2015-07-01
Key bindings don't work with groups


WovenTales:
I've just customized my xkb options to include a secondary layout with a group switch key (`grp:sclk_toggle`), and most things seem to be working well enough. However, I use bindsym in my config file, and while the shortcuts do properly update to use the new key positions in the primary group, i3 doesn't seem to notice when I've toggled to Group 2 (QWERTY), so in order to run the command I have at `Meta+h`, I need to press the key that would otherwise print `j`, as that is where `h` is located in Group 1 (Dvorak).

I haven't worked with it enough yet to know whether it matters that I haven't completely integrated the new layout into my configs yet, and so use `setxkbmap ... | xkbcomp ...` from within X, after it starts with only a single group (if i3 maybe creates a database or something that isn't properly updated), but it's probably a bug either way.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
Please provide clear steps to reproduce. Also, does the problem go away when you restart i3?

WovenTales:
* http://logs.i3wm.org/logs/5730450056151040.bz2
* Binary i3 version:  4.10.2 (2015-04-16, branch "4.10.2") © 2009-2014 Michael Stapelberg and contributors<br>
Running i3 version: 4.10.2 (2015-04-16, branch "4.10.2") (pid 20563)<br><br>
The i3 binary you just called: /usr/bin/i3<br>
The i3 binary you are running: i3
* Run`setxkbmap us,us -variant dvorak,intl -option grp:sclk_toggle,grp_led:scroll` (also works with `setxkbmap -I ~/.xkb us,us -variant dvorak,intl -option grp:sclk_toggle,grp_led:scroll -print | xkbcomp -I$HOME/.xkb - $DISPLAY` or variants thereof), switch to Group 2 (with that command, hit Scroll Lock) and try to run some keyboard shortcut using a key that changes location between the layouts (`Meta+w`/`Meta+e` is a good pair – they move to where `,` and `d` are on QWERTY, respectively).
[...]

Issue #585, opened 2011-12-10
mod4+4 makes COMMAND: *focus right* instead of COMMAND: *workspace 4*


i3bot:
[Originally reported by anrock623@…]
(subj on Gentoo Linux x64
Config was generated with i3-config-wizard with mod key set to win. No changes made by user. 

Workspace 4 bind is default:
bindsym $mod+4 workspace 4

BUT, when i press mod+4 this happens (from log):
10.12.2011 09:56:36 - COMMAND: _focus right_
10.12.2011 09:56:36 - Focusing right
single command completely parsed, dropping state...
done, json output = (null)

PR #1857, opened 2015-08-23
Use libxkbcommon for translating keysyms, support all XKB groups.

### LINKED ISSUE #1835: Neo layout not handled correctly
```
$ i3 --moreversion 2>&- || i3 --version
Binary i3 version:  4.10.3 (2015-07-30, branch "4.10.3") © 2009-2014 Michael Stapelberg and contributors
Running i3 version: 4.10.3 (2015-07-30, branch "4.10.3") (pid 5609)
```
I use the [neo ergonomic keyboard layout](http://www.neo-layout.org/)  and have problems when trying to bind keybindings with special symbols on higher layers.
There is for example an extra set of arrow keys on the 4th layer, which I would like to use in the same way as the normal arrow keys, for navigating and moving windows, but they don't seem to work with i3.
My current workaround is to explicitly specify the modifiers for the layer and then use the corresponding character from the 1st layer.
For example:
```
# change focus using neo cursor keys on layer 4
bindsym $mod+Mod3+i focus left
bindsym $mod+Mod3+a focus down
bindsym $mod+Mod3+l focus up
bindsym $mod+Mod3+e focus right

# alternatively, you can use the cursor keys:
bindsym $mod+Left focus left
bindsym $mod+Down focus down
bindsym $mod+Up focus up
bindsym $mod+Right focus right
```
I have the neo layout set via `setxkbmap de neo`.

Using `xev` shows that the neo arrow keys have the same keysym as the "normal" arrow keys, but different key codes.
First pressing the "normal" up arrow, then the up arrow on layer 4 gives:

```
KeyPress event, serial 32, synthetic NO, window 0x2600001,
    root 0x2bf, subw 0x0, time 31096176, (334,-147), root:(1621,581),
    state 0x0, keycode 111 (keysym 0xff52, Up), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 32, synthetic NO, window 0x2600001,
    root 0x2bf, subw 0x0, time 31096264, (334,-147), root:(1621,581),
    state 0x0, keycode 111 (keysym 0xff52, Up), same_screen YES,
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyPress event, serial 32, synthetic NO, window 0x2600001,
    root 0x2bf, subw 0x0, time 31096848, (334,-147), root:(1621,581),
    state 0x0, keycode 94 (keysym 0xfe11, ISO_Level5_Shift), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyPress event, serial 32, synthetic NO, window 0x2600001,
    root 0x2bf, subw 0x0, time 31097160, (334,-147), root:(1621,581),
    state 0x20, keycode 26 (keysym 0xff52, Up), same_screen YES,
    XKeysymToKeycode returns keycode: 111
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 32, synthetic NO, window 0x2600001,
    root 0x2bf, subw 0x0, time 31097328, (334,-147), root:(1621,581),
    state 0x20, keycode 26 (keysym 0xff52, Up), same_screen YES,
    XKeysymToKeycode returns keycode: 111
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 32, synthetic NO, window 0x2600001,
    root 0x2bf, subw 0x0, time 31097440, (334,-147), root:(1621,581),
    state 0x20, keycode 94 (keysym 0xfe11, ISO_Level5_Shift), same_screen YES,
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False
```
Could it be, that the ISO_Level5_Shift presses for the layer modifier are the issue?
Many thanks in advance and keep up the great work!

http://logs.i3wm.org/logs/5759180434571264.bz2

stapelberg:
fixes #1835

This commit improves the translation of keysyms to keycodes by loading
keymaps using libxkbcommon-x11 and using libxkbcommon for figuring out
the keymap, depending on each keybinding’s modifiers. This way, the
upper layers of complex layouts are now usable with i3’s bindsym
directive, such as de_neo’s layer 3 and higher.

Furthermore, the commit generalizes the handling of different XKB
groups. We formerly had support only for two separate groups, the
default group 1, and group 2. While Mode_switch is only one way to
switch to group 2, we called the binding option Mode_switch. With this
commit, the new names Group1, Group2 (an alias for Mode_switch), Group3
and Group4 are introduced for configuring bindings. This is only useful
for advanced keyboard layouts, such as people loading two keyboard
layouts and switching between them (us, ru seems to be a popular
combination).

When grabbing keys, one can only specify the modifier mask, but not an
XKB state mask (or value), so we still dynamically unbind and re-bind
keys whenever the XKB group changes.

The commit was manually tested using the following i3 config:

    bindsym Group4+n nop heya from group 4
    bindsym Group3+n nop heya from group 3
    bindsym Group2+n nop heya from group 2
    bindsym n nop heya
    bindcode Group2+38 nop fallback overwritten in group 2 only
    bindcode 38 nop fallback

…with the following layout:

    setxkbmap -layout "us,ua,ru,de" -variant ",winkeys,,neo" \
      -option "grp:shift_caps_toggle,grp_led:scroll" \
      -model pc104 -rules evdev

By default (xkb group 1, us layout), pressing “n” will result in the
“heya” message appearing. Pressing “a” will result in the “fallback”
message appearing. “j” is not triggered.

By pressing Shift+CapsLock you switch to the next group (xkb group 2, ua
layout). Pressing “a” will result in the “fallback overwritten in group
2 only” message, pressing “n” will still result in “heya”. “j” is not
triggered.

In the next group (xkb group 3, ru layout), pressing “a” will result in
the “fallback” message again, pressing “n” will result in “heya”,
“j” is not triggered.

In the last group (xkb group 4, de_neo layout), pressing “a” will still
result in “fallback”, pressing “n” will result in “heya”, pressing “j”
will result in “heya from group 4”.

The fallback behavior ensures use-cases such as ticket #1775 are still
covered.

Only binding keys when the X server is in the corresponding XKB group
ensures use-cases such as ticket #585 are still covered.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -773,8 +773,8 @@ The +binding (object)+ field contains details about the binding that was run:
 
 command (string)::
 	The i3 command that is configured to run for this binding.
-mods (array of strings)::
-	The modifier keys that were configured with this binding.
+event_state_mask (array of strings)::
+	The group and modifier keys that were configured with this binding.
 input_code (integer)::
 	If the binding was configured with +bindcode+, this will be the key code
 	that was given for the binding. If the binding is a mouse binding, it will be
@@ -792,7 +792,7 @@ input_type (string)::
  "change": "run",
  "binding": {
   "command": "nop",
-  "mods": [
+  "event_state_mask": [
     "shift",
     "ctrl"
   ],
--- a/docs/userguide
+++ b/docs/userguide
@@ -368,8 +368,8 @@ after the keys have been released.
 
 *Syntax*:
 ----------------------------------
-bindsym [--release] [<Modifiers>+]<keysym> command
-bindcode [--release] [<Modifiers>+]<keycode> command
+bindsym [--release] [<Group>+][<Modifiers>+]<keysym> command
+bindcode [--release] [<Group>+][<Modifiers>+]<keycode> command
 ----------------------------------
 
 *Examples*:
@@ -395,12 +395,13 @@ Available Modifiers:
 Mod1-Mod5, Shift, Control::
 Standard modifiers, see +xmodmap(1)+
 
-Mode_switch::
-Unlike other window managers, i3 can use Mode_switch as a modifier. This allows
-you to remap capslock (for example) to Mode_switch and use it for both: typing
-umlauts or special characters 'and' having some comfortably reachable key
-bindings. For example, when typing, capslock+1 or capslock+2 for switching
-workspaces is totally convenient. Try it :-).
+Group1, Group2, Group3, Group4::
+When using multiple keyboard layouts (e.g. with `setxkbmap -layout us,ru`), you
+can specify in which XKB group (also called “layout”) a keybinding should be
+active. By default, keybindings are translated in Group1 and are active in all
+groups. If you want to override keybindings in one of your layouts, specify the
+corresponding group. For backwards compatibility, the group “Mode_switch” is an
+alias for Group2.
 
 [[mousebindings]]
 
--- a/include/bindings.h
+++ b/include/bindings.h
@@ -31,7 +31,7 @@ Binding *configure_binding(const char *bindtype, const char *modifiers, const ch
  * Grab the bound keys (tell X to send us keypress events for those keycodes)
  *
  */
-void grab_all_keys(xcb_connection_t *conn, bool bind_mode_switch);
+void grab_all_keys(xcb_connection_t *conn);
 
 /**
  * Returns a pointer to the Binding that matches the given xcb event or NULL if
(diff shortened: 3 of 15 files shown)


Airblader:
@stapelberg The commit is already merged, isn't it? Did Github not notice because you changed it? I think we can close the PR then, right?

stapelberg:
Ah, right, I just committed it directly and forgot about the PR. Closing.

Issue #1855, opened 2015-08-21
i3 tray output all over the place


kovanovic:
I have 2 external monitors, laptop with two desk screen on each side: Those are:
* DP1 (large external monitor, left),
* eDP1 (laptop, center),  and
* VGA1 (smaller external monitor, right). 

I have four i3bars specified:
1. laptop eDP1 top: all my sys monitors and sys-tray output
2. laptop eDP1 bottom:  MPD and weather
3. DP1 bottom: only workspace names
4. VGA1 bottom: only workspace names

The issue is that my tray icons are totally messed up, icons show up on places where there is no bar at all:
* keepassX is on the top left corner of DP1.
* dropbox and GMPC are bottom right of VGA1.
* nothing is on the right of laptop's top bar where all of them should be (eDP1 top).
* skype's icon is not shown at all.

Now, when I open up a new window on DP1, the keepasx icon is not visible at all, you can only see it on the empty workspace.

Now, sometimes skype is where keepassx's icon is placed, sometimes they are all in the correct spot, sometimes they change their positions after i3 restart... it appears to be totally random.

Restarting i3 dows not resolve the problem, they just reshuffle totally weird.
![i3](https://cloud.githubusercontent.com/assets/558257/9409118/dc472a0a-480f-11e5-9558-4d889ea559db.jpg)

I use regular i3status package on arch, while for i3 I use i3-wm-iconpatch package just to add icons on window tiles

Here is the bar configuration from my i3 config file
```
set $laptop   eDP1
set $big      DP1
set $small    VGA1

bar {
    output $laptop
    tray_output $laptop
    position top
    workspace_buttons no
    status_command ~/.bin/i3bar_status.sh ~/.i3/conky_top
    font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
    font pango:DejaVu Sans Mono 10
    mode dock

    colors {
        background #000000
        statusline #ffff00
        separator #000000

                           #<border> <bg>  <text>
        focused_workspace  #ff0000 #000000 #ffffff
        active_workspace   #333333 #5f676a #ffffff
        inactive_workspace #333333 #222222 #888888
        urgent_workspace   #2f343a #900000 #ffffff
    }
}

bar {
    output $laptop
    workspace_buttons yes
    position bottom
    status_command ~/.bin/i3bar_status.sh ~/.i3/conky_bottom
    font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
[...]

PR #1863, opened 2015-08-25
Add 'tray_output primary' to the default config

Airblader:
relates to #1855

### DIFF
--- a/i3.config
+++ b/i3.config
@@ -165,6 +165,7 @@ bindsym Mod1+r mode "resize"
 # finds out, if available)
 bar {
         status_command i3status
+        tray_output primary
 }
 
 #######################################################################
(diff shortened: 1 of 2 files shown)


stapelberg:
You’ll also need to update i3.config, I think.

Airblader:
`i3.config` is only used for the initial startup, no? So it's a config the user usually wouldn't (shouldn't) edit and since it only has one bar block, there's no risk. But sure, I can put it there, too, if you want.

stapelberg:
I think your rationale is sound, but also it’s better to keep the two files in sync, just to make the mental model of dealing with the two files simpler.

Airblader:
Agreed. :)

Issue #1859, opened 2015-08-24
i3 doesn't compile on musl


nmeum:
i3 currently doesn't compile on a [musl libc](http://www.musl-libc.org/) system. The only function that seem to cause troubles is the `resolve_tilde` function from `libi3/resolve_tilde.c`. This function uses `GLOB_TILDE` which is a GNUism and (probably) only available in glibc.

This is the output of make on the next branch:

```
[libi3] CC libi3/ipc_recv_message.c
[libi3] CC libi3/get_visualtype.c
[libi3] CC libi3/mkdirp.c
[libi3] CC libi3/ipc_send_message.c
[libi3] CC libi3/fake_configure_notify.c
[libi3] CC libi3/strndup.c
[libi3] CC libi3/get_colorpixel.c
[libi3] CC libi3/resolve_tilde.c
[libi3] CC libi3/get_exe_path.c
[libi3] CC libi3/font.c
[libi3] CC libi3/dpi.c
libi3/resolve_tilde.c: In function 'resolve_tilde':
[libi3] CC libi3/ucs2_conversion.c
libi3/resolve_tilde.c:28:26: error: 'GLOB_TILDE' undeclared (first use in this function)
     int res = glob(head, GLOB_TILDE, NULL, &globbuf);
                          ^
libi3/resolve_tilde.c:28:26: note: each undeclared identifier is reported only once for each function it appears in
libi3/libi3.mk:12: recipe for target 'libi3/resolve_tilde.o' failed
make: *** [libi3/resolve_tilde.o] Error 1
make: *** Waiting for unfinished jobs....
```

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
GLOB_TILDE is not a GNU-ism, it’s also available on FreeBSD, NetBSD and other systems.

Please send a pull request if you care about getting this fixed.

Airblader:
I don't think there's any value in keeping a ticket open for this. As Michael pointed out, if you want this to be fixed, send a pull request.

lukaslozovski:
I got this to compile and run for i3status 2.13 on alpine 3.10 by adding 
`#define GLOB_TILDE GLOB_PERIOD` to the bottom of the config.h.in file

[...]

PR #1869, opened 2015-08-27
libi3: make resolve_tilde.c musl compatible

nmeum:
Since you weren't interested in providing musl compatibility yourself, this is my proposal for a more standard compatible `resolve_tilde` implementation which doesn't rely on the `GLOB_TILDE` GNU Extension.

What do you think of it?

See also: #1859

### DIFF
--- a/libi3/resolve_tilde.c
+++ b/libi3/resolve_tilde.c
@@ -8,7 +8,6 @@
 
 #include "libi3.h"
 #include <err.h>
-#include <glob.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -19,27 +18,21 @@
  *
  */
 char *resolve_tilde(const char *path) {
-    static glob_t globbuf;
-    char *head, *tail, *result;
+    char *home_dir = getenv("HOME");
+    char *result, *tilde, *npath;
 
-    tail = strchr(path, '/');
-    head = sstrndup(path, tail ? (size_t)(tail - path) : strlen(path));
-
-    int res = glob(head, GLOB_TILDE, NULL, &globbuf);
-    free(head);
-    /* no match, or many wildcard matches are bad */
-    if (res == GLOB_NOMATCH || globbuf.gl_pathc != 1)
-        result = sstrdup(path);
-    else if (res != 0) {
-        err(EXIT_FAILURE, "glob() failed");
+    tilde = strchr(path, '~');
+    if (tilde == NULL) {
+        return sstrdup(path);
     } else {
-        head = globbuf.gl_pathv[0];
-        result = scalloc(strlen(head) + (tail ? strlen(tail) : 0) + 1, 1);
-        strncpy(result, head, strlen(head));
-        if (tail)
-            strncat(result, tail, strlen(tail));
+        result = sstrndup(path, (size_t)(tilde - path));
+        result = strcat(result, home_dir);
     }
-    globfree(&globbuf);
 
-    return result;
+    npath = tilde + 1;
+    if (strchr(npath, '~')) {
+        return sstrdup(path);
+    }
+
+    return strcat(result, npath);
 }


libi3/resolve_tilde.c
```diff
@@ -19,27 +18,21 @@
...
-        result = scalloc(strlen(head) + (tail ? strlen(tail) : 0) + 1, 1);
-        strncpy(result, head, strlen(head));
-        if (tail)
-            strncat(result, tail, strlen(tail));
+        result = sstrndup(path, (size_t)(tilde - path));
+        result = strcat(result, home_dir);
+    }
+
+    npath = tilde + 1;
+    if (strchr(npath, '~')) {
```
nmeum: I was wondering if I should change that to:

```C
if (strchr(npath, '~') != NULL) {
    // …
}
```

stapelberg:
Your implementation doesn’t cover `~user` where `user != getenv("USER")`.

How about just dropping in https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/glob.c conditionally for musl?

In fact, why not add support for `GLOB_TILDE` to musl itself and thereby fix the problem for all software which uses `GLOB_TILDE`?

Airblader:
Just FYI: looking at other projects dealing with musl, a common approach seems to be to simply not resolve tilde for musl and return the path unchanged.

stapelberg:
Currently, I don’t think that’s a good approach. We should strive to keep all functionality in i3 available for all systems on which we run.

What’s the use-case for a musl-based system? Where does one typically use musl?

nmeum:
> Your implementation doesn’t cover ~user where user != getenv("USER")

I could fix that but a) would you merge the patch if I would do so? (see below) b) is this really used by anyone? I mean who stores his/her i3 configuration in someone else home directory?

> How about just dropping in https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/glob.c conditionally for musl?

Sounds good to me, you might want to use that on uclibc as well (see below).

> looking at other projects dealing with musl, a common approach seems to be to simply not resolve tilde for musl and return the path unchanged.

aha, that's news too me…could you name some examples? Either way I would agree with Michael on that one.

> What’s the use-case for a musl-based system? Where does one typically use musl?

Well…I use it on all my linux based computers (including my laptops and desktops), there are even some distributions which allow you to use musl instead of glibc by default these days. For instance gentoo, voidlinux or alpine.

Most software does infact compile on musl without too much patching, i3 is one of the rare cases where this isn't the case.

> In fact, why not add support for GLOB_TILDE to musl itself and thereby fix the problem for all software which uses GLOB_TILDE?

That might be a good idea, I don't know if the musl maintainers have a strong opinion regarding `GLOB_TILDE`, however, why not make i3 more portable by not relying on GNU extensions?

There might be other libc implementations out there which don't support `GLOB_TILDE` either, uclibc **might be** an additional example…

Airblader:
Sorry, I might've been too quick. The patch I actually found was a musl patch for i3 (the irony). http://git.alpinelinux.org/cgit/aports/plain/testing/i3wm/musl.patch

Disregard my comment. I wasn't proposing this as a way forward anyway.

stapelberg:
> That might be a good idea, I don't know if the musl maintainers have a strong opinion regarding GLOB_TILDE, however, why not make i3 more portable by not relying on GNU extensions?

GLOB_TILDE is supported by the *BSD and OS X libc, so i3 is portable across the operating systems on which it makes sense to run it. We haven’t paid too much attention to libc portability as you can tell, and I deem that far less important. After all, you can just compile i3 with a different libc.

> There might be other libc implementations out there which don't support GLOB_TILDE either, uclibc might be an additional example…

Right, but as long as nobody complains, we won’t bother with that.

Can you please suggest adding GLOB_TILDE to musl and give us the link to that communication (so that people who are interested in the progress can follow along)? Once we know what they think about it, we can figure out how to proceed here.

Side note: if you want to make the argument to include GLOB_TILDE in musl carry more weight: procps is also affected by this: https://github.com/jvvv/musl-pkg-patches/blob/master/procps_Support-libcs-without-GLOB_TILDE.patch. http://codesearch.debian.net finds 83 packages which contain the string “GLOB_TILDE”.

Airblader:
@nmeum Any updates?

nmeum:
I don't know what the best place is to file feature requests for musl…probably their mailing list, however, I started working on a patch for musl which adds support for `GLOB_TILDE`.

I will add a link to this comment as soon as I suggested my patch upstream…

Issue #1770, opened 2015-06-27
enhancement: Magic word: "__focused__"


acrisci:
The magic word `__focused__` is taken to mean the currently focused thing in the context of a command or selection criteria.

Examples:

Move all firefox windows to the current workspace:

    [class="Firefox"] move to workspace __focused__

Kill all windows of the class that match the currently focused window:

    [class="__focused__"] kill

It would be useful with #1769 to limit commands to some workspace.

    [workspace="__focused__" class="Firefox"] kill

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
Sounds good to me.

Airblader:
Granted, it's unlikely, but does this mean we accept the potential conflict of applications that for whatever reason use this word as the class, instance or whatever? The only thing where it can't happen is the workspace since they can't start with underscores.

acrisci:
When used as a value, double underscore conventionally means "private, do not use" or "secret special value, may behave weirdly". An application setting its class has no use for either of these concepts for the class string. I think it is safe to use any value that begins with two underscores as magic values.

Airblader:
Yeah, I think so, too. I'm working on at least the command criteria part of this proposal now. For the command parameters I think it makes sense to do a separate PR, maybe even more than one.

Airblader:
FYI: I've been trying to identify the usecases of command parameters:

````
move to workspace __focused__
move workspace to output __focused__
```

Didn't really find anything else where this would make sense.

Airblader:
I actually think we only need this for command criteria. We already have `move … to workspace current`, which is exactly what we want here. Seems rather odd to add `__focused__` to a *single* special path of a command.

Airblader:
@acrisci Do you insist on the special value for the command as well or can we close it now that the PR introducing this to the command criteria has been merged?

acrisci:
@Airblader this is good for now. I think this is the most useful part of the feature. If anyone else wants the other parts of this feature, they can open a separate issue and make their case. Thanks :+1: .

PR #1871, opened 2015-08-28
Introduce command criterion value __focused__

Airblader:
relates to #1770

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1604,25 +1604,35 @@ for_window [class="^evil-app$"] floating enable, move container to workspace 4
 The criteria which are currently implemented are:
 
 class::
-	Compares the window class (the second part of WM_CLASS)
+	Compares the window class (the second part of WM_CLASS). Use the
+	special value +__focused__+ to match all windows having the same window
+	class as the currently focused window.
 instance::
-	Compares the window instance (the first part of WM_CLASS)
+	Compares the window instance (the first part of WM_CLASS). Use the
+	special value +__focused__+ to match all windows having the same window
+	instance as the currently focused window.
 window_role::
-	Compares the window role (WM_WINDOW_ROLE).
+	Compares the window role (WM_WINDOW_ROLE). Use the special value
+	+__focused__+ to match all windows having the same window role as the
+	currently focused window.
 window_type::
-        Compare the window type (_NET_WM_WINDOW_TYPE). Possible values are
-        +normal+, +dialog+, +utility+, +toolbar+, +splash+, +menu+, +dropdown_menu+,
-        +popup_menu+ and +toolti+.
+	Compare the window type (_NET_WM_WINDOW_TYPE). Possible values are
+	+normal+, +dialog+, +utility+, +toolbar+, +splash+, +menu+, +dropdown_menu+,
+	+popup_menu+ and +toolti+.
 id::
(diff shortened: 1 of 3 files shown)


testcases/t/251-command-criteria-focused.t
```diff
@@ -0,0 +1,86 @@
...
+is(@{get_ws($ws)->{nodes}}, 2, 'sanity check: workspace contains two windows');
+
+cmd '[title=__focused__] move to workspace trash';
+is(@{get_ws($ws)->{nodes}}, 0, '__focused__ works for title');
+
+###############################################################################
+# 4: Test __focused__ for window role.
+###############################################################################
+
+# NOTE X11::XCB currently doesn't support setting the window role.
```
Airblader: @stapelberg Is there a good way to set the window role?

stapelberg: https://github.com/i3/i3/blob/b68a400abfeb9f17901497eb005862853656287c/testcases/t/214-layout-restore-criteria.t#L93 currently is the best way.

docs/userguide
```diff
@@ -1604,11 +1604,17 @@ for_window [class="^evil-app$"] floating enable, move container to workspace 4
 The criteria which are currently implemented are:
 
 class::
-	Compares the window class (the second part of WM_CLASS)
+        Compares the window class (the second part of WM_CLASS). Use the
+        special value +__focused__+ to match all windows having the same window
+        class as the currently focused window.
```
stapelberg: Here and below, it looks like you’re changing indentation. That doesn’t seem to be intentional?

Airblader: It wasn't intentional (I haven't noticed), but I formatted it with vim (`gq`). It's now 8 spaces, before it was a tab. What's your preference? In the code we use spaces.

testcases/t/251-command-criteria-focused.t
Airblader: Ah, thanks. I'll put that in.

docs/userguide
stapelberg: I’d prefer it if you could keep it consistent with how the file is currently formatted.

Airblader:
@stapelberg Done. The format change in there now should be correct because it was using spaces before.

stapelberg:
In its current form, the test will pass when __focused__ just matches any window. Can you open at least one more window with a different class to make sure that’s not the case, please?

Airblader:
@stapelberg Done.

docs/userguide
```diff
@@ -1604,25 +1604,35 @@ for_window [class="^evil-app$"] floating enable, move container to workspace 4
...
+	Compares the window role (WM_WINDOW_ROLE). Use the special value
+	+__focused__+ to match all windows having the same window role as the
+	currently focused window.
 window_type::
-        Compare the window type (_NET_WM_WINDOW_TYPE). Possible values are
-        +normal+, +dialog+, +utility+, +toolbar+, +splash+, +menu+, +dropdown_menu+,
-        +popup_menu+ and +toolti+.
+	Compare the window type (_NET_WM_WINDOW_TYPE). Possible values are
+	+normal+, +dialog+, +utility+, +toolbar+, +splash+, +menu+, +dropdown_menu+,
+	+popup_menu+ and +toolti+.
```
Airblader: There's a typo here. I'll fix this in a separate PR once this is merged.

PR #1877, opened 2015-08-30
Support _NET_WM_VISIBLE_NAME

### LINKED ISSUE #1872: [enhancement] Support _NET_WM_VISIBLE_NAME
We've introduced `title_format`, which changes the way we display the window title. According to the specification, this means that we must support `_NET_WM_VISIBLE_NAME`.

See http://standards.freedesktop.org/wm-spec/1.3/ar01s05.html:

> If the Window Manager displays a window name other than _NET_WM_NAME the Window Manager MUST set this to the title displayed in UTF-8 encoding.

Airblader:
fixes #1872


Airblader:
One of this and #1856 will have to be rebased due to conflicts.

src/x.c
```diff
@@ -612,7 +612,16 @@ void x_draw_decoration(Con *con) {
...
 
-    i3String *title = win->title_format == NULL ? win->name : parse_title_format(win);
+    i3String *title = NULL;
+    if (win->title_format == NULL) {
+        title = win->name;
+        /* If no format was given, we removed _NET_WM_VISIBLE_NAME during execution of
+         * the command already. */
+    } else {
+        title = parse_title_format(win);
+        ewmh_update_visible_name(win->id, i3string_as_utf8(title));
```
stapelberg: You’re calling `ewmh_update_visible_name` on the drawing (!) code path here. We should strive to keep the number of XCB calls as low as possible, and not send any calls unless something has changed. Can you move this call somewhere else (the rendering code path might make sense) and ideally make it conditional, so that it won’t send something on the wire unless the title is actually changed?

Airblader: I put it here to avoid parsing the title multiple times. However, maybe a good place to call this would actually just be whenever we receive a hint that `_NET_WM_NAME` (or `WM_NAME`) changed.

Airblader:
@stapelberg I've moved the code now so it is really only called when absolutely necessary. I think this is overall even better than before.

Issue #665, opened 2012-03-02
window-specific attributes like width_increment etc. need to be moved from Con to Window struct


stapelberg:
Probably due to confusion they are saved in the Con instead of in the Window. However, as the window can be re-attached to another Con (for example when using sticky windows and/or swallowing), they need to be stored in the Window.

Airblader:
Can we identify which properties need to be moved? I think these are pretty safe:

* `base_width`
* `base_height`
* `width_increment`
* `height_increment`

What about `mapped` and `depth`?

stapelberg:
I agree that `base_width`, `base_height`, `width_increment` and `height_increment` should be moved.

`border_width` is also set per X11 window, and `aspect_ratio` comes from the window hints, so both should be moved.

`depth` is the depth of the container window, so I currently don’t think it should be moved.

`mapped` is a corner-case, which cannot easily be moved, I think. See e.g. `tree_render()`, which changes the root node’s mapped property.

Airblader:
What about border_style?

stapelberg:
`border_style` should remain in the Con. As you mentioned, it’s a per-container attribute also used for tabbed/stacked containers.

Airblader:
Yeah, but doesn't the same apply to border_width then?

stapelberg:
I haven’t had a very close look, but I think it only applies to `current_border_width`, and `border_width` really is just determined by the X11 properties.

Airblader:
As a note, with #1966 in mind, once that is implemented we actually should move `depth` to `Window` since containers themselves no longer have any reason to have a window. That is assuming we *never* use the parent for decoration rendering anymore, including tabbed/stacked, of course.

PR #1880, opened 2015-08-30
Move data from Con to Window

Airblader:
see #665 

This does not yet include all things that can be moved (probably).

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -599,17 +599,20 @@ static void cmd_resize_floating(I3_CMD, char *way, char *direction, Con *floatin
     /* ensure that resize will take place even if pixel increment is smaller than
      * height increment or width increment.
      * fixes #1011 */
-    if (strcmp(direction, "up") == 0 || strcmp(direction, "down") == 0 ||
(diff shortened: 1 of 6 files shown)


src/commands.c
```diff
@@ -599,17 +599,19 @@ static void cmd_resize_floating(I3_CMD, char *way, char *direction, Con *floatin
...
-        if (px < 0)
-            px = (-px < focused_con->height_increment) ? -focused_con->height_increment : px;
-        else
-            px = (px < focused_con->height_increment) ? focused_con->height_increment : px;
-    } else if (strcmp(direction, "left") == 0 || strcmp(direction, "right") == 0) {
-        if (px < 0)
-            px = (-px < focused_con->width_increment) ? -focused_con->width_increment : px;
-        else
-            px = (px < focused_con->width_increment) ? focused_con->width_increment : px;
+    if (focused_con->window != NULL) {
```
stapelberg: You can use `const i3Window *window = focused_con->window`; to keep the code shorter.

Airblader:
@stapelberg Done

PR #1891, opened 2015-09-04
Handle absent window title

### LINKED ISSUE #1890: Systematic crash with firefox
I've i3 next systematically crashing when doing some actions on firefox, actions that open floating windows, for instance trying to rename a bookmark folder from the personal toolbar systematically crashes i3.

I tried to get backtrace but /tmp/i3-backtrace* are all filled with "No stack" even though my binary is unstripped:
```
file `which i3`                                                                                         
/usr/bin/i3: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.32, BuildID[sha1]=b416ccc452b9bc4757a8ac4d74d91b910f7e4d77, not stripped
```

```
i3 --moreversion                                                                                
Binary i3 version:  4.10.3-301-gee2983c (2015-09-03, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.10.3-301-gee2983c (2015-09-03, branch "next") (pid 855)
Loaded i3 config: /home/teto/.i3/config (Last modified: ven. 04 sept. 2015 17:15:08 CEST, 21586 seconds ago)

```
Wonder if anyone can reproduce ?

Airblader:
fixes #1890

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -359,7 +359,7 @@ i3String *window_parse_title_format(i3Window *win) {
     for (char *walk = format; *walk != '\0'; walk++) {
         if (STARTS_WITH(walk, "%title")) {
             if (escaped_title == NULL)
-                escaped_title = i3string_as_utf8(is_markup ? i3string_escape_markup(win->name) : win->name);
+                escaped_title = win->name == NULL ? "" : i3string_as_utf8(is_markup ? i3string_escape_markup(win->name) : win->name);
 
             buffer_len = buffer_len - strlen("%title") + strlen(escaped_title);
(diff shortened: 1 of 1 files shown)


stapelberg:
I realize you’re not changing that code, but I think the memory handling deserves a closer look here. the escape_* i3string’s are never freed, right? `g_markup_escape_text` says in its documentation that it returns a newly allocated string, so I think we should free the strings at the end of the function, right?

Airblader:
Yeah, I think you're right. I'll send a separate PR to fix that.

PR #1893, opened 2015-09-05
Added cmd_size

rr-:
Sorry for making a new pull request for this, but I was unable to salvage the previous one, since I deleted my fork...

Implements #1727.

- Doesn't move window, for this one should use `move direction` or `move position`.
- Size hints are untested, since I had no idea how to test them.
- Doesn't do anything for containers. IMO resizing containers to specific sizes is different enough to deserve a separate feature/pull request.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2007,6 +2007,7 @@ If you want to resize containers/windows using your keyboard, you can use the
 *Syntax*:
 -------------------------------------------------------
 resize grow|shrink <direction> [<px> px [or <ppt> ppt]]
+resize set <width> [px] <height> [px]
 -------------------------------------------------------
 
 Direction can either be one of +up+, +down+, +left+ or +right+. Or you can be
@@ -2015,7 +2016,8 @@ space from all the other containers. The optional pixel argument specifies by
 how many pixels a *floating container* should be grown or shrunk (the default
 is 10 pixels). The ppt argument means percentage points and specifies by how
 many percentage points a *tiling container* should be grown or shrunk (the
-default is 10 percentage points).
+default is 10 percentage points). Note that +resize set+ will only work for
+floating containers.
 
 I recommend using the resize command inside a so called +mode+:
 
@@ -2047,6 +2049,11 @@ mode "resize" {
 # Enter resize mode
 bindsym $mod+r mode "resize"
 ----------------------------------------------------------------------
+
+*Example 2 - setting urxvt size to 640x480:*
+------------------------------------------------
+for_window [class="urxvt"] resize set 640 480
+------------------------------------------------
 
 === Jumping to specific windows
 
--- a/src/commands.c
+++ b/src/commands.c
@@ -824,6 +824,37 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
 }
 
 /*
+ * Implementation of 'resize set <px> [px] <px> [px]'.
+ *
+ */
+void cmd_size(I3_CMD, char *cwidth, char *cheight) {
+    DLOG("resizing to %sx%s px\n", cwidth, cheight);
+    // TODO: We could either handle this in the parser itself as a separate token (and make the stack typed) or we need a better way to convert a string to a number with error checking
+    int x = atoi(cwidth);
+    int y = atoi(cheight);
+    if (x <= 0 || y <= 0) {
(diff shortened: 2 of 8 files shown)


docs/userguide
```diff
@@ -2007,6 +2007,7 @@ If you want to resize containers/windows using your keyboard, you can use the
 *Syntax*:
 -------------------------------------------------------
 resize grow|shrink <direction> [<px> px [or <ppt> ppt]]
+resize set <width> [px] <height> [px]
```
Airblader: The userguide should mention that `resize set` (currently) only works for floating containers.

src/commands.c
```diff
@@ -824,6 +824,35 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
...
+        ELOG("Resize failed: dimensions cannot be negative (was %sx%s)\n", cwidth, cheight);
+        return;
+    }
+
+    HANDLE_EMPTY_MATCH;
+
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        Con *floating_con;
+        if ((floating_con = con_inside_floating(current->con))) {
```
Airblader: If the matched container isn't floating, we should log the container and state that it cannot be resized because it's not floating.

Airblader:
Works good to me in all "normal" cases. However, I can crash i3 by doing this:

* Open (tiling) windows in the layout splith(urxvt, splitv(urxvt, urxvt)) with the lower right urxvt being focused in the end
* `focus parent, focus parent, floating enable`
* Focus the lower right urxvt again, execute `sleep 3; resize set 300 300` and while it sleeps, `focus parent` to focus the splitv container

i3 now crashes.

src/floating.c
```diff
@@ -826,6 +826,33 @@ void floating_reposition(Con *con, Rect newrect) {
...
+ * Sets size of the CT_FLOATING_CON to specified dimensions. Might limit the
+ * actual size with regard to size constraints taken from user settings.
+ * Additionally, the dimensions may be upscaled until they're divisible by the
+ * window's size hints.
+ *
+ */
+void floating_resize(Con *floating_con, int x, int y) {
+    DLOG("floating resize to %dx%d px\n", x, y);
+    Rect *rect = &floating_con->rect;
+    Con *focused_con = con_descend_focused(floating_con);
```
Airblader: To avoid the crash I described below, you need to ensure here that the focused_con even has a window. I think something like this would be ok:

````
int wi = 0, hi = 0;
if (focused_con->window != NULL) {
    wi = …
    hi = …
}
```

rr-:
Fixed things you've mentioned but unfortunately I failed to simulate the scenario you've described in the unit tests.

Airblader:
I think not testing that is fine, the crash scenario was pretty artifical anyway. You went down a different road of what I proposed, but I think that's OK, too, since nested floating containers aren't something we recommend or "really" support anyway.

src/commands.c
```diff
@@ -824,6 +824,37 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
...
+            floating_resize(floating_con, x, y);
+        } else {
+            ELOG("Resize failed: not a floating container\n");
```
stapelberg: Can you please include a pointer to the container in the error message?

stapelberg:
One nit (log message), otherwise this looks good to me.

@Airblader, can you please confirm that the crash you were seeing is fixed with the current code?

Airblader:
@stapelberg Yeah, it's gone.

stapelberg:
Cool. @rr- please notify me once you made the log message change, and I’ll be happy to merge this.

rr-:
Is the message good now?

src/commands.c
```diff
@@ -824,6 +824,37 @@ void cmd_resize(I3_CMD, char *way, char *direction, char *resize_px, char *resiz
...
+
+    HANDLE_EMPTY_MATCH;
+
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        Con *floating_con;
+        if ((floating_con = con_inside_floating(current->con))) {
+            floating_resize(floating_con, x, y);
+        } else {
+            DLOG("Resize failed: %p not a floating container\n", floating_con);
```
stapelberg: This should read current->con, not floating_con (since floating_con is always NULL in line 848). And please keep it ELOG, not DLOG.

rr-:
Updated

Issue #1878, opened 2015-08-30
[enhancement] Use cairo to render i3bar


Airblader:
This is similar to #1278, but for i3bar.

Airblader:
I'm going to work on a proof of concept for this. It won't be entirely prettied up yet (I won't bother too much with moving from "colorpixels" to the format we need for cairo yet), but it will show the (possible) result so we can also make sure it's pixel perfect since @stapelberg asked for this in #1278.

Airblader:
@stapelberg Actually, cairo is only optional so far in i3 (together with pango), right? Is it even acceptable to make cairo a hard dependency now? Or do we need to introduce a wrapper layer that uses `xcb_*` functions as a fallback?

stapelberg:
Let’s operate under the assumption that making it a hard dependency is okay for now.

Airblader:
@stapelberg Alright. So it'd be great if you had a look at the PR to see the overall approach and so on. If that's OK, I could continue to make an actual PR out of this and do all the little things that need to be done. Then at some point we need to make a decision on the hard dependency or fallback thing. :)

Just for comparison, here are two bars running the default i3status config. The top one is the current one, the bottom one is the new cairo version. I haven't actually done an overlay yet, but it looks pretty pixel perfect to me.

![i3bar-cairo](https://cloud.githubusercontent.com/assets/2392216/9704619/fe4ca57c-54ac-11e5-898f-43d19f8e79fe.png)

stapelberg:
I replied in #1278.

PR #1901, opened 2015-09-06
POC for i3bar-cairo

Airblader:
This is a (fairly advanced) proof of concept for using cairo to draw i3bar. All drawing calls have been replaced and a graphics context is only directly used to call font rendering routines which, in the pango case at least, will use their own cairo context anyway (this could probably be also be improved).

There are some things that would have to be done still. For example, we can move away from "colorpixels" and store colors in their own double^3 structure since this POC does it a little backwards by shifting bits back again.

This PR is not meant to be merged in this state, but should rather serve as a discussion basis for #1878 as well as a playground to see what it looks like being rendered with cairo.

I think the PR shows nicely that using cairo actually compresses the drawing code a little bit. Also note that I have ran a valgrind on it and there are some mem leaks, but I also have them with the unmodified bar. Also I haven't tested the POC extensively yet (short text, cut-offs, multiple outputs, …)

I recommend looking at commits one by one as the changes might otherwise be a little overwhelming.

### DIFF
--- a/i3bar/include/outputs.h
+++ b/i3bar/include/outputs.h
@@ -10,13 +10,34 @@
 #pragma once
 
 #include <xcb/xcb.h>
+#include <cairo/cairo-xcb.h>
 
 #include "common.h"
+#include "cairo_util.h"
 
 typedef struct i3_output i3_output;
 
 SLIST_HEAD(outputs_head, i3_output);
 struct outputs_head* outputs;
+
+struct i3_output {
+    char* name;   /* Name of the output */
+    bool active;  /* If the output is active */
+    bool primary; /* If it is the primary output */
+    bool visible; /* If the bar is visible on this output */
+    int ws;       /* The number of the currently visible ws */
+    rect rect;    /* The rect (relative to the root window) */
+
+    /* Off-screen buffer for preliminary rendering. */
+    surface_t buffer;
+    /* The actual window on which we draw. */
+    surface_t bar;
+
+    struct ws_head* workspaces;  /* The workspaces on this output */
+    struct tc_head* trayclients; /* The tray clients on this output */
+
+    SLIST_ENTRY(i3_output) slist; /* Pointer for the SLIST-Macro */
+};
 
 /*
  * Start parsing the received JSON string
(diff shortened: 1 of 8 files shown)


i3bar/include/outputs.h
```diff
@@ -10,14 +10,52 @@
...
 
+typedef struct i3_surface i3_surface;
 typedef struct i3_output i3_output;
 
 SLIST_HEAD(outputs_head, i3_output);
 struct outputs_head* outputs;
 
+/* Wrapper container that holds a window and both a graphics context
+ * and cairo access to it. */
+struct i3_surface {
```
Airblader: This kind of structure could be moved to libi3 so we can also move `init_surface` there and add other functions we will need (to destroy the surface etc.), but also so we can eventually use it to draw window decorations in i3.

i3bar/include/util.h
```diff
@@ -17,6 +17,12 @@
 
 #define STARTS_WITH(string, len, needle) ((len >= strlen(needle)) && strncasecmp(string, needle, strlen(needle)) == 0)
 
+// TODO
```
Airblader: This little hack should be resolved by moving from colorpixels to a dedicated double^3 structure.

i3bar/src/xcb.c
```diff
@@ -173,17 +172,19 @@ static void draw_separator(uint32_t x, struct status_block *block) {
     uint32_t center_x = x - sep_offset;
     if (config.separator_symbol == NULL) {
         /* Draw a classic one pixel, vertical separator. */
-        uint32_t mask = XCB_GC_FOREGROUND | XCB_GC_BACKGROUND | XCB_GC_LINE_WIDTH;
-        uint32_t values[] = {colors.sep_fg, colors.bar_bg, logical_px(1)};
-        xcb_change_gc(xcb_connection, statusline_ctx, mask, values);
-        xcb_poly_line(xcb_connection, XCB_COORD_MODE_ORIGIN, statusline_pm, statusline_ctx, 2,
-                      (xcb_point_t[]){{center_x, logical_px(sep_voff_px)},
-                                      {center_x, bar_height - logical_px(sep_voff_px)}});
+        // TODO for some reason using a line is always too thick, so we use a rectangle instead
```
Airblader: I'm not yet sure on how this can be done with an actual line. I even went to a line width less than `1.0`, but it was still too thick. A rectangle does just as well, though, so maybe this is OK.

Airblader: Note to myself: http://www.cairographics.org/FAQ/#sharp_lines

PR #1909, opened 2015-09-09
Refactor rendering noodles a bit

Airblader:


src/render.c
```diff
@@ -222,31 +225,33 @@ void render_con(Con *con, bool render_fullscreen) {
...
     /* find the height for the decorations */
-    int deco_height = render_deco_height();
+    params.deco_height = render_deco_height();
 
     /* precalculate the sizes to be able to correct rounding errors */
-    int sizes[children];
-    memset(sizes, 0, children * sizeof(int));
-    if ((con->layout == L_SPLITH || con->layout == L_SPLITV) && children > 0) {
+    int sizes[params.children];
+    memset(sizes, 0, params.children * sizeof(int));
```
Airblader: For some reason I got segfaults when trying to use `params.sizes` here directly instead of going through the trouble of a separate variable. I'm afraid it exceeded my C knowledge, so either we leave it like this or someone tells me what I was doing wrong. :)

Airblader:
@stapelberg Any reason you are hesitating about this? It's a fairly straight forward refactoring, I just pulled out some functions (and added one assert per function).

stapelberg:
It takes a comparatively long time to review this, and I don’t see a benefit, so it’s low-priority. Also, I’d prefer to merge anything rendering-related after the 4.11 release.

PR #1915, opened 2015-09-11
Indicator for tabbed / stacked containers

### LINKED ISSUE #1067: border indicators for tab and stack
[Originally reported by tanguy+i3.org@…]
(When you have a single window that has been split horizontally (or which is the last survivor of an horizontally split container, which is equivalent: an horizontally split container with only one client in it), it is indicated by a highlight of the right border. If it has been split vertically, its bottom border it highlighted instead.

When you have a single window that has been tabbed or stacked (yes, that cannot be done directly, but it can be done with two commands, and it is useful) (or again, a window which is the last survivor of a tabbed or stacked container), nothing indicates that state. I think it would be useful to indicate it somehow, and I would suggest: for tabbed windows, highlighting the right half of the top border, right border and half of the bottom border of its title bar; for stacked windows, highlighting the top and bottom border of its title bar.

The rationale for that would be that  the user could always know what will happen if he opens a new window, even if he does not remember all he did before: if he is in a container with only a single window, a border highlight will tell him that, and if he is in a container with already several windows, their current layout will thell him.

Airblader:
Two screenshots showing what this looks like (with a rather extreme border):

screenshot-2015-09-11_16-05-07
screenshot-2015-09-11_16-05-14

Note that the container in the screenshot is not just a floating window, but a floating stacked / tabbed container with only one child (since this is the only situation the indicator is shown).

fixes #1067

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -450,16 +450,24 @@ void x_draw_decoration(Con *con) {
          * opened if we are rendering a single window within a split container
          * (which is undistinguishable from a single window outside a split
          * container otherwise. */
-        if (TAILQ_NEXT(con, nodes) == NULL &&
-            TAILQ_PREV(con, nodes_head, nodes) == NULL &&
-            con->parent->type != CT_FLOATING_CON) {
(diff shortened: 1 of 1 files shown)


stapelberg:
Commented in issue #1067

Airblader:
Closing since we don't want to go for this.

PR #1917, opened 2015-09-12
Use EWMH window as focus fallback

### LINKED ISSUE #1378: Multi-montior setup: keyboard input send to unfocused window
[Originally reported by hjem@…]
(Today, I stumbled upon a focus problem in a dual montior setup, where the keyboard input is sent to an unfocused window. This can only happen with option focus_follows_mouse disabled.

Given two workspaces: one workspace on the left montior with one window and one empty workspace on the right monitor. Starting with focus on the left window; if I now would focus the empty workspace on the right, the mouse pointer is moved to the empty workspace as expected and keyboard input is handled correctly. But if I would then move the mouse pointer over the window on the left workspace while keeping focus on the empty workspace on the right, keyboard input is sent to the unfocused window on the left.

I attached the log of the above described problem, where I sent a "nop starting" to i3 to mark the start of my command chain.

Airblader:
I think this should be testable, so I'll add some tests. I gave it a quick test and the ghosting bug seems to have disappeared as expected.

fixes #1378

### DIFF
--- a/include/i3.h
+++ b/include/i3.h
@@ -37,6 +37,15 @@ extern bool debug_build;
 extern int listen_fds;
 extern xcb_connection_t *conn;
 extern int conn_screen;
+/**
+ * The EWMH support window that is used to indicate that an EWMH-compliant
+ * window manager is present. This window is created when i3 starts and
+ * kept alive until i3 exits.
+ * We also use this window as the focused window if no other window is
+ * available to be focused on the active workspace in order to prevent
+ * keyboard focus issues (see #1378).
+ */
+extern xcb_window_t ewmh_window;
 /** The last timestamp we got from X11 (timestamps are included in some events
  * and are used for some things, like determining a unique ID in startup
  * notification). */
--- a/src/ewmh.c
+++ b/src/ewmh.c
@@ -11,6 +11,8 @@
  */
 #include "all.h"
 
+xcb_window_t ewmh_window;
+
 /*
  * Updates _NET_CURRENT_DESKTOP with the current desktop number.
  *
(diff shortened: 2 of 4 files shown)


Airblader:
@stapelberg OK, I've added a test and made sure it fails before the fix. The test only asserts the correct fallback, though, not the "ghosting" effect, but that should be OK.

src/ewmh.c
```diff
@@ -222,30 +224,29 @@ void ewmh_setup_hints(void) {
 #undef xmacro
     };
 
-    /* Set up the window manager’s name. According to EWMH, section "Root Window
-     * Properties", to indicate that an EWMH-compliant window manager is
-     * present, a child window has to be created (and kept alive as long as the
-     * window manager is running) which has the _NET_SUPPORTING_WM_CHECK and
-     * _NET_WM_ATOMS. */
```
stapelberg: Why delete this comment? I think it’d be worthwhile to keep it.

include/i3.h
```diff
@@ -37,6 +37,14 @@ extern bool debug_build;
 extern int listen_fds;
 extern xcb_connection_t *conn;
 extern int conn_screen;
+/**
+ * The EWMH support window that is used to indicate that an EWMH-compliant
+ * window manager is present. This window is created when i3 starts and
+ * kept alive until i3 exits.
+ * We also use this window as the focused window if no other window is
+ * available to be focused on the active workspace.
```
stapelberg: add: “in order to prevent keyboard focus issues (see bug #1378)”

src/ewmh.c
```diff
@@ -222,30 +224,29 @@ void ewmh_setup_hints(void) {
...
-    xcb_window_t child_window = xcb_generate_id(conn);
+    ewmh_window = xcb_generate_id(conn);
     xcb_create_window(
         conn,
         XCB_COPY_FROM_PARENT,        /* depth */
-        child_window,                /* window id */
+        ewmh_window,                 /* window id */
         root,                        /* parent */
-        0, 0, 1, 1,                  /* dimensions (x, y, w, h) */
+        -100, -100, 1, 1,            /* dimensions (x, y, w, h) */
```
stapelberg: Why -100? Are these magic numbers? If not, will the numbers cause problems if I chose to arrange my outputs so that they overlap by more/less than 100 pixels?

src/ewmh.c
Airblader: I did keep it (mostly), but I put it where the variable is defined now.

src/ewmh.c
Airblader: It's just to put it off-screen. It isn't really needed (at least with the original values I didn't see any window either), but more to make it clear that this is not meant to be a visible window. As for why this and not any other value – because Openbox uses `-100`. The value doesn't really matter.

include/i3.h
Airblader: @stapelberg Done.

src/ewmh.c
stapelberg: The comment close to the variable doesn’t convey that the EWMH standard requires us to keep the window around (and where that requirement is located). I think keeping both would be good.

src/ewmh.c
stapelberg: Can you use -1, -1, 1, 1 instead and comment that the coordinates where chosen so that the window ends up offscreen, please?

src/ewmh.c
Airblader: OK, done.

src/ewmh.c
Airblader: @stapelberg  Done.

PR #1920, opened 2015-09-14
Set and unset individual atoms in _NET_WM_STATE

### LINKED ISSUE #1873: Correctly set multiple atoms in _NET_WM_STATE
We currently only ever replace `_NET_WM_STATE` for the different atoms we support. This violates the specification:

See http://standards.freedesktop.org/wm-spec/1.3/ar01s05.html:

> Atoms present in the list MUST be considered set, atoms not present in the list MUST be considered not set.

Airblader:
fixes #1873

### DIFF
--- a/src/xcb.c
+++ b/src/xcb.c
@@ -274,3 +274,48 @@ xcb_visualid_t get_visualid_by_depth(uint16_t depth) {
     }
     return 0;
 }
+
+/*
+ * Add an atom to a list of atoms the given property defines.
+ * This is useful, for example, for manipulating _NET_WM_STATE.
+ *
+ */
+void xcb_add_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_atom_t property, xcb_atom_t atom) {
+    xcb_change_property(conn, XCB_PROP_MODE_APPEND, window, property, XCB_ATOM_ATOM, 32, 1, (uint32_t[]){atom});
+}
+
+/*
+ * Remove an atom from a list of atoms the given property defines without
+ * removing any other potentially set atoms.  This is useful, for example, for
+ * manipulating _NET_WM_STATE.
+ *
+ */
+void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_atom_t property, xcb_atom_t atom) {
+    xcb_grab_server(conn);
+
+    xcb_get_property_reply_t *reply =
+        xcb_get_property_reply(conn,
+                               xcb_get_property(conn, false, window, property, XCB_GET_PROPERTY_TYPE_ANY, 0, 4096), NULL);
+    if (reply == NULL || xcb_get_property_value_length(reply) == 0)
+        goto release_grab;
+    xcb_atom_t *atoms = xcb_get_property_value(reply);
+    if (atoms == NULL) {
+        goto release_grab;
+    }
+
+    {
+        int num = 0;
+        const int current_size = xcb_get_property_value_length(reply) / (reply->format / 8);
+        xcb_atom_t values[current_size];
+        for (int i = 0; i < current_size; i++) {
+            if (atoms[i] != atom)
+                values[num++] = atoms[i];
+        }
+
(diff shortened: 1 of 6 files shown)


src/xcb.c
```diff
@@ -274,3 +274,39 @@ xcb_visualid_t get_visualid_by_depth(uint16_t depth) {
...
+}
+
+/*
+ * Remove an atom from a list of atoms the given property defines without
+ * removing any other potentially set atoms.  This is useful, for example, for
+ * manipulating _NET_WM_STATE.
+ *
+ */
+void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_atom_t property, xcb_atom_t atom) {
+    xcb_get_property_reply_t *reply =
```
stapelberg: This line introduces a memory leak because you never `free(reply);`

src/xcb.c
```diff
@@ -274,3 +274,39 @@ xcb_visualid_t get_visualid_by_depth(uint16_t depth) {
...
+        xcb_get_property_reply(conn,
+                               xcb_get_property(conn, false, window, property, XCB_GET_PROPERTY_TYPE_ANY, 0, 4096), NULL);
+    if (reply == NULL || xcb_get_property_value_length(reply) == 0)
+        return;
+    xcb_atom_t *atoms = xcb_get_property_value(reply);
+    if (atoms == NULL)
+        return;
+
+    int num = 0;
+    int current_size = xcb_get_property_value_length(reply) / (reply->format / 8);
```
stapelberg: `current_size` can be marked `const`

src/xcb.c
```diff
@@ -274,3 +274,39 @@ xcb_visualid_t get_visualid_by_depth(uint16_t depth) {
...
+    xcb_change_property(conn, XCB_PROP_MODE_APPEND, window, property, XCB_ATOM_ATOM, 32, 1, (uint32_t[]){atom});
+}
+
+/*
+ * Remove an atom from a list of atoms the given property defines without
+ * removing any other potentially set atoms.  This is useful, for example, for
+ * manipulating _NET_WM_STATE.
+ *
+ */
+void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_atom_t property, xcb_atom_t atom) {
```
stapelberg: This function is not atomic, but I think it should be. Can you add `xcb_grab_server()` calls please?

Airblader:
@stapelberg Done

src/xcb.c
```diff
@@ -274,3 +274,46 @@ xcb_visualid_t get_visualid_by_depth(uint16_t depth) {
...
+
+    xcb_get_property_reply_t *reply =
+        xcb_get_property_reply(conn,
+                               xcb_get_property(conn, false, window, property, XCB_GET_PROPERTY_TYPE_ANY, 0, 4096), NULL);
+    if (reply == NULL || xcb_get_property_value_length(reply) == 0)
+        goto release_grab;
+    xcb_atom_t *atoms = xcb_get_property_value(reply);
+    if (atoms == NULL)
+        goto release_grab;
+    {
```
stapelberg: Can you please make the previous line read the following so that people aren’t confused by the separate scope?

```c
if (atoms == NULL) {
    goto release_grab;
}
```

src/xcb.c
```diff
@@ -274,3 +274,46 @@ xcb_visualid_t get_visualid_by_depth(uint16_t depth) {
...
+        const int current_size = xcb_get_property_value_length(reply) / (reply->format / 8);
+        xcb_atom_t values[current_size];
+        for (int i = 0; i < current_size; i++) {
+            if (atoms[i] != atom)
+                values[num++] = atoms[i];
+        }
+
+        xcb_change_property(conn, XCB_PROP_MODE_REPLACE, window, property, XCB_ATOM_ATOM, 32, num, values);
+    }
+    FREE(reply);
```
stapelberg: I think the `FREE(reply);` should go into the release_grab. All jumps to release_grab are after `reply` was allocated.

Airblader: Oh c***, you're right. I even thought about that and still did it wrong.

Airblader:
@stapelberg Done.

PR #1921, opened 2015-09-14
randr: use root window in case of no randr outputs

### LINKED ISSUE #926: i3 dies if there are no available outputs.
[Originally reported by cneumann@…]
(If I disable all outputs via xrandr, i3 dies with the message "No usable outputs available.". i3 should not die but wait until a now output appears.

tcatm:
This patch introduces a root output covering the root window. It is used in two cases:

1. RandR is not available. In this case, the previous behaviour of creating a single output covering the root window is preserved.

2. RandR is available, but there is no active output. In this case, the root output is enabled and will be the only active output. If any RandR output becomes available, the root output will be disabled again. Existing mechanisms for migrating workspaces will just work without modification.

I've carefully slipped in a global variable `Output root_output` representing that output. I'm not exactly happy with it, though. We could, by convention, consider the first output in outputs to be the root output. This would work because the root output will always be created before any other output.

I also had to alter the `struct xoutput` a little by introducing a `bool root`. This could be obsoleted by introducing the aforementioned convention, too.

As a side effect, "Add a timeout: delay_exit_on_zero_displays" has been reverted.

Fixes #926 and #1489

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1011,31 +1011,6 @@ force_display_urgency_hint <timeout> ms
 force_display_urgency_hint 500 ms
 ---------------------------------
 
-=== Delaying exiting on zero displays
-
-Outputs may disappear momentarily and come back later. For example,
-using a docking station that does not announce the undock (e.g. ACPI Undock 
-event triggered through manually pushing a button before actually ejecting 
-the notebook). During the removal of the notebook from the docking station,
-all outputs disappear momentarily.
-
-To prevent i3 from exiting when no output is available momentarily, you can 
-tell i3 to delay a certain time first and check available outputs again using 
-the +delay_exit_on_zero_displays+ directive. Setting the value to 0 disables 
-this feature.
-
-The default is 500ms.
-
-*Syntax*:
-----------------------------------------
-delay_exit_on_zero_displays <timeout> ms
-----------------------------------------
-
-*Example*:
-----------------------------------
-delay_exit_on_zero_displays 500 ms
-----------------------------------
-
 === Focus on window activation
 
 [[focus_on_window_activation]]
--- a/include/config.h
+++ b/include/config.h
@@ -167,10 +167,6 @@ struct Config {
      * flag can be delayed using an urgency timer. */
     float workspace_urgency_timer;
 
-    /** Use a timer to delay exiting when no output is available.
-      * This can prevent i3 from exiting when all outputs disappear momentarily. */
-    float zero_disp_exit_timer_ms;
-
     /** Behavior when a window sends a NET_ACTIVE_WINDOW message. */
     enum {
         /* Focus if the target workspace is visible, set urgency hint otherwise. */
--- a/include/config_directives.h
+++ b/include/config_directives.h
@@ -51,7 +51,6 @@ CFGFUN(force_focus_wrapping, const char *value);
 CFGFUN(force_xinerama, const char *value);
 CFGFUN(fake_outputs, const char *outputs);
 CFGFUN(force_display_urgency_hint, const long duration_ms);
-CFGFUN(delay_exit_on_zero_displays, const long duration_ms);
 CFGFUN(focus_on_window_activation, const char *mode);
 CFGFUN(show_marks, const char *value);
 CFGFUN(hide_edge_borders, const char *borders);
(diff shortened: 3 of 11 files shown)


stapelberg:
Can you explain why this is better than the commit you’re reverting?

tcatm:
It doesn't call `die()` when doing `xrandr --output LVDS1 --off` (or disabling all outpus by other means, e.g. disconnecting all display port links). I recently switched to i3 and so far I'm liking it but I was running into this on a daily basis and it got annoying to restart the X session.

stapelberg:
Well, neither does delay_exit_on_zero_displays.

I still don’t understand why this commit is necessary. Can you please explain your use-case in more detail?

tcatm:
`delay_exit_on_zero` display will just delay the call to `die()` for a short time (500ms by default).

The scenario in which I encountered the bug is this:

1. My laptop is docked, lid closed and connected to an Displayport monitor.
2. Undocking the laptop will disconnect the external monitor.
3. This causes X to disable the output. There are now no enabled output. This is a sane condition for X as there is still the root window (which has a minimum size enforced) where applications can output things. Mouse and keyboard input does still work at this point but there's obviously no output.
4. After opening the lid my laptop will enable the internal panel again and configure it as primary output (or after redocking it would configure an external monitor as its output).

The delay between 3 and 4 may be longer than a few seconds as it's uncommon to undock a laptop and open the lid immediately. I'm more likely to put in my backpack, travel somewhere else and then waking it up by opening the lid.

Depending on timing `delay_exit_on_zero` will help in this case. It still introduces a nasty race condition. If I'm too slow, i3 may decide to kill my X session taking all running programs (think compile job running, unsaved documents, ...) with it. That's not something a window manager should do.

Increasing `delay_exit_on_zero` doesn't really solve the problem for me, either. `usleep()` will stall i3 and thus prevent it from handling events (both from windows and keyboard/mouse). I like i3 to handle keys like the suspend button and volume control and this should be possible even with no outputs connected.

Therefore, I decided to avoid the code path to `die("No output available")` altogether and even simplify the code a little bit. Now that the root window is an output, i3 is guaranteed to always have an output for migrating workspaces to thus having to reason to kill my X session anymore.

In https://github.com/i3/i3/issues/926#issuecomment-72825486 you'd agreed that this would be a better behaviour but you didn't know where to put the windows with no outputs. I believe my proposal solves this in simple and robust manner without adding anymore code complexity.

I've also mapped out all places where an active output and workspace is required in order to make i3 work without them. It turned out to be doable but being slightly more complex and touching a lot more code so I came up with a simpler fix first.

Let me know if you'd have done it differently. I can spare to work a few more hours on this issue.

Btw, I'm not sure why the testsuite fails. Running the tests locally works just fine. Could you advise? Did I change some subtle timing?

stapelberg:
Thanks for explaining your use-case, that makes sense to me.

I think the tests fail because you’ve moved `create_root_output(conn);` out of `randr_disable`, which is also called from src/xinerama.c. This breaks the case where RandR (and perhaps even Xinerama) is not available.

tcatm:
Oh, I see. Sorry!

In this case, I'd like to refactor `xinerama.c` and even `main.c` a little so that (in case of no fake outputs) a root output is always created and subsequently disabled by either `xinerama_init` or `randr_init` when other outputs are found. This should reduce code complexity a little and make i3 even more robust in regard to "strange output configurations".

Do you prefer appending commits to this PR or squashing existing commits?

stapelberg:
I’d prefer it if you could squash your commits into logical units that make sense when looking at the git history.

tcatm:
I've reworked and squashed the patch. I didn't refactor `xinerama.c` and `main.c` as that would have required rewriting most of `xinerama.c`. Instead, I'd prefer to get this merged now and work on refactoring  output handling in a seperate PR (i.e. moving code for adding, removing and overlap-detection of outputs to `output.c` and calling that from `xinerama.c` and `randr.c`. Fake outputs might then use that code too so we could actually run unittests on this later).

I've tested with Xvfb, Xephyr and the OSS Intel driver and it works fine. I also ran clang-format on all files I've touched. Thanks for your feedback! Let me know if anything else needs to be taken care of in this patch.

Issue #1929, opened 2015-09-17
enhancement: make “debuglog on” command persist via restarts


stapelberg:
The `-d all` command-line parameter should be added by the code when restarting (just like `-a` is added). See #1810 for an issue where the lack of this behavior made debugging harder.

PR #1930, opened 2015-09-17
Make “debuglog on” command persist via restarts

hwangcc23:
When restarting, add "-d all" if debuglog is on.
Reference: issue #1929

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -169,24 +169,35 @@ bool path_exists(const char *path) {
 }
 
 /*
- * Goes through the list of arguments (for exec()) and checks if the given argument
- * is present. If not, it copies the arguments (because we cannot realloc it) and
- * appends the given argument.
- *
+ * Goes through the list of arguments (for exec()) and add/replace the given option,
+ * including the option name, its argument, and the option character.
  */
-static char **append_argument(char **original, char *argument) {
+static char **add_argument(char **original, char *opt_char, char *opt_arg, char *opt_name) {
     int num_args;
-    for (num_args = 0; original[num_args] != NULL; num_args++) {
-        DLOG("original argument: \"%s\"\n", original[num_args]);
-        /* If the argument is already present we return the original pointer */
-        if (strcmp(original[num_args], argument) == 0)
-            return original;
+    for (num_args = 0; original[num_args] != NULL; num_args++)
+        ;
+    char **result = scalloc(num_args + 3, sizeof(char *));
+
+    /* copy the arguments, but skip the ones we'll replace */
+    int write_index = 0;
+    bool skip_next = false;
+    for (int i = 0; i < num_args; ++i) {
+        if (skip_next) {
+            skip_next = false;
+            continue;
+        }
+        if (!strcmp(original[i], opt_char) ||
+            (opt_name && !strcmp(original[i], opt_name))) {
+            if (opt_arg)
+                skip_next = true;
+            continue;
+        }
+        result[write_index++] = original[i];
     }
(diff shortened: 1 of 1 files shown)


src/util.c
```diff
@@ -269,6 +269,10 @@ void i3_restart(bool forget_layout) {
     LOG("restarting \"%s\"...\n", start_argv[0]);
     /* make sure -a is in the argument list or append it */
     start_argv = append_argument(start_argv, "-a");
+    /* make debuglog-on persist */
+    if (get_debug_logging()) {
+        start_argv = append_argument(start_argv, "-d all");
```
stapelberg: This won’t work, because "-d" and "all" must be separate entries in argv. I think you’ll need to use the same approach that line 280 and following are using. Maybe refactor that code into a re-usable function.

hwangcc23: Thanks for the review. I will fix it and resend the PR after work.

hwangcc23:
@stapelberg Fixed.

src/util.c
```diff
@@ -191,6 +191,37 @@ static char **append_argument(char **original, char *argument) {
...
+    int num_args;
+    for (num_args = 0; original[num_args] != NULL; num_args++)
+        ;
+    char **result = scalloc(num_args + 3, sizeof(char *));
+
+    /* copy the arguments, but skip the ones we'll replace */
+    int write_index = 0;
+    bool skip_next = false;
+    for (int i = 0; i < num_args; ++i) {
+        if (skip_next)
```
stapelberg: With the following construct, you can safe an `else` and make the control flow easier.

```c
if (skip_next) {
  skip_next = false;
  continue;
}
```

src/util.c
```diff
@@ -270,33 +301,14 @@ void i3_restart(bool forget_layout) {
     /* make sure -a is in the argument list or append it */
     start_argv = append_argument(start_argv, "-a");
```
stapelberg: I think you can delete `append_argument` and replace this call with `start_argv = add_argument(start_argv, "-a", NULL, NULL);`, right?

src/util.c
```diff
@@ -191,6 +191,37 @@ static char **append_argument(char **original, char *argument) {
     return result;
 }
 
+/*
+ * Goes through the list of arguments (for exec()) and add/replace the given option,
+ * including the option name, its argument, and the option character.
+ */
+static char **add_argument(char **original, char *opt_name, char *opt_arg, char *opt_char) {
```
stapelberg: I think the signature should be `(char **original, char *opt_char, char *opt_arg, char *opt_name)`, because every option has a char, but not every option has a name. In addition, in the call below, you actually specify “-d” as the second parameter.

hwangcc23:
@stapelberg Done

src/main.c
```diff
@@ -206,7 +206,7 @@ int main(int argc, char *argv[]) {
         {"more_version", no_argument, 0, 'm'},
         {"help", no_argument, 0, 'h'},
         {"layout", required_argument, 0, 'L'},
-        {"restart", required_argument, 0, 0},
+        {"restart", required_argument, 0, 'r'},
```
stapelberg: Why did you add -r here? Given that this option is supposed to only be set by i3, I don’t think we should provide a short option character for it. If you’re worried about the call to add_argument later, just pass --restart as opt_char?

hwangcc23: Yes. It is about the call to add_argument.
Will follow your suggestion to pass --restart as opt_char.

hwangcc23:
@stapelberg Updated.

PR #1931, opened 2015-09-17
Improvements for sticky windows

### LINKED ISSUE #1924: Border is rendered with wrong color for sticky windows on empty workspace
**i3 version 4.10.4-332-g7de15b3**

**Problem 1** - Example in the docs:
  *for_window [class="urxvt"] resize set 640 480*
should be changed to:
  *for_window [class="URxvt"] resize set 640 480*

**Problem 2** - *bind*/*bindcode* doesn't work:
For example:
  *bind Mod+24 sticky toggle*
or
  *bindcode Mod+24 sticky toggle*
*both* report an error.

**Problem 3** - unable to toggle (off) sticky property:
Steps to reproduce:
  1. *sticky toggle* on a window
  2. go to an empty workspace
  3. try to execute - *sticky toggle* (this time to off) - it does not work.

I'm pretty sure this isn't intended.

Airblader:
fixes #1924

@stapelberg I guess we need to decide whether focusing a sticky container if the target workspace is empty is what we want. On the one hand, it makes sense to focus the container on that workspace since that's the behavior we also have if one was on there before. On the other hand we have your point that accidentally switching workspaces means when the user switches back to the correct one, the sticky window will be focused.

### DIFF
--- a/include/output.h
+++ b/include/output.h
@@ -27,4 +27,4 @@ Output *get_output_from_string(Output *current_output, const char *output_str);
  * workspace on that output.
  *
  */
-void output_push_sticky_windows(void);
+void output_push_sticky_windows(Con *to_focus);
--- a/src/commands.c
+++ b/src/commands.c
@@ -1598,7 +1598,7 @@ void cmd_sticky(I3_CMD, char *action) {
 
     /* A window we made sticky might not be on a visible workspace right now, so we need to make
      * sure it gets pushed to the front now. */
-    output_push_sticky_windows();
+    output_push_sticky_windows(focused);
 
     cmd_output->needs_tree_render = true;
     ysuccess(true);
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -736,7 +736,7 @@ static void handle_client_message(xcb_client_message_event_t *event) {
                 con->sticky = !con->sticky;
 
             DLOG("New sticky status for con = %p is %i.\n", con, con->sticky);
-            output_push_sticky_windows();
+            output_push_sticky_windows(focused);
         }
 
         tree_render();
--- a/src/output.c
+++ b/src/output.c
@@ -52,31 +52,34 @@ Output *get_output_from_string(Output *current_output, const char *output_str) {
  * workspace on that output.
  *
  */
-void output_push_sticky_windows(void) {
+void output_push_sticky_windows(Con *to_focus) {
     Con *output;
-    TAILQ_FOREACH(output, &(croot->nodes_head), nodes) {
+    TAILQ_FOREACH(output, &(croot->focus_head), focused) {
         Con *workspace, *visible_ws = NULL;
         GREP_FIRST(visible_ws, output_get_content(output), workspace_is_visible(child));
 
         /* We use this loop instead of TAILQ_FOREACH to avoid problems if the
          * sticky window was the last window on that workspace as moving it in
          * this case will close the workspace. */
-        for (workspace = TAILQ_FIRST(&(output_get_content(output)->nodes_head));
-             workspace != TAILQ_END(&(output_get_content(output)->nodes_head));) {
+        for (workspace = TAILQ_FIRST(&(output_get_content(output)->focus_head));
+             workspace != TAILQ_END(&(output_get_content(output)->focus_head));) {
             Con *current_ws = workspace;
-            workspace = TAILQ_NEXT(workspace, nodes);
+            workspace = TAILQ_NEXT(workspace, focused);
 
             /* Since moving the windows actually removes them from the list of
              * floating windows on this workspace, here too we need to use
              * another loop than TAILQ_FOREACH. */
             Con *child;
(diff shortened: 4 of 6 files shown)


aksr:
Just tested this and I like this solution...
Btw, I think it would be better if `sticky toggle` would automatically, if the window isn't already floating, make it floating.

`floating toggle; sticky toggle` is not good enough:
  * If the window is already floating, it would became *tiled*.
  * Separate key-binding for `sticky toggle` would still be needed.

This way would `sticky toggle` always "work" - meaning - it would always give result.

Airblader:
What's wrong with 

    floating enable, sticky toggle

? Sure, executing that again won't tile the container, but neither would auto floating it and then making it unsticky.

aksr:
You're right! I forgot about `floating enable`.
I would just replace `,` with `;`:
  `floating enable; sticky toggle`
I think pull request is fine, ignore the rest of my comment.
As for the accidental switch - there's no way to know if it's accidental or intentional. 

Anyway, returning to the previous workspace and executing `floating toggle` isn't that *hard*.

stapelberg:
> @stapelberg I guess we need to decide whether focusing a sticky container if the target workspace is empty is what we want. On the one hand, it makes sense to focus the container on that workspace since that's the behavior we also have if one was on there before. On the other hand we have your point that accidentally switching workspaces means when the user switches back to the correct one, the sticky window will be focused.

I’d say that all workspaces should have tiling windows in normal operation, so I think _not_ focusing the window makes more sense.

aksr:
Yes, but creating a new window removes the focus from the sticky window. You don't have any tiling window(s) so it doesn't matter if the tiling window is focused, since, there's no tiling window(s). If you're on the empty workspace - you will, either, create new window(s) or edit/follow sticky window.

stapelberg:
I don’t see how that’s relevant to what I said before. As far as I understand you, you’re saying you don’t have tiling windows on that workspace, and I’m saying that’s an edge case which we can ignore when making this decision.

Airblader:
@stapelberg Sorry, but I disagree with you here as well. Opening a new workspace is a case where we don't have a tiling window (yet) and I think that's an absolute standard case, not an edge case.

aksr:
Edge case? - Window appears as if it's focused but in truth it isn't and you don't know when it's focused or not (on an empty workspace).
I think this is far away from being the edge case.
I don't have any window on an empty workspace (as is the case with everybody).
I'm just saying if you move the sticky window to do something with it and having it focused, 
doesn't trouble you because opening a window focuses a newly created window - meaning - sticky window being focused isn't a problem or an obstacle.
And if for some reasons you move the sticky window to an empty workspace and you want to edit something in a sticky window, 
you'll need to execute `focus mode_toggle` just to focus it even do it looks as if it's focused.

Anyway, I don't want to argue anymore.
I already wrote multiple times reasons and I think I made my point.
Do as you will.

Airblader:
> I’d say that all workspaces should have tiling windows in normal operation, so I think not focusing the window makes more sense.

Just to avoid misunderstandings: if the target workspace is *not* empty, which is what you describe as normal operation, the sticky window will not be focused (even after this patch). This is only about the target workspace being empty (i.e., a new workspace).

stapelberg:
> Window appears as if it's focused but in truth it isn't and you don't know when it's focused or not (on an empty workspace).

To be clear, I think it’s a bug that a window can look like it’s focused even though it’s not actually focused.

> And if for some reasons you move the sticky window to an empty workspace and you want to edit something in a sticky window, 

A sticky window is on every workspace, so “moving a sticky window to an empty workspace” doesn’t make sense to me.

I still think we should not focus sticky windows when opening a new workspace. I think having to use one focus command is an acceptable burden so that we can continue not to disturb focus while switching workspaces.

Airblader:
> To be clear, I think it’s a bug that a window can look like it’s focused even though it’s not actually focused.

Even if it is because the workspace is focused? Because this seems to be deliberate behavior in i3.

> I still think we should not focus sticky windows when opening a new workspace. I think having to use one focus command is an acceptable burden so that we can continue not to disturb focus while switching workspaces.

Alright. I'll adapt the patch then.

stapelberg:
> > To be clear, I think it’s a bug that a window can look like it’s focused even though it’s not actually focused.
>
> Even if it is because the workspace is focused? Because this seems to be deliberate behavior in i3.

What makes you think this behavior is deliberate?

I’d say that focused_inactive is the color class that should be used whenever your keypresses don’t actually go into the window, but it’s focused in the i3 tree structure (just like with not-focused workspaces right now).

Airblader:
> What makes you think this behavior is deliberate?
> just like with not-focused workspaces right now

Because there's code there which does exactly what we are seeing. This doesn't just apply to workspaces, but also to nested containers. If you select a parent, everything inside that parent will be colored with the focused group.

I'll open a bug for this.

Airblader:
@stapelberg Done.

PR #1933, opened 2015-09-19
Draw i3bar with cairo

### LINKED ISSUE #1878: [enhancement] Use cairo to render i3bar

Airblader:
fixes #1878 

@stapelberg I know we won't merge this before 4.11, just already putting the PR out there for review. I'll also test this a bunch in different situations (adding outputs, removing them, changing them etc.) and document that process here.

I tried to prepare it as much as possible to be reused when drawing decorations with cairo without actually moving it outside of i3bar yet.

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -30,6 +30,7 @@
 #include <X11/Xlib.h>
 #include <X11/XKBlib.h>
 #include <X11/extensions/XKB.h>
+#include <cairo/cairo-xcb.h>
 
 #include "common.h"
 #include "libi3.h"
@@ -72,9 +73,7 @@ int mod_pressed = 0;
 
 /* Because the statusline is the same on all outputs, we have
  * global buffer to render it on */
-xcb_gcontext_t statusline_ctx;
-xcb_gcontext_t statusline_clear;
-xcb_pixmap_t statusline_pm;
+surface_t statusline_surface;
 uint32_t statusline_width;
 
 /* Event watchers, to interact with the user */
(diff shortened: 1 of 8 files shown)


Airblader:
* Using bar on a single screen works.
* Plugging in another screen during runtime works.
* Reconfiguring a bar to be moved to another screen during runtime works.
* Turning off a second screen during runtime works.

i3bar/src/xcb.c
```diff
@@ -173,17 +170,18 @@ static void draw_separator(uint32_t x, struct status_block *block) {
     uint32_t center_x = x - sep_offset;
     if (config.separator_symbol == NULL) {
         /* Draw a classic one pixel, vertical separator. */
-        uint32_t mask = XCB_GC_FOREGROUND | XCB_GC_BACKGROUND | XCB_GC_LINE_WIDTH;
-        uint32_t values[] = {colors.sep_fg, colors.bar_bg, logical_px(1)};
-        xcb_change_gc(xcb_connection, statusline_ctx, mask, values);
-        xcb_poly_line(xcb_connection, XCB_COORD_MODE_ORIGIN, statusline_pm, statusline_ctx, 2,
-                      (xcb_point_t[]){{center_x, logical_px(sep_voff_px)},
-                                      {center_x, bar_height - logical_px(sep_voff_px)}});
+        cairo_set_line_width(statusline_surface.cr, 1.0);
```
stapelberg: I think you should be using `logical_px(1)` here instead of `1.0`. In its current state, this commit changes the rendering of i3bar for me — the separator line becomes thinner.

Airblader: @stapelberg Yeah, that's fixed now. It'll be interesting to see if on a HiDPI display the offset "trick" now still works as intended. If it doesn't we can always switch to drawing a rectangle as we did before and then it'll work fine.

stapelberg:
The rendering is still different. Compare the old rendering:
i3bar-before

…with the new rendering:
i3bar-new

When you zoom in, you’ll see the old rendering drew sharp pixels, whereas the new one draws a sharp line in the middle and two lines to the left/right with <100% opacity:
i3bar-opacity

Airblader:
OK, I'll switch to drawing a rectangle instead.

Airblader:
@stapelberg What font are you using? I'm getting that aliasing effect even without this patch. Either way I switched to the rectangle drawin now. Does that make it work better for you?

stapelberg:
Yeah, with the switch to drawing rectangles, the rendering is now pixel-perfect for me.

PR #1942, opened 2015-09-21
Added small readme to point users to website

dg-ratiodata:
Github displays the README if such a file exists. I added a small README to point users to i3's website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,14 @@
+# i3 &ndash; a tiling window manager
+
+## Users
+
+For information about i3 head over to our [website](https://i3wm.org/).
+
+## Developers & Users
+
+Read our [CONTRIBUTING.md](CONTRIBUTING.md), [DEPENDS](DEPENDS) and
+[PACKAGE-MAINTAINER](PACKAGE-MAINTAINER) file.
+
+## License
+
+Our license can be found [here](LICENSE).


README.md
```diff
@@ -0,0 +1,15 @@
+# i3 &ndash; a tiling window manager
+
+## Users
+
+i3 is a tiling window manager, completely written from scratch. For more
+information head over to our [website](https://i3wm.org/).
+
+## Developers & Users
+
+Read our [CONTRIBUTING.md](CONTRIBUTING.md)-, [DEPENDS](DEPENDS)- and
```
Airblader: Please remove the dashes, compounds aren't hyphenated like this in English.

README.md
```diff
@@ -0,0 +1,15 @@
+# i3 &ndash; a tiling window manager
+
+## Users
+
+i3 is a tiling window manager, completely written from scratch. For more
```
Airblader: s/completely written/written completely/. But I don't really think that if we have to describe i3 in 1–2 sentences that this is the important information. I'd personally just leave that part of the sentence out.

dg-ratiodata: Ok. Then you may also want to change the website ;-) I took over that sentence from there.

Airblader: Yeah, I figured. ;) I think the landing page could actually deserve an update as some things there strike me as outdated.

README.md
```diff
@@ -0,0 +1,14 @@
+# i3 &ndash; a tiling window manager
+
+## Users
+
+For information about i3 head over to our [website](https://i3wm.org/).
+
+## Developers & Users
+
+Read our [CONTRIBUTING.md](CONTRIBUTING.md), [DEPENDS](DEPENDS) and
+[PACKAGE-MAINTAINER](PACKAGE-MAINTAINER)-file.
```
Airblader: There's still a hyphen there. :9

dg-ratiodata: Fixed. :-)

README.md
dg-ratiodata: Maybe
* less text or some more icons ([FontAwesome](https://fortawesome.github.io/Font-Awesome/icons/))to make it easier to grab the unique selling points of i3
* A link "development" to make it easier to find your repository

README.md
```diff
@@ -0,0 +1,14 @@
+# i3 &ndash; a tiling window manager
+
+## Users
+
+For information about i3 head over to our [website](https://i3wm.org/).
```
stapelberg: Our website is already linked to at the top of the page, where every GitHub repository can link to a project website. I don’t think we need to repeat it here.

README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Developers & Users
+
+Read our [CONTRIBUTING.md](CONTRIBUTING.md), [DEPENDS](DEPENDS) and
```
stapelberg: People will see a link to the CONTRIBUTING file before they send a PR/file an issue.

People are never supposed to read DEPENDS, because why would you build from head instead of using the distro packages?

Likewise, only a handful of people (20, tops) are package maintainers and should read the PACKAGE-MAINTAINERS file.

I don’t think these references add a lot of value.

README.md
```diff
@@ -0,0 +1,14 @@
...
+[PACKAGE-MAINTAINER](PACKAGE-MAINTAINER) file.
+
+## License
+
+Our license can be found [here](LICENSE).
```
stapelberg: People who care about licenses will definitely find a file that’s called LICENSE.

stapelberg:
Sorry to sound so negative on this PR, I realize that you’re just trying to improve things.

Bottomline, I don’t think having a README file is necessary, see my comments on the specifics. Also keep in mind that the file would need to be updated and is just one more piece of documentation that lives outside of where we keep the rest of our docs.

PR #1945, opened 2015-09-21
Set a proper WM_CLASS on frame windows.

Airblader:
This allows matching on frame windows, for example in compton. We always say applications should set proper hints on their windows, so maybe we should set a good example. :)


stapelberg:
Can you clang-format this please?

Airblader:
Yeah, sorry, I noticed afterwards but I wasn't home yet.

Airblader:
@stapelberg Done.

Issue #1875, opened 2015-08-28
Several windows change mode from float-mode to tiling-mode


kalterfive:
Procedure on a clean workspace:
1. $mod+enter
2. $mod+enter
3. $mod+shift+space
4. $mod+a
5. $mod+shift+space
6. $mod+shift+space  
With default keybindings

The impossible takes place: several windows change mode from float-mode to tiling-mode.

Latest version (4.10.3)

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

KalterFive:
``` bash
~ $> i3 --version
i3 version 4.10.3-291-gb68a400 (2015-08-28, branch "makepkg") © 2009 Michael Stapelberg and contributors
```

stapelberg:
We will need a logfile to make sense of this. Please follow http://i3wm.org/docs/debugging.html

Airblader:
I'd actually like to see a description of what even happens. Also I think it's pretty pointless to open a ticket named "Something bug". Please change the title to something appropriate. It's rather pointless to tell us about "something".

KalterFive:
I don't know why, but I cannot get logs.  
I run i3:
``` bash
$ cat .xinitrc
exec i3 --shmlog-size=26214400
$ startx
```
Then reproduce the bug.  
Other information:
``` bash
$ file `which i3`
/usr/bin/i3: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=0edf6bd48205e83a1c8a2dfa94ec05bd168413f8, stripped
$ pm -Q gdb
gdb 7.9.1-1
```

Airblader:
I can reproduce this. Log file: http://pastebin.com/pYPejvQh
Backtrace:

````
#0  0x00007fac74993170 in __poll_nocancel () from /usr/lib/libc.so.6
No symbol table info available.
#1  0x00007fac7723a9f2 in ?? () from /usr/lib/libxcb.so.1
No symbol table info available.
#2  0x00007fac7723c56f in xcb_wait_for_event () from /usr/lib/libxcb.so.1
No symbol table info available.
[...]

PR #1952, opened 2015-09-22
Not toggle floating on a CT_FLOATING_CON

hwangcc23:
See the issue #1875.
Forbit the command to toggle floating on a CT_FLOATING_CON to avoid the crash.

### DIFF
--- a/src/floating.c
+++ b/src/floating.c
@@ -369,6 +369,12 @@ void floating_disable(Con *con, bool automatic) {
  *
  */
 void toggle_floating_mode(Con *con, bool automatic) {
+    /* forbid the command to toggle floating on a CT_FLOATING_CON */
+    if (con->type == CT_FLOATING_CON) {
+        ELOG("Cannot toggle floating mode on con = %p because it is of type CT_FLOATING_CON.\n", con);
+        return;
+    }
+
     /* see if the client is already floating */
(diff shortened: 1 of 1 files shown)


src/floating.c
```diff
@@ -369,6 +369,12 @@ void floating_disable(Con *con, bool automatic) {
  *
  */
 void toggle_floating_mode(Con *con, bool automatic) {
+    /* forbit the command to toggle floating on a CT_FLOATING_CON */
```
Airblader: s/forbit/forbid/

src/floating.c
```diff
@@ -369,6 +369,12 @@ void floating_disable(Con *con, bool automatic) {
...
+    if (con->type == CT_FLOATING_CON) {
+        LOG("not toggle floating on a CT_FLOATING_CON\n");
```
Airblader: You're probably not a native English speaker, so no worries, but can we rephrase this to 

    ELOG("Cannot toggle floating mode on %p because it is of type CT_FLOATING_CON.\n", con);

src/floating.c
hwangcc23: Sorry for the typo.

src/floating.c
hwangcc23: Thanks. Done.  :)

Issue #1926, opened 2015-09-16
Duplicate keybinding is not detected


aksr:
**i3 version 4.10.4-332-g7de15b3**

Executing:
`$ xmodmap -pke | grep 'q Q'`
I get:
`keycode  24 = q Q q Q`
The second keybinding is not detected as a duplicate:
`bindcode Mod4+24 sticky toggle`
`bindsym Mod4+q focus parent`

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
The appropriate line in the code is https://github.com/i3/i3/blob/4ae21e3a12eb7c9bbb538bbabf73571adcccb634/src/bindings.c#L517

It’s not trivial to make this work because the keysyms are only translated to keycodes _after_ the config file was parsed.

Pull requests welcome :).

Airblader:
Closing since the PR was merged.

PR #1959, opened 2015-09-25
Check duplicated bindings after translating keysym

hwangcc23:
1). See the issue #1926. For example, the second keybinding is not detected as a duplicate:
        bindcode Mod4+24 sticky toggle
        bindsym Mod4+q focus parent
2). To fix it, check duplicated bindings when translating the keysym to keycodes.

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -319,6 +319,7 @@ void translate_keysyms(void) {
         return;
     }
 
+    bool has_errors = false;
     Binding *bind;
     TAILQ_FOREACH(bind, bindings, bindings) {
         if (bind->input_type == B_MOUSE) {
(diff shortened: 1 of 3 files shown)


hwangcc23:
Tested the patch with 1st launch, restart, mode switch, and Xmodmap.

src/bindings.c
```diff
@@ -396,6 +414,31 @@ void translate_keysyms(void) {
     }
 
     xkb_state_unref(dummy_state);
+
+    if (has_errors) {
```
Airblader: Is this just duplicated code now?

hwangcc23: Must show errors using nagbar here. This is because new duplicated bindings may be found after translating keysyms (translated after the config file was parsed; translated after the keyboard mapping changes).
Or do you mean I should remove duplicated code in check_for_duplicate_bindings()?

Airblader: Yeah, it just seems bad to me that the same code is copied in two different places now. There should be only one code path checking for duplicates or at the very least only one function that handles it which is called in two places.

hwangcc23: Done.

src/bindings.c
```diff
@@ -319,6 +319,7 @@ void translate_keysyms(void) {
         return;
     }
 
+    int has_errors = 0;
```
stapelberg: nit: make this `bool has_errors = false;` and use `has_errors = true;` below, because you use it as a boolean, so there’s no point in actually counting errors.

src/bindings.c
```diff
@@ -389,6 +390,23 @@ void translate_keysyms(void) {
...
+                if (check == bind)
+                    continue;
+                if (check->symbol != NULL)
+                    continue;
+                if (check->keycode != bind->translated_to[n] ||
+                    check->event_state_mask != bind->event_state_mask ||
+                    check->release != bind->release)
+                    continue;
+                has_errors++;
+                if (has_errors) {
```
stapelberg: This condition is always true, isn’t it?

hwangcc23:
@stapelberg Done

PR #1965, opened 2015-09-27
Make the command parser stack typed

Airblader:
I'd like to also refactor the two parsers to have less duplicated code, but I figure a separate PR would be better for that.


Airblader:
@stapelberg By the way, from what I can see the TEST_PARSER flag is actually unused (or I'm misreading something). Can we remove it?

stapelberg:
Will look at this once we released 4.11.

Airblader:
@stapelberg Not sure if you wanted me to notify you again, but just in case – all four PRs can be looked at now. :)

PR #1981, opened 2015-10-06
Allow 32-bit depth visuals and RGBA in i3bar

Airblader:
This is not yet finished. I just have some points I want to bring up for clarification.

### DIFF
--- a/i3bar/src/cairo_util.c
+++ b/i3bar/src/cairo_util.c
@@ -80,3 +80,16 @@ color_t cairo_hex_to_color(const char *color) {
 void cairo_set_source_color(surface_t *surface, color_t color) {
     cairo_set_source_rgba(surface->cr, color.red, color.green, color.blue, color.alpha);
 }
+
+/**
+ * Draw the given text using libi3.
+ * This function also marks the surface dirty which is needed if other means of
+ * drawing are used. This will be the case when using XCB to draw text.
+ *
+ */
+void cairo_draw_text(i3String *text, surface_t *surface, color_t fg_color, color_t bg_color, int x, int y, int max_width) {
+    set_font_colors(surface->gc, fg_color.colorpixel, bg_color.colorpixel);
+    draw_text(text, surface->id, surface->gc, visual_type, x, y, max_width);
+
+    cairo_surface_mark_dirty(surface->surface);
+}
--- a/i3bar/src/cairo_util.c
+++ b/i3bar/src/cairo_util.c
@@ -93,3 +93,53 @@ void cairo_draw_text(i3String *text, surface_t *surface, color_t fg_color, color
 
     cairo_surface_mark_dirty(surface->surface);
 }
+
+/**
+ * Draws a filled rectangle.
+ * This function is a convenience wrapper and takes care of flushing the
+ * surface as well as restoring the cairo state.
+ * Note that the drawing is done using CAIRO_OPERATOR_SOURCE.
+ *
+ */
+void cairo_draw_rectangle(surface_t *surface, color_t color, double x, double y, double w, double h) {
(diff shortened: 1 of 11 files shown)


include/libi3.h
```diff
@@ -405,6 +405,9 @@ bool font_is_pango(void);
 void draw_text(i3String *text, xcb_drawable_t drawable,
                xcb_gcontext_t gc, int x, int y, int max_width);
 
+void draw_text2(i3String *text, xcb_drawable_t drawable, xcb_visualtype_t *visual,
```
Airblader: @stapelberg So this obviously can't stay this way. My question is how we want to deal with it – a properly named separate function or providing the visual in all tools (i3, i3-input, i3-config-wizard, …)?

Airblader:
@stapelberg Should using 32-bit visuals be a configuration option for i3bar?

i3bar/src/cairo_util.c
```diff
@@ -50,15 +51,25 @@ void cairo_surface_free(surface_t *surface) {
  *
  */
 color_t cairo_hex_to_color(const char *color) {
-    char groups[3][3] = {
+    char alpha[2];
+    if (strlen(color) == 9) {
```
stapelberg: To avoid magic numbers, you can use `strlen("#rrggbbaa")` which we recently did in i3status.

include/libi3.h
stapelberg: Could we retain the old behavior when `visual == NULL`? If yes, I’d go for changing the function and all call sites.

stapelberg:
> @stapelberg Should using 32-bit visuals be a configuration option for i3bar? I don't think it's necessary since behavior doesn't change for 24-bit visuals (shouldn't, at least).

It should not, unless we can absolutely not avoid it (e.g. in case this change breaks some older use-cases).

include/libi3.h
Airblader: That's a great idea, let's go for that.

Airblader:
@stapelberg Thanks for the feedback. I've incorporated those things, but there's still some work left to do before we can consider merging it. I'll let you know!

i3bar/src/cairo_util.c
```diff
@@ -67,5 +78,21 @@ color_t cairo_hex_to_color(const char *color) {
...
+ *
+ * TODO Remove this wrapper once all drawing is switched to cairo and instead use the same cairo
+ *      surface for drawing text as for other drawing calls.
+ *
+ */
+void cairo_draw_text(i3String *text, surface_t *surface, color_t fg_color, color_t bg_color, int x, int y, int max_width) {
+    set_font_colors(surface->gc, fg_color.colorpixel, bg_color.colorpixel);
+    draw_text(text, surface->id, surface->gc, visual_type, x, y, max_width);
+
+    cairo_surface_mark_dirty(surface->surface);
```
Airblader: @stapelberg See documentation for context. This is one of two possible solutions to the problem. The other one would be to reuse the same cairo surface already, for example by introducing a `set_font_surface` function which sets a global variable.

Given that this is temporary code anyway I think this is the better solution, but wanted to run all options by you. Just FYI, moving all drawing to cairo is what I intend to finish for 4.12 anyway.

Airblader:
Reminder for myself: test this entire thing with non-pango fonts.

i3bar/src/cairo_util.c
stapelberg: This solution is fine with me.

Airblader: We actually will need to keep this around because we support XCB text drawing anyway.

Airblader:
@stapelberg OK, I'm done. You can review it now. :)

PR #1985, opened 2015-10-10
Remove support for 32-bit visuals and RGBA colors.

### LINKED ISSUE #1984: Tray icons are always transparent
This didn't take long, sorry. :/ Even if not using any transparency, the tray icons will use transparency because they adapt to the 32-bit visual. I guess that's related to this part of the spec

> If this property is set to a visual with an alpha channel, the tray manager must use the Composite extension to composite the icon against the background using PictOpOver.

Since we don't use Composite yet, that's bad. I will try and see if I can fix this otherwise. 

(Side note, I didn't notice this myself because my wallpaper happens to be black there…)

i3 4.11

Airblader:
fixes #1984

### DIFF
--- a/i3bar/include/cairo_util.h
+++ b/i3bar/include/cairo_util.h
@@ -15,7 +15,6 @@ typedef struct color_t {
     double red;
     double green;
     double blue;
-    double alpha;
 
     /* For compatibility, we also store the colorpixel for now. */
     uint32_t colorpixel;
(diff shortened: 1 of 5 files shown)


Airblader:
@stapelberg There's something wrong with Travis. I get the same build error on i3-gaps. It's not related to the PR.

stapelberg:
I pushed a fix. Can you rebase? That should fix the travis build.

Airblader:
@stapelberg Great, thanks! Done & passing. :)

Issue #1983, opened 2015-10-08
Version 4.11 doesn't load keymap in VNC


tbnorth:
All good until the recent 4.11 upgrade, then this started happening in a VNC environment:

xkbcommon: ERROR: xkb_x11_keymap_new_from_device: illegal device ID: -1
10/07/2015 10:46:43 PM - ERROR: xkb_x11_keymap_new_from_device failed
i3: Could not load keymap
(i3 doesn't run)

compiled 4.10.4 from source and all working again (using 4.10.4)

tried a couple of different VNC servers / clients, didn't help.

Linux 3.13.0-65-generic #106-Ubuntu SMP Fri Oct 2 22:08:27 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
Ubuntu 14.04.3 LTS

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

tbnorth:
Binary i3 version:  4.11 (2015-09-30, branch "4.11") © 2009 Michael Stapelberg and contributors

http://logs.i3wm.org/logs/5761673931522048.bz2

stapelberg:
That implies `xkb_x11_get_core_keyboard_device_id()` returns `-1`, which is likely because VNC environments typically don’t have the XKB extension.

@bluetech Do you have a recommendation on how to handle this case? I.e., are you aware of other users of libxkbcommon which use a fallback?

bluetech:
Qt5 ran into the same problem, from googling I found these bug reports:
https://bugreports.qt.io/browse/QTBUG-32760
https://bugreports.qt.io/browse/QTBUG-44938

Their solution is to compile a keymap from the *local* host:
https://github.com/qtproject/qtbase/blob/924d4aefd5315c11d0637e2ba1a6b02246e89c44/src/plugins/platforms/xcb/qxcbkeyboard.cpp#L717-L734

[...]

PR #1990, opened 2015-10-11
Flush cairo surface after drawing text.

### LINKED ISSUE #1989: Cairo drawing breaks i3bar via VNC
Binary i3 version:  4.11-29-gb81435e (2015-10-11, branch "next") © 2009 Michael Stapelberg and contributors

Tested with [tightvncserver 1.3.9](https://packages.debian.org/stretch/tightvncserver).

```
Bisecting: 0 revisions left to test after this (roughly 0 steps)
[410c5da7cf0e4f4b635f22d770d24de64626981a] Use cairo for all drawing operations in i3bar.
```

With commit 410c5da7cf0e4f4b635f22d770d24de64626981a i3bar just comes up blank (aside from tray icons).

@Airblader Can you please take a look at this? Note that you need to have commit 321bba224a72966d25f5ba3f09d3ea946a4394a6 to get i3 to work with VNC.

Airblader:
This is necessary to avoid a bug where a cairo assertion fails because no snapshot is available.

fixes #1989

### DIFF
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -133,11 +133,6 @@ static void draw_text_pango(const char *text, size_t text_len,
     cairo_move_to(cr, x, y - yoffset);
     pango_cairo_show_layout(cr, layout);
 
-    /* We need to flush cairo surfaces twice to avoid an assertion bug. See #1989
-     * and https://bugs.freedesktop.org/show_bug.cgi?id=92455. */
-    cairo_surface_flush(surface);
-    cairo_surface_flush(surface);
-
     /* Free resources */
     g_object_unref(layout);
     cairo_destroy(cr);
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -133,6 +133,8 @@ static void draw_text_pango(const char *text, size_t text_len,
     cairo_move_to(cr, x, y - yoffset);
     pango_cairo_show_layout(cr, layout);
 
+    cairo_surface_flush(surface);
+
     /* Free resources */
     g_object_unref(layout);
     cairo_destroy(cr);
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -133,6 +133,9 @@ static void draw_text_pango(const char *text, size_t text_len,
     cairo_move_to(cr, x, y - yoffset);
     pango_cairo_show_layout(cr, layout);
 
+    /* We need to flush cairo surfaces twice to avoid an assertion bug. See #1989
+     * and https://bugs.freedesktop.org/show_bug.cgi?id=92455. */
+    cairo_surface_flush(surface);
     cairo_surface_flush(surface);
 
     /* Free resources */
--- a/i3bar/include/cairo_util.h
+++ b/i3bar/include/cairo_util.h
@@ -10,6 +10,14 @@
 
 #include <cairo/cairo-xcb.h>
 
+/* We need to flush cairo surfaces twice to avoid an assertion bug. See #1989
+ * and https://bugs.freedesktop.org/show_bug.cgi?id=92455. */
+#define CAIRO_SURFACE_FLUSH(surface)  \
+    do {                              \
+        cairo_surface_flush(surface); \
+        cairo_surface_flush(surface); \
+    } while (0)
+
 /* Represents a color split by color channel. */
 typedef struct color_t {
     double red;
--- a/i3bar/include/cairo_util.h
+++ b/i3bar/include/cairo_util.h
@@ -90,6 +90,13 @@ void cairo_draw_text(i3String *text, surface_t *surface, color_t fg_color, color
 void cairo_draw_rectangle(surface_t *surface, color_t color, double x, double y, double w, double h);
 
 /**
+ * Clears a surface with the given color.
+ * Note that the drawing is done using CAIRO_OPERATOR_SOURCE.
+ *
+ */
+void cairo_clear_surface(surface_t *surface, color_t color);
+
+/**
  * Copies a surface onto another surface.
  * Note that the drawing is done using CAIRO_OPERATOR_SOURCE.
  *
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -80,4 +80,4 @@ TAILQ_HEAD(statusline_head, status_block) statusline_head;
 #include "config.h"
 #include "libi3.h"
 #include "parse_json_header.h"
-#include "cairo_util.h"
+#include "draw_util.h"
--- a/i3bar/include/draw_util.h
+++ b/i3bar/include/draw_util.h
@@ -3,13 +3,16 @@
  *
  * © 2015 Ingo Bürk and contributors (see also: LICENSE)
  *
- * cairo_util.h: Utility for operations using cairo.
+ * draw.h: Utility for drawing.
  *
  */
 #pragma once
 
+#ifdef I3BAR_CAIRO
 #include <cairo/cairo-xcb.h>
+#endif
 
+#ifdef I3BAR_CAIRO
 /* We need to flush cairo surfaces twice to avoid an assertion bug. See #1989
  * and https://bugs.freedesktop.org/show_bug.cgi?id=92455. */
 #define CAIRO_SURFACE_FLUSH(surface)  \
@@ -17,6 +20,7 @@
         cairo_surface_flush(surface); \
         cairo_surface_flush(surface); \
     } while (0)
+#endif
 
 /* Represents a color split by color channel. */
 typedef struct color_t {
@@ -34,43 +38,40 @@ typedef struct surface_t {
     /* The drawable which is being represented. */
     xcb_drawable_t id;
 
-    // TODO remove this once i3 uses solely cairo for drawing operations
-    /* A classic XCB graphics context. This should not be used for
-     * drawing operations. */
+    /* A classic XCB graphics context. */
     xcb_gcontext_t gc;
 
+    int width;
+    int height;
+
+#ifdef I3BAR_CAIRO
     /* A cairo surface representing the drawable. */
     cairo_surface_t *surface;
 
     /* The cairo object representing the drawale. In general,
      * this is what one should use for any drawing operation. */
     cairo_t *cr;
+#endif
 } surface_t;
 
 /**
- * Initialize the cairo surface to represent the given drawable.
+ * Initialize the surface to represent the given drawable.
  *
  */
-void cairo_surface_init(surface_t *surface, xcb_drawable_t drawable, int width, int height);
+void draw_util_surface_init(surface_t *surface, xcb_drawable_t drawable, int width, int height);
 
 /**
  * Destroys the surface.
  *
  */
-void cairo_surface_free(surface_t *surface);
+void draw_util_surface_free(surface_t *surface);
 
 /**
  * Parses the given color in hex format to an internal color representation.
  * Note that the input must begin with a hash sign, e.g., "#3fbc59".
  *
  */
-color_t cairo_hex_to_color(const char *color);
-
-/**
- * Set the given color as the source color on the surface.
- *
- */
-void cairo_set_source_color(surface_t *surface, color_t color);
+color_t draw_util_hex_to_color(const char *color);
 
 /**
  * Draw the given text using libi3.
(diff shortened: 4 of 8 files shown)


Airblader:
@stapelberg Can you actually test this patch first, please? If it doesn't work, please comment in the issue how you're starting the VNC server. The reason I'm asking is that for me, i3bar simply crashes (contrary to just bein black).

stapelberg:
The PR makes text appear, but it’s rendered in a weird way (not as crisp as before, and apparently drawing over itself?):

With i3bar 4.11:
i3bar-vnc-4 11

With this PR:
i3bar-vnc-with_pr

Airblader:
@stapelberg This is because RENDER isn't enabled in tightvnc. You can easily verify this by using Xephyr and starting it with and without `-extension RENDER` once. The cairo xcb surface – which we have been using before my patch already – make heavy use. IMHO that makes sense, RENDER is a really common extension and afaik a de-facto standard to have. Other VNC servers support it, tightvnc apparently doesn't (at least I couldn't find a way to enable it right now).

Airblader:
Anyway, I'll go ask the cairo guys if there's anything we can do even if RENDER isn't available. But despite that I don't think it's worth supporting RENDER not being available given that we implicitly haven't been doing this in a long time and given that RENDER should imho just be made available.

Airblader:
To back this up a little, here's a post which itself is eight years old now:

https://bugs.freedesktop.org/show_bug.cgi?id=11529

> hat is, cairo (by design) tries to use the underlying graphics system where possible (win32 on windows, quartz on mac, etc). The problem here is that cairo is using an X server without modern rendering capabilities (XRender is 7 years old now). The poor performance is _primarily_ due to your X server, not cairo.
>
> Cairo was designed to match the semantics of (and thus take advantage of) the X Render extension. Compatibility with XRender-less X servers was implemented to ease the transition

psychon:
This patch will hide the assertion, but your code is still wrong. You have to flush the surface before you are drawing directly to the X11 drawable (so that cairo does all pending drawing that it didn't send to the X11 server before) and you mark it dirty after modifying it outside of cairo to tell cairo that any cached state that it may still have is invalid.

So if you mark dirty directly after flushing, the drawable/surface cannot be dirty. It could very well be that cairo just did a `PutImage` request and overwrote whatever you think you did before.

P.S.: http://lists.cairographics.org/archives/cairo/2015-October/026506.html is also me, just for reference

Airblader:
Thank you for looking into this here, I appreciate it! I'll try to change the code to see if it fixes things.

Airblader:
@psychon We only do three type of drawings in i3bar at the moment:

* Drawing text (using a separate cairo surface and object, potentially using no cairo at all)
* Drawing rectangles
* A simple paint

After each of the latter two, we flush the surface. This implies that at any given point that we draw text, the surface has been flushed anyway and should be good to go. In the text drawing function I have now moved the surface flushing to before drawing the text (although it shouldn't even be necessary there) and left marking it dirty after drawing the text. The behavior regarding the bug didn't change, however.

I then proceeded and remove all surface flushing, except for before drawing text (and marking it dirty after drawing text). In theory, as far as I understand, this should suffice as text drawing is the only way of bypassing cairo. I now get the assertion failure again. But the only place I mark it dirty is after drawing text, which can only happen after having flushed the surface before. How can this be?

psychon:
Just a random guess: Could you add a `cairo_surface_flush(src->surface);` to the end of `cairo_copy_surface()`? That function might do a `GetImage` request to download the image and attach the result as a snapshot to the source surface, so that it can avoid the `GetImage` later if needed. (IMO it would still be better to flush before drawing directly to the surface, but I can see that this is not quite trivial.) (And yes, I think it counts as a bug that cairo doesn't use `CopyArea` here.)

And since I can't find this in any bug report with the failed assertion, the assertion that the commit message refers to is line 1652 here, right? The comment before that assertion even says that this is due to a missing `cairo_surface_flush()`. ;-)
http://cgit.freedesktop.org/cairo/tree/src/cairo-surface.c#n1649

Airblader:
I've added that call, but it didn't help either.

> but I can see that this is not quite trivial.

It's perfectly trivial since we only draw to the surface directly when drawing text, which I've wrapped into a utility function that has access to the cairo surface and can flush it first (which is what I'm currently doing locally).

I've also pushed what I just tested to this PR now as we want these changes anyhow.

psychon:
I can reproduce this problem by running i3 under `xtrace --denyextensions`. However, first I run into https://github.com/i3/i3/issues/1983. So then I changed `i3bar_command` so that only i3bar gets all extensions disabled...

Anyway, the bug is in cairo, it exists ever since cairo-xcb exists (I think...) and is now reported at https://bugs.freedesktop.org/show_bug.cgi?id=92455. The work-around (which I tested with i3bar, but it's not worth providing a patch for this) is to call `cairo_surface_flush()` twice in a row every time it is used. This makes the second call detach the snapshot that the first one added.

Airblader:
Thank you so much for looking into this. Does calling it twice just solve the assertion failure or also the drawing problem (I'll try myself later, but can't at the moment)? If it fixes the drawing as well, I think we should use a wrapper function to call it twice as well.

Airblader:
@psychon I duplicated all calls to `cairo_surface_flush` that are in the code after this PR is applied, but I still get a black i3bar running under `xtrace --denyextensions` and the weird overlaying drawing running it in tightvnc.

psychon:
Oh, sorry. I only looked into the assertion failure and forgot about the rendering issue (or thought the missing flush fixes it?).
Sadly, this drawing problem cannot really be worked-around on your side:
http://cgit.freedesktop.org/cairo/commit/?id=d31dc73f9ba1facdbeee722c022d438f15e5f264
http://cgit.freedesktop.org/cairo/commit/?id=25a55822b4e22728834c96c8e73176d42bb406e3

This bug was introduced with cairo 1.12 (1.11.4 to be exact) and exists as long as these partial upload code exists:
http://cgit.freedesktop.org/cairo/commit/?id=fd613cb9f94daff0c8d4fdb27ff89894d41682a3

Airblader:
Thanks for the insights. Will the drawing problem only be a problem with missing RENDER or can it happen with RENDER as well?

psychon:
It can happen whenever a fallback occurs and SHM is not available. With the kind of drawing that you are doing, I guess that every version of RENDER should be enough (Even core rendering would be enough for this (except for the text rendering), but since this is barely ever the case, cairo doesn't implement this kind of code (yet?)).

So I think this can only happen without RENDER.

Oh and:
I would suggest the i3bar main loop print errors (events with type 0). I only learnt about cairo's bogous PutImage request via xtrace (it caused a BadLength error).

Airblader:
Great, thank you so much for your time on this. @stapelberg Given that we're coming back to there only being an issue without render, I will restate my original point about RENDER really being old enough that it should be available everywhere. So my proposal is to do the double-flush to get rid of the assertion failure and move forward.

Airblader:
> I would suggest the i3bar main loop print errors (events with type 0). I only learnt about cairo's bogous PutImage request via xtrace (it caused a BadLength error).

Yeah, that's a good idea. I'll open a ticket for this.

stapelberg:
Will the double-flush fix the issue with VNC?

Airblader:
It will fix the assertion error. The incorrect drawing is due to RENDER not being available (ultimately, technically it's due to incorrectly implemented fallback in cairo), @psychon said that given the basic drawing we do, any kind of RENDER version would be sufficient to not cause that.

So in summary we'd have incorrect drawing for people who don't have an X extension which is older than the better part of two decades now, but who at the same time run a current i3. That's a very unlikely (and not worth supporting) combination. Either way, @psychon already fixed the issue in cairo, so even if a user runs into this, they can update cairo. I don't think the combination of these things should be a showstopper for us.

stapelberg:
> So in summary we'd have incorrect drawing for people who don't have an X extension which is older than the better part of two decades now, but who at the same time run a current i3. That's a very unlikely (and not worth supporting) combination.

So, all people who use i3 with VNC. I don’t think it’s acceptable to break that use-case. Am I misunderstanding something here?

Airblader:
No, because there are plenty VNC implementations offering RENDER. Just tightvnc doesn't support it. And even then they'd just have to update cairo which I would assume will release at some point as well (though I don't know their release cycle right now).

Airblader:
One thing that still makes me wonder is how GTK and all the countless other apps using cairo work properly with VNC (if they do?). @psychon Any ideas on this?

stapelberg:
Did you test each VNC, NX, $insert_remote_solution_here implementation? :) The fact that it breaks in one of them is a good enough proxy for me that we’ll break actual users with this change.

I think we have two options here:

1. Revert the cairo i3bar rendering change.
2. Wait for cairo to release a new version and make i3 depend on that. @psychon, do you know when cairo plans to have a new release with the fix? If not, will it likely be within the next 6 months or later than that?

Airblader:
I'd really hate to see the changes be reverted. It was quite a bit of work (which, granted, isn't much of an argument) and I think it's the right direction to go. Running into a cairo bug is a bit unfortunate, however. First I think we should find out why GTK doesn't seem to break in VNC.

Either way, before we simply revert the entire thing, my "last straw" would be to adapt the implementation of `cairo_util.c` (maybe rename it) to have a build switch that switches between cairo and XCB-native drawing. We'd disable the flag in i3 for now, but I'd enable it in i3-gaps. That way we keep the changes (and work), can easily switch at any point in time and the separate code path would stay user-tested. Considering also that the build switch works on a function-local level with just a couple of lines, I think that'd be a reasonable way to go.

Depending on the fixed cairo version for the 4.12 release would be a much better approach, however.

psychon:
@stapelberg I don't know about any release plans and I would guess it takes more than 6 months. Sorry.
@Airblader I just tried reverting my cairo fix and started `xtrace -o /dev/null --denyextensions gvim`. The result is not pretty, so I guess the answer is "by not working without RENDER at all".

psychon:
Oh, sorry, I have a better answer: By using cairo-xlib instead of cairo-xcb which is an entirely different code path (and these two code paths resisted being merged for years already).

stapelberg:
Thanks, @psychon.

@Airblader Can you please make cairo-support hidden behind an off-by-default flag, then?

Airblader:
@psychon Thanks. And I guess you guys don't consider a bugfix release for this given that IMHO this is a pretty serious bug in the cairo-xcb backend?

@stapelberg Sure.

For what it's worth, I just built cairo from source and can confirm that @psychon's commits fix the issue in tightvnc.

Airblader:
@stapelberg I've said this before, but for the record – i3bar was broken without RENDER before the cairo move. Namely, when using a pango font which relies on cairo. I noticed that using markup to typeset text bold causes it to disappear. I guess it's different, though, because the user has the choice to not use pango markup (or pango at all), just saying that this isn't entirely new, just more present. ;)

I've mostly prepared the optional flag, I just need to test it a bit. I'd appreciate you testing it as well, though.

libi3/font.c
```diff
@@ -133,6 +133,11 @@ static void draw_text_pango(const char *text, size_t text_len,
     cairo_move_to(cr, x, y - yoffset);
     pango_cairo_show_layout(cr, layout);
 
+    /* We need to flush cairo surfaces twice to avoid an assertion bug. See #1989
+     * and https://bugs.freedesktop.org/show_bug.cgi?id=92455. */
+    cairo_surface_flush(surface);
```
Airblader: I'm not entirely sure this is even necessary, we didn't do it before either. @psychon Is it? Or does cairo implicitly flush when the surface/cairo object is destroyed?

psychon: Yup, when all state that it has gets freed, it can't delay pending work further. ;-) (It flushes in `cairo_surface_finish()` which is also called when the last reference to a surface is `cairo_surface_destroy()`'d).

Airblader: Thanks! I'll remove it again then.

Airblader:
Now the XCB path works fine locally and via VNC for me, the cairo code path works fine locally, but comes up black via VNC (or with denyextensions).

Airblader:
@stapelberg OK, ready for review. :)

psychon:
Sorry for reporting this here, but at least I report this at all. ;-)

Cairo 1.14.4 was released and includes the fix to cairo-xcb: http://cairographics.org/news/cairo-1.14.4/

    Fix _put_shm_image_boxes and _put_image_boxes when no SHM available with XCB.

Airblader:
@psychon Cool, thanks for the info!
@stapelberg So can we lift the dependency to cairo 1.14.4 for the 4.12 release then and switch to cairo per default (I'd still keep the XCB stuff in for a release, just in case)?

stapelberg:
@Airblader Yeah, I think that should be fine.

PR #1994, opened 2015-10-12
Make pango markup in mode names optional with a flag.

### LINKED ISSUE #1992: < in mode names does not work
From http://bugs.debian.org/801574:

> Dear Maintainer,
> In 4.10 this line was displayed when the mode was switched on:
>  set $mode_randr xrandr off (o), xrandr right (->), xrandr left (<)
> In 4.11 it does not show, I determined that it is the fault of '<' character. When you delete it, the line show again.

i3 v4.11

Airblader:
This introduces the flag "--pango" on the mode config directive to
explicitly enable pango markup for mode names. Not setting this will
cause the mode name to be rendered as is.
This fixes a regression in 4.11 where mode names containing characters
such as '<' would break user's configs as they didn't escape these
characters.

fixes #1992

### DIFF
--- a/i3bar/src/mode.c
+++ b/i3bar/src/mode.c
@@ -20,6 +20,8 @@
 struct mode_json_params {
     char *json;
     char *cur_key;
+    char *name;
+    bool pango_markup;
     mode *mode;
 };
 
@@ -31,17 +33,35 @@ static int mode_string_cb(void *params_, const unsigned char *val, size_t len) {
     struct mode_json_params *params = (struct mode_json_params *)params_;
 
     if (!strcmp(params->cur_key, "change")) {
-        /* Save the name */
-        params->mode->name = i3string_from_markup_with_length((const char *)val, len);
-        /* Save its rendered width */
-        params->mode->width = predict_text_width(params->mode->name);
+        char *copy = smalloc(sizeof(const unsigned char) * (len + 1));
+        strncpy(copy, (const char *)val, len);
+        copy[len] = '\0';
 
-        DLOG("Got mode change: %s\n", i3string_as_utf8(params->mode->name));
+        params->name = copy;
         FREE(params->cur_key);
-
         return 1;
     }
 
+    FREE(params->cur_key);
+    return 0;
+}
+
+/*
+ * Parse a boolean.
+ *
+ */
+static int mode_boolean_cb(void *params_, int val) {
+    struct mode_json_params *params = (struct mode_json_params *)params_;
+
+    if (strcmp(params->cur_key, "pango_markup") == 0) {
+        DLOG("Setting pango_markup to %d.\n", val);
+        params->pango_markup = val;
+
+        FREE(params->cur_key);
+        return 1;
+    }
+
+    FREE(params->cur_key);
     return 0;
 }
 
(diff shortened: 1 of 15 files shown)


Airblader:
@stapelberg This is quite a bit of code, but we need it to propagate it all the way through. If you want to change your mind and accept the minor breakage, I won't be sad about rejecting the PR. ;)

Either way, I've noticed that the userguide doesn't really document modes except for the resize mode example. We should change that, but in a separate PR, independently of this change.

Airblader:
By the way, the flag could also be implemented on the command, but I didn't really see what the point is in dynamically deciding whether or not to use it, so I put it on the config. I assume that's what the suggestion was anyway.

src/config_directives.c
```diff
@@ -129,6 +130,7 @@ CFGFUN(enter_mode, const char *modename) {
     DLOG("\t now in mode %s\n", modename);
     FREE(current_mode);
     current_mode = sstrdup(modename);
+    current_mode_is_markup = pango != NULL;
```
stapelberg: For consistency, please add parens around the expression, i.e. `current_mode_is_markup = (pango != NULL);`

src/bindings.c
```diff
@@ -437,7 +438,7 @@ void switch_mode(const char *new_mode) {
         grab_all_keys(conn);
 
         char *event_msg;
-        sasprintf(&event_msg, "{\"change\":\"%s\"}", mode->name);
+        sasprintf(&event_msg, "{\"change\":\"%s\",\"is_markup\":%s}", mode->name, mode->is_markup ? "true" : "false");
```
stapelberg: For consistency, please add parens around the ternary if, i.e. `(mode->is_markup ? "true" : "false")`

i3bar/src/mode.c
```diff
@@ -31,17 +33,35 @@ static int mode_string_cb(void *params_, const unsigned char *val, size_t len) {
...
-        /* Save the name */
-        params->mode->name = i3string_from_markup_with_length((const char *)val, len);
-        /* Save its rendered width */
-        params->mode->width = predict_text_width(params->mode->name);
+        char *copy = smalloc(sizeof(const unsigned char) * (len + 1));
+        strncpy(copy, (const char *)val, len);
+        copy[len] = '\0';
 
-        DLOG("Got mode change: %s\n", i3string_as_utf8(params->mode->name));
+        params->name = copy;
```
stapelberg: The following would be more in line with how we do string processing in the rest of the codebase, I think:

```c
sasprintf(&(params->name), "%.*s", len, val);
```

stapelberg:
It’s a bit unclean that in one half of the code, the flag is called `pango`, while in the rest, it’s called `is_markup`. Should we rename it to `pango_markup` in all the places, including the flag name itself?

i3bar/src/mode.c
Airblader: Not in the i3bar code. :) I agree that the other way is prettier, but let's do a separate PR. I'll create a tag for technical improvements and open an issue.

Airblader:
@stapelberg Done. I didn't rename the field in the pre-existing IPC communication as this potentially (though unlikely) can break things for users.

PR #1999, opened 2015-10-14
Revise workspace next/prev

### LINKED ISSUE #1798: "workspace next/prev" selectively skips some workspaces
i3 version 4.10.2 (2015-04-16, branch "4.10.2") © 2009-2014 Michael Stapelberg and contributors
Log:  http://logs.i3wm.org/logs/5729201025974272.bz2

I have two displays, primary is the laptop LCD, and secondary is VGA1. On primary, I have workspaces A S D F, and R. On secondary, I have workspaces 6, 7, 9, and G.

Hitting "workspace next/prev" cycles correctly between A S D F R and G, but these two commands completely pass over workspace 6, 7 and 9 on VGA1.

If however, I have a numeric workspace (usually 1-5) on the primary monitor, then the cycling works fine: It will cycle from the last numerically named workspace on the primary, into the first numbered workspace on the secondary.

It's only when there are no numbered workspaces on the primary, that the secondary monitor's numeric workspaces are skipped.

The reverse does not seem to be true: if I have only alphabetical workspaces on the secondary monitor, it will happily cycle correctly through all workspaces.

hwangcc23:
Fix #1798.

Tested w/ test cases of workspace and a created test case 528-workspace-next-prev.t (numbered workspaces and named workspaces on 2 outputs).

### DIFF
--- a/src/workspace.c
+++ b/src/workspace.c
@@ -506,12 +506,33 @@ void workspace_show_by_name(const char *num) {
  */
 Con *workspace_next(void) {
     Con *current = con_get_workspace(focused);
-    Con *next = NULL;
+    Con *next = NULL, *first = NULL, *first_opposite = NULL;
     Con *output;
 
     if (current->num == -1) {
         /* If currently a named workspace, find next named workspace. */
-        next = TAILQ_NEXT(current, nodes);
+        if ((next = TAILQ_NEXT(current, nodes)) != NULL)
+            return next;
+        bool found_current = false;
+        TAILQ_FOREACH(output, &(croot->nodes_head), nodes) {
+            /* Skip outputs starting with __, they are internal. */
+            if (con_is_internal(output))
+                continue;
+            NODES_FOREACH(output_get_content(output)) {
+                if (child->type != CT_WORKSPACE)
+                    continue;
+                if (!first)
+                    first = child;
+                if (!first_opposite && child->num != -1)
+                    first_opposite = child;
+                if (child == current) {
+                    found_current = true;
+                } else if (child->num == -1 && found_current) {
+                    next = child;
+                    return next;
+                }
+            }
+        }
     } else {
         /* If currently a numbered workspace, find next numbered workspace. */
         TAILQ_FOREACH(output, &(croot->nodes_head), nodes) {
@@ -521,6 +542,10 @@ Con *workspace_next(void) {
             NODES_FOREACH(output_get_content(output)) {
                 if (child->type != CT_WORKSPACE)
                     continue;
+                if (!first)
+                    first = child;
+                if (!first_opposite && child->num == -1)
+                    first_opposite = child;
                 if (child->num == -1)
                     break;
                 /* Need to check child against current and next because we are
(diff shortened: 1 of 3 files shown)


src/workspace.c
```diff
@@ -506,12 +506,34 @@ void workspace_show_by_name(const char *num) {
...
+                    if (child->type != CT_WORKSPACE)
+                        continue;
+                    if (!first)
+                        first = child;
+                    if (!remember && child->num != -1)
+                        remember = child;
+                    if (child == current) {
+                        found_current = 1;
+                    } else if (child->num == -1 && found_current) {
+                        next = child;
```
stapelberg: I’m having a hard time understanding this code branch. AFAICT, it iterates through all workspaces and returns the first named workspace after the current workspace. How is that different from what line 514 does, i.e. how can this ever be triggered?

testcases/t/528-workspace-next-prev.t
```diff
@@ -0,0 +1,185 @@
...
+# later in time and mess up our subsequent tests.
+sync_with_i3;
+is(focused_ws, 'E', 'workspace E focused');
+
+cmd 'workspace next';
+# We need to sync after changing focus to a different output to wait for the
+# EnterNotify to be processed, otherwise it will be processed at some point
+# later in time and mess up our subsequent tests.
+sync_with_i3;
+is(focused_ws, 'C', 'workspace C focused');
```
stapelberg: This is actually unexpected to me. Maybe I should have been clearer on the ticket.

I’d expect the focus order to be A, B, D, E, 1, 2, 3, 4, 5, C, F, A, B, …, which should be the same as if you clicked each workspace button in i3bar on each output in order.

hwangcc23: This will change the testcase 503.

> The current order should be:
output 1: 1, 5
output 2: 2

as-is: 1, 2, 5
to-be: 1, 5, 2

Is this expected to you?

src/workspace.c
hwangcc23: Line 525, get the first workspace (either numbered or named workspace) in first.
Line 527, get the first numbered workspace in remember.
Line 531, get the named workspace after current in next. 
The order is: Use next. If not found, use remember. If not found, use first.

But this is not expected to you. XD
I will rewrite the code.

testcases/t/528-workspace-next-prev.t
stapelberg: Oh, thanks for the hint about testcase 503. It actually makes sense to leave it as you have it now for consistency. (You can tell that I don’t actually use “workspace next” :)).

src/workspace.c
stapelberg: Thanks for the explanation. Maybe add a comment to `remember`, saying “remember holds the next numbered workspace (when |current| is a named workspace) or the next named workspace (when |current| is a numbered workspace)”. I also think the variable might need a more descriptive name, how about `next_opposite`?

src/workspace.c
```diff
@@ -506,12 +506,34 @@ void workspace_show_by_name(const char *num) {
...
+                    continue;
+                NODES_FOREACH(output_get_content(output)) {
+                    if (child->type != CT_WORKSPACE)
+                        continue;
+                    if (!first)
+                        first = child;
+                    if (!remember && child->num != -1)
+                        remember = child;
+                    if (child == current) {
+                        found_current = 1;
```
stapelberg: nit: `found_current = true` (this was wrong in the code before your pull request, but now is a good time to fix it)

src/workspace.c
```diff
@@ -506,12 +506,34 @@ void workspace_show_by_name(const char *num) {
  */
 Con *workspace_next(void) {
     Con *current = con_get_workspace(focused);
-    Con *next = NULL;
+    Con *next = NULL, *first = NULL, *remember = NULL;
     Con *output;
 
     if (current->num == -1) {
         /* If currently a named workspace, find next named workspace. */
         next = TAILQ_NEXT(current, nodes);
```
stapelberg: We can make this line `if ((next = TAILQ_NEXT(current, nodes)) != NULL) { return next; }`, right? That way, you can delete the next line and save one level of indention.

src/workspace.c
```diff
@@ -506,12 +506,34 @@ void workspace_show_by_name(const char *num) {
...
+                        continue;
+                    if (!first)
+                        first = child;
+                    if (!remember && child->num != -1)
+                        remember = child;
+                    if (child == current) {
+                        found_current = 1;
+                    } else if (child->num == -1 && found_current) {
+                        next = child;
+                        goto workspace_next_end;
```
stapelberg: Instead of the `goto`, why not just `return next;` here?

testcases/t/528-workspace-next-prev.t
```diff
@@ -0,0 +1,185 @@
...
+
+cmd 'workspace next';
+# We need to sync after changing focus to a different output to wait for the
+# EnterNotify to be processed, otherwise it will be processed at some point
+# later in time and mess up our subsequent tests.
+sync_with_i3;
+is(focused_ws, 'A', 'workspace A focused');
+
+cmd 'workspace next';
+# We need to sync after changing focus to a different output to wait for the
```
Airblader: Do we really need this comment copy-pasted a dozen times? Isn't it enough to write it once?

hwangcc23: Done. Thanks for the suggestion.  :)

hwangcc23:
Done

PR #2018, opened 2015-10-19
Allow multiple marks on a window

### LINKED ISSUE #2014: [enhancement] Allow for multiple marks on a single window
At the moment each window can only have one (unique) mark.  Marking an already marked window overwrites the old mark with the new one.  This is fine when you are using marks only to either manually tag a few windows or in one or two scripts/command chains. But if you want to make more extensive use of marks, for example mixing manual marks with scripts using them or having a larger number of scripts using them - things can go ugly. 

As only one mark can be placed on a window at a time, the chance that a previously set mark is inadvertently removed by another script/commands rises with the number of places where marks are used. This may break command combinations and could be especially annoying if manually set marks suddenly vanish.

I would like it to be possible to set more than one mark on a single window (similar to vim). The syntax of the currently available commands ``mark`` and ``unmark`` could remain the same for downward compatibility with older configurations. Although it would probably be useful if ``unmark`` were a bit more versatile than just remove one mark or remove all marks on all windows. For example by taking regex as argument and/or an option remove all marks on a specified window (focused or by criteria; e.g. ``unmark --window``). With active ``show_marks``, multiple marks could be shown as comma-separated list in the title bar.

Airblader:
fixes #2014 

Putting it on waiting for now because I still need to make `unmark` aware of matched windows.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1664,7 +1664,8 @@ con_mark::
         mark.
 con_id::
 	Compares the i3-internal container ID, which you can get via the IPC
-	interface. Handy for scripting.
+	interface. Handy for scripting. Use the special value +\_\_focused__+
+	to match only the currently focused window.
 
 The criteria +class+, +instance+, +role+, +title+, +workspace+ and +mark+ are
 actually regular expressions (PCRE). See +pcresyntax(3)+ or +perldoc perlre+ for
--- a/docs/userguide
+++ b/docs/userguide
@@ -2132,11 +2132,19 @@ unmark <identifier>
 ----------------------------------------------
 
 *Example (in a terminal)*:
-------------------------------
-$ i3-msg mark irssi
-$ i3-msg '[con_mark="irssi"] focus'
-$ i3-msg unmark irssi
-------------------------------
+---------------------------------------------------------
+# marks the focused container
+mark irssi
+
+# focus the container with the mark "irssi"
+'[con_mark="irssi"] focus'
+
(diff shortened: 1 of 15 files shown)


Airblader:
@stapelberg OK, you can review this now.

docs/userguide
```diff
@@ -1671,7 +1671,7 @@ workspace::
 	the special value +__focused__+ to match all windows in the currently
 	focused workspace.
 con_mark::
-	Compares the mark set for this container, see <<vim_like_marks>>.
+	Compares the marks set for this container, see <<vim_like_marks>>.
```
stapelberg: Please add that this matches when any of the container’s marks matches the specified mark.

docs/userguide
```diff
@@ -2133,21 +2133,33 @@ The additional +--toggle+ option will remove the mark if the window already has
 this mark, add it if the window has none or replace the current mark if it has
 another mark.
 
+By default, a window can only have one mark. You can use the +--add+ flag to
+put more than one mark on a window. Note that you may have to use this flag in
+combination with +--toggle+ as well.
```
stapelberg: The second sentence is not clear to me. Why do I have to use --toggle as well? Or do I have to use --add when I want to use --toggle? What are the circumstances? Can you add an example below?

stapelberg:
Looks good overall.

Airblader:
@stapelberg Done. I've also added a commit for `con_id=__focused__`.

Issue #2020, opened 2015-10-20
i3bar background color to indicate focused output


DavidMikeSimon:
I use three monitors, and it's common for me to momentarily lose track of which one has input focus, even with the window border and workspace name highlighted. I would find it useful if the background of the focused output's i3bar were also highlighted. Would a PR to add e.g. a `focused_background` parameter to i3bar's config be accepted?

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
Yep.

Airblader:
@DavidMikeSimon Sounds like you want to do the PR yourself. If so, please let us know. Otherwise I can get started on this as I want to do something else similar as well, in case Michael approves it.

@stapelberg What about background and border colors (similar to workspace buttons) for status blocks now? Some time ago, you wanted to try it out first with only urgent blocks which we've had for a while now. On top, I've had border and background (in fact more extensive than I'd do it in i3) in i3-gaps for a long time without problems now. Can we add this as well? In particular in combination with click events it makes sense to "style" certain blocks to look more like buttons.

stapelberg:
@Airblader Can you open a separate issue for that with specific use-cases?

Airblader:
@stapelberg Done.
@DavidMikeSimon Even with split tickets, please let me know wheter you intend to work on this yourself or if I can go ahead with a PR for this. :)

DavidMikeSimon:
Airblader, I'll give it a shot. This is my first i3 hack though, so I'll let you know if I get lost and want to hand over the torch. :-)

Airblader:
Sure, go ahead :) If you have any questions or pointers, feel free to ask before just giving up :)

DavidMikeSimon:
Okay, I'm noticing a few things that seem weird with the way statusline rendering is handled right now. Relevant code starts around here: https://github.com/i3/i3/blob/next/i3bar/src/xcb.c#L1888

There is a global pixmap buffer `statusline_surface` that the statusline is rendered to, separate from each bar's buffer. This looks like the optimization mentioned at https://github.com/i3/i3/pull/1657#issuecomment-94624477 which skips re-rendering the statusline for the n+1th output if the nth output's statusline had the same short_text setting.

[...]

PR #2030, opened 2015-10-23
Optionally change i3bar color on focused output, implements #2020

### LINKED ISSUE #1824: i3bar displays short text on all monitors if one monitor is too small for full text
I have three monitors, one being much smaller than the others. It appears that whenever the small screen doesn't have enough room to display the full text of my status command, all three screens display the short text of my status command, instead of just the one that is too small.

Expected behavior: the small screen displays the short text of my status commands, and the large screens display the full text of my status commands.

DavidMikeSimon:


i3bar/src/xcb.c
```diff
@@ -181,119 +186,143 @@ static void draw_separator(uint32_t x, struct status_block *block) {
...
-    /* Clear the statusline pixmap. */
-    draw_util_clear_surface(&statusline_surface, colors.bar_bg);
+    color_t bar_color = use_focus_colors ? colors.focus_bar_bg : colors.bar_bg;
+    draw_util_clear_surface(&output->statusline_buffer, bar_color);
 
-    /* Draw the text of each block. */
     uint32_t x = 0;
+    if (width > max_width) {
+        // Use unsigned integer wraparound to clip off the left side.
+        // I have been assured by FreeNode#xcb that this is safe.
```
Airblader: If it's not safe, we wouldn't merge it. I don't think this comment belongs here. Also, we typically write comments as "We", not "I".

i3bar/src/xcb.c
```diff
@@ -1147,6 +1193,7 @@ char *init_xcb_early() {
     colormap = root_screen->default_colormap;
     visual_type = get_visualtype(root_screen);
 
+    /* The various watchers to communicate with xcb */
```
Airblader: Why is this comment here now?

i3bar/src/xcb.c
```diff
@@ -1819,8 +1863,20 @@ void draw_bars(bool unhide) {
             reconfig_windows(false);
         }
 
+        bool output_needs_focus_colors = false;
+        if (focus_colors_differ) {
```
Airblader: Do we really need this flag? It seems rather unnecessary to me. Is it just for avoiding the loop over the outputs? I think that wouldn't be worth adding this flag. The colors automatically fall back anyway.

docs/ipc
```diff
@@ -520,6 +520,14 @@ statusline::
 	Text color to be used for the statusline.
 separator::
 	Text color to be used for the separator.
+focused_background::
+	Background color of the bar on the currently focused monitor output.
```
Airblader: I think we should make it clear that these are all optional and will simpl fall back to their non-focus version.

i3bar/src/xcb.c
```diff
@@ -181,119 +186,143 @@ static void draw_separator(uint32_t x, struct status_block *block) {
...
-        /* Try to use the shorter text if necessary and possible. */
-        if (use_short_text && block->short_text != NULL) {
-            I3STRING_FREE(block->full_text);
-            block->full_text = i3string_copy(block->short_text);
-        }
-
-        if (i3string_get_num_bytes(block->full_text) == 0)
-            continue;
+    bool fit = false;
+    while (!fit) {
```
Airblader: I don't really like this construct. This loop will execute either exactly once or exactly twice, right? Can we not extract this into a function with some `use_short_text` argument? It would also save us a level of indentation on a rather huge block.

i3bar/src/xcb.c
```diff
@@ -1819,8 +1863,20 @@ void draw_bars(bool unhide) {
             reconfig_windows(false);
         }
 
+        bool output_needs_focus_colors = false;
+        if (focus_colors_differ) {
+            i3_ws *ws_walk;
+            TAILQ_FOREACH(ws_walk, outputs_walk->workspaces, tailq) {
```
Airblader: I think it'd be good to move this logic into a function in `i3bar/src/outputs.c`.

Airblader:
Please make sure to format your code properly using `clang-format`. There's also some build errors you need to look at.

DavidMikeSimon:
Hold off on merging, one last change I want to make.

i3bar/src/xcb.c
```diff
@@ -232,68 +222,88 @@ void refresh_statusline(bool use_short_text) {
...
+    struct status_block *block;
+
+    color_t bar_color = use_focus_colors ? colors.focus_bar_bg : colors.bar_bg;
+    draw_util_clear_surface(&output->statusline_buffer, bar_color);
 
-    /* Draw the text of each block. */
-    uint32_t x = 0;
+    // Use unsigned integer wraparound to clip off the left side.
+    // XCB will ignore any draws that end up outside the pixmap.
+    uint32_t x = 0 - clip_left;
```
Airblader: Can you please make sure that this also works as expected when i3 is compiled with `-DI3BAR_CAIRO=1`? Just because XCB ignored it doesn't mean cairo will. In general I dislike relying on such things and prefer just solving it cleanly.

i3bar/src/xcb.c
```diff
@@ -1819,8 +1824,18 @@ void draw_bars(bool unhide) {
             reconfig_windows(false);
         }
 
+        bool use_focus_colors = false;
+        i3_ws *ws_walk;
+        TAILQ_FOREACH(ws_walk, outputs_walk->workspaces, tailq) {
+            if (ws_walk->focused) {
```
Airblader: Since my earlier comment isn't showing anymore: I'd prefer moving this into a separate function in `i3bar/src/outputs.c` as `bool outputs_is_focused(i3_output *output)`.

DavidMikeSimon: I was putting the finishing touches on that just as you wrote this. :-)

DavidMikeSimon:
Seems to work fine with cairo mode enabled. Anything else I should look into?

DavidMikeSimon:
Airblader, does everything seem okay now?

Airblader:
The final review is done by Michael and he's also the one doing the merges.

stapelberg:
@DavidMikeSimon, can you please make two meta-changes before I do the actual review?

1. Please make the commit description speak for itself, i.e. not just “fix #1824”, but rather describe what is getting fixed and put the “fixes #1824” into the description.

2. Please submit one PR per change, not a single commit which does two things (introduces colors and changes rendering). We want changes to be atomic so that we can easier pinpoint when problems were introduced/revert changes.

Thanks.

DavidMikeSimon:
Okay, I'll split this into two PRs.

DavidMikeSimon:
@stapelberg I have split the rendering change and bug fix into PR #2039. If that's merged, I'll rebase this one onto it.

DavidMikeSimon:
Okay, rebased.

DavidMikeSimon:
Whoops, the commit description had the wrong issue number, fixed now.

DavidMikeSimon:
@stapelberg Okay, I've made the changes you suggested.

PR #2039, opened 2015-10-26
Use separate buffers for i3bar statusline for each workspace, track short and long renders separately, fixes #1824

### LINKED ISSUE #1824: i3bar displays short text on all monitors if one monitor is too small for full text

DavidMikeSimon:
In i3bar currently, when the statusline is rendered in short text mode, this destroys the local copy of the long-text for each workspace block, as well as the information about the rendered width of the block in long-text mode. That causes the bug described in issue #1824, and also a related problem with sending click events to the correct statusline block.

The changes in this PR correct those problems by keeping both versions of the text and tracking separately for each block how it renders in short and long mode. Also, the function for predicting the length of the statusline is factored out.

Additionally, the global statusline buffer surface is removed, instead i3 creates a separate statusline buffer for each output. This means that for the situation described in https://github.com/i3/i3/pull/1657#issuecomment-94624477, i3bar now uses option 1 instead of option 2.  This simplifies the code by removing the need to reallocate the global buffer as output sizes change dynamically, and also will make it easier for me to implement (in another PR) the feature in #2020 to change the bar's colors on the focused output.

### DIFF
--- a/i3bar/src/outputs.c
+++ b/i3bar/src/outputs.c
@@ -144,10 +144,16 @@ static int outputs_start_map_cb(void *params_) {
     if (params->cur_key == NULL) {
         new_output = smalloc(sizeof(i3_output));
         new_output->name = NULL;
+        new_output->active = false;
+        new_output->primary = false;
+        new_output->visible = false;
         new_output->ws = 0,
+        new_output->statusline_width = 0;
+        new_output->statusline_short_text = false;
         memset(&new_output->rect, 0, sizeof(rect));
         memset(&new_output->bar, 0, sizeof(surface_t));
         memset(&new_output->buffer, 0, sizeof(surface_t));
+        memset(&new_output->statusline_buffer, 0, sizeof(surface_t));
 
         new_output->workspaces = smalloc(sizeof(struct ws_head));
         TAILQ_INIT(new_output->workspaces);
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -74,11 +74,6 @@ int bar_height;
 int xkb_base;
 int mod_pressed = 0;
 
-/* Because the statusline is the same on all outputs, we have
- * global buffer to render it on */
-surface_t statusline_surface;
-uint32_t statusline_width;
-
 /* Event watchers, to interact with the user */
 ev_prepare *xcb_prep;
 ev_check *xcb_chk;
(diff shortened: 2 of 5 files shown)


i3bar/src/outputs.c
```diff
@@ -145,9 +145,11 @@ static int outputs_start_map_cb(void *params_) {
         new_output = smalloc(sizeof(i3_output));
         new_output->name = NULL;
         new_output->ws = 0,
+        new_output->statusline_width = 0;
```
stapelberg: I think you’re missing `new_output->statusline_short_text = false;` here?

i3bar/src/xcb.c
```diff
@@ -181,119 +176,134 @@ static void draw_separator(uint32_t x, struct status_block *block) {
...
+ * Redraws the statusline to the output's statusline_buffer
+ */
+void draw_statusline(i3_output *output, uint32_t clip_left, bool use_short_text) {
+    struct status_block *block;
+
+    draw_util_clear_surface(&output->statusline_buffer, colors.bar_bg);
 
-    /* Clear the statusline pixmap. */
-    draw_util_clear_surface(&statusline_surface, colors.bar_bg);
+    // Use unsigned integer wraparound to clip off the left side.
```
stapelberg: nit: use `/* */` for comments please. Also, can you add an example of how unsigned integer wraparound will behave to the comment for clarity please?

stapelberg:
Looks good overall.

i3bar/src/outputs.c
DavidMikeSimon: Well, I saw that the other booleans in i3_output (like `active` and `primary`) weren't explicitly set either, so I figured I should just follow that pattern. Maybe I should just add all of those in?

stapelberg: Yes. If they’re not in an scalloc'ed or memset block, they should be initialized.

DavidMikeSimon: They are smalloced.

DavidMikeSimon:
@stapelberg Okay, updated with those changes

PR #2040, opened 2015-10-26
Fix crash when trying to split and float a dock container.

### LINKED ISSUE #2034: Crash with for_window
I wanted to do the following layout: turning new windows/clients into floating *containers* with stacked layout.

For this, I tried this:

```
for_window [class=".*"] focus parent, split horizontal, layout stacked, floating enable
```

But this crashes i3.

I've found this crashes i3 too:

```
for_window [class=".*"] floating enable, split horizontal, layout stacked
```

And also this:

```
for_window [class=".*"] split horizontal, layout stacked, floating enable
```

Don't know if I'm really doing "stupid" things here.

i3 version 4.11 (2015-09-30, branch "4.11")

Airblader:
fixes #2034

### DIFF
--- a/include/con.h
+++ b/include/con.h
@@ -113,6 +113,12 @@ bool con_is_internal(Con *con);
 bool con_is_floating(Con *con);
 
 /**
+ * Returns true if the container is a docked container.
+ *
+ */
+bool con_is_docked(Con *con);
+
+/**
  * Checks if the given container is either floating or inside some floating
  * container. It returns the FLOATING_CON container.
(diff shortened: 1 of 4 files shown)


Airblader:
Actually hold on, this patch is incorrect.

Airblader:
This can be reviewed now, but I asked the reporter to test this patch to see if it fixes the issue on their machine.

Airblader:
OK, it was said in the issue that this patch works.

PR #2043, opened 2015-10-28
Fix multiple memory leaks with regular expressions.

Airblader:


Airblader:
I found these while valgrinding the other PR's issue.

PR #2053, opened 2015-11-02
Raise dependency to cairo 1.14.4.

### LINKED ISSUE #2048: [enhancement] Move to cairo per default
See https://github.com/i3/i3/pull/1990#issuecomment-152448756

* Increase the dependency to cairo to 1.14.4.
* Add the `I3BAR_CAIRO` compile flag to use cairo per default.
* We keep the XCB rendering code and the compile switch for now.

Airblader:
With the recent cairo bugfix, we can now switch to using cairo for
rendering i3bar per default by raising the minimum version of cairo
to 1.14.4.

resolves #2048

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -141,6 +141,9 @@ LIBSN_LIBS   := $(call ldflags_for_lib, libstartup-notification-1.0,startup-noti
 PANGO_CFLAGS := $(call cflags_for_lib, cairo)
 PANGO_CFLAGS += $(call cflags_for_lib, pangocairo)
 I3_CPPFLAGS  += -DPANGO_SUPPORT=1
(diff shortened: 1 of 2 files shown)


common.mk
```diff
@@ -141,6 +141,7 @@ LIBSN_LIBS   := $(call ldflags_for_lib, libstartup-notification-1.0,startup-noti
 PANGO_CFLAGS := $(call cflags_for_lib, cairo)
 PANGO_CFLAGS += $(call cflags_for_lib, pangocairo)
 I3_CPPFLAGS  += -DPANGO_SUPPORT=1
+I3_CPPFLAGS  += -DI3BAR_CAIRO=1
```
stapelberg: Would it make sense to gate this on cairo being ≥ 1.14.4? See line 131 for an example.

Airblader: @stapelberg  Yeah, that makes perfect sense. I've updated it.

PR #2061, opened 2015-11-11
Only grab scrollwheel buttons if necessary

### LINKED ISSUE #2049: Window gets focus when scrolling
Hi,

I've updated i3 recently (to version 4.11) and now, when I try to scroll in a window that does not have focus, that window gets focused. Is it possible to scroll in unfocused windows and let them remain unfocused?

Airblader:
fixes #2049

### DIFF
--- a/src/xcb.c
+++ b/src/xcb.c
@@ -319,3 +319,14 @@ release_grab:
     FREE(reply);
     xcb_ungrab_server(conn);
 }
+
+/*
+ * Grab the specified buttons on a window when managing it.
+ *
+ */
+void xcb_grab_buttons(xcb_connection_t *conn, xcb_window_t window, uint8_t* buttons) {
+    for (int i = 0; i < sizeof(buttons) / sizeof(uint8_t); i++) {
+        xcb_grab_button(conn, false, window, XCB_EVENT_MASK_BUTTON_PRESS, XCB_GRAB_MODE_SYNC,
+                        XCB_GRAB_MODE_ASYNC, root, XCB_NONE, buttons[i], XCB_BUTTON_MASK_ANY);
+    }
+}
--- a/src/xcb.c
+++ b/src/xcb.c
@@ -324,8 +324,19 @@ release_grab:
  * Grab the specified buttons on a window when managing it.
  *
  */
-void xcb_grab_buttons(xcb_connection_t *conn, xcb_window_t window, uint8_t* buttons) {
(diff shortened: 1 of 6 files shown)


src/xcb.c
```diff
@@ -319,3 +319,44 @@ void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_a
...
+    Binding *bind;
+    TAILQ_FOREACH(bind, bindings, bindings) {
+        /* We are only interested in whole window mouse bindings. */
+        if (bind->input_type != B_MOUSE || !bind->whole_window)
+            continue;
+
+        char *endptr;
+        long button = strtol(bind->symbol + (sizeof("button") - 1), &endptr, 10);
+        if (button == LONG_MAX || button == LONG_MIN || button < 0 || *endptr != '\0' || endptr == bind->symbol) {
+            ELOG("Could not parse button number, skipping this binding. This is a bug.\n");
```
stapelberg: s/This is a bug/Please report this bug in i3/

src/xcb.c
```diff
@@ -319,3 +319,44 @@ void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_a
...
+            continue;
+        }
+
+        /* If the binding is for either scrollwheel button, we need to grab everything. */
+        if (button == 4 || button == 5)
+            buttons[num++] = XCB_BUTTON_INDEX_ANY;
```
stapelberg: Add a `break;` here so that the loop is terminated early.

src/xcb.c
```diff
@@ -319,3 +319,44 @@ void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_a
...
+ *
+ */
+void xcb_grab_buttons(xcb_connection_t *conn, xcb_window_t window) {
+    /* We need to figure out whether we need to grab all buttons or just 1-3.
+     * This allows binding with --whole-window to buttons 4/5, but does not
+     * grab these buttons unless such a binding exists. See #2049. */
+    uint8_t buttons[3];
+    int num = 0;
+    Binding *bind;
+    TAILQ_FOREACH(bind, bindings, bindings) {
```
stapelberg: Can you factor this loop into a separate function so that we can avoid going through all bindings over and over again when re-grabbing all the buttons?

Airblader:
@stapelberg Done.

PR #2065, opened 2015-11-15
Migrate i3 rendering to cairo

Airblader:
This patch series migrates i3 rendering of decorations to cairo. It consists of multiple commits which are logically divided and also make the review easier. I definitely recommend reviewing patch by patch here, in particular the main patch (commit #3) is quite big either way, but the others are smaller. I will also annotate some parts with explanatory comments.

I will also make test runs for

* i3 before this patch
* i3 compiled with cairo support
* i3 compiled without cairo support

and provide screenshots as well as "diff images".


Airblader:
For this test, I've set up a [script](http://pastebin.com/azpdgbEQ) and executed it for all three cases described above. Running them through [Resemble.js](https://huddle.github.io/Resemble.js/) yields "These images are the same" in all combinations.

i3 test3 old
i3 test3 new-cairo
i3 test3 new-xcb

libi3/draw_util.c
```diff
@@ -183,7 +217,7 @@ void draw_util_copy_surface(surface_t *src, surface_t *dest, double src_x, doubl
      * onto the surface rather than blending it. This is a bit more efficient and
      * allows better color control for the user when using opacity. */
     cairo_set_operator(dest->cr, CAIRO_OPERATOR_SOURCE);
-    cairo_set_source_surface(dest->cr, src->surface, dest_x - src_x, src_y);
+    cairo_set_source_surface(dest->cr, src->surface, dest_x - src_x, dest_y - src_y);
```
Airblader: This was actually a bug introduced during the i3bar to cairo move. Figuring this out took me longer than it should have… I even thought Xorg isn't sending Expose events correctly at one point. Sigh.

src/con.c
```diff
@@ -47,7 +47,7 @@ Con *con_new_skeleton(Con *parent, i3Window *window) {
         new->depth = window->depth;
         new->window->aspect_ratio = 0.0;
     } else {
-        new->depth = XCB_COPY_FROM_PARENT;
+        new->depth = root_depth;
```
Airblader: Since the root window is typically 24-bit, using `XCB_COPY_FROM_PARENT` would give us the wrong depth for container windows in v-split containers (and others), but that's exactly where we need 32-bit to fix the transparency issue. `root_depth` will be 32-bit if available and 24-bit otherwise (typically).

src/restore_layout.c
```diff
@@ -125,7 +125,7 @@ void restore_connect(void) {
 
 static void update_placeholder_contents(placeholder_state *state) {
     xcb_change_gc(restore_conn, state->gc, XCB_GC_FOREGROUND,
```
Airblader: We might want to migrate this at some point as well.

src/x.c
```diff
@@ -478,31 +503,19 @@ void x_draw_decoration(Con *con) {
     if (p->border_style != BS_NORMAL)
         goto copy_pixmaps;
 
+    /* If the parent hasn't been set up yet, skip the decoration rendering
+     * for now. */
+    if (parent->frame_buffer.id == XCB_NONE)
```
Airblader: This caused frequent XCB errors previously.

src/x.c
```diff
@@ -510,16 +523,18 @@ void x_draw_decoration(Con *con) {
...
+        sasprintf(&_title, "i3: %s", tree);
         free(tree);
 
-        draw_text_ascii(title,
-                        parent->pixmap, parent->pm_gc,
-                        con->deco_rect.x + 2, con->deco_rect.y + text_offset_y,
-                        con->deco_rect.width - 2);
-        free(title);
+        i3String *title = i3string_from_utf8(_title);
+        draw_util_text(title, &(parent->frame_buffer),
```
Airblader: As discussed in the issue, I can introduce another wrapper in `draw_util` that uses `draw_text_ascii`, but frankly I don't quite see the point (in this instance) as that one i3String isn't going to hurt in any way. Let me know if you want me to do that, though.

src/x.c
```diff
@@ -755,38 +754,40 @@ void x_push_node(Con *con) {
...
             }
 
             uint16_t win_depth = root_depth;
             if (con->window)
                 win_depth = con->window->depth;
 
-            xcb_create_pixmap(conn, win_depth, con->pixmap, con->frame, rect.width, rect.height);
+            /* Ensure we have valid dimensions for our surface. */
+            // TODO This is probably a bug in the condition above as we should never enter this path
+            //      for height == 0. Also, we should probably handle width == 0 the same way.
```
Airblader: The way I understand it, and the way one comment in this area describes it, we shouldn't even enter this code path if the pixmap we'd be creating has zero height. However, `is_pixmap_needed` overwrites that check and I don't really see why – creating such a pixmap is going to fail anyway. I also don't think zero-width should be treated any different.

If you agree, I'd submit a separate PR to fix this (and remove those `MAX` calls) after merging this.

src/x.c
```diff
@@ -755,38 +754,40 @@ void x_push_node(Con *con) {
...
+            xcb_create_pixmap_checked(conn, win_depth, con->frame_buffer.id, con->frame.id, width, height);
+            draw_util_surface_init(conn, &(con->frame_buffer), con->frame_buffer.id,
+                                   get_visualtype_by_id(get_visualid_by_depth(win_depth)), width, height);
+            draw_util_surface_set_size(&(con->frame), width, height);
 
             con->pixmap_recreated = true;
 
             /* Don’t render the decoration for windows inside a stack which are
              * not visible right now */
+            // TODO Should this work the same way for L_TABBED?
```
Airblader: Not sure why `L_TABBED` should be any different here, but I've never tried this out to see what happens.

src/x.c
```diff
@@ -755,38 +754,40 @@ void x_push_node(Con *con) {
...
+            int width = MAX(rect.width, 1);
+            int height = MAX(rect.height, 1);
 
-            /* For the graphics context, we disable GraphicsExposure events.
-             * Those will be sent when a CopyArea request cannot be fulfilled
-             * properly due to parts of the source being unmapped or otherwise
-             * unavailable. Since we always copy from pixmaps to windows, this
-             * is not a concern for us. */
-            uint32_t values[] = {0};
-            xcb_create_gc(conn, con->pm_gc, con->pixmap, XCB_GC_GRAPHICS_EXPOSURES, values);
```
Airblader: Granted, we'll be creating the GC slightly different now, but my understanding is that we only set that flag to spare us some events we don't need. I don't think it's really worth complicating the drawing API to replicate this as it's really micro-optimization.

Airblader:
@stapelberg I've filled in all the information. Let the reviewing game begin :)

include/xcb.h
```diff
@@ -141,6 +141,12 @@ void xcb_set_root_cursor(int cursor);
 uint16_t get_visual_depth(xcb_visualid_t visual_id);
 
 /**
+ * Get visual typr specified by visualid
```
stapelberg: typo: s/typr/type/

src/xcb.c
```diff
@@ -254,6 +254,27 @@ uint16_t get_visual_depth(xcb_visualid_t visual_id) {
 }
 
 /*
+ * Get visual typr specified by visualid
```
stapelberg: typo: s/typr/type/

include/libi3.h
```diff
@@ -480,3 +483,102 @@ char *get_config_path(const char *override_configpath, bool use_system_paths);
...
+    } while (0)
+#endif
+
+/* Represents a color split by color channel. */
+typedef struct color_t {
+    double red;
+    double green;
+    double blue;
+
+    /* For compatibility, we also store the colorpixel for now. */
```
stapelberg: Rather than saying “for now”, can you put the criterion here for when `colorpixel` can be removed?

include/libi3.h
```diff
@@ -480,3 +483,102 @@ char *get_config_path(const char *override_configpath, bool use_system_paths);
...
+    xcb_visualtype_t *visual_type;
+
+    int width;
+    int height;
+
+#ifdef CAIRO_SUPPORT
+    /* A cairo surface representing the drawable. */
+    cairo_surface_t *surface;
+
+    /* The cairo object representing the drawale. In general,
```
stapelberg: typo: s/drawale/drawable/

libi3/draw_util.c
```diff
@@ -15,31 +15,42 @@
...
-xcb_connection_t *xcb_connection;
+/* The default visual_type to use if none is specified when creating the surface. Must be defined globally. */
 xcb_visualtype_t *visual_type;
 
 /* Forward declarations */
-static void draw_util_set_source_color(surface_t *surface, color_t color);
+static void draw_util_set_source_color(xcb_connection_t *conn, surface_t *surface, color_t color);
+
+#define CHECK_SURFACE(surface)                                               \
+    if (surface->id == XCB_NONE) {                                           \
```
stapelberg: Use `(surface)`, see https://en.wikipedia.org/wiki/C_preprocessor#Macro_definition_and_expansion (paragraph “The second x is enclosed in its own pair of parentheses…”)

libi3/draw_util.c
```diff
@@ -15,31 +15,42 @@
...
 
-xcb_connection_t *xcb_connection;
+/* The default visual_type to use if none is specified when creating the surface. Must be defined globally. */
 xcb_visualtype_t *visual_type;
 
 /* Forward declarations */
-static void draw_util_set_source_color(surface_t *surface, color_t color);
+static void draw_util_set_source_color(xcb_connection_t *conn, surface_t *surface, color_t color);
+
+#define CHECK_SURFACE(surface)                                               \
```
stapelberg: Please use `do { } while (0)`, see https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Multi-statement_Macro

libi3/draw_util.c
```diff
@@ -15,31 +15,42 @@
...
 
 /*
  * Initialize the surface to represent the given drawable.
  *
  */
-void draw_util_surface_init(surface_t *surface, xcb_drawable_t drawable, int width, int height) {
+void draw_util_surface_init(xcb_connection_t *conn, surface_t *surface, xcb_drawable_t drawable,
+                            xcb_visualtype_t *visual, int width, int height) {
     surface->id = drawable;
+    surface->visual_type = (visual == NULL) ? visual_type : visual;
```
stapelberg: nit: please add parens around the ternary expression

src/x.c
stapelberg: I don’t remember off the top of my head. It might be good to look at the git history/git blame annotations for this file. If you want to give it a shot, feel free to send another PR to fix it. In case you break something, at least we can better document why this is necessary.

src/x.c
stapelberg: I’m not really convinced that removing this optimization is a good thing — we’ve been using for the last 4 years, and i3 prides itself on being rather fast, see goal 2 on our website (about using XCB because it’s fast).

Can you outline why it’s not possible to keep the optimization?

src/x.c
Airblader: It seems to come from a commit by you back in 2011 (937a80511a887f88e1ae3888ba533283ad146a18) to fix some X11 errors. I think it's worth a shot. The pixmap won't be created in the current behavior either anyway.

src/x.c
Airblader: It's definitely *possible*, I just don't think it's worth it. Some reasons:

* It blows up the currently rather small and contained `draw_util` for something that I strongly doubt has any noticable impact on performance.
* It somewhat makes the API dirty as well as it's a pure XCB detail while `draw_util` is abstracting XCB away.
* It will only matter if compiled without cairo support (since cairo doesn't use it), which isn't even the default.
* It will only matter for CopyArea requests which are the rarest and, except for during resizes, happen with very low frequency.
* It will only matter for CopyArea requests that *fail* for the reason stated in the comment, making it even more rare.

That said, there's no need to jump through hoops there. If those arguments don't convince you, that's fair enough for me and I'll just make the change.

src/xcb.c
Airblader: Way too many typos, though most of them have been there before this patch. I'll fix them, obviously. :)

include/libi3.h
Airblader: Really, we could do it anytime, because we can always call `get_colorpixel` when needed rather than caching it. At the very least we should keep it as long as we have the XCB rendering in here, but we might even want to keep it after that since we'll always need colorpixels one way or another.

I'll just remove the comment / rephrase it.

Airblader:
I think some more work on the transparency commit is needed. We need to synchronize resizing the frame and drawing to it correctly and should probably also clear it instead of "relying" on the pixmap being recreated.

Right now, the transparency generally works, but resizing can lead to pixmaps being visually corrupted. Still an improvement and not a showstopper since we don't officially support transparency anyway.

Airblader:
I've pushed another commit that fixes a crash I missed before and I'll address all review comments later. I'm working on getting the transparency fixed properly (without tearing) now and I *think* I've finally found the proper way to do it and will try to make it a clean patch now (though I'll amend it to the transparency commit).

One odd finding: you know how we don't draw to the buffer pixmap unless something changed? Turning that check off makes resizing a whole lot smoother and more flicker-free for me (but is obviously not an option). Maybe another thing worth investigating more on a rainy day.

src/x.c
Airblader: I think the reason only height has been considered because the major usecase is empty dockareas, where we only support horizontal ones anyway.

Airblader:
@stapelberg All done.

Airblader:
@stapelberg Please feel free to review, but hold off with merging for now.

Airblader:
The transparency fix is still not complete, unfortunately. :( I'll take another look at it. Damn you, recursion.

Airblader:
@stapelberg Let's proceed with this PR as-is. It does have some tearing when using transparency and titlebars, but that's still a *big* improvement over before where the entire background would be garbage. So one, I doubt anyone uses both transparency and titlebars, and two, if they do, we still don't support transparency.

I think actually and completely fixing transparency tearing is going to require more of a change and I'd like to work on that in a separate PR.

Another advantage is that pushing this out to the next branch would allow quicker feedback regarding any issues.

libi3/draw_util.c
```diff
@@ -15,31 +15,44 @@
...
 
-xcb_connection_t *xcb_connection;
+/* The default visual_type to use if none is specified when creating the surface. Must be defined globally. */
 xcb_visualtype_t *visual_type;
 
 /* Forward declarations */
-static void draw_util_set_source_color(surface_t *surface, color_t color);
+static void draw_util_set_source_color(xcb_connection_t *conn, surface_t *surface, color_t color);
+
+#define CHECK_SURFACE(surface)                                                   \
```
stapelberg: Can you rename `CHECK_SURFACE` to something that conveys how this macro actually contains a control flow instruction (`return`)? How about `RETURN_UNLESS_SURFACE_INITIALIZED`?

libi3/draw_util.c
```diff
@@ -48,11 +61,32 @@ void draw_util_surface_init(surface_t *surface, xcb_drawable_t drawable, int wid
  * Destroys the surface.
  *
  */
-void draw_util_surface_free(surface_t *surface) {
-    xcb_free_gc(xcb_connection, surface->gc);
+void draw_util_surface_free(xcb_connection_t *conn, surface_t *surface) {
+    xcb_free_gc(conn, surface->gc);
+#ifdef CAIRO_SUPPORT
+    if (cairo_surface_get_reference_count(surface->surface) == 0) {
```
stapelberg: Being involved this closely with the reference count system details seems a bit unclean to me. How about the following instead?

```c
cairo_surface_destroy(surface->surface);
surface->surface = NULL;
```

This is idempotent, i.e. can be called multiple times. See https://sources.debian.net/src/cairo/1.14.0-2.1/src/cairo-surface.c/?hl=1001#L949

src/x.c
stapelberg: Can we change it into a `xcb_change_gc` and put it after the `draw_util_surface_init` call? Then it doesn’t need to go into draw_util itself.

include/libi3.h
stapelberg: This wasn’t actually done AFAICT? Was that an oversight?

src/x.c
Airblader: I don't fancy it, because I intended `draw_util` to abstract away the implementation of the rendering (even more so given that it offers two implementations), so this is still a design leak, IMHO. But let's just make that compromise and add it.

Airblader:
@stapelberg All done.

stapelberg:
Can you rebase this PR please? There are unrelated commits inclided (e.g. “Use https if possible”).

Airblader:
@stapelberg Rebased

PR #2070, opened 2015-11-18
Smart option added to hide_edge_borders config param

chezstov:


chezstov:
Just like in https://github.com/ashinkarov/i3-extras/blob/master/0x2493-patches/smart-border.patch
but with new config option

src/con.c
```diff
@@ -1379,6 +1379,22 @@ Con *con_descend_direction(Con *con, direction_t direction) {
  *
  */
 Rect con_border_style_rect(Con *con) {
+    if (config.hide_edge_borders == HEBM_SMART) {
+        Con *cur = con;
+        while (cur->floating < 2 && cur) {
```
Airblader: Please don't use magic numbers.

src/con.c
```diff
@@ -1379,6 +1379,22 @@ Con *con_descend_direction(Con *con, direction_t direction) {
...
+            Con *par = cur->parent;
```
Airblader: We can afford the three letters, it'd be nicer to actually name these things `current` and `parent` to make it more readable.

src/con.c
```diff
@@ -1379,6 +1379,22 @@ Con *con_descend_direction(Con *con, direction_t direction) {
...
+            if (con_num_children(par) == 1) {
```
Airblader: If you invert this check, you can `break` early and we save one level of indentation.

Airblader:
I've added some comments. However, in general we recommend clarifying about a feature through a ticket first. This way you'll get feedback on whether such a feature would be accepted at all or whether it can be refined somehow.

chezstov:
Ok, I agree with all the comments and all the correct problem areas.

src/con.c
```diff
@@ -1379,6 +1379,24 @@ Con *con_descend_direction(Con *con, direction_t direction) {
  *
  */
 Rect con_border_style_rect(Con *con) {
+    if (config.hide_edge_borders == HEBM_SMART) {
+        Con *current = con;
+
+        while (current && !con_is_floating(current)) {
+            Con *parent = current->parent;
+
+            if (con_num_children(parent) != 1) {
```
Airblader: Here you unconditionally use `parent` and `con_num_children` never checks its input for `NULL`, so as long as `con_border_style_rect` isn't called with `NULL` to begin with, I don't think `current` will ever be `NULL`.

src/con.c
```diff
@@ -1379,6 +1379,24 @@ Con *con_descend_direction(Con *con, direction_t direction) {
  *
  */
 Rect con_border_style_rect(Con *con) {
+    if (config.hide_edge_borders == HEBM_SMART) {
```
Airblader: I'm not sure this logic works correctly for tabbed / stacked containers. Have you tried with them? I have to use similar logic in i3-gaps and there I introduced a new function in `src/con.c`:

````
/**
 * Returns the number of visible non-floating children of this container.
 * For example, if the container contains a hsplit which has two children,
 * this will return 2 instead of 1.
 */
int con_num_visible_children(Con *con) {
    if (con == NULL)
        return 0;

    int children = 0;
    Con *current = NULL;
    TAILQ_FOREACH(current, &(con->nodes_head), nodes) {
        /* Leaf nodes are a child. */
        if (con_is_leaf(current))
            children++;
        /* Stacked and tabbed containers are only considered one child. */
        else if (current->layout == L_TABBED || current->layout == L_STACKED)
            children++;
        /* Split containers need to be recursed. */
        else if (current->layout == L_SPLITH || current->layout == L_SPLITV)
            children += con_num_visible_children(current);
    }

    return children;                                                                                
}
```

Then you'd use this here:

````
if (config.hide_edge_borders == HEBM_SMART && con_num_visible_children(con_get_workspace(con))) {
    return (Rect){0, 0, 0, 0};
}
```

But maybe I'm misunderstanding it?

chezstov:
This is really good, I'll test this method soon.

Airblader:
If that function is what you'd want to work for staced/tabbed, I'd appreciate if you placed it in the same place where I have it in i3-gaps to make the merge for me easier (can't force you though :) ). See 

https://github.com/Airblader/i3/blob/gaps-next/src/con.c#L699

docs/userguide
```diff
@@ -612,7 +612,7 @@ to waste even two pixels in displayspace. Default is none.
 
 *Syntax*:
 -----------------------------------------------
-hide_edge_borders none|vertical|horizontal|both
+hide_edge_borders none|vertical|horizontal|both|smart
```
stapelberg: What does the “smart” option do? The commit doesn’t explain, and you don’t add an explanation to the userguide. Before looking at the code, I’d like to start with a textual description.

Airblader:
@chezstov What's the status here, are you going to keep working on this?

chezstov:
@Airblader IDK, I switched to i3-gaps, so I don't need in this option in i3, but I want to see it in original (TM) i3.

@stapelberg "smart" option for hide_edge_borders hides edge borders if container has only one visible window (as example if we have one window in workspace).

Airblader:
@chezstov Your answer is a bit ambigous to me. Are you going to maintain this PR and make the changes or not?

chezstov:
@Airblader And what difference does it make?

If it gets in the way, then I think you can close it.

Airblader:
There are open review comments so the difference is that if you are no longer interested in getting this merged, then we will close the pull request. And since I interpret your answer that way, I'll do that. 

If anyone wants to pick up on this, please open a new pull request.

PR #2073, opened 2015-11-20
Add i3 logo to dist tarball

tyll:


Airblader:
Can you explain why this should happen?

tyll:
On Fri, Nov 20, 2015 at 10:03:54AM -0800, Ingo Bürk wrote:
> Can you explain why this should happen?

I noticed that we ship a separate i3 logo in Fedora
(http://pkgs.fedoraproject.org/cgit/i3.git/tree/i3-logo.svg), then
wanted to upstream it and noticed that is is already in the git repo but
not the distribution tarball. The logos are somehow different, so by
using the one that is maintained by i3 directly, it will be always
updated when it is updated in i3 upstream.

Airblader:
Interesting to know. Thanks for the explanation!

stapelberg:
It’s installed into datadir, but do you know what it’s used for?

With regards to the difference in the files, I think that’s just the s/i3.zekjur.net/i3wm.org/ we did at some point.

tyll:
On Fri, Nov 20, 2015 at 11:20:28AM -0800, Michael Stapelberg wrote:
> It’s installed into datadir, but do you know what it’s used for?

I assumed it was for a desktop file but I cannot see it being used
there. I will try to find the reason for it being installed.

Airblader:
@tyll Any updates?

tyll:
I was not able to figure out, whether the logo is really used in Fedora. Therefore I close this PR for now.

PR #2074, opened 2015-11-20
Support _NET_WM_USER_TIME.

### LINKED ISSUE #2064: enhancement: implement not focusing windows when _NET_WM_USER_TIME == 0
From http://standards.freedesktop.org/wm-spec/latest/ar01s05.html:

> The special value of zero on a newly mapped window can be used to request that the window not be initially focused when it is mapped.

Applications such as Chrome use this hint that way: https://code.google.com/p/chromium/codesearch#chromium/src/ui/views/widget/desktop_aura/desktop_window_tree_host_x11.cc (line 1672)

Airblader:
With this patch, we support the special value "0" for _NET_WM_USER_TIME
on a window upon managing it, which indicates that the window shall not
be focused.

fixes #2064


src/ewmh.c
```diff
@@ -263,8 +263,8 @@ void ewmh_setup_hints(void) {
     /* I’m not entirely sure if we need to keep _NET_WM_NAME on root. */
     xcb_change_property(conn, XCB_PROP_MODE_REPLACE, root, A__NET_WM_NAME, A_UTF8_STRING, 8, strlen("i3"), "i3");
 
-    /* only send the first 31 atoms (last one is _NET_CLOSE_WINDOW) increment that number when adding supported atoms */
-    xcb_change_property(conn, XCB_PROP_MODE_REPLACE, root, A__NET_SUPPORTED, XCB_ATOM_ATOM, 32, /* number of atoms */ 32, supported_atoms);
+    /* only send the first 33 atoms (last one is _NET_CLOSE_WINDOW) increment that number when adding supported atoms */
+    xcb_change_property(conn, XCB_PROP_MODE_REPLACE, root, A__NET_SUPPORTED, XCB_ATOM_ATOM, 32, /* number of atoms */ 33, supported_atoms);
```
stapelberg: Is this actually necessary?

I’m just asking because strictly speaking, we don’t support what _NET_WM_USER_TIME actually means, aside from the special case of it being zero: https://mail.gnome.org/archives/desktop-devel-list/2004-August/msg00158.html

Airblader: I guess it's somewhat of a judgement call. You could also argue that a client could decide to set it to 0 for all its windows because the window manager doesn't claim to support it. Which application is at fault in this case?

stapelberg: I think client’s should just set the hint and not query the window manager support. It’s purely informational for the window manager, so there is no need to know whether the window manager will do anything with it.

Airblader: OK, sounds good to me. I'll remove it here then.

Airblader:
@stapelberg Done

PR #2078, opened 2015-11-22
Bug: parse con_id base 16

acrisci:
Mouse bindings that target the window that was clicked send the command
to the parser with `con_id` of the clicked window serialized base 16
for compatability with FreeBSD. See 7c2842e for explaination.

Set base to 0 for strtol to handle base 16 numbers for that reason.

This allows mouse bindings that target specific windows to work
correctly. Without this change, the focused window is always targetted
rather than the window that was actually clicked.

Regression introduced in b744c5e.

### DIFF
--- a/src/match.c
+++ b/src/match.c
@@ -280,7 +280,7 @@ void match_parse_property(Match *match, const char *ctype, const char *cvalue) {
         }
 
         char *end;
-        long parsed = strtol(cvalue, &end, 10);
+        long parsed = strtol(cvalue, &end, 0);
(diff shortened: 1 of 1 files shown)


acrisci:
@Airblader please check b744c5e for any other errors.

acrisci:
#1661

Airblader:
@acrisci Diffing the lines that have been refactored yields only exactly this one difference. I could've sworn I also did this comparison when making the commit, though. Well, had a 50:50 chance of getting it lucky, and didn't. Thanks for the fix.

PR #2103, opened 2015-12-09
Reject invalid match criteria with an error.

### LINKED ISSUE #2091: Command with `con_id` selector works on the focused window with no match
Try this in a terminal:

    i3-msg [con_id="asdf"] kill

It will kill the terminal even though "asdf" will not match any `con_id`.

It should instead do nothing and give an error message in the json response that says there is no con with that id.

Airblader:
Previously, using a command like

  [con_id=foo] kill

would kill the currently focused window because while an error for
not being able to parse the con_id was logged, no further action
was taken, which caused the criterion to be ignored. In this case,
the fallback behavior of using the focused window took over.

For con_id, id and window_type we now reject incorrect values with
an error and abort the command.

fixes #2091

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -42,6 +42,16 @@
         }                                               \
     } while (0)
 
+/** If an error occured during parsing of the criteria, we want to exit instead
+ * of relying on fallback behavior. See #2091. */
+#define HANDLE_INVALID_MATCH                                   \
+    do {                                                       \
+        if (current_match->error != NULL) {                    \
+            yerror("Invalid match: %s", current_match->error); \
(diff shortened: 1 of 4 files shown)


src/commands.c
```diff
@@ -1233,6 +1245,8 @@ void cmd_kill(I3_CMD, const char *kill_mode_str) {
         return;
     }
 
+    HANDLE_INVALID_MATCH;
```
Airblader: Unfortunately `cmd_kill` cannot use `HANDLE_EMPTY_MATCH` directly, so we need to call it ourselves here.

PR #2105, opened 2015-12-11
Reject empty swallow definitions

### LINKED ISSUE #2099: Crash with layouts
I was trying to use the append_layout feature with this simple layout (4 urxvt console):
https://transfer.sh/UyPBc/ws1.json
So just after loading the layout via i3-msg append_layout, I get this screen: 
https://transfer.sh/8KUoD/2015-12-05-162429-1920x1080-scrot.png
Funny part is that there seems to be a placeholder not drawn on the right.
Then I open/close randomly terminals and eventually I always get this backtrace:
```
Reading in symbols for ../i3/src/sighandler.c...done.
Reading in symbols for ../sysdeps/unix/sysv/linux/x86_64/sigaction.c...done.
Reading in symbols for ../i3/src/restore_layout.c...done.
#0  0x00007fae7288f8c0 in __poll_nocancel ()
at ../sysdeps/unix/syscall-template.S:81
#1  0x00007fae75124bd2 in ?? () from /usr/lib/x86_64-linux-gnu/libxcb.so.1
#2  0x00007fae7512674f in xcb_wait_for_event ()
from /usr/lib/x86_64-linux-gnu/libxcb.so.1
#3  0x0000000000408f21 in handle_signal (sig=11, info=0x7ffdbe933f30, 
      data=0x7ffdbe933e00) at ../i3/src/sighandler.c:303
#4  <signal handler called>
#5  0x0000000000438640 in update_placeholder_contents (state=0x2428130)
at ../i3/src/restore_layout.c:152
#6  0x0000000000439159 in configure_notify (event=0x242dbf0)
at ../i3/src/restore_layout.c:335
#7  0x000000000043922a in restore_handle_event (type=22, event=0x242dbf0)
at ../i3/src/restore_layout.c:351
#8  0x0000000000438180 in restore_xcb_check_cb (loop=0x7fae733ff900, 
      w=0x233a210, revents=32768) at ../i3/src/restore_layout.c:73
#9  0x00007fae731f5e45 in ev_invoke_pending ()
from /usr/lib/x86_64-linux-gnu/libev.so.4
#10 0x00007fae731f8fe7 in ev_run () from /usr/lib/x86_64-linux-gnu/libev.so.4
#11 0x0000000000439274 in ev_loop (loop=0x7fae733ff900, flags=0)
at /usr/include/ev.h:826
#12 0x000000000043d0ca in main (argc=4, argv=0x7ffdbe934d08)
at ../i3/src/main.c:876
Reading in symbols for libc-start.c...done.
Detaching from program: /usr/bin/i3, process 854
```

Airblader:
Empty swallow definitions don't make sense and can lead to crashes,
for that reason we reject them.

fixes #2099

### DIFF
--- a/src/load_layout.c
+++ b/src/load_layout.c
@@ -30,6 +30,7 @@ static bool parsing_geometry;
 static bool parsing_focus;
 static bool parsing_marks;
 struct Match *current_swallow;
+static bool swallow_is_empty;
 
 /* This list is used for reordering the focus stack after parsing the 'focus'
  * array. */
@@ -48,6 +49,7 @@ static int json_start_map(void *ctx) {
         current_swallow = smalloc(sizeof(Match));
         match_init(current_swallow);
         TAILQ_INSERT_TAIL(&(json_node->swallow_head), current_swallow, matches);
+        swallow_is_empty = true;
     } else {
         if (!parsing_rect && !parsing_deco_rect && !parsing_window_rect && !parsing_geometry) {
             if (last_key && strcasecmp(last_key, "floating_nodes") == 0) {
@@ -151,6 +153,13 @@ static int json_end_map(void *ctx) {
         json_node = json_node->parent;
     }
 
+    if (parsing_swallows && swallow_is_empty) {
+        /* We parsed an empty swallow definition. This is an invalid layout
+         * definition, hence we reject it. */
+        ELOG("Layout file is invalid: found an empty swallow definition.\n");
+        return 0;
+    }
+
     parsing_rect = false;
     parsing_deco_rect = false;
     parsing_window_rect = false;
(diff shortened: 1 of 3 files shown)


src/load_layout.c
```diff
@@ -151,6 +153,13 @@ static int json_end_map(void *ctx) {
         json_node = json_node->parent;
     }
 
+    if (parsing_swallows && swallow_is_empty) {
```
Airblader: We need a separate flag instead of just checking all fields of `current_swallow` because some of them are integers and booleans and we shouldn't / can't rely on "0" meaning "wasn't present". Those fields are mostly relevant for in-place restarts anyway.

Airblader:
I wonder whether we could instead just make `match_matches_window` return `false` if none of the criteria were specified. I think that would fix the bug as well, haven't tried it yet, though.

@stapelberg What would be your preferred solution?

stapelberg:
> I wonder whether we could instead just make match_matches_window return false if none of the criteria were specified. I think that would fix the bug as well, haven't tried it yet, though.
> 
> @stapelberg What would be your preferred solution (assuming that works)?

The problem I see with that is that it changes behavior. Users might be using commands like `[] kill`, i.e. relying on the fact that empty criteria match everything. I’d prefer it if this pull request could only look at swallows definitions and not change behavior of criteria in general.

Airblader:
@stapelberg Alright. Then this PR already does what you want since it only rejects empty swallow definitions.

TyberiusPrime:
With regards to #4185 and i3-instant-layout I politely disagree that 'empty swallows make no sense'.

PR #2108, opened 2015-12-12
Allow mouse bindings to run on the root window.

### LINKED ISSUE #2097: [enhancement] Run mouse bindings on the root window
A mouse binding like

    bindsym --whole-window $mod+button4 workspace next

will only work on windows, but not on an empty workspace. We should allow selecting mouse bindings on the root window itself. Either by default when `--whole-window` is given or via another flag such as `--include-root-window`.

Airblader:
Previously, mouse bindings could only be run when a window was present,
by using --whole-window. Such bindings would not work on empty
workspaces. However, this is a valid usecase for bindings like

    bindsym $mod+button4 workspace prev
    bindsym $mod+button5 workspace next

Hence, we need to grab the root window as well and run bindings on it.

fixes #2097

### DIFF
--- a/src/click.c
+++ b/src/click.c
@@ -363,6 +363,21 @@ int handle_button_press(xcb_button_press_event_t *event) {
         return route_click(con, event, mod_pressed, CLICK_INSIDE);
 
     if (!(con = con_by_frame_id(event->event))) {
+        /* Run bindings on the root window as well, see #2097. We only run it
+         * if --whole-window was set as that's the equivalent for a normal
+         * window. */
+        if (event->event == root) {
+            Binding *bind = get_binding_from_xcb_event((xcb_generic_event_t *)event);
+            if (bind != NULL && bind->whole_window) {
+                CommandResult *result = run_binding(bind, NULL);
+                if (result->needs_tree_render) {
+                    tree_render();
+                }
+
+                command_result_free(result);
+            }
+        }
+
         /* If the root window is clicked, find the relevant output from the
          * click coordinates and focus the output's active workspace. */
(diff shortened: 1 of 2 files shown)


Airblader:
I gotta play around with this before it can be merged. Just opening the PR already since I'm sitting at the airport…

Airblader:
@stapelberg OK, looking good as far as I can tell.

src/click.c
```diff
@@ -363,6 +363,21 @@ int handle_button_press(xcb_button_press_event_t *event) {
...
 
     if (!(con = con_by_frame_id(event->event))) {
+        /* Run bindings on the root window as well, see #2097. We only run it
+         * if --whole-window was set as that's the equivalent for a normal
+         * window. */
+        if (event->event == root) {
+            Binding *bind = get_binding_from_xcb_event((xcb_generic_event_t *)event);
+            if (bind != NULL && bind->whole_window) {
+                CommandResult *result = run_binding(bind, NULL);
+                if (result->needs_tree_render) {
```
Airblader: It actually looks to me as if calling `tree_render` on the caller side isn't necessary since `run_binding` already does that. This is valid for all calls to this function, despite the documentation of it stating that the caller should do it. I'll send a separate PR (after this one) that cleans this up and avoids the double tree render.

stapelberg:
Side note: having tests for this would be cool. Given that we have https://github.com/i3/i3/blob/next/testcases/lib/i3test/XTEST.pm now, perhaps adding support for ButtonPress events is doable? What do you think, @Airblader, would that be interesting to you?

Airblader:
@stapelberg Yeah, definitely. I'll look into it and create an issue for it as a reminder.

PR #2112, opened 2015-12-15
Avoid rendering the tree twice in a row

Airblader:
The callee already renders the tree if necessary, so despite the documentation
of the function, doing it again on the caller side is unnecessary.


Airblader:
Waiting for #2108

Airblader:
@stapelberg Good to go.

PR #2113, opened 2015-12-15
Match on all criteria even if con_id or con_mark are given.

### LINKED ISSUE #2111: con_id and mark criteria ignore all other criteria
If a command is specified with either `con_id` or `mark`, all other criteria are ignored. For example, the following will always kill the focused window:

    [con_id=__focused__ class=foobar] kill

This is due to the special handling of those criteria in `cmd_criteria_match_windows` in `src/commands.c`.

I guess the rationale for this was that `con_id` and `mark` uniquely identify a window already, so all other criteria are nops. However, this isn't the case as seen above since further criteria can rule out the window.

A real-life scenario for this breakage is described [here](https://faq.i3wm.org/question/7662/reverse-perl-matches-in-criteria-in-i3-config/?answer=7663#post-id-7663) where the user wants the kill shortcut to kill the focused window unless it's a VirtualBox window.

i3 4.11

Airblader:
Previously, if a match specification contained the con_id or con_mark criterion,
all other criteria were ignored. However, a user may want to specify one of
those two unique identifiers and still specify others as well, for example to
match the currently focused window, but only if it has a certain WM_CLASS:

    [con_id=__focused__ class=special] kill

We now check all specified criteria.

fixes #2111

### DIFF
--- a/src/match.c
+++ b/src/match.c
@@ -223,11 +223,25 @@ bool match_matches_window(Match *match, i3Window *window) {
         }
     }
 
-    /* We don’t check the mark because this function is not even called when
-     * the mark would have matched - it is checked in cmdparse.y itself */
     if (match->mark != NULL) {
-        LOG("mark does not match\n");
-        return false;
+        if ((con = con_by_window_id(window->id)) == NULL)
+            return false;
+
+        bool matched = false;
+        mark_t *mark;
+        TAILQ_FOREACH(mark, &(con->marks_head), marks) {
+            if (regex_matches(match->mark, mark->name)) {
+                matched = true;
+                break;
+            }
+        }
+
+        if (matched) {
+            LOG("mark matches\n");
+        } else {
+            LOG("mark does not match\n");
+            return false;
+        }
     }
 
     return true;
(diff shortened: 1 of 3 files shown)


Airblader:
Needs some tests.

src/match.c
```diff
@@ -223,11 +223,25 @@ bool match_matches_window(Match *match, i3Window *window) {
         }
     }
 
-    /* We don’t check the mark because this function is not even called when
-     * the mark would have matched - it is checked in cmdparse.y itself */
     if (match->mark != NULL) {
-        LOG("mark does not match\n");
-        return false;
+        if ((con = con_by_window_id(window->id)) == NULL)
```
Airblader: One little caveat is that if a command specifies a mark and the con with the mark has a window, we now check the marks twice – once in `commands.c` and once here. However, the only alternative for this that I see is passing a flag to the function on whether the mark should be checked, which doesn't seem very clean. Given that this is only called occassionally and that windows generally don't have hundreds of marks, I think this is OK.

Airblader:
@stapelberg OK, I've added tests.

Issue #2122, opened 2015-12-20
Comma enclosed in quotes


winku:
In config file:
`bindsym $mod+r exec setxkbmap "ru,us";`
In error log (after pressing $mod+r):
```
ERROR: Could not translate string to key symbol: "exec"
ERROR: Expected one of these tokens: <end>, '[', 'move', 'exec', 'exit', 'restart', 'reload', 'shmlog', 'debuglog', 'border', 'layout', 'append_layout', 'workspace', 'focus', 'kill', 'open', 'fullscreen', 'sticky', 'split', 'floating', 'mark', 'unmark', 'resize', 'rename', 'nop', 'scratchpad', 'title_format', 'mode', 'bar'
ERROR: Your command: exec setxkbmap "ru,us";
ERROR:                                  ^^^^
~
~
```
i3 version:
`i3 version 4.11 (2015-09-30, branch "4.11")`

Looks like something is wrong with comma enclosed in quotes. The problem reproduces even if I put `bindsym ...` statement in `mode` definition.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

kneitinger:
Furthermore, any combination of escaping quotes, comma, etc. all produce the same error, if that is helpful to know.

stapelberg:
You’ll need to quote the argument of the exec command, i.e.:

    bindsym $mod+x exec "setxkbmap ru,us"

I just verified that this works:
```
2015-12-25 16:03:48 - commands_parser.c:parse_command:267 - COMMAND: *exec "setxkbmap ru,us"*
2015-12-25 16:03:48 - commands.c:cmd_criteria_init:259 - Initializing criteria, current_match = 0x681c28
2015-12-25 16:03:48 - commands.c:cmd_exec:1272 - should execute setxkbmap ru,us, no_startup_id = 0
2015-12-25 16:03:48 - startup id = i3/setxkbmap/7486-0-x1_TIME313638477
2015-12-25 16:03:48 - executing: setxkbmap ru,us
2015-12-25 16:03:48 - commands.c:cmd_criteria_init:259 - Initializing criteria, current_match = 0x681c28
2015-12-25 16:03:48 - ipc.c:ipc_send_binding_event:1291 - Issue IPC binding run event (sym = x, code = 0)
[...]

PR #2125, opened 2015-12-24
Bugfix: allow exec command to properly handle quoted ';' and ','

kneitinger:
Fixes bug [#2122](https://github.com/i3/i3/issues/2122), "Comma enclosed in quotes".

All tests pass.

### DIFF
--- a/src/commands_parser.c
+++ b/src/commands_parser.c
@@ -221,10 +221,15 @@ char *parse_string(const char **walk, bool as_word) {
              * comma (,) and semicolon (;) which introduce a new
              * operation or command, respectively. Also, newlines
              * end a command. */
-            while (**walk != ';' && **walk != ',' &&
+            bool quote = false;
+            while ((**walk != ';' || quote) &&
+                   (**walk != ',' || quote) &&
                    **walk != '\0' && **walk != '\r' &&
-                   **walk != '\n')
+                   **walk != '\n') {
+                if (**walk == '"')
+                    quote = !quote;
                 (*walk)++;
+            }
         } else {
             /* For a word, the delimiters are white space (' ' or
              * '\t'), closing square bracket (]), comma (,) and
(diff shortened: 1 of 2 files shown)


Airblader:
Cool, thank you. Can you please add a line to the commit message saying "fixes #2122"?

Also, all tests passing is great, but not enough since they didn't catch this bug. Can you add a new test case that makes sure this keeps working in the future?

kneitinger:
I certainly can, but it may take a day or so.  The Travis CI build failed, did I introduce something that breaks the build?

Airblader:
Don't worry, we have time :) The build broke because the formatting is wrong. Just run clang-format on the changed file and update the commit.

kneitinger:
Okay, I fixed the commit message and corrected the formatting.  I believe I wrote a sufficient test, but I do not feel 100% confident about it.  Does it look correct?

kneitinger:
It can indeed! :D

stapelberg:
I’m hesitant to merge this. As I pointed out in issue #2122, the intended way to use the exec command is to quote the entire argument. I’d even say that it’s not i3’s job to interpret anything you pass to exec, as interpreting shell commands sounds like a road to madness :).

With the clarification in #2122, is there anything that actually _needs_ fixing?

kneitinger:
Yeah that was a super helpful clarification, and avoiding shell command parsing is a great idea! I think the docs are ambiguous as to what needs to be quoted, so that could perhaps use fixing.  Should there perhaps be a test case to verify the intended quoting of commands works?  One with internal, escaped quotes (as discussed in #2122 ) and a comma and semicolon?

stapelberg:
As mentioned in https://github.com/i3/i3/issues/2122#issuecomment-167262810, pull requests are very welcome.

A test case also can’t hurt, so if you’re inclined, please go ahead.

I’ll close this pull request, as it’s obsolete now.

PR #2133, opened 2015-12-28
Added test for #2097.

### LINKED ISSUE #2115: [enhancement] Add tests for mouse bindings on empty workspaces
See discussion on #2108:

> Side note: having tests for this would be cool. Given that we have https://github.com/i3/i3/blob/next/testcases/lib/i3test/XTEST.pm now, perhaps adding support for ButtonPress events is doable?

This refers to #2097

Airblader:
fixes #2115

### DIFF
--- a/testcases/lib/i3test/XTEST.pm
+++ b/testcases/lib/i3test/XTEST.pm
@@ -15,6 +15,8 @@ our @EXPORT = qw(
     set_xkb_group
     xtest_key_press
     xtest_key_release
+    xtest_button_press
+    xtest_button_release
     listen_for_binding
     start_binding_capture
     binding_events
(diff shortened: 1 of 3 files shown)


stapelberg:
I see that this PR is using xdotool. Are you using xdotool because it was not feasible to implement the functions you’re using (moving the cursor, sending a ButtonPress event) in https://github.com/i3/i3/blob/next/testcases/lib/i3test/XTEST.pm? If yes, can you share more details? If no, could you give that a go? That’d completely eliminate the overhead of running an external binary (we want our tests to be fast) and not tie our test suite to xdotool any more than it currently is :).

Airblader:
There's no specific reason to use xdotool other than "it was easier". xdotool uses xtest in this case anyway; I guess this is what @acrisci meant with having libxdotool. I doubt that running the binary makes this test significantly slower, but we can leave the PR open for now and I'll try changing it into xtest.

Airblader:
@stapelberg I've migrated it to xcb-xtest.

stapelberg:
Cool, thanks!

Issue #434, opened 2011-07-06
Support for 16-bit color.


i3bot:
[Originally reported by anonymous]
(Colours are wrong running in 16-bit mode.

owenversteeg:
Would it be possible to reopen this? There are a number of screens for the Raspberry Pi that people run i3 on to conserve space, and most of these have only 16bit color. As such, more people are using i3 with 16bit color and everything appears all wonky.

stapelberg:
@owenversteeg Our attitude on this has not changed. If someone sends a pull request, we’ll merge it. If you want to see this feature, the best way to achieve that is to send a pull request which implements it.

pickfire:
I seem to have this problem on RPi too.

PR #2135, opened 2015-12-28
Correct color management for non-true color displays

Alexis211:
This pull request fixes pango font color management and reintroduces fully-fledged `get_colorpixel` function in order to support non-true color displays such as 16 bpp displays.

The `get_colorpixel` function was simplified in ce97e23913b2fbd278b2079821e7c1e6d7b83d43 in order to simplify the code and do less round trips to the X11 server. However color management became wrong for 16-bit monitors and it seems that many people have been complaining about this (see issue #434). Simply fixing this function was not enough as the `set_font_colors` was not working right when non-RGB colorpixels were used for pango fonts. This was fixed by having `set_font_colors` take `color_t` as argument instead of colorpixels. All the calls to `set_font_colors` have been fixed.

The new `get_colorpixel` function may not be optimal. In particular, it has the following flaws:
- It defines `extern xcb_connection_t *conn` as if it always existed. Should this be factored into a common header file ?
- It defines the Colorpixel structure in the C file. Which header file should this be put in?
- The current (shorter) version is faster for the cases where the display is TrueColor (majority of cases). How should we detect which case we are in ? Are we sure XCB doesn't already optimize the true color case to save an X11 roundtrip ?

This pull request is not meant to be merged yet and I would be glad to have the developper's opinion on these questions.

### DIFF
--- a/libi3/get_colorpixel.c
+++ b/libi3/get_colorpixel.c
@@ -9,20 +9,27 @@
 #include <stdint.h>
 #include <string.h>
 
+#include "queue.h"
 #include "libi3.h"
 
+struct Colorpixel {
+    char *hex;
+    uint32_t pixel;
+
+    SLIST_ENTRY(Colorpixel)
+    colorpixels;
+};
+
+SLIST_HEAD(colorpixel_head, Colorpixel)
+colorpixels;
+
 /*
- * Returns the colorpixel to use for the given hex color (think of HTML). Only
- * works for true-color (vast majority of cases) at the moment, avoiding a
- * roundtrip to X11.
+ * Returns the colorpixel to use for the given hex color (think of HTML).
  *
  * The hex_color has to start with #, for example #FF00FF.
  *
  * NOTE that get_colorpixel() does _NOT_ check the given color code for validity.
  * This has to be done by the caller.
- *
- * NOTE that this function may in the future rely on a global xcb_connection_t
- * variable called 'conn' to be present.
  *
  */
 uint32_t get_colorpixel(const char *hex) {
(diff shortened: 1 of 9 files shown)


libi3/get_colorpixel.c
```diff
@@ -21,18 +35,46 @@
...
-        {hex[3], hex[4], '\0'},
-        {hex[5], hex[6], '\0'}};
-    uint8_t r = strtol(strgroups[0], NULL, 16);
-    uint8_t g = strtol(strgroups[1], NULL, 16);
-    uint8_t b = strtol(strgroups[2], NULL, 16);
-
-    return (0xFF << 24) | (r << 16 | g << 8 | b);
+    /* Lookup this colorpixel in the cache */
+    struct Colorpixel *colorpixel;
+    SLIST_FOREACH(colorpixel, &(colorpixels), colorpixels)
```
Airblader: nit: Let's please use braces here so indentation is also correct.

Airblader:
I haven't tried this out or anything, but from a quick first look this looks good to me.

stapelberg:
> It defines extern xcb_connection_t *conn as if it always existed. Should this be factored into a common header file ?

Yeah, I think moving this into `include/libi3.h` is fine. Can you do that in a separate pull request please? A number of other files also have such a declaration, even though they shouldn’t need it.

> It defines the Colorpixel structure in the C file. Which header file should this be put in?

Given that the structure is only used in that C file, it’s fine to leave it in there.

> The current (shorter) version is faster for the cases where the display is TrueColor (majority of cases). How should we detect which case we are in ? Are we sure XCB doesn't already optimize the true color case to save an X11 roundtrip ?

XCB intentionally doesn’t do optimizations like that, it’s a very low-level (auto-generated) library implementing the X11 protocol.

As per https://sources.debian.net/src/libx11/2:1.6.2-3/src/Depths.c/?hl=34#L34, I think `xcb_screen_t`’s `root_depth` property might be what you’re looking for.

Alexis211:
@stapelberg Ok, I think we're good now! See also PR #2138.

stapelberg:
Can you rebase onto next and squash your commits into logical units please?

Alexis211:
@stapelberg Done.

PR #2136, opened 2015-12-28
draw client borders in border color, not background

lotheac:
I mainly use windows without title bars (ie. 'pixel' border, not 'normal'), and sometimes have a little trouble seeing which window is focused. Because of this I like to make the window border a color that stands out a bit more, but that has the unfortunate side effect of also making focused client title bars (for clients that do have a 'normal' border) stand out because the 'background' color is used for drawing the borders. I find that a bit confusing, and think that the color called "border" should be used for client borders, instead of "background".

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -836,9 +836,9 @@ workspace "2: vim" output VGA1
 You can change all colors which i3 uses to draw the window decorations.
 
 *Syntax*:
-------------------------------------------------------
-<colorclass> <border> <background> <text> <indicator>
-------------------------------------------------------
+-------------------------------------------------------------------------
+<colorclass> <border> <background> <text> <indicator> <decoration_border>
+-------------------------------------------------------------------------
 
 Where colorclass can be one of:
 
@@ -864,19 +864,19 @@ Colors are in HTML hex format (#rrggbb), see the following example:
 
 *Examples (default colors)*:
 ---------------------------------------------------------
-# class                 border  backgr. text    indicator
-client.focused          #4c7899 #285577 #ffffff #2e9ef4
-client.focused_inactive #333333 #5f676a #ffffff #484e50
-client.unfocused        #333333 #222222 #888888 #292d2e
-client.urgent           #2f343a #900000 #ffffff #900000
-client.placeholder      #000000 #0c0c0c #ffffff #000000
+# class                 border  backgr. text    indicator decoration_border
+client.focused          #4c7899 #285577 #ffffff #2e9ef4   #285577
+client.focused_inactive #333333 #5f676a #ffffff #484e50   #5f676a
+client.unfocused        #333333 #222222 #888888 #292d2e   #222222
+client.urgent           #2f343a #900000 #ffffff #900000   #900000
+client.placeholder      #000000 #0c0c0c #ffffff #000000   #0c0c0c
 
 client.background       #ffffff
 ---------------------------------------------------------
 
 Note that for the window decorations, the color around the child window is the
-background color, and the border color is only the two thin lines at the top of
-the window.
+"decoration_border", and "border" color is only the two thin lines around the
+titlebar.
 
 The indicator color is used for indicating where a new window will be opened.
 For horizontal split containers, the right border will be painted in indicator
--- a/include/config.h
+++ b/include/config.h
@@ -54,6 +54,7 @@ struct Colortriple {
     color_t background;
     color_t text;
     color_t indicator;
+    color_t decoration_border;
 };
 
 /**
--- a/include/config_directives.h
+++ b/include/config_directives.h
@@ -59,7 +59,7 @@ CFGFUN(no_focus);
 CFGFUN(ipc_socket, const char *path);
 CFGFUN(restart_state, const char *path);
 CFGFUN(popup_during_fullscreen, const char *value);
-CFGFUN(color, const char *colorclass, const char *border, const char *background, const char *text, const char *indicator);
+CFGFUN(color, const char *colorclass, const char *border, const char *background, const char *text, const char *indicator, const char *decoration_border);
 CFGFUN(color_single, const char *colorclass, const char *color);
 CFGFUN(floating_modifier, const char *modifiers);
 CFGFUN(new_window, const char *windowtype, const char *border, const long width);
--- a/parser-specs/config.spec
+++ b/parser-specs/config.spec
@@ -282,9 +282,15 @@ state COLOR_TEXT:
 
 state COLOR_INDICATOR:
   indicator = word
-      -> call cfg_color($colorclass, $border, $background, $text, $indicator)
+      -> COLOR_DECORATION_BORDER
   end
-      -> call cfg_color($colorclass, $border, $background, $text, NULL)
+      -> call cfg_color($colorclass, $border, $background, $text, NULL, NULL)
+
+state COLOR_DECORATION_BORDER:
+  decoration_border = word
+      -> call cfg_color($colorclass, $border, $background, $text, $indicator, $decoration_border)
+  end
+      -> call cfg_color($colorclass, $border, $background, $text, $indicator, NULL)
 
 # <exec|exec_always> [--no-startup-id] command
 state EXEC:
--- a/src/config.c
+++ b/src/config.c
@@ -189,12 +189,13 @@ void load_configuration(xcb_connection_t *conn, const char *override_configpath,
     memset(&config, 0, sizeof(config));
 
 /* Initialize default colors */
-#define INIT_COLOR(x, cborder, cbackground, ctext, cindicator) \
-    do {                                                       \
-        x.border = draw_util_hex_to_color(cborder);            \
-        x.background = draw_util_hex_to_color(cbackground);    \
-        x.text = draw_util_hex_to_color(ctext);                \
-        x.indicator = draw_util_hex_to_color(cindicator);      \
+#define INIT_COLOR(x, cborder, cbackground, ctext, cindicator)     \
+    do {                                                           \
+        x.border = draw_util_hex_to_color(cborder);                \
+        x.background = draw_util_hex_to_color(cbackground);        \
+        x.text = draw_util_hex_to_color(ctext);                    \
+        x.indicator = draw_util_hex_to_color(cindicator);          \
+        x.decoration_border = draw_util_hex_to_color(cbackground); \
     } while (0)
 
     config.client.background = draw_util_hex_to_color("#000000");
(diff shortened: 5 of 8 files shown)


lotheac:
The Travis build failure is not because of my changes; the build system seems broken. https://travis-ci.org/i3/i3/jobs/99092575#L353

Airblader:
The problem with this change is that it clearly breaks behavior and will confuse users when they update. Other than that I totally welcome this change because the current behavior is pretty confusing too.

lotheac:
On Mon, Dec 28 2015 08:42:41 -0800, Ingo Bürk wrote:
> The problem with this change is that it clearly breaks behavior and will confuse users when they update. Other than that I totally welcome this change because the current behavior is pretty confusing too. 

Yeah, I realize it's a definite behavior change and can surprise
existing users. Is that a blocker for integration, and if so, how do you
propose to move forward with this?

-- 
Lauri Tirkkonen | lotheac @ IRCnet

Airblader:
> Is that a blocker for integration

We try to avoid backwards incompatible changes as much as possible and only allow them if there is a clear benefit. This decision will be up to @stapelberg.

> and if so, how do you propose to move forward with this?

If this is the case, we need to find a backwards compatible way. That includes finding a way that is justified when comparing complexity and benefit. Before thinking about this, let's wait and see what @stapelberg thinks.

stapelberg:
I think an alternative that wouldn’t break existing users would be to make the background color part of the color classes that are defined for client.focused etc., right?

lotheac:
On Tue, Dec 29 2015 03:26:02 -0800, Michael Stapelberg wrote:
> I think an alternative that wouldn’t break existing users would be to make the background color part of the color classes that are defined for client.focused etc., right?

Sorry, I don't understand what you mean by this. A color class
definition already includes 'background' and 'border' per the user
guide[0]:

    <colorclass> <border> <background> <text> <indicator>

But currently, 'border' is only used for the two lines above and below
the client title bar, and 'background' is used for the border that
surrounds the client. My change changes the latter to use 'border'.

[0]: https://i3wm.org/docs/userguide.html#_changing_colors

-- 
Lauri Tirkkonen | lotheac @ IRCnet

stapelberg:
What I’m suggesting is adding e.g. `client_background` in addition to `background`, `color`, etc.

`client_background` would default to `client.background`, but can be overridden. That way, existing behavior is not changed, and the functionality you’re looking for can be achieved.

lotheac:
On Tue, Dec 29 2015 04:52:47 -0800, Michael Stapelberg wrote:
> What I’m suggesting is adding e.g. `client_background` in addition to `background`, `color`, etc.
> 
> `client_background` would default to `client.background`, but can be overridden. That way, existing behavior is not changed, and the functionality you’re looking for can be achieved.

Well, client.background is unrelated to this entire issue as far as I
can tell -- it's a separate color class which only takes one argument
and is used for a different purpose, but I think I see what you're
saying (correct me if I'm wrong): add a fifth color to a color class
line, which would then be used for the client borders (ie. the things
surrounding the client), and if not defined, defaulting to <background>.

If that is done then I think the nomenclature needs to move around a
bit: I think borders should be called borders. That leaves coming up
with a new name for what is currently 'border' (the two lines above and
below the title bar). Also because this seems to be an area ripe for
confusion maybe <background> needs renaming too because
client.background is documented in the same place but is a completely
different thing.

I personally think that while changing the semantics of 'border' to
match what people normally consider borders is an incompatible change,
the implications of such a change to existing configurations are not so
large that it is worth it to try to be compatible. But it's obviously
your call.

If instead you are unhappy with the behavior change, I propose the
following syntax for the color changes:

    <colorclass> <separator> <titlebar> <text> <indicator> <border>

Actually the 'background' color could be renamed to 'titlebar' or
something regardless of which way this goes, because that won't break
anything and will clarify things.

So what do you think?

-- 
Lauri Tirkkonen | lotheac @ IRCnet

Airblader:
I suggest the names decoration_background and decoration_border. "decoration" is the terminology we use so far.

lotheac:
On Tue, Dec 29 2015 08:55:33 -0800, Ingo Bürk wrote:
> I suggest the names decoration_background and decoration_border. "decoration" is the terminology we use so far. 

Ok, sure. Like this then, if wishing to preserve compatibility:

    <colorclass> <decoration_separator> <decoration_background> <text> <indicator> <decoration_border>

But I still don't have a clear answer to this:

> I personally think that while changing the semantics of 'border' to
> match what people normally consider borders is an incompatible change,
> the implications of such a change to existing configurations are not
> so large that it is worth it to try to be compatible. But it's
> obviously your call.

If you are OK with this being a behavior change, then I'll amend the
current changeset to clarify the names and documentation, making the
syntax as follows:

    <colorclass> <decoration_border> <decoration_background> <text> <indicator>

So, which way should this go?

-- 
Lauri Tirkkonen | lotheac @ IRCnet

stapelberg:
While I agree that it would be better to have the terms mean what you suggested, the stronger argument to me is that changes in that area are very confusing. New versions of i3 trickle to our users over the course of years (think of Ubuntu and Debian, both of which have stable versions that ship relatively old versions of i3). We already have people who get confused by which version of the userguide applies to the version they’re running, and having to explain that a certain option was called X up until i3 version Y, when we renamed it, doesn’t make things any simpler :).

So, please don’t change the meaning of existing terminology in this case.

lotheac:
On Mon, Jan 04 2016 00:09:40 -0800, Michael Stapelberg wrote:
> While I agree that it would be better to have the terms mean what you suggested, the stronger argument to me is that changes in that area are very confusing. New versions of i3 trickle to our users over the course of years (think of Ubuntu and Debian, both of which have stable versions that ship relatively old versions of i3). We already have people who get confused by which version of the userguide applies to the version they’re running, and having to explain that a certain option was called X up until i3 version Y, when we renamed it, doesn’t make things any simpler :).
> 
> So, please don’t change the meaning of existing terminology in this case.

Of course it's up to you what is done, but now I have no idea what it is
you want to do. From your comments I gather that you want me, while
solving the problem I set out to solve (being able to color window
decoration borders differently from decoration title backgrounds), to
also:

 - not change existing behavior
 - not change existing documentation, even when that does not affect the
   way things are configured (this would be purely a clarification in
   the documentation, because the colors are not referred to by name in
   configuration)
 - come up with a different name for the window decoration borders to
   use in the documentation, because "border" is already used for
   something else

All this really stands in the way of improving things in my opinion. I
understand not wanting to change behavior (although you still have not
given me a definitive answer about whether or not you think that's OK
here, so I'm assuming), but not wanting to improve documentation I
really don't understand. How do you even add features if you can't put
them in the user guide because older versions don't support them?
Solving that problem is not really related to this change, but it's
easy: tell users to read the documentation shipped with their copy of
the software, and if you include online resource references in manpages
or elsewhere, make those references point to the same version of the
documentation as the software.

If I'm mistaken in any of the above, please tell me exactly what it is
you want to be done in order to get this integrated, because I don't
know. And please pardon my tone, it's just that this is getting a bit
frustrating.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

stapelberg:
I’m not replying in detail to the points you raised, because I think we shouldn’t get lost in detail discussion, but rather focus on what you want to accomplish.

Currently, the userguide says:

    <colorclass> <border> <background> <text> <indicator>

You suggested:

    <colorclass> <decoration_separator> <decoration_background> <text> <indicator> <decoration_border>

And I’m saying the change in terminology (from “border” to “separator”) is what we want to avoid.

So, how about we change it like this?

    <colorclass> <border> <background> <text> <indicator> <decoration_border>

lotheac:
On Mon, Jan 04 2016 00:58:40 -0800, Michael Stapelberg wrote:
> Currently, the userguide says:
> 
>     <colorclass> <border> <background> <text> <indicator>
> 
> You suggested:
> 
>     <colorclass> <decoration_separator> <decoration_background> <text> <indicator> <decoration_border>
> 
> And I’m saying the change in terminology (from “border” to “separator”) is what we want to avoid.
> 
> So, how about we change it like this?
> 
>     <colorclass> <border> <background> <text> <indicator> <decoration_border>

I have to say I think this is a missed opportunity for a much-needed
clarification of the documentation in this area, and perhaps even an
over-engineered solution, but I have implemented it this way now.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

stapelberg:
Can you also add a positive test please? Currently, the tests only verify that the option is NULL when not specified.

stapelberg:
The commit message needs an update as well to reflect the current situation.

docs/userguide
```diff
@@ -864,19 +864,19 @@ Colors are in HTML hex format (#rrggbb), see the following example:
...
+client.urgent           #2f343a #900000 #ffffff #900000   #900000
+client.placeholder      #000000 #0c0c0c #ffffff #000000   #0c0c0c
 
 client.background       #ffffff
 ---------------------------------------------------------
 
 Note that for the window decorations, the color around the child window is the
-background color, and the border color is only the two thin lines at the top of
-the window.
+'decoration_border', and 'border' color is only the two thin lines around the
```
stapelberg: Can you please use double quotes ("") for consistency with the rest of the userguide?

Airblader:
LGTM. While you're fixing @stapelberg's comment, please also rebase to the current `next`.

lotheac:
Nits fixed and rebased.

Airblader:
@lotheac @stapelberg I actually just noticed that I think the new name is also misleading. I expected `decoration_border` to be the two lines above/beneath the titlebar, but it now is the child window's border.

I think we should rename it to `window_border` or `child_border`. Opinions?

lotheac:
On Thu, Jan 07 2016 13:12:07 -0800, Ingo Bürk wrote:
> @lotheac @stapelberg I actually just noticed that I think the new name is also misleading. I expected `decoration_border` to be the two lines above/beneath the titlebar, but it now is the child window's border.

I followed what @stapelerg thought should be done. He explicitly did not
want the two lines called "border" renamed.

> I think we should rename it to `window_border` or `child_border`. Opinions?

I suggested "border" earlier, but these are fine by me as well.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

stapelberg:
I think child_border is the more descriptive one.

Airblader:
@lotheac Yeah, not renaming the existing one is fine. I just think the newly introduced one should have a better name. That was my fault, not yours, though, so I apologize. Renaming isn't a big issue, though. Especially since we haven't released this yet.

@stapelberg Alright. I'll open an issue for it now so we don't forget about it.

Issue #1814, opened 2015-07-27
Add a "split toggle" command


teto:
Hi,
I've just learned of the command "layout toggle" which allows me to merge the default 3 mappings ("stack", "tabbed", "default") into just one (wonderful).

I would like to merge the next 2 mappings taken from /etc/i3/config 
```
# split in horizontal orientation
bindsym Mod1+h split h

# split in vertical orientation
bindsym Mod1+v split v
```
into "split toggle" for the same reason (spare one mapping) but the command does not seem to exist, hence I vote for its addition.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
Are you maybe looking for `layout toggle split`?

stapelberg:
What @Airblader just said :). See also https://github.com/i3/i3/blob/next/parser-specs/commands.spec for full details about which commands are available, and feel free to send improvements of the userguide, since that should clarify questions like these.

teto:
This does not have the same effect. For instance if you have a stack of 3
windows and do "layout toggle split" then all your windows get displayed
side by side while I just want to create the next window next to the
current one. So for instance in one floor of the stack you would have 2
windows side by side (
https://transfer.sh/XBmmd/2015-07-27-221831-1920x1080-scrot.png)

2015-07-27 22:15 GMT+02:00 Michael Stapelberg <notifications@github.com>:

Airblader:
Yeah, that's why I was asking before closing the issue (*looks at Michael* :wink: ). Reopening for now to let Michael decide on it.

stapelberg:
Ah, I see. Yeah, I suppose adding a toggle parameter to the split command makes sense.

teto:
Hi,
I would like to implement this and would like some directions about the API. I tried a naive implementation and added a `static orientation_t cmd_split_get_orientation(struct Con *con, char *direction)` used by :
```
void cmd_split(I3_CMD, char *direction)
/* Implementation of 'split v|h|vertical|horizontal|toggle'. */
```
and while calling i3-msg split toggle works the first time, it doesn't afterwards, I suppose it is because tree_split creates a new container. Thus I should modify tree_split

stapelberg:
[...]

PR #2139, opened 2015-12-29
added "toggle" option to "split" command

jolange:
as requested in #1814.
I think this implements what @teto requests.

I chose this default behaviour, but it's quite a random choice:
In case the parent layout is neither `L_SPLITH` nor `L_SPLITV`, the split is done horizontally.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1228,7 +1228,7 @@ void cmd_move_workspace_to_output(I3_CMD, const char *name) {
 }
 
 /*
- * Implementation of 'split v|h|vertical|horizontal'.
+ * Implementation of 'split v|h|t|vertical|horizontal|toggle'.
  *
  */
 void cmd_split(I3_CMD, const char *direction) {
@@ -1243,7 +1243,22 @@ void cmd_split(I3_CMD, const char *direction) {
         }
 
         DLOG("matching: %p / %s\n", current->con, current->con->name);
-        tree_split(current->con, (direction[0] == 'v' ? VERT : HORIZ));
+        if (direction[0] == 't') {
+            layout_t current_layout;
+            if (current->con->type == CT_WORKSPACE) {
+                current_layout = current->con->layout;
+            } else {
+                current_layout = current->con->parent->layout;
+            }
+            /* toggling split orientation */
+            if (current_layout == L_SPLITH) {
+                tree_split(current->con, VERT);
+            } else {
+                tree_split(current->con, HORIZ);
+            }
+        } else {
+            tree_split(current->con, (direction[0] == 'v' ? VERT : HORIZ));
+        }
     }
 
     cmd_output->needs_tree_render = true;
--- a/docs/userguide
+++ b/docs/userguide
@@ -1750,20 +1750,24 @@ get placed below the current one (splitv).
 
 If you apply this command to a split container with the same orientation,
 nothing will happen. If you use a different orientation, the split container’s
-orientation will be changed (if it does not have more than one window). Use
-+layout toggle split+ to change the layout of any split container from splitv
-to splith or vice-versa.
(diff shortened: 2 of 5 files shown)


Airblader:
Please also update the userguide accordingly.

jolange:
Sure, done.

src/commands.c
```diff
@@ -1243,7 +1243,14 @@ void cmd_split(I3_CMD, const char *direction) {
         }
 
         DLOG("matching: %p / %s\n", current->con, current->con->name);
-        tree_split(current->con, (direction[0] == 'v' ? VERT : HORIZ));
+        if (direction[0] == 't') {
+            /* toggling split orientation */
+            if (current->con->parent->layout == L_SPLITH)
+                tree_split(current->con, VERT);
+            else /* default: */
```
stapelberg: nit: the `/* default: */` comment is not consistent with our coding style and doesn’t make the code any clearer I think. Could you remove it please?

Also, please add braces, even though the instructions are just one line.

jolange: Ok. I had braces first, but found that in most places in the code they were ommited for single instructions.

stapelberg:
Thanks. Could you also add a test case (see http://i3wm.org/docs/testsuite.html) please, so that the intentended behavior doesn’t break in future refactorings?

When you’re done, please squash your commits into logical units. We don’t want the pull request history to end up in the git history.

jolange:
While writing testcases I found that toggling back and forth directly on a workspace did not work (fixed now).
I added the tests to the existing `122-split.t` module, which I hope makes sense.
Everything is squashed into one commit now.

src/commands.c
```diff
@@ -1243,7 +1243,21 @@ void cmd_split(I3_CMD, const char *direction) {
         }
 
         DLOG("matching: %p / %s\n", current->con, current->con->name);
-        tree_split(current->con, (direction[0] == 'v' ? VERT : HORIZ));
+        if (direction[0] == 't') {
+            /* find current layout:
+             *   for workspace, use workspace's layout
+             *   for other containers, use parent's layout
+             */
+            layout_t current_layout = current->con->type == CT_WORKSPACE ? current->con->layout : current->con->parent->layout;
```
Airblader: I'm on mobile but I wonder if we don't already have a function that returns this information in con.c (?)

Airblader:
What happens if I use this on the parent container of a floating window (the container it is embedded  in)?

src/commands.c
jolange: I did not find any such function there, only [this](https://github.com/i3/i3/blob/98875fda762142419d557e7d99dde39eaa52f4bf/src/con.c#L1558-L1559)  similar case.

jolange:
If you are focusing a floating window and use "focus parent", you can toggle that back and forth.

If you are focusing the floating window itself and use toggle, nothing happens, due to the implementation of `tree_split`, just as with `split h|v`. I would consider this consistent with the current behaviour.

src/commands.c
Airblader: OK. :)

Airblader:
Sounds good.

src/commands.c
stapelberg: Given that this line is pretty wide, could you please favor an `if` statement over the ternary operator in this case? I.e.:

```c
layout_t current_layout;
if (current->con->type == CT_WORKSPACE) {
  current_layout = current->con->layout;
} else {
  current_layout = current->con->parent->layout;
}
```

I think this also makes the comment above obsolete, as it just repeats what the code already conveys.

stapelberg:
Looks good overall.

src/commands.c
jolange: Ok, I'll update this and rebase.

PR #2143, opened 2015-12-31
Allow title_format for all containers

### LINKED ISSUE #2120: title_format doesn't change top level tab title
i3 version 4.11 (2015-09-30, branch "4.11")

When using title format in tabbed mode, with multiple tabs and multiple windows under those tabs, I cannot change the top level tab name.

For example,  doing mod+a to select all windows under a tab, and then calling title_format, I expect that tab to have its title changed.

Airblader:
fixes #2120

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2215,7 +2215,10 @@ https://developer.gnome.org/pango/stable/PangoMarkupFormat.html[Pango markup]
 and the following placeholders which will be replaced:
 
 +%title+::
-    The X11 window title (_NET_WM_NAME or WM_NAME as fallback).
+    For normal windows, this is the X11 window title (_NET_WM_NAME or WM_NAME
+    as fallback). When used on containers without a window (e.g., a split
+    container inside a tabbed/stacked layout), this will be the tree
+    representation of the container (e.g., "H[xterm xterm]").
 +%class+::
     The X11 window class (second part of WM_CLASS). This corresponds to the
     +class+ criterion, see <<command_criteria>>.
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -504,6 +504,20 @@ char *get_config_path(const char *override_configpath, bool use_system_paths);
 int mkdirp(const char *path, mode_t mode);
 #endif
 
+/** Helper structure for usage in format_placeholders(). */
+typedef struct placeholder_t {
+    /* The placeholder to be replaced, e.g., "%title". */
+    char *name;
+    /* The value this placeholder should be replaced with. */
+    char *value;
+} placeholder_t;
+
+/**
+ * Replaces occurences of the defined placeholders in the format string.
+ *
+ */
+char *format_placeholders(char *format, placeholder_t *placeholders, int num);
+
 #ifdef CAIRO_SUPPORT
 /* We need to flush cairo surfaces twice to avoid an assertion bug. See #1989
  * and https://bugs.freedesktop.org/show_bug.cgi?id=92455. */
--- a/libi3/format_placeholders.c
+++ b/libi3/format_placeholders.c
@@ -1,0 +1,67 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3 - an improved dynamic tiling window manager
+ * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
+ *
+ */
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "libi3.h"
+
+#ifndef STARTS_WITH
+#define STARTS_WITH(string, needle) (strncasecmp((string), (needle), strlen((needle))) == 0)
+#endif
+
+/*
+ * Replaces occurences of the defined placeholders in the format string.
+ *
+ */
+char *format_placeholders(char *format, placeholder_t *placeholders, int num) {
+    if (format == NULL)
+        return NULL;
+
+    /* We have to first iterate over the string to see how much buffer space
+     * we need to allocate. */
+    int buffer_len = strlen(format) + 1;
+    for (char *walk = format; *walk != '\0'; walk++) {
+        for (int i = 0; i < num; i++) {
+            if (!STARTS_WITH(walk, placeholders[i].name))
+                continue;
+
+            buffer_len = buffer_len - strlen(placeholders[i].name) + strlen(placeholders[i].value);
+            walk += strlen(placeholders[i].name) - 1;
+            break;
+        }
+    }
+
+    /* Now we can parse the format string. */
+    char buffer[buffer_len];
+    char *outwalk = buffer;
+    for (char *walk = format; *walk != '\0'; walk++) {
+        if (*walk != '%') {
+            *(outwalk++) = *walk;
+            continue;
+        }
+
+        bool matched = false;
+        for (int i = 0; i < num; i++) {
+            if (!STARTS_WITH(walk, placeholders[i].name)) {
+                continue;
+            }
+
+            matched = true;
+            outwalk += sprintf(outwalk, "%s", placeholders[i].value);
+            walk += strlen(placeholders[i].name) - 1;
+            break;
+        }
+
+        if (!matched)
+            *(outwalk++) = *walk;
+    }
+
+    *outwalk = '\0';
+    return sstrdup(buffer);
+}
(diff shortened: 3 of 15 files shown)


Airblader:
Not ready to be merged yet. I'll let you know.

src/con.c
```diff
@@ -1970,3 +1970,54 @@ char *con_get_tree_representation(Con *con) {
...
+            .name = "%class",
+            .value = ESCAPE_MARKUP(sstrdup(win->class_class))};
+        placeholders[2] = (placeholder_t){
+            .name = "%instance",
+            .value = ESCAPE_MARKUP(sstrdup(win->class_instance))};
+    }
+
+#undef ESCAPE_MARKUP
+
+    i3String *formatted = i3string_from_utf8(format_placeholders(con->title_format, &placeholders[0], 3));
```
Airblader: I think there's a memory leak here because we need to free the returned string from `format_placeholders`.

src/window.c
```diff
@@ -66,8 +66,13 @@ void window_update_name(i3Window *win, xcb_get_property_reply_t *prop, bool befo
     i3string_free(win->name);
     win->name = i3string_from_utf8_with_length(xcb_get_property_value(prop),
                                                xcb_get_property_value_length(prop));
-    if (win->title_format != NULL)
-        ewmh_update_visible_name(win->id, i3string_as_utf8(window_parse_title_format(win)));
```
Airblader: I think this was a memory leak before.

src/commands.c
```diff
@@ -1917,27 +1917,30 @@ void cmd_title_format(I3_CMD, const char *format) {
...
+                /* We can remove _NET_WM_VISIBLE_NAME since we don't display a custom title. */
+                ewmh_update_visible_name(current->con->window->id, NULL);
+            }
         }
 
-        /* Make sure the window title is redrawn immediately. */
-        current->con->window->name_x_changed = true;
+        if (current->con->window != NULL) {
+            /* Make sure the window title is redrawn immediately. */
+            current->con->window->name_x_changed = true;
```
Airblader: For windowless containers we should free the deco render params here to force the redraw.

Airblader:
Userguide and test are missing.

Airblader:
@stapelberg OK, ready.

src/con.c
```diff
@@ -2000,3 +2000,56 @@ char *con_get_tree_representation(Con *con) {
...
+i3String *con_parse_title_format(Con *con) {
+    assert(con->title_format != NULL);
+
+    i3Window *win = con->window;
+
+    /* We need to ensure that we only escape the window title if pango
+     * is used by the current font. */
+    const bool pango_markup = font_is_pango();
+
+#define ESCAPE_MARKUP(str) ((pango_markup) ? g_markup_escape_text(str, -1) : str)
```
Airblader: I think usages of this are all memory leaks because `g_markup_escape_text` returns a newly allocated string. I'll follow up on it.

Airblader:
@stapelberg OK, I've fixed the memory leak. Since there seem to be multiple leaks on master that are fixed with this PR, we should think about delivering this in the next bugfix, even though it's a feature extension.

include/libi3.h
```diff
@@ -504,6 +504,20 @@ char *get_config_path(const char *override_configpath, bool use_system_paths);
 int mkdirp(const char *path, mode_t mode);
 #endif
 
+/** Helper structure for usage in format_placeholders(). */
```
stapelberg: Why did you factor out `format_placeholders`? There’s only one call site currently?

libi3/format_placeholders.c
```diff
@@ -0,0 +1,67 @@
...
+ *
+ */
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "libi3.h"
+
+#ifndef STARTS_WITH
+#define STARTS_WITH(string, needle) (strncasecmp(string, needle, strlen(needle)) == 0)
```
stapelberg: nit: Use `(string)` and `(needle)` within the macro contents.

See https://en.wikipedia.org/wiki/C_preprocessor#Macro_definition_and_expansion starting at “The second x is enclosed in its own pair of parentheses”.

src/con.c
```diff
@@ -2000,3 +2000,47 @@ char *con_get_tree_representation(Con *con) {
 
     return complete_buf;
 }
+
+/*
+ * Returns the container's title considering the current title format.
+ * If no format is set, this will simply return the container's name.
```
stapelberg: Is this line still accurate? In the function body, there’s `assert(con->title_format != NULL)`, so I don’t see how no format can be set.

src/window.c
stapelberg: Can you send the fix for the memory leak in a separate PR, so that we can separate merge it into master please?

docs/userguide
```diff
@@ -2211,7 +2211,9 @@ https://developer.gnome.org/pango/stable/PangoMarkupFormat.html[Pango markup]
 and the following placeholders which will be replaced:
 
 +%title+::
-    The X11 window title (_NET_WM_NAME or WM_NAME as fallback).
+    The X11 window title (_NET_WM_NAME or WM_NAME as fallback). If used on
+    windowless containers (split containers within a tabbed/stacked container),
+    this will be the tree representation of the children.
```
stapelberg: From reading this, I’m not 100% certain what happens.

1. I think we should give an example for the tree representation, possibly something like `(e.g. the “i3: V[urxvt firefox]” description of a vsplit container within a stacked container)`

2. Does `%title` contain “urxvt firefox” or “i3: V[urxvt firefox]“ in the example above?

include/libi3.h
Airblader: Because even with only one usage this is cleaner than one big function :) And it seems like the right kind of logical packet to be refactoring. 

Another thing I had in mind is that we use this in i3status and while we have no directly shared code between them, we might be able to reuse it there.

libi3/format_placeholders.c
Airblader: I had copied this from another place in the code, so I'll change it in both places.

docs/userguide
Airblader: It contains "V[urxvt firefox]", without the i3, which I figured makes more sense to the user. I'll update the description to be clearer.

src/con.c
Airblader: Yeah, it's wrong now. Willdo.

Airblader:
@stapelberg I've fixed everything you mentioned. As for the memory leaks, there were several ones. I can send a separate PR doing nothing but fixing those (it's like 3–4, I think) and then rebase this PR, though the rebase will be a simple `--ours`, omitting the fix on master since it's fixed here anyway.

Airblader:
See #2148 .

src/con.c
```diff
@@ -2000,3 +2000,46 @@ char *con_get_tree_representation(Con *con) {
...
+        title = pango_escape_markup(con_get_tree_representation(con));
+        class = sstrdup("i3-frame");
+        instance = sstrdup("i3-frame");
+    } else {
+        title = pango_escape_markup(sstrdup((win->name == NULL) ? "" : i3string_as_utf8(win->name)));
+        class = pango_escape_markup(sstrdup((win->class_class == NULL) ? "" : win->class_class));
+        instance = pango_escape_markup(sstrdup((win->class_instance == NULL) ? "" : win->class_instance));
+    }
+
+    placeholder_t placeholders[3];
```
stapelberg: The following is slightly more readable and easier to extend:

```c
const placeholder_t placeholders[] = {
  {.name = "%title", .value = title},
  // …
};
```

src/con.c
```diff
@@ -2000,3 +2000,46 @@ char *con_get_tree_representation(Con *con) {
...
+    placeholders[0] = (placeholder_t){.name = "%title", .value = title};
+    placeholders[1] = (placeholder_t){.name = "%class", .value = class};
+    placeholders[2] = (placeholder_t){.name = "%instance", .value = instance};
+
+    char *formatted_str = format_placeholders(con->title_format, &placeholders[0], 3);
```
stapelberg: replace the magic number 3 by `sizeof(placeholders) / sizeof(placeholder_t)` as you did below.

src/con.c
Airblader: `const` cannot be used here, but I'll do the rest.

Airblader:
@stapelberg Done and rebased.

PR #2145, opened 2016-01-02
Fix segfault when calling "i3 -C".

### LINKED ISSUE #2144: "-C" switch causes segfault
`i3 -C` causes a segfault because of 287a0b4c3cc24a18f5063b7d5743f0b85d6852a1

@Alexis211 FYI.

i3 4.11

Airblader:
Commit 287a0b4 introduced a segfault when validating the i3 config
as the root_screen will not be set in this case, causing a null
pointer dereference.

fixes #2144

### DIFF
--- a/testcases/complete-run.pl
+++ b/testcases/complete-run.pl
@@ -30,6 +30,7 @@ BEGIN {
 
 # these are shipped with the testsuite
 use lib $dirname . 'lib';
+use i3test::Util qw(slurp);
 use StartXServer;
 use StatusLine;
 use TestWorker;
@@ -156,7 +157,7 @@ for my $display (@displays) {
 
 # Read previous timing information, if available. We will be able to roughly
 # predict the test duration and schedule a good order for the tests.
-my $timingsjson = StartXServer::slurp('.last_run_timings.json');
+my $timingsjson = slurp('.last_run_timings.json') if -e '.last_run_timings.json';
 %timings = %{decode_json($timingsjson)} if length($timingsjson) > 0;
 
 # Re-order the files so that those which took the longest time in the previous
@@ -245,7 +246,7 @@ printf("\t%s with %.2f seconds\n", $_, $timings{$_})
 if ($numtests == 1) {
     say '';
     say 'Test output:';
-    say StartXServer::slurp($logfile);
+    say slurp($logfile);
 }
 
 END { cleanup() }
--- a/testcases/lib/SocketActivation.pm
+++ b/testcases/lib/SocketActivation.pm
@@ -96,8 +96,13 @@ sub activate_i3 {
         # the interactive signalhandler to make it crash immediately instead.
         # Also disable logging to SHM since we redirect the logs anyways.
         # Force Xinerama because we use Xdmx for multi-monitor tests.
-        my $i3cmd = abs_path("../i3") . q| -V -d all --disable-signalhandler| .
-                                        q| --shmlog-size=0 --force-xinerama|;
+        my $i3cmd = abs_path("../i3") . q| --shmlog-size=0 --disable-signalhandler --force-xinerama|;
+        if (!$args{validate_config}) {
+            # We only set logging if i3 is actually started, but not if we only
+            # validate the config file. This is to keep logging to a minimum as
+            # such a test will likely want to inspect the log file.
+            $i3cmd .= q| -V -d all|;
+        }
 
         # For convenience:
         my $outdir = $args{outdir};
@@ -105,6 +110,10 @@ sub activate_i3 {
 
         if ($args{restart}) {
             $i3cmd .= ' -L ' . abs_path('restart-state.golden');
+        }
+
+        if ($args{validate_config}) {
+            $i3cmd .= ' -C';
         }
 
         if ($args{valgrind}) {
@@ -148,6 +157,11 @@ sub activate_i3 {
     # close the socket, the child process should be the only one which keeps a file
     # descriptor on the listening socket.
     $socket->close;
+
+    if ($args{validate_config}) {
+        $args{cv}->send(1);
+        return $pid;
+    }
 
     # We now connect (will succeed immediately) and send a request afterwards.
     # As soon as the reply is there, i3 is considered ready.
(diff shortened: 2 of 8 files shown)


Airblader:
Putting it on `waiting` until I've also added a test.

Airblader:
OK, I've added a test now.

Airblader:
@stapelberg And now with passing tests.

testcases/lib/SocketActivation.pm
```diff
@@ -96,8 +96,12 @@ sub activate_i3 {
         # the interactive signalhandler to make it crash immediately instead.
         # Also disable logging to SHM since we redirect the logs anyways.
         # Force Xinerama because we use Xdmx for multi-monitor tests.
-        my $i3cmd = abs_path("../i3") . q| -V -d all --disable-signalhandler| .
-                                        q| --shmlog-size=0 --force-xinerama|;
+        my $i3cmd = abs_path("../i3") . q| --shmlog-size=0 --disable-signalhandler|;
+        if (!$args{validate}) {
+            # We only do these if i3 is actually started, but not if -C is passed,
+            # since this contains nothing useful for us in this case.
+            $i3cmd .= q| -V -d all --force-xinerama|;
```
Airblader: The reasoning for putting `--force-xinerama` in here is that it causes an error to be logged.

testcases/t/262-config-validation.t
```diff
@@ -0,0 +1,37 @@
...
+
+invalid
+EOT
+
+my $pid = launch_with_config($config, validate => 1);
+my $log = get_i3_log;
+
+# We don't care so much about the output (there are tests for this), but rather
+# that we got correct output at all instead of, e.g., a segfault.
+ok($log =~ /Expected one of these tokens/, 'an invalid config token was found');
```
Airblader: I'll admit that this isn't perfect, but at least an improvement.

testcases/lib/SocketActivation.pm
stapelberg: Can you make the comment a bit more specific? I.e., can you replace “do these” and “this” with what they refer to? I’m having a bit of trouble following them from just glancing over the diff, and I think that’s the level of understanding that we should aim for with commits :).

testcases/lib/SocketActivation.pm
```diff
@@ -96,8 +96,12 @@ sub activate_i3 {
         # the interactive signalhandler to make it crash immediately instead.
         # Also disable logging to SHM since we redirect the logs anyways.
         # Force Xinerama because we use Xdmx for multi-monitor tests.
-        my $i3cmd = abs_path("../i3") . q| -V -d all --disable-signalhandler| .
-                                        q| --shmlog-size=0 --force-xinerama|;
+        my $i3cmd = abs_path("../i3") . q| --shmlog-size=0 --disable-signalhandler|;
+        if (!$args{validate}) {
```
stapelberg: Maybe s/validate/validate_config/ for clarity?

testcases/lib/i3test.pm
```diff
@@ -857,8 +859,21 @@ sub launch_with_config {
...
+    # If we called i3 with -C, we wait for it to exit and then return as
+    # there's nothing else we need to do.
+    if ($args{validate}) {
+        $cv->recv;
+        waitpid $i3_pid, 0;
+
+        # We need this since exit_gracefully will not be called in this case.
+        undef $i3_pid;
+
+        return undef;
```
stapelberg: Wouldn’t it make sense to return `$?` (or `${^CHILD_ERROR_NATIVE}` if you prefer the expressive forms) so that we can check the exit code as well?

testcases/lib/i3test.pm
```diff
@@ -871,6 +886,23 @@ sub launch_with_config {
...
 }
 
+=head2 get_i3_log
+
+Returns the content of the log file for the current test.
+
+=cut
+sub get_i3_log {
+    my $logfile = "$ENV{OUTDIR}/i3-log-for-$ENV{TESTNAME}";
+    my $content = do {
```
stapelberg: Given that this is the third definition of `slurp` in our code base (see testcases/t/171-config-migrate.t and testcases/lib/StartXServer.pm), how about we properly define `slurp` in i3test once and refer to it? :)

Airblader:
@stapelberg Done.

testcases/lib/SocketActivation.pm
```diff
@@ -96,8 +96,14 @@ sub activate_i3 {
         # the interactive signalhandler to make it crash immediately instead.
         # Also disable logging to SHM since we redirect the logs anyways.
         # Force Xinerama because we use Xdmx for multi-monitor tests.
-        my $i3cmd = abs_path("../i3") . q| -V -d all --disable-signalhandler| .
-                                        q| --shmlog-size=0 --force-xinerama|;
+        my $i3cmd = abs_path("../i3") . q| --shmlog-size=0 --disable-signalhandler|;
+        if (!$args{validate_config}) {
+            # We only set logging and force xinerama if i3 is actually started,
```
stapelberg: I understand the rationale for logging, but isn’t the log file bigger if we don’t specify --force-xinerama, because it will contain an error message about RandR being absent?

testcases/complete-run.pl
```diff
@@ -30,6 +30,7 @@ BEGIN
 
 # these are shipped with the testsuite
 use lib $dirname . 'lib';
+use i3test::Test;
```
stapelberg: nit: here and below, use `use i3test::Test qw(slurp);` so that only the `slurp` function is imported.

stapelberg:
Looks good overall.

testcases/complete-run.pl
Airblader: I tried, but then I get this:

````
plan() doesn't understand slurp at ./complete-run.pl line 33.
BEGIN failed--compilation aborted at ./complete-run.pl line 33.
```

Airblader:
@stapelberg Updated the xinerama thing. See my comment for the `qw` point.

testcases/complete-run.pl
stapelberg: Indeed. I think that’s because the package is based on Test::Builder::Module. It seems like we don’t have a better place already set up, so how about putting the slurp function into a new `testcases/lib/i3util.pm`?

Airblader:
@stapelberg Done.

PR #2161, opened 2016-01-14
Add pledge support for OpenBSD

botovq:
[pledge(2)](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/pledge.2?query=pledge) is a new system call in OpenBSD that forces a process into restricted-service operation mode. The following three small patches adding pledge support to i3, i3-msg and i3-nagbar [have been tested](https://marc.info/?l=openbsd-ports-cvs&m=145069308429529&w=2) on OpenBSD-current for a few weeks now. See also the [discussion on the mailing list](https://marc.info/?t=145061676600001&r=1&w=2).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -802,6 +802,11 @@ int main(int argc, char *argv[]) {
         xcb_free_pixmap(conn, pixmap);
     }
 
+#if defined(__OpenBSD__)
+    if (pledge("stdio rpath wpath cpath proc exec unix", NULL) == -1)
+        err(EXIT_FAILURE, "pledge");
+#endif
+
     struct sigaction action;
 
     action.sa_sigaction = handle_signal;
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -119,6 +119,10 @@ static yajl_callbacks reply_callbacks = {
 };
 
 int main(int argc, char *argv[]) {
+#if defined(__OpenBSD__)
+    if (pledge("stdio rpath unix", NULL) == -1)
+        err(EXIT_FAILURE, "pledge");
+#endif
     char *env_socket_path = getenv("I3SOCK");
     if (env_socket_path)
         socket_path = sstrdup(env_socket_path);
--- a/i3-nagbar/main.c
+++ b/i3-nagbar/main.c
@@ -470,6 +470,11 @@ int main(int argc, char *argv[]) {
     font = load_font(pattern, true);
     set_font(&font);
 
+#if defined(__OpenBSD__)
+    if (pledge("stdio rpath wpath cpath getpw proc exec", NULL) == -1)
+        err(EXIT_FAILURE, "pledge");
+#endif
+
     xcb_rectangle_t win_pos = get_window_position();
 
     xcb_cursor_t cursor;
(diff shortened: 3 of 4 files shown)


Airblader:
The PR is somewhat broken. Those merge commits shouldn't be in there.

botovq:
Sorry about that, I only use git/github once in a blue moon. What is the best way for me to remove them?

botovq:
Sorry about that.  I removed the merge commits and squashed the three remaining commits. Is this better?

Airblader:
Yeah, the PR looks fine now. As for the actual change, we'll see what @stapelberg says.

src/main.c
```diff
@@ -802,6 +802,13 @@ int main(int argc, char *argv[]) {
         xcb_free_pixmap(conn, pixmap);
     }
 
+#if defined(__OpenBSD__)
+    /* Unable to use pledge(2) with shm* system calls. */
+    if (shmlog_size == 0)
```
stapelberg: So this breaks the “shmlog” command, with which you can enable debug logging during runtime, correct?

botovq: Indeed, this is correct. "i3-msg 'shmlog on'" will crash i3. Technically it's not the shm_open call but rather the 'sysconf(_SC_PHYS_PAGES)' in src/log.c that will be blocked.

stapelberg: So, how can we avoid that? One solution would be to not use pledge() in the i3 binary itself. Any other ideas?

botovq: Let me be frank: If we decide to not use pledge in the i3 binary itself then I'd say we'd better drop the entire patch.  It seems pointless to just do this for the short-lived i3-msg and i3-nagbar.

But before withdrawing the patch, I'd like to discuss whether we should consider making a trade-off here. How important it is to have the possibility of enabling shmlog at runtime? Would we be willing to trade the shmlog feature for pledge on OpenBSD? After all, nobody ran into this in a number of weeks, so it seems to be a feature that isn't used all that often.

stapelberg: Enabling shmlog at runtime is very useful when debugging, and I’d hate it if OpenBSD users were at a disadvantage when reporting issues with i3.

Given that the attack surface is rather small, I think I’d prefer not having `pledge` support over being unable to debug issues at runtime.

botovq: I see. I pointed the maintainer dcoppa to this discussion, and I'd appreciate if we could give him a chance to chime in before closing the matter entirely, but I don't see a way of having both things at the same time and I understand your way of weighing priorities.

semarie:
@botovq  does the current patch disable `shmlog` under OpenBSD or would crash with the current diff ? Reading it, it seems it is just that pledge(2) isn't enforced if `shmlog` is in use, isn't it ? Does I miss something ?

botovq:
@stapelberg could you please check whether this updated version is acceptable? @semarie figured out a way that doesn't break enabling shmlog at runtime and that allows us to add pledge calls unconditionally. I was under the mistaken impression that there was a shm_at(2) call hidden later on in i3 itself, but this turned out not to be the case.

stapelberg:
This looks fine to me. Thanks!

PR #2162, opened 2016-01-15
Handle the EWMH atom _NET_WM_DESKTOP.

### LINKED ISSUE #2153: Support _NET_WM_DESKTOP correctly
We currently have `_NET_WM_DESKTOP` in `_NET_SUPPORTED`. This seems to have been the case since around 2009 (1a0817eb3912f25b75cfe61854920c52c78493d2), although I don't really see the reason for it even in that commit.

As a `git grep NET_WM_DESKTOP` reveals, we don't do *anything* with this atom – we don't respect it upon map requests, we don't keep it updated and we don't respond to client messages, all of which should be the case according to [the spec](http://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472635472).

We discussed before in #1507 and #938 that we don't want to support this atom, so we should also not claim to support it in the first place – unless we have a good reason to. @stapelberg Are you aware of any reasons?

However, I'd also like to propose supporting it instead. At least giving it (another) shot. I think we can have a canonical way to enumerate workspaces (even only-named ones) according to the spec and we do have sticky windows now, so we could respect the special value by floating the window and making it sticky.

I'd only actually do this, though, if we can verify that proper support of this atom makes pagers (like gnome-panel) work correctly, see [this post](https://www.reddit.com/r/i3wm/comments/402e2v/i3_workspaces_munged_together_by_desktop_tools/?ref=share&ref_source=link) and similar requests. Otherwise we shouldn't worry about the effort. @stapelberg Thoughts?

Airblader:
We already claim _NET_WM_DESKTOP support in _NET_SUPPORTED since around 2009,
but haven't actually done anything with it. However, especially pagers like
gnome-panel rely on this property to be updated and many tools, like GTK, want
to use the corresponding client messages to make a window sticky, move it
around etc.

This patch implements full support according to the EWMH spec. This means:

* We set the property on all windows when managing it.
* We keep the property updated on all windows at all times.
* We read and respect the property upon managing a window if it was set before
  mapping the window.
* We react to client messages for it.
* We remove the property on withdrawn windows.

Note that the special value 0xFFFFFFFF, according to the spec, means that the
window shall be shown on all workspaces. We do this by making it sticky and
float it. This shows it on all workspaces at least on the output it is on.

Furthermore, the spec gives us the freedom to ignore _NET_WM_DESKTOP when
managing a window if we have good reason to. In our case, we give window
swallowing a higher priority since the user would likely expect that and we
want to keep placeholder windows only around for as long as we have to.
However, we do prioritize this property over, for example, startup
notifications.

fixes #2153

### DIFF
--- a/include/ewmh.h
+++ b/include/ewmh.h
@@ -37,6 +37,13 @@ void ewmh_update_desktop_names(void);
 void ewmh_update_desktop_viewport(void);
 
 /**
+ * Updates _NET_WM_DESKTOP for all windows.
+ * A request will only be made if the cached value differs from the calculated value.
+ *
+ */
+void ewmh_update_wm_desktop(void);
+
+/**
  * Updates _NET_ACTIVE_WINDOW with the currently focused window.
  *
  * EWMH: The window ID of the currently active window or None if no window has
@@ -96,3 +103,21 @@ void ewmh_setup_hints(void);
  *
  */
 void ewmh_update_workarea(void);
+
+/**
+ * Returns the workspace container as enumerated by the EWMH desktop model.
+ * Returns NULL if no workspace could be found for the index.
+ *
+ * This is the reverse of ewmh_get_workspace_index.
+ *
+ */
+Con *ewmh_get_workspace_by_index(uint32_t idx);
+
+/**
+ * Returns the EWMH desktop index for the workspace the given container is on.
+ * Returns NET_WM_DESKTOP_NONE if the desktop index cannot be determined.
+ *
+ * This is the reverse of ewmh_get_workspace_by_index.
+ *
+ */
+uint32_t ewmh_get_workspace_index(Con *con);
--- a/include/workspace.h
+++ b/include/workspace.h
@@ -14,6 +14,14 @@
 #include "tree.h"
 #include "randr.h"
 
+/* We use NET_WM_DESKTOP_NONE for cases where we cannot determine the EWMH
+ * desktop index for a window. We cannot use a negative value like -1 since we
+ * need to use uint32_t as we actually need the full range of it. This is
+ * technically dangerous, but it's safe to assume that we will never have more
+ * than 4294967279 workspaces open at a time. */
+#define NET_WM_DESKTOP_NONE 0xFFFFFFF0
+#define NET_WM_DESKTOP_ALL 0xFFFFFFFF
+
 /**
  * Returns a pointer to the workspace with the given number (starting at 0),
  * creating the workspace if necessary (by allocating the necessary amount of
--- a/src/ewmh.c
+++ b/src/ewmh.c
@@ -21,24 +21,9 @@ xcb_window_t ewmh_window;
  *
  */
 void ewmh_update_current_desktop(void) {
-    Con *focused_ws = con_get_workspace(focused);
-    Con *output;
-    uint32_t idx = 0;
-    /* We count to get the index of this workspace because named workspaces
-     * don’t have the ->num property */
-    TAILQ_FOREACH(output, &(croot->nodes_head), nodes) {
-        Con *ws;
-        TAILQ_FOREACH(ws, &(output_get_content(output)->nodes_head), nodes) {
-            if (STARTS_WITH(ws->name, "__"))
-                continue;
-
-            if (ws == focused_ws) {
-                xcb_change_property(conn, XCB_PROP_MODE_REPLACE, root,
-                                    A__NET_CURRENT_DESKTOP, XCB_ATOM_CARDINAL, 32, 1, &idx);
-                return;
-            }
-            ++idx;
-        }
+    const uint32_t idx = ewmh_get_workspace_index(focused);
+    if (idx != NET_WM_DESKTOP_NONE) {
+        xcb_change_property(conn, XCB_PROP_MODE_REPLACE, root, A__NET_CURRENT_DESKTOP, XCB_ATOM_CARDINAL, 32, 1, &idx);
     }
 }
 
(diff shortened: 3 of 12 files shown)


Airblader:
Tests missing.

src/ewmh.c
```diff
@@ -138,6 +123,72 @@ void ewmh_update_desktop_viewport(void) {
...
+        xcb_delete_property(conn, window, A__NET_WM_DESKTOP);
+    }
+}
+
+/*
+ * Updates _NET_WM_DESKTOP for all windows.
+ * A request will only be made if the cached value differs from the calculated value.
+ *
+ */
+void ewmh_update_wm_desktop(void) {
```
Airblader: We could also do this in `src/x.c` like we do for `_NET_WM_STATE_HIDDEN`. We'd have to recurse outside of the recursion already in place there because the recursion here is aimed at keeping the loops efficient by looping over workspaces rather than windows (figuring out the desktop index for each window means looping over the workspaces again for each window).

Airblader: Instead of calling this in all the places where windows are "changed", we could also just call it in `x.c` instead (and maybe where we update the current desktop, just so there's not much of a delay). We do this for `_NET_WM_STATE_HIDDEN`, too, as it avoids the problem of potentially missing a code path.

However, we would probably still want it outside of the recursion in `x.c`. Doing this for every window means that for every window we loop over all workspaces to figure out its index. That has higher complexity than what I do now, which is work over the workspaces and then recurse all windows with that index. We're not talking crazy numbers here, of course, but it's a matter of principle. :)

Just let me know whether you prefer calling this in the appropriate places (as now) or once in `x.c` (x_push_changes or something).

Airblader:
@stapelberg OK, ready for review. Note that the change to the tree rendering is not in this yet.

include/ewmh.h
```diff
@@ -36,6 +36,13 @@ void ewmh_update_desktop_names(void);
  */
 void ewmh_update_desktop_viewport(void);
 
+/*
```
stapelberg: nit: `/**`

include/ewmh.h
```diff
@@ -96,3 +103,21 @@ void ewmh_setup_hints(void);
  *
  */
 void ewmh_update_workarea(void);
+
+/*
```
stapelberg: nit: `/**`

include/ewmh.h
```diff
@@ -96,3 +103,21 @@ void ewmh_setup_hints(void);
...
+ * Returns the workspace container as enumerated by the EWMH desktop model.
+ * Returns NULL if no workspace could be found for the index.
+ *
+ * This is the reverse of ewmh_get_workspace_index.
+ *
+ */
+Con *ewmh_get_workspace_by_index(uint32_t idx);
+
+/*
```
stapelberg: nit: `/**`

include/workspace.h
```diff
@@ -14,6 +14,14 @@
 #include "tree.h"
 #include "randr.h"
 
+/* We use this for cases where we cannot determine the EWMH desktop index for a
```
stapelberg: nit: s/this/NET_WM_DESKTOP_NONE/?

src/ewmh.c
```diff
@@ -138,6 +123,70 @@ void ewmh_update_desktop_viewport(void) {
...
+
+    Con *output;
+    TAILQ_FOREACH(output, &(croot->nodes_head), nodes) {
+        Con *workspace;
+        TAILQ_FOREACH(workspace, &(output_get_content(output)->nodes_head), nodes) {
+            if (con_is_internal(workspace))
+                continue;
+
+            ewmh_update_wm_desktop_recursively(workspace, desktop);
+            desktop++;
```
stapelberg: nit: for consistency with the two functions below, please make this `++desktop;`

src/ewmh.c
```diff
@@ -138,6 +123,70 @@ void ewmh_update_desktop_viewport(void) {
...
+        }
+    }
+
+    xcb_flush(conn);
```
stapelberg: I think the `xcb_flush(conn);` call should go outside this function. The caller might want to batch up this change with others. No other function in src/ewmh.c calls `xcb_flush`.

src/ewmh.c
stapelberg: I think calling it in the appropriate places is preferable, so let’s keep it as-is for now.

src/ewmh.c
```diff
@@ -138,6 +123,70 @@ void ewmh_update_desktop_viewport(void) {
...
 
+static void ewmh_update_wm_desktop_recursively(Con *con, const uint32_t desktop) {
+    /* Recursively call this to descend through the entire subtree. */
+    Con *child;
+    TAILQ_FOREACH(child, &(con->nodes_head), nodes) {
+        ewmh_update_wm_desktop_recursively(child, desktop);
+    }
+    /* If con is a workspace, we also need to go through the floating windows on it. */
+    if (con->type == CT_WORKSPACE) {
+        TAILQ_FOREACH(child, &(con->floating_head), nodes) {
```
stapelberg: s/nodes/floating_windows/

src/ewmh.c
```diff
@@ -138,6 +123,70 @@ void ewmh_update_desktop_viewport(void) {
...
+        }
+    }
+
+    if (!con_has_managed_window(con))
+        return;
+
+    xcb_window_t window = con->window->id;
+
+    uint32_t wm_desktop = desktop;
+    /* Sticky windows are only actually sticky when they are floating or inside a floating container.
```
stapelberg: nit: here and below, it looks like the comment is not line-wrapped at 80 characters. Can you fix that please?

src/ewmh.c
```diff
@@ -138,6 +123,70 @@ void ewmh_update_desktop_viewport(void) {
...
+        wm_desktop = NET_WM_DESKTOP_ALL;
+    }
+
+    /* If this is the cached value, we don't need to do anything. */
+    if (con->window->wm_desktop == wm_desktop)
+        return;
+    con->window->wm_desktop = wm_desktop;
+
+    if (wm_desktop != NET_WM_DESKTOP_NONE) {
+        DLOG("Setting _NET_WM_DESKTOP = %d for window %d.\n", wm_desktop, window);
```
stapelberg: nit: for consistency with other log messages, please make it `for window 0x%08x`

src/ewmh.c
```diff
@@ -21,24 +21,9 @@ xcb_window_t ewmh_window;
...
-                continue;
-
-            if (ws == focused_ws) {
-                xcb_change_property(conn, XCB_PROP_MODE_REPLACE, root,
-                                    A__NET_CURRENT_DESKTOP, XCB_ATOM_CARDINAL, 32, 1, &idx);
-                return;
-            }
-            ++idx;
-        }
+    uint32_t idx = ewmh_get_workspace_index(focused);
```
stapelberg: nit: add `const`

src/ewmh.c
```diff
@@ -138,6 +123,70 @@ void ewmh_update_desktop_viewport(void) {
...
+    if (con->type == CT_WORKSPACE) {
+        TAILQ_FOREACH(child, &(con->floating_head), nodes) {
+            ewmh_update_wm_desktop_recursively(child, desktop);
+        }
+    }
+
+    if (!con_has_managed_window(con))
+        return;
+
+    xcb_window_t window = con->window->id;
```
stapelberg: nit: add `const`

src/handlers.c
```diff
@@ -839,32 +844,47 @@ static void handle_client_message(xcb_client_message_event_t *event) {
...
+        Con *con = con_by_window_id(event->window);
+        if (con == NULL) {
+            DLOG("Couldn't find con for window %d, ignoring the request.\n", event->window);
+            return;
+        }
 
-                    workspace_show(ws);
-                    tree_render();
+        if (index == NET_WM_DESKTOP_ALL) {
+            /* The window is requesting to be visible on all workspaces, so
```
stapelberg: Please add a `LOG` message here so that curious users can see why that happens.

src/manage.c
```diff
@@ -194,6 +195,16 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     char *startup_ws = startup_workspace_for_window(cwindow, startup_id_reply);
     DLOG("startup workspace = %s\n", startup_ws);
 
+    /* Get _NET_WM_DESKTOP if it was set. We use -1 otherwise since that's never a legal value. */
```
stapelberg: While the comment says that we set wm_desktop to -1, I can’t see code which does that? Is the comment obsolete, or is that an oversight in the code?

stapelberg:
Looks good overall.

src/ewmh.c
Airblader: We don't *really* need to call it anyway (others don't either). The rationale was that unlike other functions here where we update a single value, this function might in theory update all currently open windows. I'll remove it, though.

src/ewmh.c
Airblader: Oops.

Airblader:
@stapelberg Updated.

PR #2164, opened 2016-01-16
Update license year range to 2016

pra85:


stapelberg:
The fact that there’s a range in that file is an oversight. To make it consistent with what we did in commit 884214f14fdbd0a4a368d2a36d5e50324fa1d52a, can you please remove the range altogether?

pra85:
@stapelberg I have updated the pull request and removed the year range

PR #2169, opened 2016-01-20
Don't grab the buttons on the root window.

### LINKED ISSUE #2097: [enhancement] Run mouse bindings on the root window

### LINKED ISSUE #2168: Can't use the mouse after reloading the config
Hello,

Sorry to be here again. Every time I reload my config, I stop being able to right/left click and use touchpad gestures until I restart i3wm. I'm using the next branch. I've been able to track it down to commit 6f12f02 - Allow mouse bindings to run on the root window.

Have you experienced anything similar? I don't think it has anything to do with my config. I would be more than happy to provide more information.

Airblader:
We don't actually need to grab the buttons to fix #2097, but doing so
will cause a freeze due to unreleased events.
    
We partially revert 6f12f02 which avoids the freeze, but doesn't break
functionality.

relates to #2097
fixes #2168

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -164,10 +164,6 @@ void regrab_all_buttons(xcb_connection_t *conn) {
         xcb_grab_buttons(conn, con->window->id, grab_scrollwheel);
     }
 
-    /* Also grab the root window to allow bindings to work on there as well. */
-    xcb_ungrab_button(conn, XCB_BUTTON_INDEX_ANY, root, XCB_BUTTON_MASK_ANY);
-    xcb_grab_buttons(conn, root, grab_scrollwheel);
-
     xcb_ungrab_server(conn);
 }
 


Airblader:
This doesn't actually work quite yet.

Airblader:
OK, it works now, as far as I can tell.

stapelberg:
I haven’t completely understood this yet: how is the reload relevant for mouse button replaying? In the description, you say “during reloads”, so does this only affect reloads during which the user managed to click in the same time (sounds rather rare)?

Airblader:
My understanding of this is that with the reload, we grab the pointer. This is independent of whether the user clicked anything. However, with this active grab, when processing events on the root window, we now no longer release those events, which blocks it from all clients. So we need to replay the events in this case just as we do for clicks onto "actual" windows.

The problem that this only occurs during reloads actually seems to be a bug of implementing the pointer grab on the root window. I think we only do this on config reloads, not initially when i3 is started. Which is why there we also don't see the freeze (but why mouse bindings on the root window probably won't work in that case either).

Airblader:
Actually we do seem to grab the root window (tests wouldn't be working otherwise anyway). I'm not yet sure where.

Airblader:
OK, I don't really understand where we grab the pointer on a fresh start of i3. Also, I think we need to use REPLAY only if we didn't run a binding. Otherwise we want to eat the event, as we do with bindings on non-root windows.

Airblader:
It seems that we don't need to grab the root window, ever. It seems to work fine without it. My best guess right now that this has to do with the EWMH support window focus patch, but I haven't verified it.

I'll look at this again tomorrow. If it confirms, all we need to do is stop grabbing the buttons on the root window.

Airblader:
It seems that we don't need to grab the root window, ever. So the problem sort of disappears. I think this is because when I wrote the patch, I first grabbed the root window buttons which seemed like a good idea. When it still didn't work, I noticed that bindings must be triggered in a different code path.

The problem is that afterwards I never cross-checked that grabbing was even necessary.

@stapelberg I've updated the PR.

Airblader:
Just FYI, the reason we don't need to grab it is because we already do. XCB_EVENT_MASK_BUTTON_PRESS is contained in ROOT_EVENT_MASK.

PR #2170, opened 2016-01-21
fix i3-nagbar invocation

dcoppa:
i3-nagbar already launch the command inside a terminal, thus the "pageraction" string in commands.c is wrong. While here, also fix the manpage example.


Airblader:
i3-nagbar launches it with /bin/sh, not $EDITOR, though, doesn't it?

dcoppa:
What about "i3-sensible-editor ~/.config/i3/config" then?

stapelberg:
Do you realize the code is commented out? :) I think it might be better to finally remove that code.

dcoppa:
omg! I badly need some days off, it seems... facepalm

The manpage fix is still valid though ;)

stapelberg:
Can you update the pull request to either not touch the commented-out code or remove it please?

dcoppa:
I've removed the commented-out code.

stapelberg:
Thanks! One more thing before I can merge this: can you please squash your commits? Currently, the pull request contains 5 commits, and we really need just 2 in our history (remove dead code, update manpage).

PR #2172, opened 2016-01-22
Don't ignore focus on _NET_WM_DESKTOP client messages.

Airblader:
Ignoring the focus leads to a broken focus if the pointer is, e.g.,
over i3bar at the point in time the window is moved by the client
message. It also causes i3bar's workspace display to break.

Thanks to GermainZ for reporting.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -881,7 +881,7 @@ static void handle_client_message(xcb_client_message_event_t *event) {
                 return;
             }
 
-            con_move_to_workspace(con, ws, false, false, true);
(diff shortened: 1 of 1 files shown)


src/handlers.c
```diff
@@ -881,7 +881,7 @@ static void handle_client_message(xcb_client_message_event_t *event) {
                 return;
             }
 
-            con_move_to_workspace(con, ws, false, false, true);
+            con_move_to_workspace(con, ws, true, false, false);
```
Airblader: (This is now the same as `move con to workspace`)

PR #2178, opened 2016-01-27
Allow dragging tiled windows with the mouse.

MForster:
A container can be dragged by the title bar to one of the four sides of
another container. That container will then be split either horizontally
or vertically.

This does not yet consider stacked/tabbed containers or dragging to
empty outputs.

### DIFF
--- a/include/move.h
+++ b/include/move.h
@@ -15,3 +15,13 @@
  *
  */
 void tree_move(Con *con, int direction);
+
+typedef enum { BEFORE,
+               AFTER } position_t;
+
+/**
+ * This function detaches 'con' from its parent and inserts it either before or
+ * after 'target'.
+ *
+ */
+void insert_con_into(Con *con, Con *target, position_t position);
--- a/include/move.h
+++ b/include/move.h
@@ -15,13 +15,3 @@
  *
  */
 void tree_move(Con *con, int direction);
-
-typedef enum { BEFORE,
-               AFTER } position_t;
-
-/**
- * This function detaches 'con' from its parent and inserts it either before or
- * after 'target'.
- *
- */
-void insert_con_into(Con *con, Con *target, position_t position);
--- a/src/click.c
+++ b/src/click.c
@@ -321,10 +321,14 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
             return 1;
     }
     /* 8: otherwise, check for border/decoration clicks and resize */
-    else if ((dest == CLICK_BORDER || dest == CLICK_DECORATION) &&
-             is_left_or_right_click) {
+    else if (dest == CLICK_BORDER && is_left_or_right_click) {
         DLOG("Trying to resize (tiling)\n");
         tiling_resize(con, event, dest);
+    }
+    /* 9: otherwise, check for dragging a tiled window */
+    else if (dest == CLICK_DECORATION && is_left_or_right_click) {
+        DLOG("Trying to drag (tiling)\n");
+        tiling_drag(con, event);
     }
 
 done:
--- a/src/click.c
+++ b/src/click.c
@@ -320,15 +320,15 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
         if (floating_mod_on_tiled_client(con, event))
             return 1;
     }
-    /* 8: otherwise, check for border/decoration clicks and resize */
+    /* 8: floating modifier pressed, initiate a drag */
+    else if (dest == CLICK_INSIDE && mod_pressed && event->detail == XCB_BUTTON_INDEX_1) {
+        DLOG("Trying to drag (tiling)\n");
+        tiling_drag(con, event);
+    }
+    /* 9: otherwise, check for border/decoration clicks and resize */
     else if (dest == CLICK_BORDER && is_left_or_right_click) {
         DLOG("Trying to resize (tiling)\n");
         tiling_resize(con, event, dest);
-    }
-    /* 9: otherwise, check for dragging a tiled window */
-    else if (dest == CLICK_DECORATION && is_left_or_right_click) {
-        DLOG("Trying to drag (tiling)\n");
-        tiling_drag(con, event);
     }
 
 done:
--- a/src/tiling_drag.c
+++ b/src/tiling_drag.c
@@ -126,7 +126,7 @@ static xcb_window_t create_drop_indicator(Rect rect) {
  *
  */
 void tiling_drag(Con *con, xcb_button_press_event_t *event) {
-    DLOG("con = %p\n", con);
+    DLOG("Start dragging tiled container: con = %p\n", con);
 
     /* Don't change focus while dragging. */
     x_mask_event_mask(~XCB_EVENT_MASK_ENTER_WINDOW);
--- a/src/tiling_drag.c
+++ b/src/tiling_drag.c
@@ -134,46 +134,6 @@ static xcb_window_t create_drop_indicator(Rect rect) {
 }
 
 /*
- * Moves the dropped container to its new position. This makes the the container
- * and its target the only two children of a split container with appropriate
- * layout. If the parent container of the target has correct layout and no
- * additional children, then no new split container is created.
- *
- */
-static void drop_con(Con *con, Con *target, direction_t direction) {
-    DLOG("Dropping container: con = %p, target = %p, direction = %d\n", con, target, direction);
-
-    /* Split the target's parent if we need to. */
-    orientation_t orientation = (direction == D_LEFT || direction == D_RIGHT) ? HORIZ : VERT;
-    if (con_orientation(target->parent) != orientation || con_num_children(target->parent) > 1) {
-        tree_split(target, orientation);
-    }
-
-    /* Detach the container from its old parent. */
-    Con *old_parent = con->parent;
-    Con *parent = target->parent;
-
-    con_detach(con);
-    con_fix_percent(con->parent);
-
-    /** Attach the container to the target's parent. */
-    con->parent = parent;
-    con->percent = target->percent = .5;
-
-    if (direction == D_LEFT || direction == D_UP) {
-        TAILQ_INSERT_BEFORE(target, con, nodes);
-        TAILQ_INSERT_HEAD(&(parent->focus_head), con, focused);
-    } else {
-        TAILQ_INSERT_AFTER(&(parent->nodes_head), target, con, nodes);
-        TAILQ_INSERT_HEAD(&(parent->focus_head), con, focused);
-    }
-
-    /* Clean up. */
-    CALL(old_parent, on_remove_child);
-    tree_flatten(croot);
-}
-
-/*
  * Initiates a mouse drag operation on a tiled window.
  *
  */
@@ -197,7 +157,7 @@ void tiling_drag(Con *con, xcb_button_press_event_t *event) {
 
     /* Move the container to the drop position. */
     if (drag_result != DRAG_REVERT && target != NULL && target != con) {
-        drop_con(con, target, direction);
+        con_move_to_side_of_con(con, target, direction);
         tree_render();
     }
 }
--- a/src/tiling_drag.c
+++ b/src/tiling_drag.c
@@ -1,0 +1,157 @@
+#undef I3__FILE__
+#define I3__FILE__ "tiling_drag.c"
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3 - an improved dynamic tiling window manager
+ * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
+ *
+ * tiling_drag.c: Reposition tiled windows by dragging.
+ *
+ */
+#include "all.h"
+
+/*
+ * Returns the leaf container at the given coordinates or NULL if no such
+ * container exists.
+ *
+ */
+static Con *con_by_coordinates(uint32_t x, uint32_t y) {
+    Con *con;
+    TAILQ_FOREACH(con, &all_cons, all_cons)
+    if (con->window != NULL && rect_contains(con->rect, x, y))
+        return con;
+    return NULL;
+}
+
+struct callback_params {
+    xcb_window_t indicator;
+    Con **target;
+    direction_t *direction;
+};
+
+/*
+ * The callback that is executed on every mouse move while dragging. On each
+ * invocation we determin the drop target and the direction in which to insert
+ * the dragged container. The indicator window is updated to show the new
+ * position of the dragged container. The target container and direction are
+ * passed out using the callback params.
+ *
+ */
+DRAGGING_CB(drag_callback) {
+    const struct callback_params *params = extra;
+
+    Con *target = con_by_coordinates(new_x, new_y);
+    direction_t direction;
+
+    DLOG("new x = %d, y = %d, con = %p, target = %p\n", new_x, new_y, con, target);
+    if (target == NULL)
+        return;
+
+    /* If the target is the dragged container itself then we want to highlight
+     * the whole container. Otherwise we determine the direction of the nearest
+     * border and highlight only that half of the target container.
+     *
+     * TODO(MForster): Support dropping on containers with tabbed or stacked
+     * layout as well as on empty outputs.
+     */
+    Rect rect = target->rect;
+    if (target != con) {
+        uint32_t d_left = new_x - rect.x;
+        uint32_t d_top = new_y - rect.y;
+        uint32_t d_right = rect.x + rect.width - new_x;
+        uint32_t d_bottom = rect.y + rect.height - new_y;
+        uint32_t d_min = min(min(d_left, d_right), min(d_top, d_bottom));
+
+        if (d_left == d_min) {
+            rect.width /= 2;
+            direction = D_LEFT;
+        } else if (d_top == d_min) {
+            rect.height /= 2;
+            direction = D_UP;
+        } else if (d_right == d_min) {
+            /* This is done in three steps to get symmetric rounding behavior
+             * with the first two cases. */
+            rect.x += rect.width;
+            rect.width /= 2;
+            rect.x -= rect.width;
+            direction = D_RIGHT;
+        } else if (d_bottom == d_min) {
+            /* This is done in three steps to get symmetric rounding behavior
+             * with the first two cases. */
+            rect.y += rect.height;
+            rect.height /= 2;
+            rect.y -= rect.height;
+            direction = D_DOWN;
+        }
+    }
+
+    xcb_configure_window(conn, params->indicator,
+                         XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y | XCB_CONFIG_WINDOW_WIDTH | XCB_CONFIG_WINDOW_HEIGHT,
+                         &(rect.x));
+    xcb_flush(conn);
+
+    *(params->target) = target;
+    *(params->direction) = direction;
+}
+
+/*
+ * Returns a new drop indicator window with the given initial coordinates.
+ *
+ */
+static xcb_window_t create_drop_indicator(Rect rect) {
+    uint32_t mask = 0;
+    uint32_t values[2];
+
+    mask = XCB_CW_BACK_PIXEL;
+    values[0] = config.client.focused.indicator.colorpixel;
+
+    mask |= XCB_CW_OVERRIDE_REDIRECT;
+    values[1] = 1;
+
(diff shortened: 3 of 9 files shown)


MForster:
Context:
* http://thread.gmane.org/gmane.comp.window-managers.i3.general/1270/focus=2023
* https://youtu.be/YGlACHImjhw

I do realize that this change is not fit for submission as is. But I'd would like to discuss how to proceed with the implementation, and there are a couple of open questions.

src/tiling_drag.c
```diff
@@ -0,0 +1,157 @@
...
+#include "all.h"
+
+/*
+ * Returns the leaf container at the given coordinates or NULL if no such
+ * container exists.
+ *
+ */
+static Con *con_by_coordinates(uint32_t x, uint32_t y) {
+    Con *con;
+    TAILQ_FOREACH(con, &all_cons, all_cons)
```
Airblader: Please use braces here so that we also get proper indentation.

src/tiling_drag.c
```diff
@@ -0,0 +1,157 @@
...
+    if (con->window != NULL && rect_contains(con->rect, x, y))
```
Airblader: I think this will misbehave for tabbed containers. I'm also not sure how it'll behave for windows that have been split without opening a new window.

src/tiling_drag.c
```diff
@@ -0,0 +1,157 @@
...
+
+    return indicator;
+}
+
+/*
+ * Initiates a mouse drag operation on a tiled window.
+ *
+ */
+void tiling_drag(Con *con, xcb_button_press_event_t *event) {
+    DLOG("con = %p\n", con);
```
Airblader: We should state what exactly we are doing, not just a container name.

src/click.c
```diff
@@ -321,11 +321,15 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
...
     /* 8: otherwise, check for border/decoration clicks and resize */
-    else if ((dest == CLICK_BORDER || dest == CLICK_DECORATION) &&
-             is_left_or_right_click) {
+    else if (dest == CLICK_BORDER && is_left_or_right_click) {
         DLOG("Trying to resize (tiling)\n");
         tiling_resize(con, event, dest);
     }
+    /* 9: otherwise, check for dragging a tiled window */
+    else if (dest == CLICK_DECORATION && is_left_or_right_click) {
+        DLOG("Trying to drag (tiling)\n");
```
Airblader: My main problem here is that I think there will be users confused by this because they expect the tiling resize here. We should think about how exactly to trigger the dragging resize.

src/tiling_drag.c
```diff
@@ -0,0 +1,157 @@
...
+    uint32_t mask = 0;
+    uint32_t values[2];
+
+    mask = XCB_CW_BACK_PIXEL;
+    values[0] = config.client.focused.indicator.colorpixel;
+
+    mask |= XCB_CW_OVERRIDE_REDIRECT;
+    values[1] = 1;
+
+    xcb_window_t indicator = create_window(conn, rect, XCB_COPY_FROM_PARENT, XCB_COPY_FROM_PARENT,
```
Airblader: We should set a proper WM_CLASS on this window so that users can configure their compositor.

Airblader:
I think this is testable with XTEST. We also need to look into the usual edge case behaviors: urgent flags, tabbed,...

src/click.c
MForster: Yes, I agree.

Instead of dragging the title bar, I could trigger on $mod+button2. That is unused for tiled windows right now, right?

src/tiling_drag.c
```diff
@@ -0,0 +1,157 @@
...
+    if (drag_result != DRAG_REVERT && target != NULL && target != con) {
+        orientation_t orientation = (direction == D_LEFT || direction == D_RIGHT) ? HORIZ : VERT;
+        position_t position = (direction == D_LEFT || direction == D_UP) ? BEFORE : AFTER;
+
+        DLOG("Moving dragged container: target = %p, orientation = %d, position = %d\n", target, orientation, position);
+
+        /* TODO(MForster): We shouldn't split unconditionally. This creates
+         * unnecessary chains of split containers. But for now splitting means
+         * that we don't have to deal with percentages. */
+        tree_split(target, orientation);
```
MForster: There are a couple of options what I could do here:
* Just don't split if not necessary. That makes it difficult to insert the container with exactly half the size of the target without duplicating code.
* Extend `insert_con_into` to optionally take percentage values or similar.
* After `tree_split` do something like `tree_flatten` that looks at the new container and its parent and merges them if they have the same layout, adjusting percentage values on the way.

My tendency goes to the last options, but I'm open for suggestions.

src/tiling_drag.c
MForster: I followed the stile in [`con.c`](https://github.com/i3/i3/blob/next/src/con.c#L534). I can update the functions there as well.

BTW: Should I move this to `con.c`?

include/move.h
```diff
@@ -15,3 +15,13 @@
  *
  */
 void tree_move(Con *con, int direction);
+
+typedef enum { BEFORE,
+               AFTER } position_t;
```
MForster: The names of these enum values sound a bit too generic to be made public. Should I rename them to `POS_BEFORE`/`POS_AFTER`?

src/tiling_drag.c
Airblader: Yeah, we have some places where this is still wrong, but we prefer the braces by now. You can update those as well if you want. And yes, I think moving it there makes sense.

src/tiling_drag.c
Airblader: Currently this behavior makes sense because it fits the indicator. Any change of this behavior needs to be reflected in the size of the indicator and in fact means we'd have to make the indicator window a subwindow of the parent window (or just the root window itself).

Note that we already have a `tree_flatten`, but I don't think it does what you describe here.

include/move.h
Airblader: We usually have the pattern of using the first letter of the enum type's name as a prefix, i.e., `P_BEFORE` / `P_AFTER`.

src/tiling_drag.c
MForster: Done

src/tiling_drag.c
MForster: I've improved the logic. I think it should be correct now.

src/tiling_drag.c
MForster: Done

src/click.c
MForster: I changed the code to trigger on floating modifier and left button. I think this was not used before and is a nice event to trigger this as well.

src/tiling_drag.c
MForster: Sorry, I don't quite understand this. I'm not too familiar with the protocol here. Could you elaborate? I couldn't find other code where i3 creates a window with a proper WM_CLASS. My code is inspired by the code that draws the resize indicator [in resize.c](https://github.com/i3/i3/blob/next/src/resize.c#L119).

MForster:
I've addressed most of Airblader's comments. Please take another look.

Open issues:
* Window class for the indicator
* Tests

src/tiling_drag.c
Airblader: Have a look at `src/x.c` and search for `XCB_ATOM_WM_CLASS`. The reason we should set it is to user can match on it in their compositor configuration. The important thing is to change the class before mapping the window.

src/tiling_drag.c
```diff
@@ -0,0 +1,191 @@
...
+/*
+ * Returns the visible leaf container at the given coordinates or NULL if no
+ * such container exists.
+ *
+ */
+static Con *find_drop_target(uint32_t x, uint32_t y) {
+    Con *con;
+    TAILQ_FOREACH(con, &all_cons, all_cons) {
+        if (con->window != NULL && rect_contains(con->rect, x, y) &&
+            con->floating < FLOATING_AUTO_ON &&
```
Airblader: I'd prefer `!con_is_floating(con)`. It'll do exactly the same, but keep the logic encapsulated.

MForster: Yes, that's better. I try to use established patterns, but I'm not familiar enough with the code yet to not sometimes miss them.

Done.

src/tiling_drag.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/*
+ * Moves the dropped container to its new position. This makes the the container
+ * and its target the only two children of a split container with appropriate
+ * layout. If the parent container of the target has correct layout and no
+ * additional children, then no new split container is created.
+ *
+ */
+static void drop_con(Con *con, Con *target, direction_t direction) {
```
Airblader: I'm wondering whether this can just use `_con_move_to_con`. The reason for it is that we should avoid duplicating all this logic. For example, I doubt (haven't tried it) that the current solution would move urgency hints correctly.

Airblader:
Thanks :) I appreciate the effort you're putting into this, this will be a great feature. Before I take another close look at it, I'll also have @stapelberg comment on it.

src/tiling_drag.c
MForster: Ah, nice. I didn't even know that there are compositors that are not window managers. I'll need to see whether I want to use one.

Anyway, I've added code to set the class to i3-drag/i3-drag and verified that it works with `compton --opacity-rule "50:class_g = 'i3-drag'"`

Airblader: This is now violating the specs because the window is mapped before we set the WM_CLASS on it (see https://tronche.com/gui/x/icccm/sec-4.html#WM_CLASS). But let's worry about getting this right later as it's such a minor detail.

MForster: Ah, sorry. I missed this in your previous comment. Done.

src/tiling_drag.c
MForster: Ok, I reworked the code to use `_con_move_to_con`. This made it necessary to add yet another boolean flag to that function.

BTW: `move.c` seems to have some duplication as well.

include/move.h
MForster: I've undone this change, because I'm not using this code any more.

src/tiling_drag.c
MForster: With the latest implementation I've extended `tree_flatten` to handle the case I'm interested in.

src/tree.c
```diff
@@ -687,6 +732,12 @@ void tree_next(char way, orientation_t orientation) {
  * split container then and if you move containers this way multiple times,
  * redundant chains of split-containers can be the result.
  *
+ * A second example is this
+ *   [v-split]
+ *   [v-split]
+ * [child1] [child2]
+ * The two v-split containers can be collapsed to a single one. Such a situation
+ * is created when dragging tiled containers.
```
MForster: Do you agree that these containers are also redundant? Or are there cases where this collapsing isn't correct?

MForster:
@stapelberg, do you have any comments?

Airblader:
@MForster Since Michael hasn't responded to any tickets in the last few days I'm guessing he's busy with work. Just give him some time. :)

src/tree.c
stapelberg: I think the example you gave should indeed be collapsed (or flattened, as we call it). If it passes the test suite, that’s a rather good signal that nothing will break.

stapelberg:
Sorry for the late reply, I’ve indeed been busy, and this feature deserves a bigger chunk of uninterrupted time.

The code looks good overall, and we can get to minor style details/other nits later in the process.

One thing I’d like to see done next (before spending time on tests etc.) is support for tabbed/stacked containers. There are users (me included) who spend most of their time with `workspace_layout stacked`, i.e. dealing with stacked containers exclusively.

Further, I think we should also spend some time on the user interface. It seems like currently, the performed operation when dragging a window is to always split the target container. I wonder, and this ties into the above point about stacked/tabbed containers, whether moving the windows would be a better choice, at least for some layouts? When dragging a window from one stacked container to another, I certainly wouldn’t expect new splits.

Another action that currently isn’t possible AFAICT is moving a window out of a container, thereby creating a new container on the workspace level. E.g., in this layout:

Should there be a way to achieve what “move right” does? (The MPlayer window will end up between the vim/terminal windows and the Vimperator window.)

Airblader:
Out of curiosity, what happens currently if the window is dragged onto a dock window?

MForster:
@stapelberg wrote:
> Sorry for the late reply, I’ve indeed been busy, and this feature deserves a bigger chunk of uninterrupted time.

No worries. I agree, and I didn't want to pressure you. I just wanted to make sure that you didn't miss the pull request.

> One thing I’d like to see done next (before spending time on tests etc.) is support for tabbed/stacked containers.

Yes. Absolutely. I didn't spend time on this yet because I wanted to get some feedback first. Seeing that the general approach is fine, I'm happy to work on tabbed/stacked containers next.

> Further, I think we should also spend some time on the user interface. It seems like currently, the performed operation when dragging a window is to always split the target container. I wonder, and this ties into the above point about stacked/tabbed containers, whether moving the windows would be a better choice, at least for some layouts? When dragging a window from one stacked container to another, I certainly wouldn’t expect new splits.

My suggestion would be to make this dependent on the drop location. Dropping on one side of the window content would still split the window, but dropping on the title bar or possibly a bit below would insert the window into a stack or tab bar, creating a new stacked or tabbed container if it doesn't exist yet.

> Another action that currently isn’t possible AFAICT is moving a window out of a container, thereby creating a new container on the workspace level.

Yes. I thought about this. Currently it is only possible to insert a dragged window next to a leaf in the tree. For all practical purposes I found this good enough. But I can see that additional drop locations could be useful.

The case you're describing is adding it as a child of the root. In theory we could also consider adding at an arbitrary intermediate level in the tree. This could be selected by the distance from the border. Basically dropping right at the window border would insert the window where you're suggesting. Further to the middle of the window would insert it next to the window itself. Positions in between would select some intermediate level. My concern is that this would become too complex. I could try to prototype it to try it out.

Or, we could just support leaves and children of the root. My gut feeling is that this is a bit inconsistent. But it could be a compromise.

> Should there be a way to achieve what “move right” does? (The MPlayer window will end up between the vim/terminal windows and the Vimperator window.)

The second part of your question is whether to always split, or whether to insert in an existing container, if possible. I've tried out not to split and found it inferior for two reasons:

1. Always splitting is very predictable for the user. The window ends up exactly on one side of the drop location. In many cases the layout does not "jump" at all. 

2. The code for not splitting would be more complex. Predicting where the window would end up in an existing split container makes it way more difficult to compute the coordinates for the drag indicator. It's certainly doable, but I think it would make it necessary to refactor the existing layout code to extract shared logic.

**Next steps:**

My suggestion would be that I start implementing support for stacked/tabbed containers and show a few more screencast about how this behaves. Or you just patch the change and try it out. In any case I think it's useful to have a prototype to play with to decide what the best user interface is.

A few process questions as the development of the feature will certainly take more time:
* Do you prefer merging in intermediate milestones, or should we keep it fully on its own branch until it is complete?
* Should I regularly rebase the branch to `next` to keep it more up-to-day, or do you prefer not rewriting history?

@Airblader wrote:
> Out of curiosity, what happens currently if the window is dragged onto a dock window?

i3 crashes :-). I had not considered this case, because my dock is hidden. I'll fix this.

src/tree.c
MForster: Great. It passes the test suite.

Airblader:
Other interesting drop locations are floating windows and empty workspaces (across screens) unless you've handled those already.

As for rebasing, we do generally want you to rebase it. History matters once it's merged. It's up to you when you want to rebase, we only need you to before merging, but doing so every now and then might be easier for you.

MForster:
Floating windows and empty workspaces I had tested, and those are are currently ignored. For floating windows this is likely the right behavior. For empty workspaces I want to enable dropping there.

I'll rebase then from time to time.

Airblader:
@MForster @stapelberg With 4.12 released, shall we continue with this? Let's not have this PR die. :)

MForster:
@Airblader, I definitely want to finish this. I'm using my prototype all of the time. Only, this time it was me being busy at work :-/.

Also, it turned out that support for split containers is more straight-forward than for tabbed/stacked containers. I'm still deliberating about the best UX for tabbed/stacked containers. For split containers I'm reasonably convinced that the best UX is to only look at leaf containers. Otherwise it becomes too confusing. But for tabbed/stacked containers that restriction seems arbitrary. 

And then I haven't really found out how to best get access to the geometry of container headers. I'll need these to determine the correct drop location for tabbed containers.

Anyway, I'll have some vacation in two weeks. I hope to make progress then.

lmao-emerald:
Is this project dead? As an i3 use I would love to have this.

MForster:
@lmao-emerald I hope not. I'm using the feature daily (my own patched version of i3). But I wasn't able to make progress either...

xtaran:
I'd also be happy to see that feature in i3 as I'm a former awesome user and used dragging windows to exchange their position all the time. (As I understand it here, it's though no exchanging of window positions. Nevertheless, having such an user interface would be nice anyways.)  Actually I still occassionally try to do that with i3 and only then notice that the tried-to-use feature is not (yet) there.

pnth:
I change my i3 layout often. It would be very useful if there is mouse support for layout. How could I tried this awesome feature? I tried the following but doesn't work:

```bash
% git clone https://github.com/i3/i3.git
% git checkout -b i3mouse e9ad03613d4d9f214de97026dfa9b24d9b9cc7fc
fatal: reference is not a tree: e9ad03613d4d9f214de97026dfa9b24d9b9cc7fc
% git branch -a
* next
  remotes/origin/HEAD -> origin/next
  remotes/origin/master
  remotes/origin/next
  remotes/origin/randr-15
% git checkout master
% git checkout -b i3mouse e9ad03613d4d9f214de97026dfa9b24d9b9cc7fc
fatal: reference is not a tree: e9ad03613d4d9f214de97026dfa9b24d9b9cc7fc
```

MForster:
@pnth You can check out directly from https://github.com/MForster/i3.git, use branch `tiling-drag`. To try it out press the floating modifier button while dragging anywhere in the window.

pnth:
$mod+shift+drag works. Thanks a lot. My previous install actually works.

Deevad:
I'm an artist and I can't use both hands on a keyboard to control a WM like i3 (one of my hand keeps a stylus to draw/digital-paint, most of the time ; consider it as a mouse cursor). I really like i3 design, features and performances. I also like tiling windows management. 

That's why, I am interested a lot in the work proposed here by @MForster ( I discovered his demo  https://www.youtube.com/watch?v=YGlACHImjhw , then came here via the mailing-list linked in the video description ).

Does this work will be merged in main 'master' source code ?

MForster:
Depends on whether I find the time to continue the work :-/.

David Revoy <notifications@github.com> schrieb am Mi., 15. Juni 2016 um
18:29 Uhr:

Deevad:
Thank you @MForster for the reply. No problem if you can't find time for it: I totally understand. It's already really generous to propose the code in your fork/branch.

pnth:
Hi David,
I have been using this feature a lot and really love it. I really want to move a window from one leaf of a tree to be a sibling of another window in another tree, instead of being its child, by dropping to the target window's title bar. Could you give some hints whether this is an easy modification and where to look into the source code so that I may try to do it.

Much appreciated

philipp-classen:
Very nice patch. Looking forward to seeing it integrated. :-)

For all Arch Linux users that would like to try it out, here is an AUR package:
https://aur.archlinux.org/packages/i3-mouse-dragging-git

yaourt -S i3-mouse-dragging-git
i3 restart

Then you should be able to move windows by holding $mod.

Airblader:
We're nearing this PR's first birthday. What's the status?

MForster:
I think I'll have to accept the sad truth that I won't finish this. At least I won't find the time to implement support for stacked/tabbed containers (which I hardly use). I've been running with this patch for a year, and it's good enough for me, which lowers the incentive. I'm open for integrating as is, or somebody else can pick it up.

Airblader:
Fair enough, thank you for the feedback. I'd propose to open an issue for this topic and linking this PR in there, but closing the PR for now. That way people can still find it and anybody can pick up your work and continue it. I'd prefer not to merge it as-is with nobody being committed to working on it at the moment.

Sound good? @stapelberg Your thoughts?

stapelberg:
Sounds good.

orestisf1993:
Watchers of this PR: I opened a continuation in #3085.

nawtrobar:
Have you looked at sway's implementation? You're able to drag and resize tiled windows in sway by using the mod key. Maybe the feature is easily ported into i3.

orestisf1993:
You can already resize using the mod key. Dragging containers like in sway is something I consider implementing.

Issue #268, opened 2010-11-23
Parsing spaces/comments at end of line in config file


i3bot:
[Originally reported by B-Con <brad@…>]
(In my i3 config file, if I have the following lines:

bindsym Mod1+s T   # container tabbing
bindsym Mod1+comma bn   # normal border

the first command will create a working binding, but the second one will not. It seems that some bindsym lines may end with a comment, while others may not. 

Specifically, if I attempt to use "Mod1+s", I get the following error message in my log file (via -V):
- --- parsing command "T   # container tabbing" ---
- Error: unknown direction: T

However if I use Mod1+comma, the window bordering changes appropriately.

So, it seems that the parsing of spaces and/or comments at the end of config lines is inconsistent.

PR #2181, opened 2016-01-30
Do not count '\' in comment as line continuation 

### LINKED ISSUE #2176: config: Comment lines can be continued with '\'
I noticed today that comment lines ending in the line-continuation character `\`, which precede a non-comment line, cause the non-comment line to be ignored.  Presumably because the line continuation is honored, and the next line is considered part of the comment.

For example, the following would spawn only one of the two terminal windows a user might expect, because the first one is probably parsed as being commented out:

```shell
# comment with a line continuation character at the end \
exec gnome-terminal

# comment without line continuation character at the end
exec gnome-terminal
```

Of course now that I know this, I can avoid it, but it's fairly obtuse to discover because in some cases like the above there are no errors -- the line just silently fails to execute.  In other cases (namely, if there's a data dependency on the skipped directive later in the config) there's an error, but it's removed from the location of the actual problem.

Ideally, the config parser should apply line continuations after comments have been ignored.

---

This is reproducible for me with i3 4.11 on Ubuntu 15.10 using the packages from sur5r's repository:
```
$ i3 --moreversion 2>&- || i3 --version
Binary i3 version:  4.11 (2015-09-30, branch "4.11") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.11 (2015-09-30, branch "4.11") (pid 26498)
Loaded i3 config: /home/jpk/.config/i3/config (Last modified: Tue 26 Jan 2016 01:03:17 AM EST, 1492 seconds ago)

$ apt-cache policy i3
i3:
  Installed: 4.11-1~~wily1
  Candidate: 4.11-1~~wily1
  Version table:
 *** 4.11-1~~wily1 0
        500 http://debian.sur5r.net/i3/ wily/universe amd64 Packages
        100 /var/lib/dpkg/status
     4.10.3-1build1 0
        500 http://us.archive.ubuntu.com/ubuntu/ wily/universe amd64 Packages
```

Here's a [minimal config](https://gist.github.com/thatjpk/0e5481610bf1044630d4) that reproduces the behavior (just the default `i3-config-wizard` config with the quoted snippet above appended).  
 - Current behavior: The session starts with one terminal open.  
 - Expected behavior: The session starts with two terminals open.

jolange:
fixes #2176 

I'll squash this into a single commit once you approved the changes.

@thatjpk: FYI

### DIFF
--- a/src/config_parser.c
+++ b/src/config_parser.c
@@ -846,17 +846,26 @@ bool parse_file(const char *f, bool use_nagbar) {
         if (buffer[strlen(buffer) - 1] != '\n' && !feof(fstr)) {
             ELOG("Your line continuation is too long, it exceeds %zd bytes\n", sizeof(buffer));
         }
+
+        /* sscanf implicitly strips whitespace. */
+        const bool skip_line = (sscanf(buffer, "%511s %511[^\n]", key, value) < 1 || strlen(key) < 3);
+        const bool comment = (key[0] == '#');
+
         continuation = strstr(buffer, "\\\n");
         if (continuation) {
-            continue;
+            if (!comment) {
+                continue;
+            }
+            DLOG("line continuation in comment is ignored: \"%.*s\"\n", (int)strlen(buffer) - 1, buffer);
+            continuation = NULL;
         }
 
         strncpy(buf + strlen(buf), buffer, strlen(buffer) + 1);
 
-        /* sscanf implicitly strips whitespace. Also, we skip comments and empty lines. */
-        if (sscanf(buffer, "%511s %511[^\n]", key, value) < 1 ||
-            key[0] == '#' || strlen(key) < 3)
+        /* Skip comments and empty lines. */
+        if (skip_line || comment) {
             continue;
+        }
 
         if (strcasecmp(key, "set") == 0) {
             if (value[0] != '$') {
--- a/docs/userguide
+++ b/docs/userguide
@@ -1115,11 +1115,15 @@ show_marks yes
 Config files support line continuation, meaning when you end a line in a
 backslash character (`\`), the line-break will be ignored by the parser. This
 feature can be used to create more readable configuration files.
+Commented lines are not continued.
 
 *Examples*:
 -------------------
 bindsym Mod1+f \
 fullscreen toggle
+
+# this line is not continued \
+bindsym Mod1+F fullscreen toggle
 -------------------
 
 == Configuring i3bar
--- a/testcases/t/247-config-line-continuation.t
+++ b/testcases/t/247-config-line-continuation.t
@@ -198,4 +198,20 @@ EOT
 is(launch_get_border($config), 'none', 'no border');
 
 
+#####################################################################
+# test ignoring of line continuation within a comment
+#####################################################################
+
+$config = <<'EOT';
+# i3 config file (v4)
+font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
+
+set $vartest \"special title\"
+for_window [title="$vartest"] border pixel 1
+# this line is not continued, so the following is not contained in this comment\
+for_window [title="$vartest"] border none
+EOT
+
+is(launch_get_border($config), 'none', 'no border');
+
 done_testing;


src/config_parser.c
```diff
@@ -847,8 +847,14 @@ bool parse_file(const char *f, bool use_nagbar) {
             ELOG("Your line continuation is too long, it exceeds %zd bytes\n", sizeof(buffer));
         }
         continuation = strstr(buffer, "\\\n");
-        if (continuation) {
-            continue;
+        if (continuation){
+            /* do not count '\' in a comment as continuation */
+            if (strchr(buffer,'#')) {
```
Airblader: What if the # is inside of a string etc? I don't think this sufficiently checks for a comment

jolange: You are right. Color specifications is another such case...
Any others one has to consider?

Airblader: We shouldn't be wondering what cases we have to consider. Checking for a comment here should be exactly the same as where we actually identify comments, using the same code.

jolange: Sure, sounds reasonable. But I did not yet figure out how to do that. As far as I understand, this is completely defined via `parser-specs/config.spec`. Any hint or suggestion how to make use of this in this case would be helping.

Airblader: Yes, you're right. I forgot that it's handled in the parser already (I was on mobile and didn't check). Anyway, I think the pattern for comments should be `^\s*#`.

jolange: This would only match whole comment lines, but not `some-sensible-command # comment`, right?

jolange:
For now, just "comment-only" lines are not continued:
```
# the following line should be continued
exec "sleep\
 1"

# this line should not be continued \
client.focused          #ff4040 #ff0000 \
#1793D1 #2e9ef4   #ff0000

exec "sleep 2" # don't continue\
exec "sleep 3"
```
becomes:
```
# the following line should be continued
exec "sleep 1"

# this line should not be continued \
client.focused          #ff4040 #ff0000 #1793D1 #2e9ef4   #ff0000

exec "sleep 2" # don't continueexec "sleep 3"
```

I think this is not consistent, but I don't see a nice way of checking if a line contains a comment at this point yet.

jolange:
Sorry for the long silence here, I didn't have the time to take a closer look at this until now. Unfortunately, after finally putting some thought into it, I still don't see a good way to make use of parser information to check if the current line contains a comment.

IMO the options are
- Only disallow continuation of comment-only lines, while allowing it for lines just ending with a comment. (As currently implemented in this PR) [inconsistent]
- Figure out all possible occurences of `#` not introducing a comment (strings, colors, ...). [not a nice solution]
- Stay with the current behaviour. [= unexpected behaviour]

Of these, I'd prefer the last one meanwhile... If you see another (good) option, please let me know.

Airblader:
No, I don't really see a better option either.

thatjpk:
Yeah, I see the problem.  The thing that prompted me to file the issue was I didn't expect the behavior -- but now that I know about it, I can avoid it.  So if the best compromise is to leave the current behavior as is, then perhaps a note and an example about it in the documentation would be helpful.

Airblader:
Feel free to provide a pull request to add that hint. :-)

stapelberg:
Issue #268 is relevant — we currently don’t support comments at the end of lines at all.

So, strictly speaking, only lines which begin with a `#` sign are considered comments. If comment signs at the end of a line work in some places, that is a happy coincidence.

I would be okay with only skipping continuations in comment-only lines. I think people run into this issue rarely enough that making that situation strictly better is good enough.

jolange:
> we currently don’t support comments at the end of lines at all.

Ah, thanks, I was not aware of that! (Never tried it before, first when working on this issue)

In that case I totally agree. So consider the current state of this PR as my proposal.

src/config_parser.c
```diff
@@ -848,7 +849,13 @@ bool parse_file(const char *f, bool use_nagbar) {
         }
         continuation = strstr(buffer, "\\\n");
         if (continuation) {
-            continue;
+            /* do not count '\' in a comment as continuation */
+            if (regex_matches(re_comment, buffer)) {
```
stapelberg: While the regular expression that you’re using *should* be doing the same thing as the `sscanf` call below, I think it’d be cleaner to actually re-use the `sscanf` call, both from a duplication point of view and also because it avoids compiling a likely rarely-used regular expression every time the configuration file is read.

I think we could move the `sscanf` call further up, possibly like so (untested):

```c
/* sscanf implicitly strips whitespace. */
const bool comment = (sscanf(buffer, "%511s %511[^\n]", key, value) < 1 ||
                      key[0] == '#' || strlen(key) < 3)
continuation = strstr(buffer, "\\\n");
if (continuation) {
  if (!comment) {
    continue;
  }
  DLOG("line continuation in comment is ignored: \"%.*s\"\n", (int)strlen(buffer) - 1, buffer);
  continuation = NULL;
}

strncpy(buf + strlen(buf), buffer, strlen(buffer) + 1);

/* Skip comments and empty lines. */
if (comment) {
  continue;
}
```

jolange: Yes, you have a point there. But I would only use `key[0] == '#'` for the comment bool. In particular this suggestion does not work for comments without spaces: `#comment\`

Please see my proposed variation of this.

src/config_parser.c
```diff
@@ -847,23 +846,26 @@ bool parse_file(const char *f, bool use_nagbar) {
         if (buffer[strlen(buffer) - 1] != '\n' && !feof(fstr)) {
             ELOG("Your line continuation is too long, it exceeds %zd bytes\n", sizeof(buffer));
         }
+
+        /* sscanf implicitly strips whitespace. */
+        const bool skipLine = (sscanf(buffer, "%511s %511[^\n]", key, value) < 1 || strlen(key) < 3);
```
stapelberg: nit: for consistency, can you rename `skipLine` to `skip_line` please?

src/config_parser.c
```diff
@@ -847,23 +846,26 @@ bool parse_file(const char *f, bool use_nagbar) {
...
+        const bool comment = key[0] == '#';
```
stapelberg: nit: for consistency, can you please add parens, i.e. `const bool comment = (key[0] == '#');`?

stapelberg:
Looks good overall, will merge once the nits are addressed.

jolange:
@stapelberg done & rebased

Issue #2184, opened 2016-01-31
i3bar sends invalid XEMBED_EMBEDDED_NOTIFY message


nicklan:
At https://github.com/i3/i3/blob/next/i3bar/src/xcb.c#L827 which is:
```c
ev->data.data32[1] = atoms[XEMBED_EMBEDDED_NOTIFY];
```

The xembed message isn't an atom, it's just an id. (see [here](http://standards.freedesktop.org/xembed-spec/xembed-spec-latest.html#messages)) This line causes an invalid value to be sent in ev->data.data32[1].

The line should simply be

```c
 ev->data.data32[1] = XEMBED_EMBEDDED_NOTIFY;
```

i3 version 4.11 (2015-09-30, branch "4.11") © 2009 Michael Stapelberg and contributors

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
Yeah, this seems to be wrong indeed. I also checked it against awesome's code. Do you want to send a PR fixing this?

nicklan:
sure, i'll do so shortly, thanks

nicklan:
okay, PR #2185 sent

PR #2185, opened 2016-01-31
Fix issue #2184.  don't send atom, but raw code

nicklan:
XEMBED messages aren't atoms, but just codes, so i3bar should just send
XEMBED_EMBEDDED_NOTIFY rather than indexing the atom array.


stapelberg:
Thanks!

PR #2190, opened 2016-02-03
Userguide: "hiding vertical borders" -> "hiding outer borders"

jolange:
Rephrased to fit with the options `horizontal` and `both` (introduced in 57effd6). Esp. the heading should be this general.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -605,9 +605,10 @@ new_window pixel 3
 ---------------------
 
 
-=== Hiding vertical borders
+[[_hiding_vertical_borders]]
+=== Hiding borders adjacent to the screen edges
 
-You can hide vertical borders adjacent to the screen edges using
+You can hide container borders adjacent to the screen edges using
 +hide_edge_borders+. This is useful if you are using scrollbars, or do not want
 to waste even two pixels in displayspace. Default is none.
 


docs/userguide
```diff
@@ -605,9 +605,9 @@ new_window pixel 3
 ---------------------
 
 
-=== Hiding vertical borders
+=== Hiding outer borders
 
-You can hide vertical borders adjacent to the screen edges using
+You can hide the outer borders of containers adjacent to the screen edges using
```
Airblader: Containers don't have "inner borders adjacent to the screen edge", so I think "the outer" is misleading here. I'd also prefer the heading to be "Hiding borders adjacent to the screen edge", accordingly. Just my 2c :) Anyway +1 for updating this.

jolange: Right, that's kind of redundant. I don't really care about the exact formulation, as long as "vertical" is gone. I'll take your suggestions for now and see what @stapelberg says.

docs/userguide
```diff
@@ -605,9 +605,9 @@ new_window pixel 3
 ---------------------
```
stapelberg: Can you add `_hiding_vertical_borders` here please so that links don’t break?

jolange: Sure, done.

PR #2197, opened 2016-02-07
Update i3-sensible-terminal

norrland:
Adding 'termit' to the list of terminals.

### DIFF
--- a/i3-sensible-terminal
+++ b/i3-sensible-terminal
@@ -8,7 +8,7 @@
 # We welcome patches that add distribution-specific mechanisms to find the
 # preferred terminal emulator. On Debian, there is the x-terminal-emulator
 # symlink for example.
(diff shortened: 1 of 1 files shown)


Airblader:
What is "termit"? I only know "termite" and that's already in the list.

stapelberg:
https://packages.debian.org/stretch/termit

norrland:
https://github.com/thestinger/termite

Airblader:
@stapelberg Shouldn't we only be appending new entries at the end? Otherwise we can break `i3-sensible-terminal` for users who have both installed.

stapelberg:
No, people should set the `TERMINAL` environment variable or replace calls to i3-sensible-terminal with executing their favorite terminal. On distributions such as Debian, `x-terminal-emulator` takes care of user preferences, and other distributions are welcome to follow suit.

Airblader:
Just seems like a no-brainer as it adds no complexity to avoid breakage, but I'm all for pushing users towards reasonable setups.

PR #2200, opened 2016-02-08
Simplify multi-monitor presentations

wentasah:
It seems that setting up i3 for multi-monitor presentations (e.g. with
LibreOffice Impress) is not as easy as it should be [[1], [2]]. I use the
method [1], but it is not ideal when one is in hurry or under stress
before the presentation.

This commit is my first attempt to solve this issue. I'm not sure wheter
this is the right way to implement this. Suggestions welcome :)

Programs like LO Impress or pdfpc configure the presentation windows to
have the same geometry as the outputs where the window should be. With
this change, when such a window is detected, it is open on the
appropriate workspace. Later, when the program sets its fullscreen
status, the window appear at the right output and the presentation works
out of the box.

Tested with LO Impress and pdfpc. The testsuite passes.

[1]: https://faq.i3wm.org/question/2114/libreoffice-presentation-done-righttm-with-i3.1.html
[2]: https://faq.i3wm.org/question/3965/dual-screen-presentation-does-not-work.1.html


stapelberg:
Can you please change the commit description to first and foremost include what you’re changing, not what you’re trying to accomplish? I.e., AFAICT, make fullscreen windows open on the output which is indicated by their geometry.

Also, the travis tests fail. Let me know once both are addressed and I’ll take a look at the change itself.

include/randr.h
```diff
@@ -79,6 +79,15 @@ Output *get_output_by_name(const char *name);
  */
 Output *get_output_containing(unsigned int x, unsigned int y);
 
+/**
+ * Returns the active (!) output which has the top left corner at
```
Airblader: If you need to warn about something in the documentation, the better solution likely is to give the function a better name.

src/manage.c
```diff
@@ -236,6 +236,8 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
         }
     }
 
+    Output *output_for_fullscreen = get_output(geom->x, geom->y, geom->width, geom->height);
```
Airblader: Reading this, the function might as well just take a `Rect` as an input.

src/manage.c
```diff
@@ -271,6 +273,18 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
                 nc = tree_open_con(nc, cwindow);
             else
                 nc = tree_open_con(nc->parent, cwindow);
+        } else if (output_for_fullscreen) {
```
Airblader: I don't really like that we are talking about fullscreen. The fact is that at this point we have *no* idea whether the window will request to go into fullscreen later on.

include/randr.h
wentasah: Yes, I'm bad at inventing names :-) Would "find_output()" be better?

Airblader: That's just as general, you just replaced one placeholder word with another. :) How about `get_output_with_rect(Rect rect)`?

wentasah:
This pull request was open against master branch. I've rebased the changes to next and I don't know how to change the PR to next. So I'm closing this PR and will open a new one.

PR #2203, opened 2016-02-09
Ensure that the "border" command uses logical pixels.

### LINKED ISSUE #2202: Custom DPI breaks pixel borders
i3 version 4.11 (2015-09-30, branch "4.11"), 64-bit Manjaro

1280x800 12,1" screen. The standard setting of 96 DPI works fine, but I wanted to fix it by setting proper millimeter values in 90-monitor.conf as well as DPI in .Xresources.

Now whenever I tile windows in i3, they are assigned a thick pixel border even though it's set to 1pixel. When I forcibly Mod-y them, they turn into 1 pixel borders. But every newly created window has thick borders.

The Workspace frames and the i3status separators are thick as well.

Airblader:
Until now, only the config directive for borders (new_window,
new_float) respected the DPI setting (using logical_px). This
patch makes sure we also do so for runtime "border" commands.

fixes #2202

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2255,6 +2255,10 @@ and +border none+ to make the client borderless.
 
 There is also +border toggle+ which will toggle the different border styles.
 
+Note that "pixel" refers to logical pixel. On HiDPI displays, a logical pixel
+may be represented by multiple physical pixels, so +pixel 1+ might not
+necessarily translate into a single pixel row wide border.
+
 *Syntax*:
 -----------------------------------------------
 border normal|pixel [<n>]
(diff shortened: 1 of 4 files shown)


stapelberg:
Thanks! Do we have other places in the commands parser spec where converting to `&` numbers makes sense?

Airblader:
@stapelberg  When I introduced the number type to the command parser I thought I had migrated everything that could be migrated to numbers, must've missed that. :) I can't see anything else right now, at least there's no other `strtol` in `commands.c`.

PR #2210, opened 2016-02-11
Don't allow floating workspaces anymore.

### LINKED ISSUE #2166: Strange behavior when making a group of stacked windows floating
Hello,

This is very unimportant, but I figured I would report it anyway.

If you get into a blank workspace and open four windows in stacked mode, then you focus parent twice, and then you make them float, the windows are displayed in a strange way. Doesn't really matter which programs I open. I get something like the following in my computer:
http://imgur.com/Jkq5x92

I don't think it's related to any of my configuration, since I've been able to reproduce this with the default configuration and with no composite manager
[5748291350298624.txt](https://github.com/i3/i3/files/93584/5748291350298624.txt)

Airblader:
This is essentially a revert of daf00a9 which introduced a mechanism
to float workspace containers by wrapping all children into a new
container and floating it.

This caused a bug and there's no good reason (anymore) to allow doing
this in the first place as we don't support floating nested containers.

fixes #2166

### DIFF
--- a/testcases/t/155-floating-split-size.t
+++ b/testcases/t/155-floating-split-size.t
@@ -21,6 +21,9 @@ use i3test;
 
 my $tmp = fresh_workspace;
 
+open_window;
+cmd 'split v';
+
 #####################################################################
 # open a window with 200x80
 #####################################################################
@@ -29,6 +32,8 @@ my $first = open_window({
         rect => [ 0, 0, 200, 80],
         background_color => '#FF0000',
     });
+
+cmd 'split h';
 
 #####################################################################
 # Open a second window with 300x90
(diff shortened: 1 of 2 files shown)


testcases/t/155-floating-split-size.t
```diff
@@ -21,6 +21,9 @@ use i3test;
 
 my $tmp = fresh_workspace;
 
+open_window;
```
Airblader: @stapelberg I think it's worth keeping this test around anyway. However, it also shows that a simple `open, open, focus parent, floating enable` won't work anymore, but `open, split v, open, split h, open, focus parent, floating enable` still will (since that's not a workspace container). Given that we don't want floating nested containers, that should be OK, though, right?

stapelberg: Agreed.

Issue #2116, opened 2015-12-15
_NET_WM_FULLSCREEN_MONITORS not handled properly.


fabrox:
i3 is not handling the _NET_WM_FULLSCREEN_MONITORS (http://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472531472) hint properly. There are several applications that use this hint to span a fullscreen window over several monitors.

One example is freerdp - If started with /multimon /f the resulting session window should span over all connected monitors. In i3 the resulting window is only shown in fullscreen on the monitor from which freerdp is started and thus the space on monitors to the top/left/right is invisible.

I don't know if that helps but some window managers that get it right are for example xfwm4, dwm and awesome.

i3 versions i tested it with are 4.11 and 4.10.4.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

fabrox:
i3 version 4.11 (2015-09-30, branch "4.11")

Airblader:
> i3 is not handling the _NET_WM_FULLSCREEN_MONITORS ([…]) hint properly.

i3 is not handling it all. However, i3 also doesn't claim to handle it and hence any application should be able to deal with that.

> There are several applications that use this hint to span a fullscreen window over several monitors.

i3 only allows to span a fullscreen container over one or all monitors, nothing in between anyway.

> I don't know if that helps but some window managers that get it right are for example xfwm4, dwm and awesome.

I've only looked at the awesome source code, but awesome doesn't handle this hint at all either.

fabrox:
>> i3 is not handling it all. However, i3 also doesn't claim to handle it and hence any application should be >> able to deal with that.

Yes they should, absolutely. But unfortunately they don't and thus I've thought it would be easyer to get i3 
handling this hint than implementing "workarounds" for applications that make use of this hint.

>> i3 only allows to span a fullscreen container over one or all monitors, nothing in between anyway.

What do I need to tell i3 to span a window over all screens? I did not knew that it is possible and it this would solve my problem at least to a certain degree.

>> I've only looked at the awesome source code, but awesome doesn't handle this hint at all either. That >>suggests that it might be something other than this atom that is relevant for your use case.
>>
[...]

PR #2213, opened 2016-02-14
Simplify multi-monitor presentations

wentasah:
This is a followup to #2200. I addressed the comments there and also fixed problems I encountered with the previous version (look for Okular in the commit message). I also rebased the patches to next (hence the new PR). I run the testsuite, but it failed even without my commits.

### DIFF
--- a/include/randr.h
+++ b/include/randr.h
@@ -79,6 +79,13 @@ Output *get_output_by_name(const char *name);
  */
 Output *get_output_containing(unsigned int x, unsigned int y);
 
+/**
+ * Returns the active output which spans exactly the area specified by
+ * rect or NULL if there is no output like this.
+ *
+ */
+Output *get_output_with_dimensions(Rect rect);
+
 /*
  * In contained_by_output, we check if any active output contains part of the container.
  * We do this by checking if the output rect is intersected by the Rect.
--- a/src/con.c
+++ b/src/con.c
@@ -1158,6 +1158,19 @@ void con_move_to_workspace(Con *con, Con *workspace, bool fix_coordinates, bool 
 }
 
 /*
+ * Moves the given container to the currently focused container on the
+ * visible workspace on the given output.
+ *
+ */
+void con_move_to_output(Con *con, Output *output) {
+    Con *ws = NULL;
+    GREP_FIRST(ws, output_get_content(output->con), workspace_is_visible(child));
+    assert(ws != NULL);
+    DLOG("Moving con %p to output %s\n", con, output->name);
+    con_move_to_workspace(con, ws, false, false, false);
+}
+
+/*
  * Returns the orientation of the given container (for stacked containers,
  * vertical orientation is used regardless of the actual orientation of the
  * container).
--- a/src/manage.c
+++ b/src/manage.c
@@ -359,8 +359,16 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     if (xcb_reply_contains_atom(state_reply, A__NET_WM_STATE_FULLSCREEN)) {
         /* If this window is already fullscreen (after restarting!), skip
          * toggling fullscreen, that would drop it out of fullscreen mode. */
-        if (fs != nc)
+        if (fs != nc) {
+            Output *output = get_output_with_dimensions((Rect){geom->x, geom->y, geom->width, geom->height});
+            /* If the requested window geometry spans the whole area
+             * of an output, move the window to that output. This is
+             * needed e.g. for LibreOffice Impress multi-monitor
+             * presentations to work out of the box. */
+            if (output != NULL)
+                con_move_to_output(nc, output);
             con_toggle_fullscreen(nc, CF_OUTPUT);
+        }
         fs = NULL;
     }
 
--- a/include/con.h
+++ b/include/con.h
@@ -269,6 +269,13 @@ void con_move_to_workspace(Con *con, Con *workspace, bool fix_coordinates,
                            bool dont_warp, bool ignore_focus);
 
 /**
+ * Moves the given container to the currently focused container on the
+ * visible workspace on the given output.
+ *
+ */
+void con_move_to_output(Con *con, Output *output);
+
+/**
  * Moves the given container to the given mark.
  *
  */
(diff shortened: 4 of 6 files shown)


wentasah:
This also seems to fix the problem with freerdp mentioned in  #2116.

include/randr.h
```diff
@@ -79,6 +79,13 @@ Output *get_output_by_name(const char *name);
  */
 Output *get_output_containing(unsigned int x, unsigned int y);
 
+/**
+ * Returns the active output which spans exactly the area specified by
+ * rect or NULL if there is no output like this.
+ *
+ */
+Output *get_output_by_rect(Rect rect);
```
stapelberg: I think the name of this function could be a bit clearer. Currently, from reading just the name, the function might also return the first output which contains (parts of, or the entirety of) the given rect.

How about `get_output_with_dimensions(Rect rect)` instead?

src/con.c
```diff
@@ -1158,6 +1158,21 @@ void con_move_to_workspace(Con *con, Con *workspace, bool fix_coordinates, bool
...
 
 /*
+ * Moves the given container to the currently focused container on the
+ * visible workspace on the given output.
+ *
+ */
+void con_move_to_output(Con *con, Output *output) {
+    /* fs is either the visible workspace or fullscreen window on the
+     * given output */
+    Con *fs = con_get_fullscreen_con(output_get_content(output->con), CF_OUTPUT);
```
stapelberg: The following pattern is more consistent with the i3 code base:

```c
Con *ws = NULL;
GREP_FIRST(ws, output_get_content(output->con), workspace_is_visible(child));
```

src/manage.c
```diff
@@ -359,8 +359,16 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     if (xcb_reply_contains_atom(state_reply, A__NET_WM_STATE_FULLSCREEN)) {
         /* If this window is already fullscreen (after restarting!), skip
          * toggling fullscreen, that would drop it out of fullscreen mode. */
-        if (fs != nc)
+        if (fs != nc) {
+            Output *output = get_output_by_rect((Rect){geom->x, geom->y, geom->width, geom->height});
+            /* If the requested window geometry spans the whole area
+             * of an output, move the window to that output. This is
+             * needed for multi-monitor presentations to work out of
```
stapelberg: Can you include the name of the program (LibreOffice?) with which you tested this? “multi-monitor presentations” in general could be implemented using override_redirect, hence bypassing the window manager entirely. So, strictly speaking, it’s not necessary for the window manager to cooperate :).

src/handlers.c
```diff
@@ -719,6 +719,21 @@ static void handle_client_message(xcb_client_message_event_t *event) {
...
+                     * output, move the window to that output. This is
+                     * needed for multi-monitor presentations to work
+                     * out of the box. */
+                    Output *output = get_output_by_rect(con->geometry);
+                    struct timeval second_ago;
+
+                    gettimeofday(&second_ago, NULL);
+                    second_ago.tv_sec--;
+
+                    if (output && _i3_timercmp(con->map_time, second_ago, >= ))
```
stapelberg: I dislike the fact that time is involved in this piece of code.

Can we remove this code entirely instead? Applications should map their window with the proper geometry and hints, I’d say.

include/randr.h
Airblader: :) We talked about the name before, this name was my quick suggestion. I do like yours even better.

src/handlers.c
Airblader: Yes, this definitely can't be in here. If LibreOffice doesn't set the hints properly, it's a bug against LibreOffice. It sets the geometry to signal fullscreen, so if it wants fullscreen, it should set that hint before mapping.

src/manage.c
```diff
@@ -359,8 +359,16 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
...
+             * the box. */
+            if (output)
```
Airblader: @stapelberg I'm personally a fan of being explicit and writing `!= NULL`. I think this is currently not consistent in the codebase. What do you prefer? I'd like to get this more consistent.

Airblader:
We'll also need tests for this.

src/handlers.c
wentasah: LibreOffice (at least recent versions) does not need this. It sets the fullscreen hint before mapping its windows. I need this for pdfpc. I'll file a bug against it.

wentasah:
Somewhere, I read that Xephyr does not support multiple outputs. How to
test this feature then? If there is some way, can you recommend me a
similar existing test from which I can start?

Airblader:
i3 has "fake outputs" used for tests. Just search for "fake-outputs" in the testcases. The `new-test` script in `testcases/` also takes a `-m` flag for this.

wentasah:
This is an updated version addressing the above comments. It includes a new test and is rebased onto current "next".

src/con.c
```diff
@@ -1158,6 +1158,22 @@ void con_move_to_workspace(Con *con, Con *workspace, bool fix_coordinates, bool
...
+ *
+ */
+void con_move_to_output(Con *con, Output *output) {
+    Con *ws = NULL;
+    GREP_FIRST(ws, output_get_content(output->con), workspace_is_visible(child));
+    if (ws != NULL) {
+        DLOG("Moving con %p to output %s\n", con, output->name);
+        con_move_to_workspace(con, ws, false, false, false);
+    } else {
+        DLOG("No visible workspace on output %s\n", output->name);
```
Airblader: I think we should use `assert(ws != NULL)` here, or at least log this as an error. We should *always* have a visible workspace on an active output.

wentasah: I also think that the else part cannot happen, but in all other places in the code, the else branch was present. Anyway, I'm pushing a new version with assert.

wentasah:
Ping. I there anything I can do to move this patch further?

wentasah:
Ping. I there anything I can do to move this patch further?

stapelberg:
Sorry, this fell through the cracks. Merged now, thanks!

PR #2219, opened 2016-02-21
Don't add tray_output primary for generated example i3 config

rtlanceroad:
With "tray_output primary" in config file generated by i3-config-wizard, the 3rd apps, like nm-applet,hexchat,dropbox doesn't show in tray. So let's users choose it.

### DIFF
--- a/i3.config
+++ b/i3.config
@@ -165,7 +165,6 @@ bindsym Mod1+r mode "resize"
 # finds out, if available)
 bar {
         status_command i3status
-        tray_output primary
 }
 
 #######################################################################
--- a/i3.config.keycodes
+++ b/i3.config.keycodes
@@ -152,5 +152,4 @@ bindcode $mod+27 mode "resize"
 # finds out, if available)
 bar {
         status_command i3status
-        tray_output primary
 }


Airblader:
What do you mean with that they don't show up in the tray? They will show, if the user configured their primary screen correctly. They should also show on the first output if no primary screen is set. 

Everything else is a bug, but a bug we should not fix by removing this line.

rtlanceroad:
If a user installs i3 4.11 for the first time, i3-config-wizard will ask you to generate ~/.config/i3/config, here are some of my autostart programs:

## Autostart apps
exec --no-startup-id nitrogen --restore
exec --no-startup-id hexchat
exec --no-startup-id nm-applet
....

These programs should autostart with icons in the tray. But with  tray_output primary , it ooesn't show.

Airblader:
Like I said, `tray_output primary` should show the tray icons on the output marked as the primary output in xrandr. If no primary output exists, it should fall back to the first available output. So the correct behavior in both cases would be to show icons.

If this doesn't happen, please report it as a bug. Make sure you reproduce the problem on the actual `next` branch (*not* using the 4.11 release) since we had big changes in the tray output handling since.

I'm going to close this PR as it does not properly fix anything.

rtlanceroad:
Sure

Airblader:
@rtlanceroad I've looked into the code and I think I know what causes the bug. If you open the ticket I'll try to provide a patch for you to see whether it fixes the issue.

rtlanceroad:
@Airblader I'll try your patch.

Airblader:
Thinking more about it, I actually think that we should be removing the (non-working) fallback behavior to the first available output if "primary" is set, but no primary output is present. The userguide currently explicitly says that users should use `tray_output primary` in all their bar blocks if they have more than one of them. However, this would in fact lead to a race condition since every bar instance will try to fall back to the first available output it manages.

In this case, the proper fix might in fact be removing `tray_output primary` from the default config, but also remove the "dead" part of the fallback behavior. See also #1855 for this.

I'm going to reopen the PR to remove the config part. We'll also need to remove the dead code.

@stapelberg Does that sound right to you? We could also actually make the fallback work correctly, but this would be a breaking change (for anyone with multiple bar blocks which all specify "primary") and get us a lot of duplicate bug reports with the next release. So it's probably not a good idea.

Airblader:
@rtlanceroad I am going to open the ticket now so we have this properly documented.

rtlanceroad:
@Airblader I see the changes in User's Guide. Thank you very much.

PR #2221, opened 2016-02-21
Revert e2e7b7 and remove unreachable fallback code

### LINKED ISSUE #2220: "tray_output primary" does not fall back to the first available output
This issue originally stems from #2219 as reported by @rtlanceroad.

In #1855 we discussed that the i3bar code says it falls back to the first available output for the system tray in case `tray_output primary` was specified but there is no primary output. As it turns out, this code appears to be misleading.

We have the "sanity check" code that decides whether the tray should even be configured for a bar instance. If `tray_output primary` is set, it will only initialize the tray if the primary output is contained in the list of outputs managed by this bar instance. This effectively prevents the fallback behavior from working since the code will never be reached.

There are two ways we can fix this:

* We remove the dead fallback behavior code. If a user specified `tray_output primary` and no primary output exists, then no tray will be shown.
* We make the fallback behavior work by initializing the tray if `tray_output primary` is set, ignoring whether the primary output even exists in the list of managed outputs.

However, we are basically forced to go with the first solution. The userguide currently states that users with multiple bar blocks should use `tray_output primary` in all of them (or explicitly set `tray_output none` in all but one) to avoid race conditions. If we implemented the fallback correctly, users with such a configuration would in fact see the very race condition we wanted to avoid by that as every bar instance will be trying to either put the tray on the primary or its first available output.

Hence, this bug should be fixed with the following changes:

* Remove the `tray_output primary` from the default config added by e2e7b70d in #1855. We should remove it since it's very common for users not to have a primary output and it leads to no tray being shown.
* Remove the "dead" fallback code to avoid further confusion as it happened in #1855. Note that the code referenced there has since (after 4.11) changed and now looks different.

Airblader:
fixes #2220

### DIFF
--- a/i3.config
+++ b/i3.config
@@ -165,7 +165,6 @@ bindsym Mod1+r mode "resize"
 # finds out, if available)
 bar {
         status_command i3status
-        tray_output primary
 }
 
 #######################################################################
--- a/i3.config.keycodes
+++ b/i3.config.keycodes
@@ -152,5 +152,4 @@ bindcode $mod+27 mode "resize"
 # finds out, if available)
 bar {
         status_command i3status
-        tray_output primary
 }
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -1761,16 +1761,35 @@ void reconfig_windows(bool redraw_bars) {
             }
 
             /* Unless "tray_output none" was specified, we need to initialize the tray. */
-            const char *first = (TAILQ_EMPTY(&(config.tray_outputs))) ? SLIST_FIRST(outputs)->name : TAILQ_FIRST(&(config.tray_outputs))->output;
-            if (!tray_configured && strcasecmp(first, "none") != 0) {
-                /* We do a sanity check here to ensure that this i3bar instance actually handles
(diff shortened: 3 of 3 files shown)


Airblader:
@rtlanceroad Just for good measure, can you please apply this patch and make sure everything still works as expected?

Airblader:
I've locally tested myself that the following things still work:

* Specific tray_outputs make the tray appear.
* Not specifying tray_output makes the tray appear (first available output).
* Specifying tray_output primary with no primary output causes no tray to show up.
* Specifying tray_output primary with a primary output makes the tray appear.

Airblader:
… And just for informational purposes: The dead fallback code was introduced in 2012 before the "sanity check" to avoid race conditions existed. Introducing the sanity check killed the fallback behavior for primary.

rtlanceroad:
@Airblader  Sorry for my late reply. I tested on WM with pure Arch Linux & i3 4.11 built from abs && patches from you. Now it looks good to me. The tray just shows up. I dont have other monitors to test that in multiple monitor case. Would be nice to see the testing result from other people. : ) Thanks for the patch.

PR #2230, opened 2016-02-23
Update the number of atoms in _NET_SUPPORTED.

Airblader:
When _NET_WM_WINDOW_TYPE_NOTIFICATION support was introduced in
33f017daa9f6c48e0d9c5bcb6135b47a0b32bf76, updating the number of
atoms to be written into _NET_SUPPORTED has been forgotten.


Airblader:
@ccryx A little late, but just FYI. :)

PR #2231, opened 2016-02-23
Introduce I3_FLOATING_WINDOW.

### LINKED ISSUE #2223: [enhancement] Allow compositors to identify floating windows
While we don't officially support compositing, I think it's an indisputable fact that many users use i3 with a compositor. For a while I tried to see whether there is anything in the specs that would allow a compositor to tell floating and tiling windows apart, but with little luck.

I propose that we introduce a "proprietary" atom that we maintain on windows to indicate this state. Maybe something like `I3_FLOATING_WINDOW`. This would allow users to select such windows conditionally in their compositor configurations.

Example usecases for this are drop shadows on floating windows or transparency changes for floating windows.

I think this would be a nice-to-have feature for a large group of people while at the same time cheap to maintain.

Airblader:
This patch introduces a proprietary atom I3_FLOATING_WINDOW which will be
set and maintained for floating windows and removed on tiling containers.

This allows users to select on this atom, e.g., in their compositor
configuration or in utility scripts (without using the IPC).

fixes #2223


include/atoms.xmacro
```diff
@@ -46,6 +46,7 @@ xmacro(I3_CONFIG_PATH)
 xmacro(I3_SYNC)
 xmacro(I3_SHMLOG_PATH)
 xmacro(I3_PID)
+xmacro(I3_FLOATING_WINDOW)
```
Airblader: Note that we don't update the number of atoms in `src/ewmh.c` since this is nothing that belongs in `_NET_SUPPORTED`.

src/floating.c
```diff
@@ -257,22 +285,20 @@ void floating_enable(Con *con, bool automatic) {
     if (set_focus)
         con_focus(con);
 
+    bool needs_center = true;
```
stapelberg: I think this change belongs into a separate PR?

stapelberg:
Can you rebase this please? I’ve pushed fdbbae56b2b5f8049380b9456c01c3966b3b18f2 in the meantime.

src/floating.c
Airblader: @stapelberg Why? I did this refactoring here so there's only one return path and I only need to add the call to the new function once. I could've added it above all the return statements, but I liked it better right before sending the IPC message just like in the floating_disable case.

Airblader:
Rebased

src/floating.c
stapelberg: Didn’t realize this is a no-op refactoring while quickly scanning it earlier.

I think using “goto done” would be more consistent with other places in the code — I think it makes it a bit more obvious how the control flow works.

Airblader:
@stapelberg Done.

PR #2239, opened 2016-03-02
window_update_motif_hints(): uint32_t for fields, fixes #2238

### LINKED ISSUE #2238: Patch from #832 adding _MOTIF_WM_HINTS support is incorrect
Hi, I have a suspicion that the patch from issue #832 adding the `_MOTIF_WM_HINTS` support is incorrect, works by chance and might break with different endianness of the system. At https://github.com/i3/i3/blob/f14a94031c6b33a11715bfc4e0273cc05d981ae6/src/window.c#L341 there is:
```
uint64_t *motif_hints = (uint64_t *)xcb_get_property_value(prop);
[...]
if (motif_hints[1] & MWM_DECOR_ALL || motif_hints[1] & MWM_DECOR_TITLE)
```
It suggests that decorations are stored in element [1], but sources (even the one quoted by the window.c file) show that it's actually the [2] element of the property:
https://people.gnome.org/~tthurman/docs/metacity/xprops_8h-source.html
https://git.gnome.org/browse/gtk+/tree/gdk/x11/MwmUtil.h
Also experiments with GTK+ and xprop confirm that.

Although X11 on amd64 can internally use 64-bit unsigned long for fields described as 32-bit in the old documentation and 64-bit fields are valid for calls such as Xlib `XChangeProperty()` (https://git.gnome.org/browse/gtk+/tree/gdk/x11/gdkwindow-x11.c#n4415), the XCB `xcb_get_property_value()` function returns data in protocol format, not in local representation format. See:
https://lists.gnu.org/archive/html/emacs-diffs/2015-11/msg00199.html
In the XWayland code the `motif_wm_hints` structure consists of `uint32_t` (instead of `uint64_t`) elements to work directly with the `xcb_get_property_value()` function. See:
https://cgit.freedesktop.org/wayland/weston/tree/xwayland/window-manager.c#n505

wodny:
xcb_get_property_value() returns 32-bit property fields

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -320,6 +320,9 @@ void window_update_motif_hints(i3Window *win, xcb_get_property_reply_t *prop, bo
      * https://people.gnome.org/~tthurman/docs/metacity/xprops_8h-source.html
      * http://stackoverflow.com/questions/13787553/detect-if-a-x11-window-has-decorations
      */
+#define MWM_HINTS_FLAGS_FIELD 0
+#define MWM_HINTS_DECORATIONS_FIELD 2
+
 #define MWM_HINTS_DECORATIONS (1 << 1)
 #define MWM_DECOR_ALL (1 << 0)
 #define MWM_DECOR_BORDER (1 << 1)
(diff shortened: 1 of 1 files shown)


src/window.c
```diff
@@ -333,17 +333,20 @@ void window_update_motif_hints(i3Window *win, xcb_get_property_reply_t *prop, bo
...
-    uint64_t *motif_hints = (uint64_t *)xcb_get_property_value(prop);
+     * a particular decoration. Notice that X11 (Xlib) often mentions 32-bit fields
+     * which in reality are implemented using unsigned long variables (64-bits long
+     * on amd64 for example). On the other hand, xcb_get_property_value() behaves
+     * strictly according to documentation, i.e. returns 32-bit data fields. */
+    uint32_t *motif_hints = (uint32_t *)xcb_get_property_value(prop);
 
     if (motif_border_style != NULL && motif_hints[0] & MWM_HINTS_DECORATIONS) {
-        if (motif_hints[1] & MWM_DECOR_ALL || motif_hints[1] & MWM_DECOR_TITLE)
+        if (motif_hints[2] & MWM_DECOR_ALL || motif_hints[2] & MWM_DECOR_TITLE)
```
Airblader: Could we please define named values for the indices (0 and 2) here, similar to how we did it for `MWH_HINTS_DECORATIONS`? That way the code is a bit more self-explanatory.

Airblader:
It'd be great if you could change the commit message to include a "fixes #2238" in the end. That way the issue is closed automatically after merge and we have a reference to the ticket.

wodny:
@Airblader I've force-updated the last commit to "change the commit message to include a »fixes #2238«". I don't know if that's the best practice or if I should have added another commit. There are also the additional defines.

Airblader:
@wodny Yes, force-pushing is the right approach for the PR, we want to keep the history clean and not a bunch of commis for a single change.

One last thing (from my side), the code is currently not line-breaking after 80 characters. Can you just run clang3.5-format (note the version) on the changed files? The Travis check will probably fail because of that as well.

wodny:
@Airblader I'm unable to force clang3.5-format to rewrap the comment. Does it require any special options?

wodny:
@Airblader I've updated the ColumnLimit setting to 80, but it results in function parameters rewrapping. Is it a desired formatting?

Airblader:
@wodny You need to run `clang3.5-format -i src/window.c`, that will in-place reformat the file.

@stapelberg Travis seems to have failed for an unrelated reason.

Airblader:
@wodny No, please don't reformat the unrelated places. Those should be formatted correctly as they are. `clang3.5-format -i` is what will do the trick, this is what Travis will do as well. Sorry that this is taking up most of the time… :)

wodny:
@Airblader Unfortunately I've never used clang-format before. A couple of runs of `clang-format-3.5 -i -lines 323:367 window.c` with different `-style` variant hopefully did the trick. Repushed forcefully.

BTW, the included `.clang-format` file doesn't force the wrap at the 80th column.

Airblader:
Either way, it now looks correctly formatted to me now. :) Thanks for your patience.

PR #2242, opened 2016-03-07
fix link after change of id

pniederlag:
fix broken link inside documentation


stapelberg:
Can you also add the old id please so that external links keep working?

Airblader:
Superseded by #2462.

PR #2244, opened 2016-03-08
Make i3.applications.desktop file more complete

jubalh:
Add Encoding and GenericName to i3.applications.desktop file.

### DIFF
--- a/i3.applications.desktop
+++ b/i3.applications.desktop
@@ -2,6 +2,7 @@
 Type=Application
 Name=i3
 NoDisplay=true
+GenericName=A dynamic tiling window manager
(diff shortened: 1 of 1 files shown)


i3.applications.desktop
```diff
@@ -1,7 +1,9 @@
 [Desktop Entry]
+Encoding=UTF-8
```
stapelberg: I don’t think we should be adding this:

```
$ desktop-file-validate i3.applications.desktop
i3.applications.desktop: warning: key "Encoding" in group "Desktop Entry" is deprecated
```

jubalh:
Alright! Removed that one

PR #2245, opened 2016-03-08
Update information on where to find #i3 on IRC

Osse:
I was unsure on whether to update the information in the slides since maybe the should remain as how they were presented, but I saw in the log that they have received similar updates before.


stapelberg:
I’ll gladly merge the change to the `PACKAGE-MAINTAINER` file, but the presentations are historical documents and should be treated as such — they aren’t kept up to date and contain some other mistakes/inaccurate information.

Can you change the PR to only touch the `PACKAGE-MAINTAINER` file please?

Osse:
That was my suspicion; no problem.

PR #2264, opened 2016-03-19
Fix typos in documentation.

ixjlyons:
Trivial fix but at least I'm able to contribute somehow :)

edit: Actually, I went ahead and checked the spelling on all of the top-level docs and caught a few more things. Would you prefer them all in one PR or separate ones for each file?

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1372,7 +1372,7 @@ NetworkManager, VLC, Pidgin, etc. can place little icons.
 You can configure on which output (monitor) the icons should be displayed or
 you can turn off the functionality entirely.
 
-You can use mutliple +tray_output+ directives in your config to specify a list
+You can use multiple +tray_output+ directives in your config to specify a list
 of outputs on which you want the tray to appear. The first available output in
 that list as defined by the order of the directives will be used for the tray
 output.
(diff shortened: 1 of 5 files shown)


jolange:
I guess one PR for doc typos should be fine.

ixjlyons:
Ok, I edited the title to reflect more general changes and squashed down to one commit.

docs/userguide
```diff
@@ -1423,7 +1423,7 @@ tray_padding <px> [px]
 
 *Example*:
 -------------------------
-# Obey Fitts's law
+# Obey Fitts' law
```
jolange: Actually, Wikipedia says
> Fitts's law (often cited as Fitts' law) is...

ixjlyons: Interesting. I actually do research related to human computer interfaces and I've never seen it written as `Fitts's law` (I've seen it written as `Fitt's law`, but that's definitely incorrect). There doesn't seem to be a definitive consensus one way or the other regarding possessive form of proper names ending in "s" [[1]], but I'm in favor of fewer changes.

[1]: http://www.dailywritingtips.com/possessive-of-proper-names-ending-in-s/

ixjlyons:
I don't think there are any controversial changes remaining :) Should be good to go.

PR #2265, opened 2016-03-19
i3-dmenu-desktop: Remove quotes around exec cmd

mikkeloscar:
The reason for this patch is discussed more in depth here: SirCmpwn/sway#521

The real issue is that the `exec` command in i3 doesn't treat quoted arguments the same way as other commands do. For instance `exec "firefox        " http://google.com` is the same as `exec firefox http://google.com` in i3. This is an unfortunate design, but something that most likely won't change because of backwards compatibility.

Sway aims to be compatible with i3, but we would like to avoid implementing this behavior if it doesn't break too much for the users.

This patch solves some of the issue by making `i3-dmenu-desktop` compatible with sway, without affecting how it works with i3.

### DIFF
--- a/i3-dmenu-desktop
+++ b/i3-dmenu-desktop
@@ -444,7 +444,7 @@ EOT
     if (exists($app->{StartupNotify}) && !$app->{StartupNotify}) {
         $nosn = '--no-startup-id';
     }
-    $cmd = qq|exec $nosn "$exec"|;
+    $cmd = qq|exec $nosn $exec|;
 }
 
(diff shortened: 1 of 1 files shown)


stapelberg:
I don’t follow the discussion in the referenced issue (might be because I’m tired from jet lag). Could you do me and other readers of this commit a favor and document the rationale for this commit in the commit description itself please? Currently it just says that the quotes are unnecessary but doesn’t expand on this. Why are they unnecessary? What’s broken currently? With this change, what could break because of the change?

mikkeloscar:
I have realized that this won't work for some edge cases in i3, so I'll close it again. Sorry about that.

Issue #2141, opened 2015-12-30
Irregular Behavior with Mod Key in FreeBSD


spython01:
Note:  This issue was first reported on the i3wm FAQ page over on [reddit](https://www.reddit.com/r/i3wm/comments/3ykfw0/irregular_mod_key_behavior_with_freebsd_in/).  I'm running i3wm (version 4.10.4) on FreeBSD 10.2 as a guest on VirtualBox 5.0.2 and am seeing some irregular behavior with my Mod key, which is set to `Mod1` (the `<Alt>` key). (Right `<Ctrl>` is my VirtualBox Host Key.)  The steps I took to install i3 on my virtual machine can be found [here](http://bottlenix.wikidot.com/installing-i3wm).

Some functions work properly:

`$mod+Return` brings up a new terminal window

`$mod+j` focuses left, `$mod+;` focuses right

`$mod+Shift+q` closes a window

Other functions, however, do not do anything:

`$mod+d` does not bring up the dmenu

`$mod+Shift+e` does not exit i3wm (in fact, it brings up the "Edit" menu in Firefox instead -- but this was eventually fixed as described below).

My original `~/.i3/config` file, which was created by default, can be found [here](http://pastebin.com/raw/01VtJr3k).  The reddit thread confirmed that this was only a partial file.  I then ran this command:

    cp /usr/local/etc/i3/config ~/.i3/

which gave me this more complete `config` [file](http://pastebin.com/raw/ccW571zk)

So I believe that there are at least two issues:
1. `$mod+d` does not bring up the dmenu, even after I copied over the more complete one from `/usr/local/etc/i3`
2. The `config` file that is generated by default in FreeBSD is not complete.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

LuaKT:
Just checking the obvious, you have installed dmenu?

samirparikh:
I am the **biggest** idiot in the world.  No I did not so I just ran

    pkg install dmenu

which seems to have fixed that issue.  I still think there is a bug with the creation of the `config` file however.  Thank you @LuaKT !

spython01:
I am the **biggest** idiot in the world. No I did not so I just ran

    pkg install dmenu

[...]

PR #2267, opened 2016-03-21
Pevent that i3-config-wizard generates incomplete config file

Chr1stoph:
Dear i3 team,
i3-config-wizard is a tool of x11-wm/i3 to generate a default configuration file. The last lines of a newly generated file end with

\# alternatively, you can use the cursor keys:
bindsym $mod+ISO_Level3_Shift focus left
bindsym $mod+Super_R focus down
bindsym $mod+Print focus up

The pull request does not stop at non existing keysymbols. it replace what is possible and leaves the codes for non existing symbols untouched. Please find the patch as text file.
[patch-i3-config-wizard_main.txt](https://github.com/i3/i3/files/182974/patch-i3-config-wizard_main.txt)

Thank you for your kind consideration,
Christoph


Airblader:
Please open it as a proper pull request, not an attached patch file. Also, make sure to reference ticket #2141 in the commit message. Lastly, in the ticket discussion we discussed how we want to fix this. It seems like your patch doesn't do that at all.

PR #2269, opened 2016-03-22
Debugging procedure hanging i3 add.

joepvd:
Credits go to Ingo in message
https://www.freelists.org/post/i3-discuss/Seeing-a-hang-on-i3-version-412,3

### DIFF
--- a/docs/debugging
+++ b/docs/debugging
@@ -147,6 +147,8 @@ After pressing "b" in the crash dialog, you will get a file called
 id (PID) and the second one is incremented each time you generate a backtrace,
 starting at 0.
 
+== On hangs: Obtaining a backtrace from a virtual console
+
 When the crash dialogue is not shown, for example in the case of a hang, you
 need to attach +gdb+ to the running i3.  To do so, switch to another TTY 
 (Ctrl + Alt + F2), log in and run
--- a/docs/debugging
+++ b/docs/debugging
@@ -147,6 +147,28 @@ After pressing "b" in the crash dialog, you will get a file called
 id (PID) and the second one is incremented each time you generate a backtrace,
 starting at 0.
 
+When the crash dialogue is not shown, for example in the case of a hang, you
+need to attach +gdb+ to the running i3.  To do so, switch to another TTY 
+(Ctrl + Alt + F2), log in and run
+
+------------------------------------------------------------------------------
+$ gdb $(which i3) $(pidof i3)
+------------------------------------------------------------------------------
+
+(Note that this may require sudo)
+This should attach gdb to the (only running) i3 instance. Now enable logging
+to a file and then get the current backtrace by entering this in your gdb
+shell:
+
+------------------------------------------------------------------------------
+set logging file /tmp/i3hang.log
+set logging on
+bt full
+------------------------------------------------------------------------------
+
+You can now quit gdb again and switch back to your main TTY (with the hanging
+i3 session).
+
 == Sending bug reports/debugging on IRC
 
 When sending bug reports, please attach the *whole* log file. Even if you think


docs/debugging
```diff
@@ -147,6 +147,28 @@ After pressing "b" in the crash dialog, you will get a file called
...
 
+When the crash dialogue is not shown, for example in the case of a hang, you
+need to attach +gdb+ to the running i3.  To do so, switch to another TTY 
+(Ctrl + Alt + F2), log in and run
+
+------------------------------------------------------------------------------
+$ gdb $(which i3) $(pidof i3)
+------------------------------------------------------------------------------
+
+(Note that this may require sudo)
```
stapelberg: Under which circumstances specifically does this require sudo? gdb doesn’t need elevated privileges AFAIK and i3 runs as your user…?

docs/debugging
```diff
@@ -147,6 +147,28 @@ After pressing "b" in the crash dialog, you will get a file called
...
+This should attach gdb to the (only running) i3 instance. Now enable logging
+to a file and then get the current backtrace by entering this in your gdb
+shell:
+
+------------------------------------------------------------------------------
```
stapelberg: Is there a way to do this without describing an interactive procedure? gdb has a `--batch` option and `--command`, perhaps these could be used?

docs/debugging
```diff
@@ -147,6 +147,28 @@ After pressing "b" in the crash dialog, you will get a file called
 id (PID) and the second one is incremented each time you generate a backtrace,
 starting at 0.
 
+When the crash dialogue is not shown, for example in the case of a hang, you
```
stapelberg: I think it might be clearer to put this into its on section, e.g. `== On hangs: Obtaining a backtrace from a virtual console`

docs/debugging
Airblader: Frankly I don't know the mechanics here, but if I run gdb without sudo for the running session, my backtrace will only say "No stack."

docs/debugging
joepvd: `--batch` and `--command` both need a file as an argument. The standard way to send lines from the current terminal to `gdb` is with usual shell redirection, so something like this:

    cat <<EOF | sudo gdb $(which i3) $(pidof i3)
    set logging file /tmp/i3hang_$(date +%F-%T).log
    set logging on
    backtrace full
    EOF

Not sure if such an unwieldy command would be handy here; I think the feedback that shell / gdb would provide after each single line would be help with fixing typos. 

Please let me know if you disagree.

docs/debugging
stapelberg: Also, I just spotted that your spelling is “dialogue”, whereas elsewhere in the document we’re using “dialog”. Can you change it to “dialog” for consistency please?

docs/debugging
stapelberg: `gdb --batch $(which i3) $(pidof i3) -ex 'backtrace full' > /tmp/i3hang_$(date +%F-%T).log` works for me.

docs/debugging
stapelberg: It works fine for me without sudo. Can you provide the full output without sudo please?

Airblader: Sorry, I thought I had replied already, but must've been distracted. Here it is:

````
gdb $(which i3) $(pidof i3)
GNU gdb (GDB) 7.11
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/bin/i3...done.
Attaching to program: /usr/bin/i3, process 815
ptrace: Operation not permitted.
/home/ingo/815: No such file or directory.
(gdb) 
```

And having googled the failure I feel like I should point out that I do *not* use Ubuntu. I run Manjaro on a 4.5 kernel.

stapelberg: I’m guessing `/proc/sys/kernel/yama/ptrace_scope` contains `1` on your machine.

How about the following explanation instead of “this may require sudo”?

```markdown
On modern Linux systems,
[the Yama Linux Security Module](https://www.kernel.org/doc/Documentation/security/Yama.txt)
prevents processes from tracing each other. Use `sudo` to get the above gdb invocation to work.
```

Airblader: SGTM

Airblader:
@joepvd Are you going to keep working on this?

PR #2277, opened 2016-04-02
Point out the default ipc message type in i3-msg.man

xuhdev:


man/i3-msg.man
```diff
@@ -29,7 +29,7 @@ from the root window and then try /tmp/i3-ipc.sock before exiting
 with an error.
 
 *-t* 'type'::
-Send ipc message, see below.
+Send ipc message, see below. Default to "command".
```
stapelberg: nit: I suggest “This option defaults to "command".” — “Default to” reads a bit strange to me.

xuhdev:
@stapelberg Done!

Issue #1993, opened 2015-10-12
Release tarballs on github not buildable due to missing {I3_}VERSION


sur5r:
When $(TOPDIR) does not contain the {I3_}VERSION files, the build system assumes it's building from a git release and tries to infer version information from git. The tarballs on github however, do not contain those files which results in a rather strage build error:

    fatal: Not a git repository (or any parent up to mount point /tmp)
    Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
    fatal: Not a git repository (or any parent up to mount point /tmp)
    Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
    fatal: Not a git repository (or any parent up to mount point /tmp)
    Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
    fatal: Not a git repository (or any parent up to mount point /tmp)
    Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
    [i3] CC src/window.c
    [i3] CC src/util.c
    [i3] CC src/ipc.c
    In file included from ../i3-4.11/src/ipc.c:13:0:
    ../i3-4.11/src/ipc.c: In function ‘handle_get_version’:
    include/yajl_utils.h:17:50: error: expected expression before ‘)’ token
     #define y(x, ...) yajl_gen_##x(gen, ##__VA_ARGS__)
                                                      ^
    ../i3-4.11/src/ipc.c:845:5: note: in expansion of macro ‘y’
         y(integer, MAJOR_VERSION);
         ^
    include/yajl_utils.h:17:50: error: expected expression before ‘)’ token
     #define y(x, ...) yajl_gen_##x(gen, ##__VA_ARGS__)
                                                      ^
    ../i3-4.11/src/ipc.c:848:5: note: in expansion of macro ‘y’
         y(integer, MINOR_VERSION);
         ^
    src/i3.mk:40: recipe for target 'src/ipc.o' failed
    make: *** [src/ipc.o] Error 1

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
[...]

PR #2278, opened 2016-04-03
Version extraction

jolange:
Strategy to handle #1993:
- When creating a new release with the script, `VERSION` and `I3_VERSION` are updated and checked in, so these are always available in (github) release tarballs. After creating the release tag, `I3_VERSION` is updated to `X.xx-post-dev`, which will be used when downloading a non-release tarball.
- When building, git version information is used if possible. Otherwise (building from a tarball), the version files are used.

I manually added the current version files to make this work directly for non-release tarballs.


release.sh
```diff
@@ -56,6 +56,9 @@ cp "${STARTDIR}/RELEASE-NOTES-${RELEASE_VERSION}" "RELEASE-NOTES-${RELEASE_VERSI
 git add RELEASE-NOTES-${RELEASE_VERSION}
 git rm RELEASE-NOTES-${PREVIOUS_VERSION}
 sed -i "s,<refmiscinfo class=\"version\">[^<]*</refmiscinfo>,<refmiscinfo class=\"version\">${RELEASE_VERSION}</refmiscinfo>,g" man/asciidoc.conf
+echo "${RELEASE_VERSION}" > VERSION
+echo "${RELEASE_VERSION} ($(date +%F), branch \"$(git rev-parse --abbrev-ref HEAD)\")" > I3_VERSION
```
stapelberg: The branch isn’t important when doing a release, it’ll always be release-<version>. I’ve only kept it because it would’ve been more effort to change it, but given that we’re changing the mechanism anyway, can you just remote the branch here please?

release.sh
```diff
@@ -68,14 +71,13 @@ diff -u \
...
 
-if ! tar xf i3-${RELEASE_VERSION}.tar.bz2 --to-stdout --strip-components=1 i3-${RELEASE_VERSION}/I3_VERSION | grep -q "^${RELEASE_VERSION} "
-then
-	echo "I3_VERSION file does not start with ${RELEASE_VERSION}"
-	exit 1
-fi
 
 gpg --armor -b i3-${RELEASE_VERSION}.tar.bz2
 
+echo "${RELEASE_VERSION}-post-dev" > I3_VERSION
```
stapelberg: Nit: The “post-dev” suffix sounds too definitive to me. I’d like to use something which conveys that a version couldn’t be determined. Maybe the “unknown” suffix I originally suggested, or perhaps “from-unknown-tarball” to be more descriptive?

jolange: what about `-non-git`?

stapelberg: `-non-git` is okay with me.

jolange:
Ok, done. I'll squash the last two commits and rebase everything if you think this is good to go @stapelberg.

I3_VERSION
```diff
@@ -0,0 +1 @@
+4.12-post-dev
```
stapelberg: This still says post-dev, but should be non-git.

release.sh
```diff
@@ -68,14 +71,13 @@ diff -u \
...
-then
-	echo "I3_VERSION file does not start with ${RELEASE_VERSION}"
-	exit 1
-fi
 
 gpg --armor -b i3-${RELEASE_VERSION}.tar.bz2
 
+echo "${RELEASE_VERSION}-non-git" > I3_VERSION
+git add I3_VERSION
+git commit -a -m "post ${RELEASE_VERSION} development version"
```
stapelberg: The commit description could be a bit clearer. What about this:

Set non-git version to ${RELEASE_VERSION}-non-git.

stapelberg:
Commented on two nits, should be good to go after you’ve addressed these.

I3_VERSION
jolange: Sorry, forgot about this.

jolange:
@stapelberg done.

stapelberg:
Thanks!

PR #2281, opened 2016-04-06
Remove unmaintained CHANGELOG file from i3bar

Airblader:
@stapelberg On a related note, the LICENSE in i3bar also has a year range. I'm not sure whether we can just change that to "2009" like in all the other places, too, though, since the LICENSE is attributed to Axel Wagner?


stapelberg:
> I'm not sure whether we can just change that to "2009" like in all the other places, too, though, since the LICENSE is attributed to Axel Wagner?

We should change it to “© 2010 Axel Wagner”, I think. @Merovius, does that sound reasonable? See https://github.com/i3/i3/pull/1636 for context, we’ve changed copyright notices to use the starting year, not a range of years.

Merovius:
SGTM

Airblader:
#2284

PR #2286, opened 2016-04-11
Load X resources

Airblader:


Airblader:
I will have to adapt Travis for the time being to make it work. I'll also add another test.

Airblader:
I'm not entirely sure what's happening with Travis, but likely I'm doing something wrong. Unfortunately I have no experience with docker whatsoever. I'd appreciate some help. :)

Airblader:
@stapelberg I'm lost now. From what I can see, all packages and util-xrm are built and compiled just fine. Do you have any idea why the tests are stalling?

stapelberg:
It’s not obvious to me why that’s happening. Here are a couple of debug strategies that come to mind:

* Comment out all code which actually uses xcb-util-xrm to see if this might be a linking-related issue. Try commenting out all code/makefile changes mentioning xcb-util-xrm just to establish a sane test environment.
* Modify the complete-run.pl command to not run all tests, but only run a specific, simple test case.
* In case that still hangs, try building the docker container locally and running the tests in it.
* In case that works, create a separate GitHub repository for testing travis, configure credentials for Docker (use the `travis encrypt` command). That way, travis will push the docker container under your credentials and you can download it to see if you can reproduce it within the travis-built docker container (I’d be surprised).

Airblader:
@stapelberg  OK, so removing only the actual code using the library doesn't change anything, but removing the Makefile stuff fixes it (of course the test now fails).

Unfortunately I'm still without any idea as to what this means. I'm now adding it all back in and have it run with only a simple test case as proposed.

Airblader:
OK, with a single test it fails like this:

````
Test output:
[:100] t/002-i3-sync.t: starting
[:100] t/002-i3-sync.t: finished
output for t/002-i3-sync.t:
not ok 1 - No such atom (I3_SOCKET_PATH) at /usr/src/i3/testcases/lib/i3test.pm line 796.
# BEGIN failed--compilation aborted at ./t/002-i3-sync.t line 19.
# 
#   Failed test 'No such atom (I3_SOCKET_PATH) at /usr/src/i3/testcases/lib/i3test.pm line 796.
# BEGIN failed--compilation aborted at ./t/002-i3-sync.t line 19.
# '
#   at /usr/src/i3/testcases/lib/TestWorker.pm line 56.
+ xvfb-run ./complete-run.pl --parallel=1
xvfb-run: error: Xvfb failed to start
```

Which looks like i3 isn't running, i.e., it probably crashed. The tests run fine on my machine, however.

Airblader:
I'm thinking the issue is the linking. I assume the shared library simply can't be found when i3 is started.

stapelberg:
Try throwing in an ldconfig call.

Airblader:
I think I've fixed it. I simply had to specify the `--prefix` on autogen to get it installed in a path the loader already knows. Thanks for all the help. Let's hope the next push comes back green so we can talk about the actual change here. :-)

Airblader:
@stapelberg Victory at last.

src/config_parser.c
```diff
@@ -812,12 +815,56 @@ void start_config_error_nagbar(const char *configpath, bool has_errors) {
...
+static char *get_resource(char *name) {
+    if (conn == NULL) {
+        return NULL;
+    }
+
+    /* Load the resource database lazily. */
+    if (database == NULL) {
+        database = xcb_xrm_database_from_default(conn);
+
+        if (database == NULL) {
```
Airblader: Here's one thing that definitely needs improving; we don't want to try loading the database again and again for every resource if it's simply not available.

stapelberg: You could add another static variable (e.g. `database_attempted_load`) which gets set to `false` in `parse_config`, then to `true` in `get_resource`.

stapelberg:
Looks good overall.

I want to wait with merging this until the Debian package made it to testing, so that users will be able to install the auto-built packages.

src/config_parser.c
Airblader: I was going to just load an empty string database. That'll result in the fallback being used, too.

stapelberg: That’s also fine.

Airblader:
This currently breaks normal variables, I'll fix it.

Airblader:
OK, fixed now.

Airblader:
Reloading the config causes a crash, I'll look at it.

Airblader:
Done.

Airblader:
@stapelberg Out of curiosity, how long does it typically take a package proposal to make it into Debian testing?

stapelberg:
It’s pretty variable and depends mostly on how busy the ftp-masters are I think. Maybe https://ftp-master.debian.org/stat.html gives you a better feeling? I’d wait another week or two before I start pinging people about it :).

Airblader:
@stapelberg It has made it into Debian testing now. :-)

docs/userguide
```diff
@@ -700,6 +700,32 @@ absolutely no plans to change this. If you need a more dynamic configuration
...
+resource database to achieve an easily maintainable, consistent color theme
+across many X applications.
+
+Defining a resource will load this resource from the resource database and
+assign its value to the specified variable. A fallback must be specified in
+case the resource cannot be loaded from the database.
+
+*Syntax*:
+-------------------------------------------
+resource <resource_name> $<name> <fallback>
```
stapelberg: I’m thinking it might make sense to turn the order of arguments around, so that the directive is closer to `set`. How about `set_from_resource $<name> <resource_name> <fallback>`? Also, is fallback optional or required?

docs/userguide
```diff
@@ -700,6 +700,32 @@ absolutely no plans to change this. If you need a more dynamic configuration
...
+loaded from the X resource database.
+
+[[xresources]]
+=== X resources
+
+<<variables>> can also be created using a value configured in the X resource
+database. This is useful, for example, to avoid configuring color values within
+the i3 configuration. Instead, the values can be configured, once, in the X
+resource database to achieve an easily maintainable, consistent color theme
+across many X applications.
```
stapelberg: Could the example below be amended to show a minimum example of re-using the color theme across more than just i3?

travis/travis-base-386.Dockerfile
```diff
@@ -27,3 +27,15 @@ COPY debian/control /usr/src/i3-debian-packaging/control
 RUN linux32 apt-get update && \
     DEBIAN_FRONTEND=noninteractive mk-build-deps --install --remove --tool 'apt-get --no-install-recommends -y' /usr/src/i3-debian-packaging/control && \
     rm -rf /var/lib/apt/lists/*
+
+# Install xcb-util-xrm. This is a workaround until it is available in the
```
stapelberg: Here and in `travis/travis-base.Dockerfile`, this can be removed since libxcb-util-xrm made it to Debian testing, right?

stapelberg:
Can you squash https://github.com/i3/i3/pull/2286/commits/d11be18604e4e4bb6a8f98106ea6c037d438f064? It looks to me like it doesn’t need to be a separate commit.

travis/travis-base-386.Dockerfile
Airblader: I removed it in those files and added it to `debian/control`. Anything else I need to do?

Airblader:
I have squashed it all into a single commit now and made the changes.

docs/userguide
```diff
@@ -700,6 +700,38 @@ absolutely no plans to change this. If you need a more dynamic configuration
...
+----------------------------------------------------
+
+*Example*:
+--------------------------------------------------------------------------
+# The ~/.Xresources should contain a line such as
+#     *color0: #121212
+# and must be loaded properly, e.g., by using
+#     xrdb ~/.Xresources
+# Then this value can be used both by other clients (such as your terminal
+# emulator) and in i3 via
```
stapelberg: The wording is a bit too vague for my taste: it can be used, but how do I do that?

I think something like “This value is picked up by other applications (e.g. the URxvt terminal emulator) and you can use it in i3 as well by using:”?

travis/travis-base-ubuntu-386.Dockerfile
```diff
@@ -27,3 +27,15 @@ COPY debian/control /usr/src/i3-debian-packaging/control
...
     DEBIAN_FRONTEND=noninteractive mk-build-deps --install --remove --tool 'apt-get --no-install-recommends -y' /usr/src/i3-debian-packaging/control && \
     rm -rf /var/lib/apt/lists/*
+
+# Install xcb-util-xrm. This is a workaround until it is available in the
+# distribution packages.
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends xutils-dev
+RUN git clone --recursive https://github.com/Airblader/xcb-util-xrm.git && \
+    cd xcb-util-xrm && \
+    git submodule update --init && \
```
stapelberg: nit: `git submodule update --init` can be skipped because you’re using `git clone --recursive`

stapelberg:
Looks good overall.

Airblader:
@stapelberg Done and rebased

PR #2290, opened 2016-04-12
Only grab the mouse buttons that need to be grabbed.

### LINKED ISSUE #2271: Mouse buttons 11-14 bindsym not working
Since I updated to version 4.12 from 4.11 all my mouse bindings for buttons 11-14  stopped working. Basic mouse buttons bindings works, only those additional doesn't. Log here is with new config file with just `bindsym --whole-window button11 exec --no-startup-id xterm` added. What I mean stopped working is that xterm is not started.

```
Running i3 version: 4.12 (2016-03-06, branch "4.12") (pid 11412)
Loaded i3 config: /home/cynerd/.config/i3/config (Last modified: Thu 24 Mar 2016 05:14:00 PM CET, 1681 seconds ago)

```

`http://logs.i3wm.org/logs/5738275486564352.bz2`

Airblader:
This is a followup to #2049. While we had fixed that bug by only grabbing
buttons 4 and 5 if there is a whole-window binding for that button, this
did not consider buttons higher than 5 as found on many mice.

Therefore, we now ditch the special handling for scrollwheel buttons and
instead do the same for all buttons higher than 3.

fixes #2271

### DIFF
--- a/include/bindings.h
+++ b/include/bindings.h
@@ -104,10 +104,10 @@ CommandResult *run_binding(Binding *bind, Con *con);
 bool load_keymap(void);
 
 /**
- * Returns true if the current config has any binding to a scroll wheel button
- * (4 or 5) which is a whole-window binding.
- * We need this to figure out whether we should grab all buttons or just 1-3
- * when managing a window. See #2049.
- *
+ * Returns a list of buttons that should be grabbed on a window.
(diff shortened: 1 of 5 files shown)


Airblader:
Despite the tests we have for #2049 already I want to give this a manual test.

Airblader:
It has now been tested.

Issue #2247, opened 2016-03-09
Bug or feature?: title indent on `split vertical|horizontal`


Adaephon-GH:
Output of `i3 --moreversion 2>&- || i3 --version`:

Binary i3 version:  4.12 (2016-03-06, branch "4.12") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12 (2016-03-06, branch "4.12") (pid 603)
Loaded i3 config: /home/adaephon/.i3/config (Last modified: Mi 09 Mär 2016 08:20:50 CET, 2516 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3

URL to a logfile as per http://i3wm.org/docs/debugging.html:

http://logs.i3wm.org/logs/5680316245082112.bz2

I noticed that some of my window titles are indented, that is they are shifted slightly to the right. It seems that it mostly affects windows that are inside nested containers.

To confirm, I opened multiple multiple windows while creating a new split container `split vertical|horizontal` after each new window, alternating `horizontal` and `vertical` (although the results remain the same if only `horizontal` or `vertical` is used. 

The result can be seen in this screen shot:

![splitvhvhvetc](https://cloud.githubusercontent.com/assets/6870588/13629308/8e6ba420-e5d8-11e5-9811-401d0609cbb9.png)

As can be seen the indention grows with the level of nesting. The indention is added immediately when `split horizontal|vertical` is called.

When changing the layout of the containers to stacked with `layout stacking` the indention vanishes:

![stacked](https://cloud.githubusercontent.com/assets/6870588/13629419/53440f4e-e5d9-11e5-80f2-b4c936829a3a.png)

But it remains mostly when switching to tabbed layout (`layout tabbed`):

![mostly_tabbed](https://cloud.githubusercontent.com/assets/6870588/13629453/86b2b95c-e5d9-11e5-97b4-86b4f568f28c.png)

Unless, I change the layout on the outermost container to tabbed, too; then all indents vanish:

![fully_tabbed](https://cloud.githubusercontent.com/assets/6870588/13629473/a80310e8-e5d9-11e5-9a78-ffbf4f29792a.png)

On the other hand, if I create the nested containers moving a container with split layout into a another container with split layout, there is no indention. This can be recreated by utlilizing only `layout splitv`, `focus parent`, `focus left` and `move right` to manipulate layout: 

1. Open two windows, 
2. `layout splitv`
3. open window
4. `move right`, 
5. `layout splitv`, 
[...]

PR #2291, opened 2016-04-12
Multiple fixes for title rendering

Airblader:
@stapelberg The change #2247 is actually about is not yet in here. As discussed there, we believe removing the weird indentation of titles is favorable. If we convinced you, I'll make another commit for this in this PR.

The commit messages (hopefully) elaborate the changes in these commits rather well.

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -455,13 +455,14 @@ void x_draw_decoration(Con *con) {
                             w->x + w->width, 0, r->width - (w->x + w->width), r->height);
     }
 
+    Rect br = con_border_style_rect(con);
+
     /* 3: draw a rectangle in border color around the client */
     if (p->border_style != BS_NONE && p->con_is_leaf) {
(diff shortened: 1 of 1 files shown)


Airblader:
Just for reference, here's what the fixed version looks like:

screenshot-2016-04-12_23-10-23

Airblader:
The fix is not correct for tabbed layouts. Will adapt it.

Airblader:
I'm going to have to revert part of this. As much as I think this looks better, for tabbed and stacked layouts it just doesn't work (at this point I wish border width wasn't something that could be different for every container). I've thought about this for a while and just can't figure out a reasonable combination that is consistent.

I'll instead change this PR to fix the rendering bug alone, but not align title and marks on the client window (i.e., I won't indent them by the border width).

Of course, removing the weird indentation from #2247 is still to be decided.

PR #2293, opened 2016-04-13
Hex in struct Colorpixel.hex to contiguous memory

EvilPudding:
No reason for hex not to be of a constant size, and
no reason to introduce a cache miss by allocating it
separated from the rest of the structure.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -9,6 +9,7 @@
  * commands.c: all command functions (see commands_parser.c)
  *
  */
+#include <stdint.h>
 #include <float.h>
 #include <stdarg.h>
 
@@ -1625,7 +1626,7 @@ void cmd_open(I3_CMD) {
     ystr("success");
     y(bool, true);
     ystr("id");
-    y(integer, (long int)con);
+    y(integer, (uintptr_t)con);
     y(map_close);
 
     cmd_output->needs_tree_render = true;
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -12,6 +12,7 @@
 #include "all.h"
 #include "yajl_utils.h"
 
+#include <stdint.h>
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <fcntl.h>
@@ -217,7 +218,7 @@ static void dump_binding(yajl_gen gen, Binding *bind) {
 void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     y(map_open);
     ystr("id");
-    y(integer, (long int)con);
+    y(integer, (uintptr_t)con);
 
     ystr("type");
     switch (con->type) {
@@ -444,7 +445,7 @@ void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     ystr("focus");
     y(array_open);
     TAILQ_FOREACH(node, &(con->focus_head), focused) {
-        y(integer, (long int)node);
+        y(integer, (uintptr_t)node);
     }
     y(array_close);
 


EvilPudding:
Not really understanding what is failing, I could not find any other reference to the hex member, what am I missing?

Airblader:
@EvilPudding I think the failing test is just a blinking test (i.e., some race condition). I'll restart the test (and take a look at that one later).

libi3/get_colorpixel.c
```diff
@@ -74,7 +74,7 @@ uint32_t get_colorpixel(const char *hex) {
 
     /* Store the result in the cache */
     struct Colorpixel *cache_pixel = scalloc(1, sizeof(struct Colorpixel));
-    cache_pixel->hex = sstrdup(hex);
+    strncpy(cache_pixel->hex, hex, 8);
```
Airblader: If I configure a color value of `#123456789`, this will now crash due to a missing NULL byte, won't it? We should explicitly set the last character to be '\0' .

We also need to make this big enough to allow RGBA, so we actually need `char hex[10]`.

Airblader: Actually never mind the part about RGBA. `Colorpixel` is only used for non-true color displays where alpha transparency doesn't work anyway (and frankly I wouldn't know how to get it to work).

The NULL byte is still a concern, though.

EvilPudding: Shouldn't it throw an error in case the last byte is not null? Because just capping it may confuse users who mistakenly got an extra number in by mistake.

Airblader: I don't mind logging an error, but it shouldn't crash the window manager. The case I have in mind, though, is a user using the same configuration on two machines; one true color display, one non-true color display. They specify an alpha channel in their config which works fine on the first machine and is ignored on the other one. I don't think there's any need to break that.

EvilPudding: Then it probably should just warn the user the alpha channel is being ignored.

Airblader: I personally don't really care whether we log a warning or not (since we don't *offically* support transparency). But we can't crash the window manager because of it, so we need the NULL byte.

EvilPudding: Would you like me to change it, or has any commit been done for it's effect?

Airblader: I don't see any changes to the original commit yet. Yes, we definitely do need to ensure that the last byte is a NULL byte.

libi3/get_colorpixel.c
```diff
@@ -74,7 +74,8 @@ uint32_t get_colorpixel(const char *hex) {
 
     /* Store the result in the cache */
     struct Colorpixel *cache_pixel = scalloc(1, sizeof(struct Colorpixel));
-    cache_pixel->hex = sstrdup(hex);
+    /* copy 7 bytes, leaving the 8th NULL */
+    strncpy(cache_pixel->hex, hex, 7);
```
Airblader: I don't think this is correct (and definitely not idiomatic). This will cause undefined behavior in the last byte. We should really set the last byte explicitly:

    cache_pixel->hex[7] = '\0';

Airblader:
By the way, please don't merge the next branch into this but instead use rebasing. That way we don't end up with meaningless merge commits (force pushing to your feature branch is fine).

libi3/get_colorpixel.c
EvilPudding: What undefined behavior are you talking about? Memory was allocated with calloc, not malloc.

Airblader: D'oh, you're right. Still, in all other 80 places in the current codebase we explicitly set it that way and I think we should stay consistent here. :) Sorry we have to talk for so long about a rather small change, I appreciate your patience.

PR #2295, opened 2016-04-13
Fix rendering of marks (and smaller fixed)

Airblader:
Attempt #2. This time we really only fix the cut off text and an edge case, but don't change how much titles are intended from the left resp. marks from the right.


stapelberg:
Can you provide a before/after screenshot which shows the issue that’s being addressed please?

Airblader:
Sure:

screenshot-2016-04-15_18-50-46
screenshot-2016-04-15_18-49-07

PR #2296, opened 2016-04-13
Correctly count the number of windows for no_focus.

### LINKED ISSUE #2292: Can't get no_focus working: always says "This is the first window on this workspace..."

    Running i3 version: 4.12 (2016-03-06, branch "4.12") (pid 2193)
    Loaded i3 config: /home/florian/.config/i3/config (Last modified: mer. 13 avril 2016 11:23:09 CEST, 1567 seconds ago)

http://logs.i3wm.org/logs/5741830645743616.bz2

**What I did:**

I configured I3 to don't focus VLC when opening: `no_focus [class="Vlc"]`

**What I saw:**

VLC always grabs focus in workspace with another windows.

**What I expected instead:**
VLC doesn't grab focus.

**Comment**
I read the logs and it seems that https://github.com/i3/i3/blob/next/src/manage.c#L578 always returns `true` (and thus print `"This is the first window on this workspace, ignoring no_focus.\n"` in logs) even if VLC is not the first window in the workspace.

For example:

before
after

**Note:** I open VLC with [Synapse](https://farm8.staticflickr.com/7131/7552939146_45e8abc6e5_b.jpg)

Thanks.

Airblader:
Previously we counted the number of (direct) children of the workspace to
decide whether no_focus should be applied or not. However, this doesn't
work correctly if there's a single container with multiple windows on the
workspace.

This patch correctly counts all windows on the workspace.

fixes #2292

### DIFF
--- a/include/con.h
+++ b/include/con.h
@@ -201,6 +201,12 @@ Con *con_for_window(Con *con, i3Window *window, Match **store_match);
 int con_num_children(Con *con);
 
 /**
+ * Count the number of windows (i.e., leaf containers).
+ *
+ */
+int con_num_windows(Con *con);
+
+/**
  * Attaches the given container to the given parent. This happens when moving
  * a container or when inserting a new container at a specific place in the
  * tree.
--- a/src/con.c
+++ b/src/con.c
@@ -728,6 +728,26 @@ int con_num_children(Con *con) {
 }
 
 /*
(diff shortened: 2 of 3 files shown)


Airblader:
Waiting for feedback from the reporter first.

Chocobozzz:
It seems to work perfectly. Thanks a lot.

PR #2298, opened 2016-04-15
Cast con id as uintptr_t

### LINKED ISSUE #2283: [Question] Integral type for con_id
Currently, I have a problem with overflowing con_ids in an IPC client (https://github.com/acrisci/i3ipc-glib/issues/8 and https://github.com/cornerman/i3-easyfocus/issues/6), because they were wrongly represented as an int there.

In i3, the id for the container is the address to the corresponding Con struct and is treated as a ```long``` when being read or written (https://github.com/i3/i3/blob/next/src/match.c#L298 and https://github.com/i3/i3/blob/next/src/ipc.c#L220). I was under the impression that long is only guaranteed to be 4 bytes. So, wouldn't an ```intptr_t```/```uintptr_t``` make more sense? Also, shouldn't it be unsigned?

Just asking to find out, what type should be chosen to represent the con_id in an IPC client and why it is currently signed?

acrisci:
Use `uintptr_t` to cast the con id to int instead of `long int`. This
type is guaranteed to hold the pointer as an int regardless of platform.

fixes #2283

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -9,6 +9,7 @@
  * commands.c: all command functions (see commands_parser.c)
  *
  */
+#include <stdint.h>
 #include <float.h>
 #include <stdarg.h>
 
@@ -1625,7 +1626,7 @@ void cmd_open(I3_CMD) {
     ystr("success");
     y(bool, true);
     ystr("id");
-    y(integer, (long int)con);
+    y(integer, (uintptr_t)con);
     y(map_close);
 
     cmd_output->needs_tree_render = true;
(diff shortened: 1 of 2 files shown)


acrisci:
Did I miss any?

stapelberg:
I can’t spot any other places either. Thanks!

cornerman:
Cool, thanks! Shouldn't we also take care of this when [reading the con_id in a criteria](https://github.com/i3/i3/blob/next/src/match.c#L298)? Currently, this only accepts positive long values. I don't know of a string to uintptr_t function; but maybe using `strtoul` or `strtoull` and checking the bounds could work.

acrisci:
@cornerman yeah that seems like good reasoning to me.

stapelberg:
Agreed.

PR #2301, opened 2016-04-15
Minor optimization

EvilPudding:
Did not want to create a new pull request for this, but alas, I amended my last commit, so it forced me to, sorry.

### DIFF
--- a/libi3/get_colorpixel.c
+++ b/libi3/get_colorpixel.c
@@ -13,7 +13,7 @@
 #include "libi3.h"
 
 struct Colorpixel {
-    char *hex;
+    char hex[8];
     uint32_t pixel;
 
     SLIST_ENTRY(Colorpixel)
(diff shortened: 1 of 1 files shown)


Airblader:
You just have to force push your branch. :) That's not a good idea on main branches but on pull request branches it's perfectly fine.

EvilPudding:
Sorry, I don't understand what you mean, I rebased and force pushed into my branch.

Airblader:
@stapelberg What's the status for this PR? Anything we are waiting for? Of course it needs a rebase by now.

stapelberg:
Nothing from my end, I think this is good-to-go once rebased.

EvilPudding:
You guys are still here? I changed like 3 lines, why you didn't just delete this pull request and copy the changes is beyond me.

Airblader:
@EvilPudding I can understand the frustration, it's unfortunate that this PR has been forgotten for this long. Both Michael and I have busy day jobs, as do you, I assume, and speaking at least for myself the past few months have been pretty busy.

As for just taking over the changes: when possible, we avoid this, because we want to properly attribute the changes to the author by using their commit. We're also community-driven, which means we expect the community to make changes they care about rather than taking them away from them and doing the work ourselves instead (though we of course also contribute code ourselves). Sometimes we take over orphaned PRs, for example bug fixes, because those are worth adding; but if possible, we want the author of the code to do the work so their work is attributed to them.

Long story short, thanks for updating the change. Unfortunately you did a merge, which caused a merge commit to appear in the history. To keep a clean history we want to avoid this, so you should revert that merge and do a rebase instead (and then force-push the branch).

EvilPudding:
I understand all of that, and I'm not frustrated, I'm admired that you care about such a small pull request.
I think I did it right now.

stapelberg:
Thanks!

PR #2302, opened 2016-04-15
Medium optimization

EvilPudding:
Removed all I3_IPC_MAGIC's calls to strlen.

I3_IPC_MAGIC is a constant string literal, so why not store the size
instead of even inside the same function calling strlen so many times?

Also, strlen loops through the string until it finds the NULL char,
but we know I3_IPC_MAGIC is a string literal, so calling sizeof, which
is calculated at compile time, saves a lot of loops.


stapelberg:
I’ll close this as well, see #2303 for discussion.

PR #2303, opened 2016-04-15
Removed pointless strlens.

EvilPudding:
There is no reason to use strlen on a literal...
I must sound like a broken record, but the code is full of this...

### DIFF
--- a/i3-config-wizard/main.c
+++ b/i3-config-wizard/main.c
@@ -333,10 +333,11 @@ static char *rewrite_binding(const char *input) {
 
             /* A literal. */
             if (token->name[0] == '\'') {
-                if (strncasecmp(walk, token->name + 1, strlen(token->name) - 1) == 0) {
+                uint32_t token_name_len = strlen(token->name);
+                if (strncasecmp(walk, token->name + 1, token_name_len - 1) == 0) {
                     if (token->identifier != NULL)
                         push_string(token->identifier, token->name + 1);
-                    walk += strlen(token->name) - 1;
+                    walk += token_name_len - 1;
                     if ((result = next_state(token)) != NULL)
                         return result;
                     break;
@@ -565,7 +566,7 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
                                 A__NET_WM_NAME,
                                 A_UTF8_STRING,
                                 8,
-                                strlen("i3: generate config"),
+                                sizeof("i3: generate config") - 1,
                                 "i3: generate config");
             xcb_flush(conn);
         } else
@@ -693,7 +694,7 @@ static void finish() {
 #endif
         /* skip the warning block at the beginning of the input file */
         if (head_of_file &&
-            strncmp("# WARNING", line, strlen("# WARNING")) == 0)
+            strncmp("# WARNING", line, sizeof("# WARNING") - 1) == 0)
             continue;
 
         head_of_file = false;
@@ -707,7 +708,7 @@ static void finish() {
         }
 
         /* Set the modifier the user chose */
-        if (strncmp(walk, "set $mod ", strlen("set $mod ")) == 0) {
+        if (strncmp(walk, "set $mod ", sizeof("set $mod ") - 1) == 0) {
             if (modifier == MOD_Mod1)
                 fputs("set $mod Mod1\n", ks_config);
             else
@@ -717,7 +718,7 @@ static void finish() {
 
         /* Check for 'bindcode'. If it’s not a bindcode line, we
          * just copy it to the output file */
-        if (strncmp(walk, "bindcode", strlen("bindcode")) != 0) {
+        if (strncmp(walk, "bindcode", sizeof("bindcode") - 1) != 0) {
             fputs(walk, ks_config);
             continue;
         }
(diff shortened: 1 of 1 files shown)


Airblader:
I don't know much about compiler optimization, but I would assume that this will be optimized by the compiler anyway. I don't think there's an actual performance hit here.

stapelberg:
It’s my understanding that `strlen` on literals gets optimized away, yes. Optimizations of `i3-config-wizard` specifically are likely a waste of time — that program is run exactly once per i3 user (on average), and it doesn’t take long to run.

EvilPudding:
One should never rely on compiler optimizations. strlen means loop the string and find the NULL char, no reason not to merge this, when the original code is silly.

stapelberg:
Given that there are subtle differences between `strlen` and `sizeof`, and both result in the same assembler code, I prefer to keep using `strlen` as it seems more generic to me. See also http://stackoverflow.com/a/15000792/712014

EvilPudding:
Thank you for linking me to a stackoverflow that agrees with me.

> So, only if the string spans the entire array (e.g. char s[] = "stuff"), would using sizeof for a statically defined array return what you want (and be faster as it wouldn't need to loop through to find the null-terminator) (if the last character is a null-terminator, you will need to subtract 1). If it doesn't span the entire array, it won't return what you want.

I was going to give you links that prove my point, but you did it for me.

EvilPudding:
And by the way, one of the strlens I removed was on a heap string, and was being repeated inside the same loop, which can't ever be optimized by the compiler.

Airblader:
We're still talking improvements that are likely not even really measurable, certainly not ever noticable by the user. Given that alone, I prefer the obvious `strlen` over the slightly more clunky `sizeof -1`. Not that I feel strongly about it, though.

EvilPudding:
And somehow, strlen seems more generic... It's a function call, while sizeof is a compiler operation. Some compilers know that using calling the function on a literal is terrible, so they correct you, but the code is still calling string.h's strlen.

EvilPudding:
@Airblader There were 3 strlens on a heap string on the same loop, which I improved by storing them onto a variable, and if you people want to show **any** expert saying you should use strlen over sizeof on literals, go ahead, until now, I've seen terrible code, and a link to a stackoverflow that agrees with me. Please do more research on this. Inside a loop, having 3 strlens on the same non constant string is extremely inefficient, but you won't merge the actual optimization, because you prefer telling C to loop the string it's length (that is how any C programmer should read `strlen`), only to be corrected by the compiler.

Airblader:
First of all, you should calm down a little.

I'm not even arguing your point that `sizeof` is more efficient. I'm arguing the fact that it's relevant in any way. Do you have any reason to believe this is a bottleneck? In the config wizard? Because I don't think so. And considering that, I, personally, just find `strlen` to be nicer to read. I really don't care about a millisecond of execution time in a wizard that runs once in the entire lifetime of the machine. But as I also said, I don't *fight* this improvement either. I'm mostly indifferent about it.

> only to be corrected by the compiler.

If we should avoid being corrected by the compiler on all costs, we should be writing very different code. I think it's perfectly fine to let the compiler do its job, which includes optimizing the code. Code that we want to write for humans to be readable and understandable.

Of course we don't want to introduce bottlenecks for the sake of readability either, but this does not strike me as anything relevant to the performance of the application.

EvilPudding:
First of all, I am calm.
It's not the job of the compiler to optimize strlen into sizeof-1, that is a compiler choice, and if you want good C code, you should not just expect the compiler to do that optimization, it is not part of the C standard.
I did not say we should avoid being corrected by the compiler at all costs, changing strlen to sizeof is not a huge cost. Second of all, when you're coding, you have to keep in mind if any function you use is going to have loops inside it, having strlen meaning something for some cases and another for other cases is readability problem. When I was reading the code, I didn't even know `I3_IPC_MAGIC` was a string literal, I had to go to the definition, all I saw was a function that calls a loop being used multiple times on the same function. Whereas if I had what I introduced, `I3_IPC_MAGIC_LEN`, I would assume it was correct, because even if it wasn't a string literal, I would know it's being stored in an effective way. To conclude, I do not agree having `strlen` mean different things in different parts of the code is readable.

EvilPudding:
Sorry for the way I spoke, I should not have called the code terrible, it's just I'm used to looking at `strlen` as a generally bad practice, and this doesn't even change much.

Airblader:
> it is not part of the C standard.

I generally see your point, but I think compilers do quite a lot of optimizations that are "not in the C standard" that people (rightfully) rely on. Nobody compiles software these days without optimizations and my personal view is that it's totally fine to rely on that in code like this which is not performance critical whatsoever. But again, I do understand where you're coming from.

> To conclude, I do not agree having strlen mean different things in different parts of the code is readable.

And it doesn't, so I don't understand this point. It always means "the length of the string", whether or not that is constant. In that regard, using `strlen` actually makes more sense to me, as it'd be a consistent way to achieve this rather than introducing a separate way that only works for specific strings.

As for readability, we could have a "compromise" by introducing a macro like

````
#define STRLEN_CONST(str) (sizeof(str) - 1)
```

I'd be OK with that, but ultimately @stapelberg decides whether we want any of this or not.

EvilPudding:
@Airblader Just ignore my pull request, I love i3, and the work you guys do here.
This whole discussion is so minor it's really not worth your time, I mean it.
I've been a real prick complaining about it, just because it triggered my OCD, when all it amounts to is aesthetics basically, which should be completely up to you guys.
Sorry again, keep up the great work!

Airblader:
@EvilPudding No need to apologize for anything. Even small contributions are contributions and we generally value every single one (and don't mind discussing them). It did sort of go awry here and that's unfortunate, but sometimes opinions differ. Please don't let it stop you from further contributions and thanks for both the time you put in and your words. :)

PR #2306, opened 2016-04-17
Modify construction of config variables list to ensure they match by longest lexeme.

### LINKED ISSUE #2235: Config-file variable expansion when variable name is a prefix of another variable

    i3 version 4.11-269-g6fe8d30 (2016-02-25, branch "next") © 2009 Michael Stapelberg and contributors

**Summary**

Defining a variable which name is a prefix of an earlier defined variable, will make the earlier defined variable unusable as the lexer/parser will expand the latter defined (prefix) variable first.

**What I did:**

If I define the following in my config file:

    set $footer       #ff0000
    set $foo          #00ff00

    client.focused    $footer

**What I saw:**

This will be expanded to:

    client.focused     #00ff00ter

Since `$foo` will be expanded to `#00ff00` first.

**What I expected instead:**

I would have expected that it was expanded to:

    client.focused     #ff0000

**Work-around**

By defining the longest variable first, it will function as _I_ would expect:

    set $foo          #00ff00
    set $footer       #ff0000
   
    client.focused    $footer

Expands to:

    client.focused    #ff0000

**Fix**

I haven't really looked at the lexing/parser code, so don't know. Maybe this is OK and not really an issue. But it did cause some confusion for me at least, and guess it may trouble someone else some day down the line :)

kneitinger:
Does not change default behavior when two variable declarations use the exact same name.

fixes #2235

### DIFF
--- a/src/config_parser.c
+++ b/src/config_parser.c
@@ -879,9 +879,23 @@ bool parse_file(const char *f, bool use_nagbar) {
                 v_value++;
 
             struct Variable *new = scalloc(1, sizeof(struct Variable));
+            struct Variable *test = NULL, *loc = NULL;
             new->key = sstrdup(v_key);
             new->value = sstrdup(v_value);
-            SLIST_INSERT_HEAD(&variables, new, variables);
+            /* ensure that the correct variable is matched in case of one being
+             * the prefix of another */
+            SLIST_FOREACH(test, &variables, variables) {
+                if (strlen(new->key) >= strlen(test->key))
+                    break;
+                loc = test;
+            }
+
+            if (loc == NULL) {
+                SLIST_INSERT_HEAD(&variables, new, variables);
+            } else {
+                SLIST_INSERT_AFTER(loc, new, variables);
+            }
+
             DLOG("Got new variable %s = %s\n", v_key, v_value);
             continue;
         }
--- a/testcases/t/183-config-variables.t
+++ b/testcases/t/183-config-variables.t
@@ -79,6 +79,22 @@ EOT
 
 is(launch_get_border($config), 'none', 'no border');
 
+#####################################################################
+# test that longest matching variable name is substituted
+#####################################################################
+
+$config = <<'EOT';
+# i3 config file (v4)
+font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
+
+set $var normal title
+set $vartest special title
+set $vart mundane title
+for_window [title="$vartest"] border none
+EOT
+
+is(launch_get_border($config), 'none', 'no border');
+
 
 
 done_testing;


src/config_parser.c
```diff
@@ -879,9 +879,23 @@ bool parse_file(const char *f, bool use_nagbar) {
                 v_value++;
 
             struct Variable *new = scalloc(1, sizeof(struct Variable));
+            struct Variable *test = NULL, *loc = NULL;
             new->key = sstrdup(v_key);
             new->value = sstrdup(v_value);
-            SLIST_INSERT_HEAD(&variables, new, variables);
+            /* ensure that correct variable is matched in case of one being the
```
stapelberg: nit: s/correct/the correct/

stapelberg:
Thanks for the pull request. Can you amend it to also include a test case which exercises this code path to make sure we don’t introduce a regression later?

Airblader:
@kneitinger Please use rebasing and not merging to avoid polluting the history with merge commits. Also, when you made the changes, drop a comment so that we get notified. :)

kneitinger:
That all makes sense!

I believe I rebased the commits in the intended way, but I'm somewhat inexperienced with that operation so let me know if it needs to be done in a different way.  Thanks :)

Airblader:
@kneitinger Unfortunately, something is still wrong – it shows commits that don't belong in your pull request. Typically, this is what you want to do:

* Fork i3 on Github and clone your fork to your machine.
* Add the upstream remote via `git remote add upstream https://github.com/i3/i3.git`
* Create your feature branch off of `next` and create your commit(s).

Now, if changes have been pushed to upstream meanwhile, you can do

    # first fetch the changes from upstream
    git fetch --all
    # rebase your feature branch onto the updated upstream branch
    git rebase upstream/next

Most of the time, this should work without any problems and you can then push your branch again via `git push --force` (the force is needed because you rebased). If you do run into a conflict, git will tell you the necessary steps to resolve them.

Now, this is of course a bit too late here and your branch is now in a dirty state. You can fix it like so (or many other ways, but I figure this one is easy to understand):

````
# First, let's save your hard work by copying it into a patch file
git format-patch -1 520b13

# Next up, we simply reset your branch (which nukes your commit, but that's why we saved it)
git remote add upstream https://github.com/i3/i3.git
git fetch upstream
git reset --hard upstream/next

# Let's reapply your commit
git am *.patch
rm *.patch

# All set! Now check the git log once to make sure that your commit is on the very top and then push it
git log
git push --force
```

Thanks for your patience and, of course, your contribution. :-)

kneitinger:
Yes! Thank you so much for explaining that to me, it makes a lot of sense (and simplifies things)!  Anyway, I believe that this is *now* in the correct state :D

PR #2313, opened 2016-04-23
i3-input: Proper position in non-standard cases.

### LINKED ISSUE #2312: i3-input positioned off-screen
`Binary i3 version:  4.12 (2016-03-06, branch "4.12") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12 (2016-03-06, branch "4.12") (pid 4419)
`
Not sure whether this is an i3 or i3-input bug:
When selecting a parent container and starting i3-input, i3-input gets positioned off-screen.

Steps to reproduce:
Go to new workspace, create layout as follows:
<pre>
+-------------------+
|  Term             |
|                   |
|                   |
+---------+---------+
|         |         |
| Term    | Term    |
|         |         |
+---------+---------+
</pre>

Focus bottom right terminal, run "sleep 1; ./i3-input" with debugging enabled and focus parent container (Mod-A).
For some reason: `i3-input/main.c:get_window_position:352 - Determined coordinates of window with input focus at x = 0 / y = 1070.` My screen is 1680x1050

i3-input shouldn't position itself off-screen, but why is the focused window off-screen anyhow?

I think a a log isn't necessary.

Airblader:
This commit fixes two issues:
* We detect the EWMH support window from the root window. If this window
  currently has the input focus, we ignore this. We do this because this
  window is not a window the user is aware of and positioning relative to
  it makes no sense.
* We also detect whether the current input focus is in an i3-frame window.
  This can happen, e.g., when selecting a parent (split) container. Since
  frame windows are direct children of the root window, we must not
  translate its coordinates or we get weird results and i3-input ends up
  off-screen (see #2312). For all other windows, including those without
  any WM_CLASS, we proceed as before.

fixes #2312

### DIFF
--- a/i3-input/main.c
+++ b/i3-input/main.c
@@ -317,9 +317,30 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
 static xcb_rectangle_t get_window_position(void) {
     xcb_rectangle_t result = (xcb_rectangle_t){logical_px(50), logical_px(50), logical_px(500), font.height + logical_px(8)};
 
+    xcb_get_property_reply_t *supporting_wm_reply = NULL;
     xcb_get_input_focus_reply_t *input_focus = NULL;
     xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_get_property_reply_t *wm_class = NULL;
     xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    xcb_atom_t A__NET_SUPPORTING_WM_CHECK;
+    xcb_intern_atom_cookie_t nswc_cookie = xcb_intern_atom(conn, 0, strlen("_NET_SUPPORTING_WM_CHECK"), "_NET_SUPPORTING_WM_CHECK");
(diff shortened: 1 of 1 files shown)


i3-input/main.c
```diff
@@ -317,10 +317,36 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
 static xcb_rectangle_t get_window_position(void) {
     xcb_rectangle_t result = (xcb_rectangle_t){logical_px(50), logical_px(50), logical_px(500), font.height + logical_px(8)};
 
+    xcb_get_property_reply_t *supporting_wm_reply = NULL;
     xcb_get_input_focus_reply_t *input_focus = NULL;
     xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_get_property_reply_t *wm_class = NULL;
     xcb_translate_coordinates_reply_t *coordinates = NULL;
 
+#define INTERN_ATOM(name)                                                                 \
```
stapelberg: Why are you defining a macro that you only use once? Avoiding a macro would make the code a bit easier to understand.

stapelberg:
Looks good overall.

Airblader:
@stapelberg Done.

Issue #1784, opened 2015-07-06
New windows from RubyMine are not drawn


danuker:
1. Open a project with RubyMine (or probably any other JetBrains IDE), and switch i3 to either **tabbed** or **stacked** mode.

2. Select two files from the Project panel (left)
![diff-p1](https://cloud.githubusercontent.com/assets/74835/8523039/52c5ab7a-23f3-11e5-9553-4ec55c613e55.png)

3. Press Ctrl+D (for the difference between the files). The new window title **appears and is focused**, but the window itself is **not drawn** (the old one remains).
![diff-p2](https://cloud.githubusercontent.com/assets/74835/8523038/52c3821e-23f3-11e5-9695-eb5c9afbe224.png)

4. Just moving the mouse on top of the new tab (not even clicking) makes the window re-draw and show up properly.
![diff-p3](https://cloud.githubusercontent.com/assets/74835/8523040/52c6b8c6-23f3-11e5-9f41-632d99fb19a8.png)

Step 4 should not be necessary - the window should show up by itself.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

danuker:
i3 version 4.10.2 (2015-04-16, branch "4.10.2") © 2009-2014 Michael Stapelberg and contributors

I will provide a log file later today if needed.

Airblader:
Yes, please provide a log file. Your version doesn't include the hidden state handling we will introduce yet, so that can't be it. Since i3 isn't involved in the client rendering, this could be an application bug. The only thing i3 could be doing wrong is not correctly mapping the clients.

Airblader:
Please make sure to read [the debugging guide](https://github.com/i3/i3/blob/next/docs/debugging) from the current dev branch (not the website) to make it a bit easier on us.

danuker:
http://logs.i3wm.org/logs/5739238230327296.bz2

At around 16:00:00 I opened "file1 vs file2 (/home/danuth/RubymineProjects/untitled)" in the default mode (and the window was drawn).
At around 16:00:10 I switched to tabbed mode, and opened it again. (and the window was not drawn).
At around 16:00:20 I moved my mouse up on the tab, and the window was rendered.

Airblader:
[...]

PR #2314, opened 2016-04-24
Feature/net wm state focused

### LINKED ISSUE #2273: Support _NET_WM_STATE_FOCUSED

i3 version 4.12 (2016-03-06, branch "4.12") © 2009 Michael Stapelberg and contributors

**What I did:**

Run any gtk3 app with client side decoration like Nautilus or Evince. Make window floating, Move focus to this window and move out.

**What I saw:**

Window decoration stays like window is focused.

**What I expected instead:**

Window decoration should change appearance depending of is window focused.

Xfwm4 patch https://bugzilla.xfce.org/show_bug.cgi?id=11067
Gnome bug https://bugzilla.gnome.org/show_bug.cgi?id=734159

acrisci:
_NET_WM_STATE_FOCUSED is set on _NET_WM_STATE to indicate that the
window is focused. It must be set when the window is newly focused and
removed once the window no longer has focus.

> _NET_WM_STATE_FOCUSED indicates whether the window's decorations are
> drawn in an active state. Clients MUST regard it as a read-only hint.
> It cannot be set at map time or changed via a _NET_WM_STATE client
> message.

For example, this is used by GTK applications to show the decoration in
an active or inactive state. This change can be tested by opening a GTK
application (like evince), focusing the window and unfocusing the
window, and observing a change in the window decorations.

Fixes #2273


acrisci:
@Airblader I made logic changes to `t/253-multiple-net-wm-state-atoms.t`.

testcases/t/264-net-wm-state-focused.t
```diff
@@ -0,0 +1,65 @@
...
+
+$windowB = open_window;
+
+ok(!is_focused($windowA), 'when a another window is focused, the old window should not have _NET_WM_STATE_FOCUSED set');
+
+fresh_workspace;
+
+ok(!is_focused($windowB), 'when focus moves to the root window, neither window should have _NET_WM_STATE_FOCUSED set');
+
+# TODO: test to make sure the atom is properly set/unset when WM_TAKE_FOCUS is used to focus the window
```
acrisci: Is it possible?

testcases/t/264-net-wm-state-focused.t
```diff
@@ -0,0 +1,65 @@
...
+
+ok(is_focused($windowA), 'a newly opened window that is focused should have _NET_WM_STATE_FOCUSED set');
+
+$windowB = open_window;
+
+ok(!is_focused($windowA), 'when a another window is focused, the old window should not have _NET_WM_STATE_FOCUSED set');
+
+fresh_workspace;
+
+ok(!is_focused($windowB), 'when focus moves to the root window, neither window should have _NET_WM_STATE_FOCUSED set');
```
Airblader: nit: focus doesn't move to the root window, but to the EWMH support window.

testcases/t/264-net-wm-state-focused.t
Airblader: It should be. You can set up a window with the appropriate input model and use `xdotool` to trigger the `XSetInputFocus` call. My guess is that this test would currently fail since the FocusIn event is handled such that it prevents the focus logic to be triggered.

Airblader: Addendum: actually the test would probably succeed, but for the wrong reasons. See #1784. We incorrectly update these information after sending `WM_TAKE_FOCUS` when really it should happen once the client actually set input focus. But that's not the problem of this PR, of course.

acrisci: So remove the todo then since this will no longer be required once we correct that mistake?

Airblader: No, they're orthogonal issues. Ideally we'd want to test this case here as well (not that I'd insist on it). It's just a little ugly to write.

acrisci: I guess your right, although it wouldn't add to coverage since that path would already be tested.

Airblader: I think we should leave this test to #1784 where it will implicitly happen anyway. There we also have a better reason to write a test for `WM_TAKE_FOCUS`. There's little sense in testing something here that we know is currently broken anyway. So I think you can remove the TODO for now.

acrisci:
Ok I updated this so it passes the build so it should be good to go.

testcases/t/253-multiple-net-wm-state-atoms.t
```diff
@@ -36,12 +36,25 @@ sub get_wm_state {
...
-    return \@atoms;
+    return @atoms;
 }
 
 my $wm_state_sticky = $x->atom(name => '_NET_WM_STATE_STICKY')->id;
 my $wm_state_fullscreen = $x->atom(name => '_NET_WM_STATE_FULLSCREEN')->id;
 
+sub state_contains_atom {
+    my $atom = pop;
+    my @state = @_;
```
stapelberg: more idiomatic:
```perl
sub state_contains_atom {
    my ($atom, @state) = @_;
    # …
```

(But see below for further comments.)

testcases/t/253-multiple-net-wm-state-atoms.t
```diff
@@ -51,18 +64,24 @@ my $wm_state_fullscreen = $x->atom(name => '_NET_WM_STATE_FULLSCREEN')->id;
 fresh_workspace;
 my $window = open_window;
 cmd 'sticky enable';
-is_deeply(get_wm_state($window), [ $wm_state_sticky ], 'sanity check: _NET_WM_STATE_STICKY is set');
+my @state = get_wm_state($window);
+ok(state_contains_atom(@state, $wm_state_sticky), 'sanity check: _NET_WM_STATE_STICKY is set');
```
stapelberg: Here and below, the following pattern is more idiomatic:

```perl
ok(scalar grep { $_ == $wm_state_sticky } @state > 0, 'sanity check: _NET_WM_STATE_STICKY is set');
```

stapelberg:
Looks good overall.

testcases/t/264-net-wm-state-focused.t
```diff
@@ -0,0 +1,65 @@
...
+
+$windowB = open_window;
+
+ok(!is_focused($windowA), 'when a another window is focused, the old window should not have _NET_WM_STATE_FOCUSED set');
+
+fresh_workspace;
+
+ok(!is_focused($windowB), 'when focus moves to the ewmh support window, neither window should have _NET_WM_STATE_FOCUSED set');
+
+# TODO: test to make sure the atom is properly set/unset when WM_TAKE_FOCUS is used to focus the window
```
Airblader: TODO here?

src/x.c
```diff
@@ -1121,7 +1136,7 @@ void x_push_changes(Con *con) {
                     xcb_change_window_attributes(conn, focused->window->id, XCB_CW_EVENT_MASK, values);
                 }
 
-                ewmh_update_active_window((con_has_managed_window(focused) ? focused->window->id : XCB_WINDOW_NONE));
```
Airblader: Is this not an unused method now that should be removed?

orestisf1993: `ewmh_update_active_window` is used in `change_ewmh_focus`

PR #2317, opened 2016-04-24
Focus a newly managed container only if it doesn't use the globally a…

### LINKED ISSUE #1784: New windows from RubyMine are not drawn

Airblader:
…ctive input model.

fixes #1784

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -602,9 +602,20 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
         xcb_discard_reply(conn, wm_user_time_cookie.sequence);
     }
 
+    if (set_focus) {
+        /* Even if the client doesn't want focus, we still need to focus the
+         * container to not break focus workflows. Our handling towards X will
+         * take care of not setting the input focus. However, one exception to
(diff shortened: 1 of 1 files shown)


Airblader:
Just FYI, this is how Openbox handles it as well, see `openbox/client.c`:

````
            /* Don't steal focus from globally active clients.
               I stole this idea from KWin. It seems nice.
            */
            else if (!(focus_client->can_focus || focus_client->focus_notify))
            {    
                steal = FALSE;
                ob_debug("Not focusing the window because a globally "
                         "active client has focus");
            } 
```

Airblader:
@stapelberg Good to go here.

Issue #2316, opened 2016-04-24
Redraw problems on Raspberry (Raspbian, Debian Jessie)


fhaun:
Output of `i3 --moreversion 2>&- || i3 --version`:

```
Binary i3 version:  4.12-44-g42a3371 (2016-04-23, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12-44-g42a3371 (2016-04-23, branch "next") (pid 13143)
Loaded i3 config: /home/fhaun/.i3/config (Last modified: Sun 24 Apr 2016 05:18:35 CEST, 28868 seconds ago)

The i3 binary you just called: /home/fhaun/opt/i3-next/usr/bin/i3
The i3 binary you are running: /home/fhaun/opt/i3-next/usr/bin/i3
```

URL to a logfile as per http://i3wm.org/docs/debugging.html:

http://logs.i3wm.org/logs/5693341270278144.bz2

**What I did:**

Running i3 on Raspberry Pi 3, OS is Raspbian (Debian Jessie). I've tried color depth of 16, 24 and 32. With and without an compositor (compton).

**What I saw:**

I have redraw problems in **all programs**, e.g. an Emacs buffer:
http://i.imgur.com/krE7C20.png

This happens after scrolling via mouse or keyboard. Terminal outputs, e.g. while compiling are also not correctly updated.

After a git bisect I found the commit when the redraw problems begins.

```
780cb8d15d07febbb9be47cb1bc1201072b31838 is the first bad commit
commit 780cb8d15d07febbb9be47cb1bc1201072b31838
Author: Ingo Bürk <ingo.buerk@tngtech.com>
Date:   Tue Nov 17 12:50:06 2015 +0100

    Use 32-bit visual by default if available.
    
    With this patch, we use 32-bit visuals per default whenever it is
    available. Otherwise, we fall back to the actual root window's
    depth, which will typically be 24-bit.
    
    Before this patch, we already used 32-bit depth for containers with
    a window that uses 32-bit. However, this means that we didn't use
    32-bit for split parent containers on which decoration is drawn.
    For 32-bit windows using transparency, this caused a graphical glitch
    because the decoration pixmap behind it would show through. This
    behavior is fixed with this change.
    
    relates to #1278

:040000 040000 9cbc0df993e1b5b2aca17718d8956158182eb278 4b053c965aefcf38b2fa15bfe1773f6de9592b6e M      include
[...]

PR #2323, opened 2016-04-25
Remove redundant depth argument from x_con_init().

Airblader:


stapelberg:
Is this going to make the possible revert discussed in #2316 harder? If yes, I’d like to wait with merging.

Airblader:
Since we don't need to revert the entire commit I think there's no overlap here, but given that this is just a tiny cleanup, let's wait just to be on the safe side.

Airblader:
@stapelberg We know now that this won't conflict even if we were to do the partial revert, so we can merge it, I think.

yshui:
Uhm, sorry. As the author of the original change, I'm asking why are you removing this?

This is needed if we are not going to use 32-bit depth for all windows (which is the revert you are talking about, right?).

Airblader:
No, see the comment above. Since we wouldn't need to revert entirely, this won't conflict.

yshui:
What part are you planning to revert? You are going to revert to a state where no frame window is using the 32bit depth?

Airblader:
We're currently not planning on reverting anything. We are keeping the option open if more problems arise, but as of now, after many months of 4.12, only a single Raspberry user has reported issues. So IMHO there's no need for the revert right now. The patch that "reverts" support is described in the issue linked above.

yshui:
That makes sense.

Thanks.

Issue #2082, opened 2015-11-24
i3 behavior in a multiple xrandr/xscreen situation


woodensquares:
I have X configured as follows, left to right:

monitor0: 1680x1050 card1, X screen 0, randr left
monitor1: 2560x1440 card2, X screen 0, randr right
monitor2: 1280x1024 card3, X screen 1

- if I move the mouse on the xscreen 0 monitors, everything is fine
- if I cross the boundary between xscreen 0 and xscreen 1 the keyboard focus does not switch *UNLESS* I enter the title bar of a window already in xscreen 1 or in general I cross a container boundary on xscreen 1
- if there are no windows in xscreen 1 the keyboard does not switch at all from xscreen 0
- if there are multiple windows in xscreen 1 then the keyboard switches when I cross a cointainer boundary in xscreen 1

simply moving the mouse across from monitor1(xs0) to monitor2(xs1) with windows both in monitor1(xs0) and monitor2(xs1) gives me this in the log

move from monitor1(xs0) to monitor2(xs1) 
```
24/11/15 10:06:04 AM - handlers.c:handle_event:1353 - event type 7, xkb_base 85
24/11/15 10:06:04 AM - handlers.c:handle_enter_notify:131 - enter_notify for 00800058, mode = 0, detail 2, serial 11448
24/11/15 10:06:04 AM - handlers.c:handle_enter_notify:132 - coordinates 2558, 551
```
move back from monitor2(xs1) to monitor1(xs0)
```
24/11/15 10:06:12 AM - handlers.c:handle_event:1353 - event type 7, xkb_base 85
24/11/15 10:06:12 AM - handlers.c:handle_enter_notify:131 - enter_notify for 000004fb, mode = 0, detail 4, serial 11448
24/11/15 10:06:12 AM - handlers.c:handle_enter_notify:132 - coordinates 4239, 575
24/11/15 10:06:12 AM - handlers.c:handle_enter_notify:153 - Getting screen at 4239 x 575
24/11/15 10:06:12 AM - randr.c:get_output_containing:102 - comparing x=4239 y=575 with x=1680 and y=0 width 2560 height 1440
24/11/15 10:06:12 AM - workspace.c:_workspace_show:385 - Not switching, already there.
24/11/15 10:06:12 AM - con.c:con_focus:200 - con_focus = 0x245a490
24/11/15 10:06:12 AM - con.c:con_focus:200 - con_focus = 0x24c7e50
[...]

PR #2331, opened 2016-05-01
Multiple X screens and keyboard focus (issue #2082)

woodensquares:
Let me know if you think this approach would be clean enough, if so I will investigate about seeing if I can get a test case working with multiple Xephyrs (as the test would have to simulate moving the mouse/keyboard in-between screens) and I will resubmit once I have that. The normal test suite runs clean, but of course it does not cover multiple X screen issues.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -22,6 +22,7 @@
 int randr_base = -1;
 int xkb_base = -1;
 int xkb_current_group;
+extern uint8_t previous_screen;
 
 /* After mapping/unmapping windows, a notify event is generated. However, we don’t want it,
    since it’d trigger an infinite loop of switching between the different windows when
@@ -88,7 +89,7 @@ bool event_is_ignored(const int sequence, const int response_type) {
  * current workspace, if so.
  *
  */
-static void check_crossing_screen_boundary(uint32_t x, uint32_t y) {
+static void check_crossing_screen_boundary(uint32_t x, uint32_t y, bool force_treerender) {
     Output *output;
 
     /* If the user disable focus follows mouse, we have nothing to do here */
@@ -114,8 +115,27 @@ static void check_crossing_screen_boundary(uint32_t x, uint32_t y) {
     con_focus(next);
 
     /* If the focus changed, we re-render to get updated decorations */
-    if (old_focused != focused)
+    if (old_focused != focused || force_treerender)
         tree_render();
+}
+
+/*
+ * This is called when the pointer leaves the root window
+ *
+ */
+static void handle_leave_notify(xcb_enter_notify_event_t *event) {
+    DLOG("leave_notify for %08x, mode = %d, detail %d, serial %d, focus %d\n",
+         event->event, event->mode, event->detail, event->sequence,
+         event->same_screen_focus);
+
+    if (event->mode != XCB_NOTIFY_MODE_NORMAL) {
+        DLOG("This was not a normal notify, ignoring\n");
+        return;
+    }
+
+    /* When leaving we don't care about the previous value */
+    previous_screen = event->same_screen_focus;
+    return;
 }
 
 /*
@@ -124,11 +144,11 @@ static void check_crossing_screen_boundary(uint32_t x, uint32_t y) {
  */
 static void handle_enter_notify(xcb_enter_notify_event_t *event) {
     Con *con;
-
     last_timestamp = event->time;
 
-    DLOG("enter_notify for %08x, mode = %d, detail %d, serial %d\n",
-         event->event, event->mode, event->detail, event->sequence);
+    DLOG("enter_notify for %08x, mode = %d, detail %d, serial %d, focus %d\n",
+         event->event, event->mode, event->detail, event->sequence,
+         event->same_screen_focus);
     DLOG("coordinates %d, %d\n", event->event_x, event->event_y);
     if (event->mode != XCB_NOTIFY_MODE_NORMAL) {
         DLOG("This was not a normal notify, ignoring\n");
@@ -141,6 +161,18 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
         return;
     }
 
+    bool focus_crossed_screen = false;
+    if (event->same_screen_focus != previous_screen) {
+        /* Note we should not skip on previous_screen being
+         * UINT8_MAX because we don't want the first display
+         * switch to be missed (say we have i3 on :0.0 and :0.1,
+         * the :0.1 i3 will not get an event in general until
+         * the first time the user tries to use it, since the
+         * mouse pointer is likely on :0.0 to start with */
+        DLOG("The user changed screens %d -> %d\n", previous_screen, event->same_screen_focus);
+        focus_crossed_screen = true;
+    }
+
     bool enter_child = false;
     /* Get container by frame or by child window */
     if ((con = con_by_frame_id(event->event)) == NULL) {
@@ -151,7 +183,12 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
     /* If not, then the user moved their cursor to the root window. In that case, we adjust c_ws */
     if (con == NULL) {
         DLOG("Getting screen at %d x %d\n", event->root_x, event->root_y);
-        check_crossing_screen_boundary(event->root_x, event->root_y);
+        /* The user entered the root window coming from another X
+         * screen, this means that the above will not have called
+         * tree_render (leaving the keyboard focus on the other
+         * screen). Force a rerender depending on focus_crossed_screen.
+         */
+        check_crossing_screen_boundary(event->root_x, event->root_y, focus_crossed_screen);
         return;
     }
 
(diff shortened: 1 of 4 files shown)


src/handlers.c
```diff
@@ -122,13 +123,28 @@ static void check_crossing_screen_boundary(uint32_t x, uint32_t y) {
  * When the user moves the mouse pointer onto a window, this callback gets called.
  *
  */
+static void handle_leave_notify(xcb_enter_notify_event_t *event) {
```
Airblader: Please don't place this *between* `handle_enter_notify` and its documentation.

src/handlers.c
```diff
@@ -122,13 +123,28 @@ static void check_crossing_screen_boundary(uint32_t x, uint32_t y) {
...
+    DLOG("leave_notify for %08x, mode = %d, detail %d, serial %d, focus %d\n",
+         event->event, event->mode, event->detail, event->sequence,
+         event->same_screen_focus);
+
+    if (event->mode != XCB_NOTIFY_MODE_NORMAL) {
```
Airblader: Can you elaborate on what other modes there are and why we ignore them?

src/handlers.c
```diff
@@ -141,6 +157,18 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
         return;
     }
 
+    bool refresh = false;
```
Airblader: nit: I think this name does not reflect what this variable is for.

src/handlers.c
```diff
@@ -152,6 +180,15 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
     if (con == NULL) {
         DLOG("Getting screen at %d x %d\n", event->root_x, event->root_y);
         check_crossing_screen_boundary(event->root_x, event->root_y);
+
+        if (refresh && !config.disable_focus_follows_mouse) {
+            /* The user entered the root window coming from another X
+             * screen, this means that the above will not have called
+             * tree_render (leaving the keyboard focus on the other
+             * screen) let's force a rerender.
```
Airblader: nit: This seems to be its own sentence.

src/handlers.c
```diff
@@ -152,6 +180,15 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
...
+             */
+            tree_render();
```
Airblader: So shouldn't this just be included in `check_crossing_screen_boundary`? On a side note, that function seems to be quite a misnomer in the X terminology since really what it handles is outputs, not screens (and now we do handle actual screens). Not your change, of course, just something I've noticed.

src/handlers.c
```diff
@@ -141,6 +157,18 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
...
 
+    bool refresh = false;
+    if (event->same_screen_focus != previous_screen) {
+        /* Note we should not skip on previous_screen being
+         * UINT8_MAX because we don't want the first display
+         * switch to be missed (say we have i3 on :0.0 and :0.1,
+         * the :0.1 i3 will not get an event in general until
+         * the first time the user tries to use it, since the
+         * mouse pointer is likely on :0.0 to start with */
+        DLOG("The user changed screens\n");
```
Airblader: nit: The log message should state the old and new value.

Airblader:
I've only given it a quick look, but in principle this looks good to me (I haven't tested it). I think you'll find writing a test to be rather difficult, I don't think the test infrastructure is set up to do this at the moment. I'd personally accept not having a test for it, though, of course, it would be quite nice if we did.

Airblader:
One more thing, please be a little bit more specific in the commit message and describe the nature of the change in detail. Also, if you add a *separate* line to the commit message saying "fixes #2082", Github will connect the issue and PR and close the issue once the PR is merged.

src/handlers.c
woodensquares: to be hones on this one I just pasted the same logic as enter (figuring that leave/enter would need to behave the same way), I am not sure if it would be required in this case or not, but it hasn't caused me problems up to now

src/handlers.c
woodensquares: sorry about that, will change

src/handlers.c
woodensquares: I named it this way because the idea behind everything is that when the pointer comes back onto the X screen the keyboard focus is "refreshed" to where it's supposed to be, what do you suggest I call it?

src/handlers.c
woodensquares: sure, will add

src/handlers.c
Airblader: Ah, I didn't realize the handle_notify case had this, too. A quick look at the documentation shows that other modes are grabbing cases. Makes sense to ignore those.

src/handlers.c
woodensquares: that is already in check_crossing_screen_boundary, but it's executed only if old_focused != focused (which is not the case here, as from the i3 perspective there was no focus change on the screen we are on), I didn't want to have an unconditional tree_render for normal operation, just when the screen is changed

src/handlers.c
Airblader: Yeah, I get the intention. You're obviously looking at this code through the glasses of having made the change. Someone reading this in a year probably won't be able to imply that "refresh" has something to do with input focus :-) How about something like `focus_crossed_screen`?

src/handlers.c
Airblader: Yeah, I mean move the `(refresh && !disable_…) { tree_render() }` logic into `check_crossing_screen_boundary`. It's just a new case to be handled within that (albeit unfortunately named) function.

woodensquares:
I have addressed your comments, squashed and changed the commit message, let me know if this is better, this said I would prefer to keep testing this a bit more if you don't mind just in case I have missed something (I run two X screens full time and use them quite a bit so if there are issues I should theoretically hit them hopefully)

src/handlers.c
woodensquares: ah, sure, I can definitely do that

src/handlers.c
woodensquares: heh, I know what you mean, I will rename

Airblader:
>  this said I would prefer to keep testing this a bit more if you don't mind

Absolutely. We're in no hurry, so take however much time you want to test it (Michael will be the one deciding on the merge anyway – I'm just here to leave my two cents). :-)

woodensquares:
will do, anyways in the meantime I have updated the pull with your latest comments, not sure if anybody here is using i3 on multiple X screens, but if anybody feels like testing this let me know if you see any issues

src/handlers.c
```diff
@@ -185,21 +222,25 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
...
     Con *ws = con_get_workspace(con);
-    if (ws != con_get_workspace(focused))
+    if (ws != con_get_workspace(focused) || focus_crossed_screen)
         workspace_show(ws);
 
     focused_id = XCB_NONE;
     con_focus(con_descend_focused(con));
     tree_render();
 
+    previous_screen = event->same_screen_focus;
```
woodensquares: Note this has to be here for things to work correctly, otherwise it seems to work only for the root window case (switching screens with no windows on either) and not for cases where the screen is switched and there are containers on both

woodensquares:
Just chiming in after a week everything seems to be working well, the only unintuitive corner case-ish behavior I was able to trigger was this:

have a terminal open in both :0.0 and :0.1, in :0.1's terminal run sleep 10 ; exit, press enter then move the mouse to the one in :0.0, confirm that keyboard focus is there and wait, when the terminal in :0.1 will close the keyboard focus will leave and go to that screen

this is the log from i3 (both logs, from the i3 on :0.0 and :0.1) when this happens

https://gist.github.com/woodensquares/5aab31f7f05ddc3ae5c423cb5047f4bf

none of my code changes seem to be exercised by this, it's not a big deal, moving the mouse to another monitor and back fixes it, but it is surprising that the focus gets stolen by the other screen when a window there closes. Besides this corner case everything has been working well for me.

Airblader:
What's the status on this?

woodensquares:
sorry for the delay, was on vacation, I am still using this patch and it's been working fine for me over the months with the issue listed above with a window closing on a different screen that will steal the focus, which I was not able to fix. 

Given that it is pretty niche to have more than one X screen it's really up to you if you think if it's worth mainlining this or leaving it as an open issues for the (likely very few) interested parties that would then patch i3 themselves if needed

Airblader:
I'm not opposed to merging this, but it's up to @stapelberg. In any case before a merge this needs a rebase.

stapelberg:
I’m leaning towards not merging this PR for now:

* i3 doesn’t aim to support multi-screen setups. It’s not clear to me whether this PR is really all that’s necessary, or whether there are other corner cases that need to be addressed as well.
* Even if the PR is all that’s needed for now, there might be regressions in the future because our core developers don’t use a multi-screen setup. This point could be addressed by extending the test suite, which would only be acceptable under the condition that doing so doesn’t significantly slow down the test suite. Even if the test suite was modified that way, it’s not clear who would maintain this feature and fix issues — I’m not keen on signing the core developers up to support a feature they don’t need.

I also don’t have a good intuition on user demand for this change. I’m guessing only a tiny minority of i3 users would benefit from this change.

Airblader:
I didn't really intend to take this PR as a "we now support it" change, but just as an improvement to the situation. Given that the changes aren't significantly complex, it would've been OK for me.

Since I'm not affected myself I of course don't feel strongly. We could also open a "voting issue" where we ask people coming across this to leave a vote if they're affected and then revisit with 4.14?

stapelberg:
> I didn't really intend to take this PR as a "we now support it" change, but just as an improvement to the situation.

We currently don’t have that concept in i3 (at least AFAIR). We only have things that are supported, and things which are explicitly not supported but required for testing. We don’t currently have things that should work but if they break it’s too bad.

> We could also open a "voting issue" where we ask people coming across this to leave a vote if they're affected and then revisit with 4.14?

That seems like a good way to judge user interest, yes.

Airblader:
> We currently don’t have that concept in i3 (at least AFAIR).

I think we do – with compositing. Many i3 users use compositing and quite often i3 is only »usable« for them with a compositor.

stapelberg:
I’d say compositing is explicitly not supported. It’s a sad fact that using a compositor sometimes seems required to work around bugs that are believed to reside outside of i3. People who run a compositor are not using a supported setup, but nevertheless it might work better for them than a supported setup.

I wonder if we’re getting too far down the rabbit hole with regards to definitions in this discussion, though, so let’s try to step back: even if one wants to see compositors as a precedent for not-supported-but-not-entirely-unsupported, I’d still argue we should try hard to not increase the number of such cases :).

Airblader:
Fair enough. Shall we close this then?

stapelberg:
Yep.

PR #2332, opened 2016-05-01
Use the last known timestamp when calling xcb_set_input_focus().

### LINKED ISSUE #2322: Set input focus with last known timestamp
When calling `xcb_set_input_focus`, we should do so with `last_timestamp` rather than `XCB_CURRENT_TIME`.

I'm only opening this issue as a reminder becasue there currently are a few pending pull requests and issues in that area and keeping multiple pull requests up to date is just more stressful. I'll provide the patch as soon as those are through.

i3 4.12

Airblader:
In most cases this won't make a difference, but consider the following
scenario:

* Some application acquires input focus on its own (which the window manager
  is not involved in, so it will "always" work).
* Before the FocusIn / FocusOut events are processed by i3, the
  workflow to push X changes is run, resulting in xcb_set_input_focus()
  being called.

Using XCB_CURRENT_TIME, this means i3 will "win" and override the focus,
even though it simply wasn't aware (yet) that focus had already been taken
over by someone else. When processing the FocusIn / FocusOut events, i3
assumes that the focus is already set which results in a broken focus
state as i3 assumes some container has the input focus when it doesn't.

With the new behavior, i3's attempt to set the input focus will fail if
the FocusIn / FocusOut events have not yet processed. Once they are processed,
the focus state will be automatically corrected.

fixes #2322

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -1120,7 +1120,7 @@ void x_push_changes(Con *con) {
                     values[0] = CHILD_EVENT_MASK & ~(XCB_EVENT_MASK_FOCUS_CHANGE);
                     xcb_change_window_attributes(conn, focused->window->id, XCB_CW_EVENT_MASK, values);
                 }
-                xcb_set_input_focus(conn, XCB_INPUT_FOCUS_POINTER_ROOT, to_focus, XCB_CURRENT_TIME);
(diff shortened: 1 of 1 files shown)


Airblader:
On a side note, both awesome and Openbox do this as well. I've also checked that i3 updates `last_timestamp` in exactly the same cases as awesome.

PR #2336, opened 2016-05-02
Introduce title_alignment directive for window title alignment.

### LINKED ISSUE #1750: Make title alignment configurable
The alignment of the title bar text should be configurable to `center`, `left`, or `right`.

Multiple formats can be given (for such as `left` and `right`) that will display window info on both sides of the titlebar.

For context, the current default for the `left` alignment is `%title` and the default for `right` is `[%marks]`.

See also #1723

Airblader:
This commit introduces a new config directive, title_alignment, which
controls how window titles are rendered (left, center, right).

fixes #1750

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -573,6 +573,26 @@ workspace_layout default|stacking|tabbed
 workspace_layout tabbed
 ---------------------
 
+[[title_alignment]]
+=== Window title alignment
+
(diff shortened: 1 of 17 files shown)


Airblader:
I'm closing this since the discussion isn't at a point I thought it was at yet. The PR is referenced in the issue, so we can revisit it if we want to. In the meantime there's no point in keeping it open.

PR #2340, opened 2016-05-05
do not match docks in config and command criteria

cornerman:
When currently matching windows with criteria in the config or in a command, the dock clients are also matched. For example, running `i3-msg [class="i3bar"] fullscreen` crashes i3 and `i3-msg [class=".*"] move workspace current` moves all windows including the i3bar windows as normal containers to the current workspace.

This seems wrong, as I cannot focus dock clients and therefore do not see a reason why criteria should match them to issue commands on them. So, I have changed the match struct to not match dock clients for command and config criteria. The matching for loading and restoring layout is left unchanged, as I was not sure whether it should apply here, too.

### DIFF
--- a/testcases/t/264-dock-criteria.t
+++ b/testcases/t/264-dock-criteria.t
@@ -1,0 +1,71 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • http://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • http://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • http://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Verifies that command or config criteria does not match dock clients
+# Bug still in: 4.12-38-ge690e3d
+use i3test i3_autostart => 0;
+
+my $config = <<EOT;
+# i3 config file (v4)
+for_window [class="dock"] move workspace current
+EOT
+
+my $pid = launch_with_config($config);
+
+my $ws = fresh_workspace();
+
+
+## command criteria should not match dock windows
+open_window({
+    window_type => $x->atom(name => '_NET_WM_WINDOW_TYPE_DOCK'),
+    wm_class => "x"
+});
+
+is(get_dock_clients, 1, "created one docked client");
+is_num_children($ws, 0, 'no container on the current workspace');
+
+cmd '[class="^x$"] move workspace current';
+
+does_i3_live
+is(get_dock_clients, 1, "one docked client after move");
+is_num_children($ws, 0, 'no container on the current workspace');
+
+cmd '[class="^x$"] fullscreen';
+
+does_i3_live
+is(get_dock_clients, 1, "one docked client after fullscreen");
+is_num_children($ws, 0, 'no container on the current workspace');
+
+cmd '[class="^x$"] kill';
+
+does_i3_live
+is(get_dock_clients, 1, "one docked client after kill");
+is_num_children($ws, 0, 'no container on the current workspace');
+
+
+## config criteria should not match dock windows
+open_window({
+    window_type => $x->atom(name => '_NET_WM_WINDOW_TYPE_DOCK'),
+    wm_class => "dock"
+});
+
+does_i3_live
+is(get_dock_clients, 2, "created second docked client");
+is_num_children($ws, 0, 'no container on the current workspace');
+
+
+exit_gracefully($pid);
+done_testing;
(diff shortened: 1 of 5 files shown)


Airblader:
I have mixed feelings about unconditionally excluding dock windows from match criteria. I think I prefer each command deciding on its own which window / container it can apply to, with repetition being avoided by proper refactoring.

On the other hand, we've had problems with this multiple times in the past and the only command that I could see be useful for dock windows at the moment is `kill`, and even that only in scripts and as a very theoretical usecase.

Airblader:
Addendum: I think the best argument in favor of this change at the moment is the `class=.*` match. But this could also be solved otherwise, for example with a `nodock` criterion, or even `con_type=leaf|nested|dock`.

It should also be noted that `class=.*` will not actually apply to all windows, but only to those that set `WM_CLASS`.

cornerman:
I understand the problem with unconditionally excluding dock clients, but, as you said, especially the `class=.*` match is probably undesired by the user. I also only see a usecase for the `kill` command on dock clients, but that could also be achieved with `xkill`.

Not sure whether adding an option to the crtiteria would be worth it. But if you decide for an extra option, I would be in favor of excluding dock clients by default. Then, it might be a good idea to revisit the commands and check which ones can actually operate on docks.

Airblader:
> especially the class=.* match is probably undesired by the user

But as I explained above, this match doesn't actually do what the user expects it to do *anyway*. I don't think it's a good idea for a user to match this way.

> but that could also be achieved with xkill.

But that's not an i3 command.

> But if you decide for an extra option, I would be in favor of excluding dock clients by default. 

All match criteria are ANDed, there's no way to OR them, so this wouldn't help.

Airblader:
I retract my first point in the previous comment. `WM_CLASS` is a mandatory property of any mapped window, so we should take the position that we can in fact expect it to be present on windows. Which means a user should be able to expect `class=.*` to match "all windows".

The other point still stands, though, excluding docks by default would render an additional criterion useless unless we implement a full conjunctive normal form.

cornerman:
> All match criteria are ANDed, there's no way to OR them, so this wouldn't help.

I do not see the problem with excluding dock clients by default. As you said, one could add an additional criterion for docks and whether you switch dock clients off or on shouldn't make a difference, does it?

cornerman:
> I'm not decisively against this change, by the way; as I said, I just have mixed feelings at the moment.

Sure, no need to rush things. Also, I am not really familiar with layout loading but should this be done there as well?

acrisci:
If i3bar is the only dock client we are talking about, then of course it makes sense to exclude dock clients because i3bar should imo never be the target of commands. `kill` might be the only close to valid command, but we support all the reasons why you would do that with features like `reload`, `restart`, and the commands to hide the bar.

Are there any other dock windows in common use to take into consideration? Can we imagine a dock window for which it would be valid to move between workspaces, or other commands like that?

If the "select all windows" use case is the only thing we want to support, instead of nerfing criteria, how about simply introducing a new kind of selection criteria just for this purpose?

```
i3-msg '[all_windows] kill'
```

And then we can define a "window" for this purpose to exclude dock windows.

I'm just thinking aloud here.

acrisci:
So it appears that dock windows actually should not be moved (unless they themselves request to be moved).

[_NET_WM_STRUT_PARTIAL](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472578320) determines the location of the dock window at the edge of the screen so the wm cannot move the dock window.

So my thinking now is that dock windows are essentially like unmanaged windows, since they cannot accept focus or be moved by the window manager. So maybe criteria should only be applied to proper managed windows.

I think that is strong reasoning for excluding dock windows from criteria.

Airblader:
@stapelberg This PR has been around for a while now. Should we go ahead with it and get the PR rebased?

stapelberg:
Yes.

@cornerman Could you rebase please?

cornerman:
@stapelberg Done!

Though one test is failing on travis, all tests pass on my machine.

stapelberg:
I’ll have to insist on getting the tests passing on travis, as travis is instrumental for judging whether contributions are safe to merge. Given that the test failed in both runs (gcc and clang) and over two separate attempts (when you pushed, and after I restarted the test just now), it does not seem to be a one-time flake.

See https://github.com/i3/i3/blob/next/.travis.yml for details on our travis setup. You should be able to run the Docker containers from https://hub.docker.com/r/i3wm/ to get the same environment that we use on travis.

Let me know if you still have trouble reproducing the issue.

testcases/t/264-dock-criteria.t
```diff
@@ -0,0 +1,70 @@
...
+## config criteria should not match dock windows
+open_window({
+    window_type => $x->atom(name => '_NET_WM_WINDOW_TYPE_DOCK'),
+    wm_class => "dock"
+});
+
+does_i3_live
+is(get_dock_clients, 2, "created second docked client");
+is_num_children($ws, 0, 'no container on the current workspace');
+
```
Airblader: @cornerman I think all you're missing is the call to `exit_gracefully` here.

cornerman: Great, this fixes it!

stapelberg:
Can you rebase please?

cornerman:
@stapelberg yes!

PR #2343, opened 2016-05-05
i3-input@main.c: free a duplicated string if we exit early

neosilky:


stapelberg:
This is not a memleak. When returning from main, the process exits and the operating system reclaims the memory. We don’t `free` memory before exiting in other parts of i3 either.

PR #2344, opened 2016-05-05
Use posixly correct getopt in i3-msg

### LINKED ISSUE #2339: i3-msg doesn't parse double dash options like --toggle
```
$ i3-msg -v
i3-msg 4.12 (2016-03-06, branch "4.12")
```

**What I did:**
```
$ i3-msg mark --toggle test
i3-msg: unknown option -- toggle
[{"success":true}]
$ i3-msg mark --toggle test
i3-msg: unknown option -- toggle
[{"success":true}]
```
**What I saw:**

On the first call i3 added the tag ``test`` to the current container/window. On the second call nothing happened.

**What I expected instead:**
I expected the tag to be removed after the second call and no ``unknown option`` message from ``i3-msg``.

madroach:
This allows calling 'i3-msg mark --toggle test' without '--' or quoting.
resolves #2339

### DIFF
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -141,7 +141,7 @@ int main(int argc, char *argv[]) {
         {"help", no_argument, 0, 'h'},
         {0, 0, 0, 0}};
 
-    char *options_string = "s:t:vhq";
+    char *options_string = "+s:t:vhq";
 
     while ((o = getopt_long(argc, argv, options_string, long_options, &option_index)) != -1) {
         if (o == 's') {


Airblader:
If this is all it takes then I'll gladly change my opinion :-)

stapelberg:
I’m not sure this is a good idea, actually.

I think I’d prefer it if users who like this behavior would set the `POSIXLY_CORRECT` environment variable instead of forcing all users of i3-msg to the posixly correct behavior.

madroach:
I'm used to the 'first positional argument cancels option processing' behaviour similar commands like ``sudo``, ``doas``, ``xargs`` use. It is the reasonable behaviour for this kind of command. 
Ever tried ``touch foo -bar``? or ``mv foo -foobar``. Seems like few commands actually use the non-posix behaviour. The non-posix commands usually pay close attention to the order and association of options with positional parameters. ``sox`` comes to mind. i3-msg does not belong in that class of commands, but rather fits well with ``xargs`` and ``sudo``.
Where's the downside of it? In what way do you think are people without POSIXLY_CORRECT forced to type different commands?

On May 6, 2016 9:45:03 AM CEST, Michael Stapelberg <notifications@github.com> wrote:
>I’m not sure this is a good idea, actually.
>
>I think I’d prefer it if users who like this behavior would set the
>`POSIXLY_CORRECT` environment variable instead of forcing all users of
>i3-msg to the posixly correct behavior.
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/i3/i3/pull/2344#issuecomment-217372404


--
http://gmerlin.de
OpenPGP: http://gmerlin.de/christopher.pub
F190 D013 8F01 AA53 E080  3F3C F17F B0A1 D44E 4FEE

jolange:
I don't have a strong preference here, but it certainly changes behaviour: `i3-msg floating toggle --quiet` would no longer work.

stapelberg:
> Where's the downside of it?

My biggest concern is what @jolange mentioned: we’d be changing i3-msg’s behavior in such a way that it might break existing user’s scripts. Currently, I’m favoring not risking any breakage over changing the behavior to better suit an individual user.

Airblader:
It's been a month without any further arguments on why to favor this change, so I'll close the PR.

PR #2350, opened 2016-05-12
Fix building and testing on OpenBSD

madroach:
Those changes allow building on OpenBSD whose `pthread` implementation in not feature complete.
in `sd-daemon.c` there is this `!defined(__linux__)` check. I couldn't find any Linuxism in the following code and it builds just fine on OpenBSD. Maybe remove the check altogether instead of just checking for OpenBSD?

### DIFF
--- a/i3-dump-log/main.c
+++ b/i3-dump-log/main.c
@@ -28,8 +28,10 @@
 #include "shmlog.h"
 #include <i3/ipc.h>
 
-static uint32_t offset_next_write,
-    wrap_count;
+#if !defined(__OpenBSD__)
+static uint32_t offset_next_write;
+#endif
+static uint32_t wrap_count;
 
 static i3_shmlog_header *header;
 static char *logbuffer,
@@ -57,17 +59,26 @@ static void print_till_end(void) {
 
 int main(int argc, char *argv[]) {
     int o, option_index = 0;
-    bool verbose = false,
-         follow = false;
+    bool verbose = false;
+#if !defined(__OpenBSD__)
+    bool follow = false;
+#endif
 
     static struct option long_options[] = {
         {"version", no_argument, 0, 'v'},
         {"verbose", no_argument, 0, 'V'},
+#if !defined(__OpenBSD__)
         {"follow", no_argument, 0, 'f'},
+#endif
         {"help", no_argument, 0, 'h'},
-        {0, 0, 0, 0}};
+        {0, 0, 0, 0}
+    };
 
+#if !defined(__OpenBSD__)
     char *options_string = "s:vfVh";
+#else
+    char *options_string = "vVh";
+#endif
 
     while ((o = getopt_long(argc, argv, options_string, long_options, &option_index)) != -1) {
         if (o == 'v') {
@@ -75,11 +86,17 @@ int main(int argc, char *argv[]) {
             return 0;
         } else if (o == 'V') {
             verbose = true;
+#if !defined(__OpenBSD__)
         } else if (o == 'f') {
             follow = true;
+#endif
         } else if (o == 'h') {
             printf("i3-dump-log " I3_VERSION "\n");
-            printf("i3-dump-log [-f] [-s <socket>]\n");
+#if !defined(__OpenBSD__)
+            printf("i3-dump-log [-fhVv]\n");
+#else
+            printf("i3-dump-log [-hVv]\n");
+#endif
             return 0;
         }
     }
@@ -162,6 +179,7 @@ int main(int argc, char *argv[]) {
     walk = logbuffer + sizeof(i3_shmlog_header);
     print_till_end();
 
+#if !defined(__OpenBSD__)
     if (follow) {
         /* Since pthread_cond_wait() expects a mutex, we need to provide one.
          * To not lock i3 (that’s bad, mhkay?) we just define one outside of
@@ -177,6 +195,7 @@ int main(int argc, char *argv[]) {
             }
         }
     }
+#endif
 
     return 0;
 }
--- a/src/sd-daemon.c
+++ b/src/sd-daemon.c
@@ -45,9 +45,6 @@
 #include "sd-daemon.h"
 
 int sd_listen_fds(int unset_environment) {
-#if defined(DISABLE_SYSTEMD) || !defined(__linux__)
-    return 0;
-#else
     int r, fd;
     const char *e;
     char *p = NULL;
@@ -121,7 +118,6 @@ finish:
     }
 
     return r;
-#endif
 }
 
 int sd_is_fifo(int fd, const char *path) {
--- a/testcases/t/180-fd-leaks.t
+++ b/testcases/t/180-fd-leaks.t
@@ -20,6 +20,9 @@ use i3test;
 use POSIX qw(mkfifo);
 use File::Temp qw(:POSIX tempfile);
 
+SKIP: {
+skip "Procfs not available on $^O", 1 if $^O eq 'openbsd';
+
 my $i3 = i3(get_socket_path());
 
 my $tmp = tmpnam();
@@ -58,5 +61,7 @@ for my $fd (keys %fds) {
 }
 
 is(scalar keys %fds, 0, 'No file descriptors leaked');
+
+}
 
 done_testing;
--- a/include/shmlog.h
+++ b/include/shmlog.h
@@ -11,7 +11,9 @@
 #pragma once
 
 #include <stdint.h>
+#if !defined(__OpenBSD__)
 #include <pthread.h>
+#endif
 
 /* Default shmlog size if not set by user. */
 extern const int default_shmlog_size;
@@ -37,8 +39,10 @@ typedef struct i3_shmlog_header {
      * and don’t matter — clients use an equality check (==). */
     uint32_t wrap_count;
 
+#if !defined(__OpenBSD__)
     /* pthread condvar which will be broadcasted whenever there is a new
      * message in the log. i3-dump-log uses this to implement -f (follow, like
      * tail -f) in an efficient way. */
     pthread_cond_t condvar;
+#endif
 } i3_shmlog_header;
--- a/src/i3.mk
+++ b/src/i3.mk
@@ -6,7 +6,10 @@ i3_SOURCES           := $(filter-out $(i3_SOURCES_GENERATED),$(wildcard src/*.c)
 i3_HEADERS_CMDPARSER := $(wildcard include/GENERATED_*.h)
 i3_HEADERS           := $(filter-out $(i3_HEADERS_CMDPARSER),$(wildcard include/*.h))
 i3_CFLAGS             = $(XKB_COMMON_CFLAGS) $(XKB_COMMON_X11_CFLAGS) $(XCB_CFLAGS) $(XCB_KBD_CFLAGS) $(XCB_WM_CFLAGS) $(XCB_CURSOR_CFLAGS) $(XCB_XRM_CFLAGS) $(PANGO_CFLAGS) $(YAJL_CFLAGS) $(LIBEV_CFLAGS) $(PCRE_CFLAGS) $(LIBSN_CFLAGS)
-i3_LIBS               = $(XKB_COMMON_LIBS) $(XKB_COMMON_X11_LIBS) $(XCB_LIBS) $(XCB_XKB_LIBS) $(XCB_KBD_LIBS) $(XCB_WM_LIBS) $(XCB_CURSOR_LIBS) $(XCB_XRM_LIBS) $(PANGO_LIBS) $(YAJL_LIBS) $(LIBEV_LIBS) $(PCRE_LIBS) $(LIBSN_LIBS) -lm -lpthread
+i3_LIBS               = $(XKB_COMMON_LIBS) $(XKB_COMMON_X11_LIBS) $(XCB_LIBS) $(XCB_XKB_LIBS) $(XCB_KBD_LIBS) $(XCB_WM_LIBS) $(XCB_CURSOR_LIBS) $(XCB_XRM_LIBS) $(PANGO_LIBS) $(YAJL_LIBS) $(LIBEV_LIBS) $(PCRE_LIBS) $(LIBSN_LIBS) -lm
+ifneq ($(UNAME),OpenBSD)
+i3_LIBS               += -lpthread
+endif
 
 # When using clang, we use pre-compiled headers to speed up the build. With
 # gcc, this actually makes the build slower.
--- a/src/log.c
+++ b/src/log.c
@@ -20,7 +20,9 @@
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <errno.h>
+#if !defined(__OpenBSD__)
 #include <pthread.h>
+#endif
 
 #include "util.h"
 #include "log.h"
@@ -157,11 +159,13 @@ void open_logbuffer(void) {
 
     header = (i3_shmlog_header *)logbuffer;
 
+#if !defined(__OpenBSD__)
     pthread_condattr_t cond_attr;
     pthread_condattr_init(&cond_attr);
     if (pthread_condattr_setpshared(&cond_attr, PTHREAD_PROCESS_SHARED) != 0)
         fprintf(stderr, "pthread_condattr_setpshared() failed, i3-dump-log -f will not work!\n");
     pthread_cond_init(&(header->condvar), &cond_attr);
+#endif
 
     logwalk = logbuffer + sizeof(i3_shmlog_header);
     loglastwrap = logbuffer + logbuffer_size;
@@ -277,8 +281,10 @@ static void vlog(const bool print, const char *fmt, va_list args) {
 
         store_log_markers();
 
+#if !defined(__OpenBSD__)
         /* Wake up all (i3-dump-log) processes waiting for condvar. */
         pthread_cond_broadcast(&(header->condvar));
+#endif
 
         if (print)
             fwrite(message, len, 1, stdout);
--- a/src/startup.c
+++ b/src/startup.c
@@ -103,8 +103,8 @@ static int _prune_startup_sequences(void) {
  */
 void startup_sequence_delete(struct Startup_Sequence *sequence) {
     assert(sequence != NULL);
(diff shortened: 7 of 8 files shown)


madroach:
travis complains about the code formatting. What's wrong with it in `i3-dump-log/main.c`?

Airblader:
@madroach Travis actually tells you what's wrong :)

````
+ git diff --exit-code
diff --git a/i3-dump-log/main.c b/i3-dump-log/main.c
index b63e90b..665dc4a 100644
--- a/i3-dump-log/main.c
+++ b/i3-dump-log/main.c
@@ -75,7 +75,8 @@ int main(int argc, char *argv[]) {
         {"follow", no_argument, 0, 'f'},
 #endif
         {"help", no_argument, 0, 'h'},
-        {0, 0, 0, 0}};
+        {0, 0, 0, 0}
+    };
 
 #if !defined(__OpenBSD__)
     char *options_string = "s:vfVh";
```

Airblader:
Also, not speaking about the change itself, but for a merge please squash the commits and keep the patch clean without merge commits.

i3-dump-log/main.c
```diff
@@ -57,29 +61,45 @@ static void print_till_end(void) {
 
 int main(int argc, char *argv[]) {
     int o, option_index = 0;
+#if !defined(__OpenBSD__)
     bool verbose = false,
          follow = false;
+#else
+    bool verbose = false;
```
Airblader: Instead of having the same line twice, I'd prefer

````
    bool verbose = false;
#if !defined(__OpenBSD__)
    bool follow = false;
#endif
```

testcases/t/180-fd-leaks.t
```diff
@@ -20,43 +20,45 @@ use i3test;
 use POSIX qw(mkfifo);
 use File::Temp qw(:POSIX tempfile);
 
-my $i3 = i3(get_socket_path());
+if (-d '/proc/self') {
```
Airblader: There's an actual SKIP directive which would be better, I think. I'm also not sure about how it is decided that the test is skipped.

madroach: You are concerned the test will be silently skipped when you actually would like to be told something is wrong/changed with your procfs?
I see two options:
* skip with `if (qx/uname/ eq 'OpenBSD')`
* use `fstat` instead, which would be portable to OpenBSD

Airblader: Actually never mind my comment about using `/proc/self`. I should have read the testcase better first. :-)

stapelberg:
@madroach The tests still fail: 

> Global symbol "$i3" requires explicit package name (did you forget to declare "my $i3"?) at ./t/180-fd-leaks.t line 25.

Can you fix the test case and give us a note when the pull request is ready to review please?

madroach:
As far as I am concerned this is ready to be merged. Feel free to commit it as a single commit. Or should I prepare a single diff and commit message?

src/sd-daemon.c
```diff
@@ -45,7 +45,7 @@
 #include "sd-daemon.h"
 
 int sd_listen_fds(int unset_environment) {
-#if defined(DISABLE_SYSTEMD) || !defined(__linux__)
+#if defined(DISABLE_SYSTEMD) || !defined(__linux__) && !defined(__OpenBSD__)
```
Airblader: Actually I don't quite like this. A `a || b && c` expression is rather difficult to interpret correctly intuitively. Can we use some parentheses to make it clearer how this expression is evaluated?

Airblader:
It'd be great if you could squash the commits into a single commit with a proper commit message, thank you. :-) I've also left another comment which I found just now.

madroach:
``` diff
From 1fae8b0f4c6355ea7faa145c090f3af8caad025e Mon Sep 17 00:00:00 2001
From: Christopher Zimmermann <madroach@gmerlin.de>
Date: Fri, 27 May 2016 18:28:05 +0200
Subject: [PATCH] Fixes for building and testing on OpenBSD

- common.mk: add -I/usr/local/include
- shmlog: don't use pthread on OpenBSD since it does not provide all required features
- sd-daemon.c: no special feature is required from the OS for socket activation. Why depend on systemd or linux?!?
- startup.c: 64 bit time_t
- StartXServer.pm: perl forks for glob() on OpenBSD. Therefore attach SIGCHLD handler later.
- disable 180-fd-leaks test, because it requires procfs. (could use fstat on OpenBSD)
---
 common.mk                     |  4 ++--
 i3-dump-log/main.c            | 31 +++++++++++++++++++++++++------
 include/shmlog.h              |  4 ++++
 src/i3.mk                     |  5 ++++-
 src/log.c                     |  6 ++++++
 src/sd-daemon.c               |  2 +-
 src/startup.c                 |  8 ++++----
 testcases/lib/StartXServer.pm | 18 +++++++++---------
 testcases/t/180-fd-leaks.t    |  5 +++++
 9 files changed, 60 insertions(+), 23 deletions(-)

diff --git common.mk common.mk
index 0215d35..a1678cf 100644
--- common.mk
+++ common.mk
@@ -177,9 +177,9 @@ I3_LDFLAGS += -Wl,-rpath,/usr/local/lib -Wl,-rpath,/usr/pkg/lib
 endif
 
 ifeq ($(UNAME),OpenBSD)
-I3_CFLAGS += -I${X11BASE}/include
+I3_CFLAGS += -I${X11BASE}/include -I${LOCALBASE}/include
 LIBS += -liconv
-I3_LDFLAGS += -L${X11BASE}/lib
+I3_LDFLAGS += -L${X11BASE}/lib -L${LOCALBASE}/lib
 endif
 
 ifeq ($(UNAME),FreeBSD)
diff --git i3-dump-log/main.c i3-dump-log/main.c
index 9d4eefc..d9b0613 100644
--- i3-dump-log/main.c
+++ i3-dump-log/main.c
@@ -28,8 +28,10 @@
 #include "shmlog.h"
 #include <i3/ipc.h>
 
-static uint32_t offset_next_write,
-    wrap_count;
+#if !defined(__OpenBSD__)
+static uint32_t offset_next_write;
+#endif
+static uint32_t wrap_count;
 
 static i3_shmlog_header *header;
 static char *logbuffer,
@@ -57,17 +59,26 @@ static void print_till_end(void) {
 
 int main(int argc, char *argv[]) {
     int o, option_index = 0;
-    bool verbose = false,
-         follow = false;
+    bool verbose = false;
+#if !defined(__OpenBSD__)
+    bool follow = false;
+#endif
 
     static struct option long_options[] = {
         {"version", no_argument, 0, 'v'},
         {"verbose", no_argument, 0, 'V'},
+#if !defined(__OpenBSD__)
         {"follow", no_argument, 0, 'f'},
+#endif
         {"help", no_argument, 0, 'h'},
-        {0, 0, 0, 0}};
+        {0, 0, 0, 0}
+    };
 
+#if !defined(__OpenBSD__)
     char *options_string = "s:vfVh";
+#else
+    char *options_string = "vVh";
+#endif
 
     while ((o = getopt_long(argc, argv, options_string, long_options, &option_index)) != -1) {
         if (o == 'v') {
@@ -75,11 +86,17 @@ int main(int argc, char *argv[]) {
             return 0;
         } else if (o == 'V') {
             verbose = true;
+#if !defined(__OpenBSD__)
         } else if (o == 'f') {
             follow = true;
+#endif
         } else if (o == 'h') {
             printf("i3-dump-log " I3_VERSION "\n");
-            printf("i3-dump-log [-f] [-s <socket>]\n");
+#if !defined(__OpenBSD__)
+            printf("i3-dump-log [-fhVv]\n");
+#else
+            printf("i3-dump-log [-hVv]\n");
+#endif
             return 0;
         }
     }
@@ -162,6 +179,7 @@ int main(int argc, char *argv[]) {
     walk = logbuffer + sizeof(i3_shmlog_header);
     print_till_end();
 
+#if !defined(__OpenBSD__)
     if (follow) {
         /* Since pthread_cond_wait() expects a mutex, we need to provide one.
          * To not lock i3 (that’s bad, mhkay?) we just define one outside of
@@ -177,6 +195,7 @@ int main(int argc, char *argv[]) {
             }
         }
     }
+#endif
 
     return 0;
 }
diff --git include/shmlog.h include/shmlog.h
index 01fea8b..0216c8d 100644
--- include/shmlog.h
+++ include/shmlog.h
@@ -11,7 +11,9 @@
 #pragma once
 
 #include <stdint.h>
+#if !defined(__OpenBSD__)
 #include <pthread.h>
+#endif
 
 /* Default shmlog size if not set by user. */
 extern const int default_shmlog_size;
@@ -37,8 +39,10 @@ typedef struct i3_shmlog_header {
      * and don’t matter — clients use an equality check (==). */
     uint32_t wrap_count;
 
+#if !defined(__OpenBSD__)
     /* pthread condvar which will be broadcasted whenever there is a new
      * message in the log. i3-dump-log uses this to implement -f (follow, like
      * tail -f) in an efficient way. */
     pthread_cond_t condvar;
+#endif
 } i3_shmlog_header;
diff --git src/i3.mk src/i3.mk
index 6cb4abf..de74782 100644
--- src/i3.mk
+++ src/i3.mk
@@ -6,7 +6,10 @@ i3_SOURCES           := $(filter-out $(i3_SOURCES_GENERATED),$(wildcard src/*.c)
 i3_HEADERS_CMDPARSER := $(wildcard include/GENERATED_*.h)
 i3_HEADERS           := $(filter-out $(i3_HEADERS_CMDPARSER),$(wildcard include/*.h))
 i3_CFLAGS             = $(XKB_COMMON_CFLAGS) $(XKB_COMMON_X11_CFLAGS) $(XCB_CFLAGS) $(XCB_KBD_CFLAGS) $(XCB_WM_CFLAGS) $(XCB_CURSOR_CFLAGS) $(XCB_XRM_CFLAGS) $(PANGO_CFLAGS) $(YAJL_CFLAGS) $(LIBEV_CFLAGS) $(PCRE_CFLAGS) $(LIBSN_CFLAGS)
-i3_LIBS               = $(XKB_COMMON_LIBS) $(XKB_COMMON_X11_LIBS) $(XCB_LIBS) $(XCB_XKB_LIBS) $(XCB_KBD_LIBS) $(XCB_WM_LIBS) $(XCB_CURSOR_LIBS) $(XCB_XRM_LIBS) $(PANGO_LIBS) $(YAJL_LIBS) $(LIBEV_LIBS) $(PCRE_LIBS) $(LIBSN_LIBS) -lm -lpthread
+i3_LIBS               = $(XKB_COMMON_LIBS) $(XKB_COMMON_X11_LIBS) $(XCB_LIBS) $(XCB_XKB_LIBS) $(XCB_KBD_LIBS) $(XCB_WM_LIBS) $(XCB_CURSOR_LIBS) $(XCB_XRM_LIBS) $(PANGO_LIBS) $(YAJL_LIBS) $(LIBEV_LIBS) $(PCRE_LIBS) $(LIBSN_LIBS) -lm
+ifneq ($(UNAME),OpenBSD)
+i3_LIBS               += -lpthread
+endif
 
 # When using clang, we use pre-compiled headers to speed up the build. With
 # gcc, this actually makes the build slower.
diff --git src/log.c src/log.c
index e8a08b5..e9a1160 100644
--- src/log.c
+++ src/log.c
@@ -20,7 +20,9 @@
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <errno.h>
+#if !defined(__OpenBSD__)
 #include <pthread.h>
+#endif
 
 #include "util.h"
 #include "log.h"
@@ -157,11 +159,13 @@ void open_logbuffer(void) {
 
     header = (i3_shmlog_header *)logbuffer;
 
+#if !defined(__OpenBSD__)
     pthread_condattr_t cond_attr;
     pthread_condattr_init(&cond_attr);
     if (pthread_condattr_setpshared(&cond_attr, PTHREAD_PROCESS_SHARED) != 0)
         fprintf(stderr, "pthread_condattr_setpshared() failed, i3-dump-log -f will not work!\n");
     pthread_cond_init(&(header->condvar), &cond_attr);
+#endif
 
     logwalk = logbuffer + sizeof(i3_shmlog_header);
     loglastwrap = logbuffer + logbuffer_size;
@@ -277,8 +281,10 @@ static void vlog(const bool print, const char *fmt, va_list args) {
 
         store_log_markers();
 
+#if !defined(__OpenBSD__)
         /* Wake up all (i3-dump-log) processes waiting for condvar. */
         pthread_cond_broadcast(&(header->condvar));
+#endif
 
         if (print)
             fwrite(message, len, 1, stdout);
diff --git src/sd-daemon.c src/sd-daemon.c
index f0b5ca0..04270a1 100644
--- src/sd-daemon.c
+++ src/sd-daemon.c
@@ -45,7 +45,7 @@
 #include "sd-daemon.h"
 
 int sd_listen_fds(int unset_environment) {
-#if defined(DISABLE_SYSTEMD) || !defined(__linux__)
+#if defined(DISABLE_SYSTEMD) || !(defined(__linux__) || defined(__OpenBSD__))
     return 0;
 #else
     int r, fd;
diff --git src/startup.c src/startup.c
index b7950c2..2a6bc00 100644
--- src/startup.c
+++ src/startup.c
@@ -103,8 +103,8 @@ static int _prune_startup_sequences(void) {
  */
 void startup_sequence_delete(struct Startup_Sequence *sequence) {
     assert(sequence != NULL);
-    DLOG("Deleting startup sequence %s, delete_at = %ld, current_time = %ld\n",
-         sequence->id, sequence->delete_at, time(NULL));
+    DLOG("Deleting startup sequence %s, delete_at = %lld, current_time = %lld\n",
+         sequence->id, (long long)sequence->delete_at, (long long)time(NULL));
 
     /* Unref the context, will be free()d */
     sn_launcher_context_unref(sequence->context);
@@ -239,8 +239,8 @@ void startup_monitor_event(SnMonitorEvent *event, void *userdata) {
             /* Mark the given sequence for deletion in 30 seconds. */
             time_t current_time = time(NULL);
             sequence->delete_at = current_time + 30;
-            DLOG("Will delete startup sequence %s at timestamp %ld\n",
-                 sequence->id, sequence->delete_at);
+            DLOG("Will delete startup sequence %s at timestamp %lld\n",
+                 sequence->id, (long long)sequence->delete_at);
 
             if (_prune_startup_sequences() == 0) {
                 DLOG("No more startup sequences running, changing root window cursor to default pointer.\n");
diff --git testcases/lib/StartXServer.pm testcases/lib/StartXServer.pm
index 86a37d9..d3042c0 100644
--- testcases/lib/StartXServer.pm
+++ testcases/lib/StartXServer.pm
@@ -69,15 +69,6 @@ sub start_xserver {
     my @displays = ();
     my @childpids = ();
 
-    $SIG{CHLD} = sub {
-        my $child = waitpid -1, POSIX::WNOHANG;
-        @pids = grep { $_ != $child } @pids;
-        return unless @pids == 0;
-        print STDERR "All X server processes died.\n";
-        print STDERR "Use ./complete-run.pl --parallel 1 --keep-xserver-output\n";
-        exit 1;
-    };
-
     # Yeah, I know it’s non-standard, but Perl’s POSIX module doesn’t have
     # _SC_NPROCESSORS_CONF.
     my $num_cores;
@@ -101,6 +92,15 @@ sub start_xserver {
 
     say "Starting $parallel Xephyr instances, starting at :$displaynum...";
 
+    $SIG{CHLD} = sub {
+        my $child = waitpid -1, POSIX::WNOHANG;
+        @pids = grep { $_ != $child } @pids;
+        return unless @pids == 0;
+        print STDERR "All X server processes died.\n";
+        print STDERR "Use ./complete-run.pl --parallel 1 --keep-xserver-output\n";
+        exit 1;
+    };
+
     my @sockets_waiting;
     for (1 .. $parallel) {
         my $socket = fork_xserver($keep_xserver_output, $displaynum,
diff --git testcases/t/180-fd-leaks.t testcases/t/180-fd-leaks.t
index 454bfe7..16272d3 100644
--- testcases/t/180-fd-leaks.t
+++ testcases/t/180-fd-leaks.t
@@ -20,6 +20,9 @@ use i3test;
 use POSIX qw(mkfifo);
 use File::Temp qw(:POSIX tempfile);
 
+SKIP: {
+skip "Procfs not available on $^O", 1 if $^O eq 'openbsd';
+
 my $i3 = i3(get_socket_path());
 
 my $tmp = tmpnam();
@@ -59,4 +62,6 @@ for my $fd (keys %fds) {
 
 is(scalar keys %fds, 0, 'No file descriptors leaked');
 
+}
+
 done_testing;
-- 
2.7.4
```

Airblader:
I assume you pasted this as a patch file because you don't know this, but: you can just squash the commits and then force push the branch. Github will then update the pull request automatically.

madroach:
Ok, heres the sqashed commit; but what's wrong with the clang build? To me this looks like an error in the docker framework.

Airblader:
@madroach I just restarted the build which fixed it. Happens sometimes. We only have to wait for Michael now who is probably just busy at the moment (he's the only one doing merges). In the meantime do you mind rebasing your commit to the current head? Thank you :)

stapelberg:
While squashing commits to keep the history clean is the correct practice, I’d like for the individual changes in this pull request to be individual commits (one commit per logical change).

The changes themselve look good, but can you please address the following nits before I merge them?

> shmlog: don't use pthread on OpenBSD since it does not provide all required features

Is there an OpenBSD issue which we can reference so that we can revert this commit once OpenBSD’s pthread implementation catches up?

> sd-daemon.c: no special feature is required from the OS for socket activation. Why depend on systemd or linux?!?

Instead of asking a question, I’d prefer if we could tell people what we are doing with some level of confidence. I suggest something like “sd-daemon.c: successfully tested on OpenBSD”.

src/sd-daemon.c
```diff
@@ -45,7 +45,7 @@
 #include "sd-daemon.h"
 
 int sd_listen_fds(int unset_environment) {
-#if defined(DISABLE_SYSTEMD) || !defined(__linux__)
```
madroach: Can this check be removed alltogether?

stapelberg: Yes, I think so. I can’t see platform-specific code in that function, at least not any more specific than what i3 already uses.

Airblader:
@stapelberg I've pushed Travis and the build succeeded now.

Issue #860, opened 2012-10-28
interactively-chosen arbitrary title strings


i3bot:
[Originally reported by i3-new-ticket]
(when a presses mod+t (or any key combination that is deemed superior), he should be prompted to set an arbitrary title for the currently focused client (application).

why? allowing the end-user to fix the title clutter that is resulted by applications that suggest long titles, and achieving better organization. e.g. instead using firefox's title, we can set a title that semantically is more meaningful to our own use (e.g. setting it to "work stuff", "coding sites", "personal stuff", "social websites"...etc).

i suggest adding something like "bindsym $mod+t set_title", which can in-turn (somehow) result in prompting the user for input.

Airblader:
As per @einalex's request: this has since been implemented with `title_format`.

PR #2359, opened 2016-05-21
adding custom container names (for composite and window containers)

einalex:
This should help with issue #860
It adds an additional attribute 'custom_name' to the Con struct and adjusts the rename command to allow setting and unsetting it for the currently selected container. (rename window[ \<name\>])
If a custom name is set it will be displayed in places where the container is represented somehow.
Custom names also get saved in JSON with the rest of the container and so will survive in place restarting of i3.

I chose to add an additional attribute instead of using the 'name' attribute because this allows to still display the class instance in the tree representation when no custom name has been set.
A prior attempt at using the 'name' attribute ended in the tree representation getting littered with window names. The container name seems to get set when the window changes its name.

I'd be happy to adjust this patch if needed.

### DIFF
--- a/include/commands.h
+++ b/include/commands.h
@@ -295,6 +295,12 @@ void cmd_scratchpad_show(I3_CMD);
 void cmd_title_format(I3_CMD, const char *format);
 
 /**
+ * Implementation of 'rename window <name>'
+ *
+ */
+void cmd_rename_window(I3_CMD, const char *name);
+
+/**
  * Implementation of 'rename workspace <name> to <name>'
  *
  */
--- a/include/con.h
+++ b/include/con.h
@@ -449,6 +449,12 @@ void con_set_urgency(Con *con, bool urgent);
 char *con_get_tree_representation(Con *con);
 
 /**
+ * Returns the container title.
+ *
+ */
+char *con_get_title(Con *con);
+
+/**
  * force parent split containers to be redrawn
  *
  */
--- a/include/data.h
+++ b/include/data.h
@@ -594,6 +594,7 @@ struct Con {
     struct Rect geometry;
 
     char *name;
+    char *custom_name;
 
     /** The format with which the window's name should be displayed. */
     char *title_format;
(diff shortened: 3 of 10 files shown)


Airblader:
What does this do that title_format doesn't do?

Airblader:
On another note, changing the windows name is not a good idea and violates the spec. The spec offers the visible name property for that which we already support with title_format.

einalex:
It allows you to set names for containers - including containers that do not have a window - individually.
For example it allows you to set a name for a composite container in which you would normaly display a tree.
You could also set the name of any container in a tree, leaf or node and have your chosen name appear in the tree at the appropriate place.

In my experience setting the name of the window doesn't help you very much as windows will keep changing their name overriding your settings in the process.

I agree that the window manager shouldn't arbitrarily change window names. I do not agree that the window's choice should take priority over the user's choice though.

Airblader:
> It allows you to set names for containers - including containers that do not have a window - individually.

Again, how is this not covered by `title_format` already? This directive can also be specified on nested containers. See here: http://i.imgur.com/8or0plP.png

> In my experience setting the name of the window doesn't help you very much as windows will keep changing their name overriding your settings in the process.

That's what `_NET_WM_VISIBLE_NAME` is for and `title_format` sets this property (and also displays the custom title). Plus `title_format` supports placeholders for, e.g., the original title.

einalex:
I didn't realize you could use title format without the placeholders.Yes it works well. Still the custom title set with title_format doesn't show up in the tree. That is a downside in my eyes.
I would appreciate it if you could set the corresponding bugs to solved
I didn't realize you could use title format without the placeholders.Yes it works well. Still the custom title set with title_format doesn't show up in the tree. That is a downside in my eyes.

I would appreciate it if you could mark the corresponding issues as fixed.

Airblader:
> That is a downside in my eyes.

I understand where you are coming from and that you'd want this. Of course it could be an idea to use the title in the tree representation, but I think this would too easily lead to "too long" representations. A config directive seems too much complexity for too little benefit to me. I could think of a bunch of other compromises, but they all boil down to not being worth it in my eyes. But if you have a proposal and feel strongly about it, please feel free to open this as a feature request so it can be discussed more.

Airblader:
> I would appreciate it if you could mark the corresponding issues as fixed.

I take that as confirmation that it's OK to close this PR.

PR #2370, opened 2016-06-07
i3-nagbar: added timeout arg

samsaga2:
I've added a timeout (in seconds) to i3-nagbar. It's useful to show some messages without the user interaction.

> i3-nagbar -m "Your toast is ready" -t warning -x 4


Airblader:
i3-nagbar isn't a notification daemon, so I oppose this change. Why not use something like dunst for this which implements a standardized specification for notifications and a much better user experience?

samsaga2:
You are right. I did not know dunst.

Issue #2378, opened 2016-06-18
Restore previous floating window size (and possibly position) when switching from tiling to floating


i336:
I switch apps between tiling and floating mode reasonably frequently (for example it's currently the only way to drag tabs between Chromium windows without occasional deadlocks; I also use it to work in a lot of terminals at once) and I've been puzzling over i3's mildly confusing floating window management for a little while now - every application seems to have its own idea of what size window to use when I switch to floating mode!

After pondering the source code for a bit, I finally understand: i3 is using the window's requested geometry to reset the window size. This does not work, in _many_ circumstances. For example, Geany saves window geometry on exit, and I most frequently use it on my (larger) center display, but I have it on a side display at the moment, so putting it into floating mode means it's too big. Chrome behaves very similarly as well. Also, setting a larger-than-normal font in `xterm` at runtime is a disaster - its window geometry will be based off a smaller font, producing a cropped display (48x15, to be exact) every time I switch from tiled to floated. :(

In short: I completely understand the rationale and elegant simplicity behind simply depending on the window geometry to restore the size, but in practice this breaks with a lot of things.

Configurable options to save/restore window size and position when entering/leaving floating mode would be awesome.

i3 is really cool, btw. :D

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

i336:
I'm using i3 4.12, but I was reading the code on GitHub when I drew the conclusions above, so *technically*, `master`.

Airblader:
In general I agree and would welcome such a change. However, one could argue that it doesn't really play nice with the intention of floating windows in i3 which is for them to be used for dialogs and menus and such only.
Personally, I think there's plenty of usecases for floating windows in everyday usage even in a tiled window manager and I think you also made a few good points on those.

However, there's definitely some edge cases to keep in mind. For example when moving a tiled window  to another output. Moving it to the last floating position would move it to another output, which is bad. Even when considering that, it might put it off-screen. So it's not a trivial problem.

i336:
Hmmm. I see.

[...]

Issue #1874, opened 2015-08-28
[enhancement] Support _NET_WM_STATE_MAXIMIZED_*


Airblader:
We always give tiled windows as much space as possible. Hence, we could set `_NET_WM_STATE_MAXIMIZED_VERT` and `_NET_WM_STATE_MAXIMIZED_HORZ` on tiled windows. Floating windows should not have them set.

According to the spec, we (w|c|sh)ould use this to restore the geometry of a previously floated window which was tiled and is now being floated again.

See http://standards.freedesktop.org/wm-spec/1.3/ar01s02.html

> Maximizing a window should give it as much of the screen area as possible (this may not be the full screen area, but only a smaller 'workarea', since the window manager may have reserved certain areas for other windows). A window manager is expected to remember the geometry of a maximized window and restore it upon de-maximization.

However, we could also argue that this doesn't fit into our model of limited floating support.

stapelberg:
I’m undecided on whether this makes sense. Do any other window managers set the hint? Do any toolkits actually look at the hint?

acrisci:
It might be nice to have some way to differentiate floating windows from tiling windows with some sort of property. I think something like this could give compton users a way to have shadows on floating windows and not tiling windows, but I'm not exactly sure how that would work.

But that's the only use case I can think of and it's not very strong.

Airblader:
I'm not sure if others set the hint or which tools actively use it. I do think that restoring the original geometry makes sense, but we technically don't need those atoms for that at all. I'll try to have a look to see if others set this, though I'm not sure when I get a chance to.

Airblader:
I've just verified that Kubuntu does set these values.

stapelberg:
Based on the fact that others don’t seem to set these hints either, I don’t think we should set them. i3’s user-base is not big enough to convince application authors to use the hint, and the use-case which @acrisci feels unclean — if we wanted to enable that, we should rather set a different (custom?) hint.

Airblader:
@stapelberg Are you sure you didn't misread my last comment? Kubuntu *does* set those values.

stapelberg:
Sorry, I indeed misread that. Okay, so the hint being set is a good start, but do applications actually use it? Perhaps using http://codesearch.debian.net/ is helpful for this.

Airblader:
I'll try to see if I can find something. I'll reopen for the meantime.

Airblader:
[...]

PR #2380, opened 2016-06-20
Add support for _NET_WM_STATE_MAXIMIZED_{HORZ, VERT}

yshui:
Some applications will adjust their window layout according to its maximized state.

For example, in Google Chrome, there's no space between the tabs and the top of the window when its maximized, saving some vertical space (see attached screenshot if you still can't understand).

When maximized:
screenshot_20160620_044909

When not:
screenshot_20160620_044923


Airblader:
Relevant: https://github.com/i3/i3/issues/1874
reddit with this user: https://www.reddit.com/r/i3wm/comments/4oh58a/troubles_with_chrome/

I'm a little surprised to see you provide a PR after you never replied back to my last comment on reddit. Of course, providing a PR is fine, but I think there's just still a discussion open about _how_ this should be supported. We just generally recommend discussing the issue first to avoid rejecting PRs.

But it's too late for that now, therefore:

* Can you please describe in words (and in the commit message) how these atoms are used in i3 with your implementation?
* Can you give a rationale for other scenarios so we can get a sense of whether this makes sense?

As for the change itself, we definitely require tests for this. I haven't looked at the code itself yet because I see no point until we decided whether this should be supported at all and if so, how exactly. We can't tailor the solution to work with a single application, we need to consider the big picture.

Airblader:
Also, I'm not convinced the Chrome case isn't a configuration issue. I don't have this gap:

screenshot-2016-06-20_11-12-00

yshui:
@Airblader 

> Also, I'm not convinced the Chrome case isn't a configuration issue
The gap is probably theme related. I can observe same behavior under different WM/DE.

> Can you please describe in words (and in the commit message) how these atoms are used in i3 with your implementation?
This is explained quite clearly by comments in con_is_maximized() (as well as by the code itself).

> Can you give a rationale for other scenarios so we can get a sense of whether this makes sense?
I can't really find another scenarios right now. But this implements part of the standard, and does not add much penalty performance-wise.

yshui:
@Airblader I guess you probably set the 'Use system title bar and borders' in Chrome?

Airblader:
> The gap is probably theme related. I can observe same behavior under different WM/DE.

The point is, I would argue that this change may in fact be _wrong_. We don't want to set an atom because it makes Chrome look nicer for someone. We would want to set the atom because we feel the situation appropriately represents what i3 considers "maximized". And that requires that we properly define what we should consider maximized and whether a concept of maximized even makes sense for a tiling window manager like i3 at all.

And I'm not convinced that this concept applies the way you implement it, cmp. with the spec:

> A window manager is expected to remember the geometry of a maximized window and restore it upon de-maximization.

There is no such thing for the windows you consider maximized. We only have this for fullscreen windows (which already use another, dedicated atom). How would you argue that your implementation represents this part of the spec?

On the other hand, we could also interpret 

> Maximizing a window should give it as much of the screen area as possible (this may not be the full screen area, but only a smaller 'workarea', since the window manager may have reserved certain areas for other windows)

to mean that _all_ tiled windows should be considered maximized (that's really what the tiling paradigm does).

So overall, again, I feel that your change doesn't implement the spec, but workarounds an issue you have with a specific application. It doesn't seem to me that it appropriately reflects the intention of the concept described in EWMH.

> This is explained quite clearly by comments in con_is_maximized() (as well as by the code itself).

Code can be wrong. I prefer a textual description of the intention of the change, again also in the commit message, so we can check whether that is how we want to represent this concept and whether the implementation matches that intention.

yshui:
@Airblader 

> Code can be wrong.

I could argue natural languages vague and ambiguous so reading the code is better, but anyway. My intention is that, if a window is put in container that takes the maximal amount of vertical/horizontal space, it should be considered vertically/horizontally maximized.

A maximized container does not share vertical/horizontal space with another one. In other words, a maximized container is not in a splith/splitv container with more than one child.

Airblader:
> My intention is that, if a window is put in container that takes the maximal amount of vertical/horizontal space, it should be considered vertically/horizontally maximized.

I think intuitively this makes sense because it's what users are used from floating window managers. My concerns above are still valid, however. These atoms are really only part of a larger maximization feature which i3, being a tiling window manager, simply doesn't have.

My concern lies less within the (indeed not all too significant) complexity of the change, but the fact that we implement a concept partially in a specific way to accommodate the insignificant needs of a single application; needs we know can be fulfilled otherwise and needs I'm not sure one shouldn't blame on the application for relying on something the window manager doesn't claim to support. Effectively, this could cause problems if we have better, actual usecases for these atoms down the road as it would then potentially become a backwards incompatible change.

That's why I think we should have more examples of applications using this hint (and looking at how they use it) and that's also the reason I closed my ticket proposing to support these atoms.

Anyhow, that's my opinion. I think I've said what I want to say; I'm happy to support this as long as we know that we support it in a sensible way. I'll leave this decision up to @stapelberg now.

yshui:
> My concern lies less within the (indeed not all too significant) complexity of the change, but the fact that we implement a concept partially in a specific way 

I think your concern is the "restore geometry on de-maximization" part, right? I can implement that. We just need to agree on what "de-maximization" means. I think it's reasonable to think making a container float as de-maximization.

Airblader:
> We just need to agree on what "de-maximization" means. I think it's reasonable to think making a container float as de-maximization.

That implies that floating windows are the "normal" state and tiled windows are just maximized windows. That fits neither to your implementation (which allows tiled windows to be non-maximized), nor does it fit the i3 philosophy (which is that tiled windows are the default).

I think that the concept of maximaztion, and therefore maximized windows, just simply doesn't apply to a tiling window manager such as i3. I'm open to change that view, I just haven't hard any argument yet that convinced me.

yshui:
> That implies that floating windows are the "normal" state and tiled windows are just maximized windows. 

Er, no. tiled windows are maximized only when it takes max amount of vertical/horizontal space. So when a maximized tiled window goes to floating, it's geometry last time it's floating is restored. Otherwise it will keep it's geometry when it's tiled.

> nor does it fit the i3 philosophy (which is that tiled windows are the default).

Why can't you think i3 as "maximized by default"? "maximized" is not a conflicting concept with "normal".

Airblader:
> Why can't you think i3 as "maximized by default"? "maximized" is not a conflicting concept with "normal".

Mh, I think I could get behind that kind of argument. I think it would even justify not having a "de-maximization" mechanism (I really think making it floating is not what we should consider de-maximized). I guess strictly speaking the spec doesn't demand that the concept of maximized windows requires a mechanism to (de-)maximize windows by the user.

yshui:
@Airblader 

> I really think making it floating is not what we should consider de-maximized

Yes, but I think restoring floating geometry might be something people would want.

Airblader:
Yes, but we already have #2378 for that, let's worry about it there. :-)

src/con.c
```diff
@@ -315,6 +315,44 @@ bool con_is_hidden(Con *con) {
     return false;
 }
```
Airblader: Please add the documentation of the function here as well.

src/con.c
```diff
@@ -315,6 +315,44 @@ bool con_is_hidden(Con *con) {
...
 
+bool con_is_maximized(Con *con, orientation_t o) {
```
Airblader: Please use a more descriptive parameter name than `o` (`orientation` is fine).

src/con.c
```diff
@@ -315,6 +315,44 @@ bool con_is_hidden(Con *con) {
...
+    Con *current = con;
+    layout_t l;
```
Airblader: Please use a more descriptive name.

src/con.c
```diff
@@ -315,6 +315,44 @@ bool con_is_hidden(Con *con) {
     return false;
 }
 
+bool con_is_maximized(Con *con, orientation_t o) {
+    Con *current = con;
```
Airblader: Both of these aren't actually needed until after the fullscreen condition.

src/con.c
```diff
@@ -315,6 +315,44 @@ bool con_is_hidden(Con *con) {
...
+
+    switch (o) {
+        case HORIZ:
+            l = L_SPLITH;
+            break;
+        case VERT:
+            l = L_SPLITV;
+            break;
+        default:
+            return false;
```
Airblader: This sounds like the function has been called incorrectly. Please use an `assert(false)` with a log message.

src/con.c
```diff
@@ -315,6 +315,44 @@ bool con_is_hidden(Con *con) {
...
+    }
+
+    /* go through all parents, if anyone of them is in split container
```
Airblader: * nit: s/go/Go/
* nit: /is in split/is in a split/
* nit: s/member/child/

src/con.c
```diff
@@ -315,6 +315,44 @@ bool con_is_hidden(Con *con) {
...
+     * with more than 1 member, return false. */
+    while (current != NULL && current->type != CT_WORKSPACE) {
+        Con *parent = current->parent;
+        if (parent && parent->layout == l) {
+            if (TAILQ_FIRST(&parent->nodes_head) != TAILQ_LAST(&parent->nodes_head, nodes_head))
```
Airblader: Can we please use `con_num_children` here to rely on existing abstractions?

src/x.c
```diff
@@ -663,6 +664,30 @@ static void set_hidden_state(Con *con) {
...
+ *
+ */
+static void set_maximize_state(Con *con) {
+    if (!con->window)
+        return;
+
+    con_state *state = state_for_frame(con->frame.id);
+    bool con_maximized = con_is_maximized(con, HORIZ);
+    if (con_maximized != state->is_maximized_horz) {
+        DLOG("setting MAXIMIZED_HORZ for con %p(%s) to %d\n", con, con->name, con_maximized);
```
Airblader: s/MAXIMIZED_HORZ/_NET_WM_STATE_MAXIMIZED_HORZ/

src/x.c
```diff
@@ -663,6 +664,30 @@ static void set_hidden_state(Con *con) {
...
+        xcb_add_or_remove_property_atom(conn, con->window->id, A__NET_WM_STATE, A__NET_WM_STATE_MAXIMIZED_HORZ, con_maximized);
+        state->is_maximized_horz = con_maximized;
+    }
+
+    con_maximized = con_is_maximized(con, VERT);
+    if (con_maximized != state->is_maximized_vert) {
+        DLOG("setting MAXIMIZED_VERT for con %p(%s) to %d\n", con, con->name, con_maximized);
```
Airblader: s/MAXIMIZED_HORZ/_NET_WM_STATE_MAXIMIZED_VERZ/

src/xcb.c
```diff
@@ -343,6 +343,19 @@ void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_a
 }
 
 /*
+ * Wrapper function for add or remote an atom to the given property, based on the
+ * 'add' parameter.
+ *
+ */
+
+void xcb_add_or_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_atom_t property, xcb_atom_t atom, bool add) {
+    if (add)
```
Airblader: Please use explicit parentheses to be consistent with at least most of the rest of the codebase.

src/xcb.c
```diff
@@ -343,6 +343,19 @@ void xcb_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_a
 }
 
 /*
+ * Wrapper function for add or remote an atom to the given property, based on the
+ * 'add' parameter.
+ *
+ */
+
+void xcb_add_or_remove_property_atom(xcb_connection_t *conn, xcb_window_t window, xcb_atom_t property, xcb_atom_t atom, bool add) {
```
Airblader: I'm pretty sure that some time ago we decided we don't want this wrapper function. We also generally don't want functions which decide their behavior based on a boolean argument.

Airblader:
I've reviewed the code now. Tests also need to be written still, though you may want to wait for Michael to decide whether this will be merged at all.

src/con.c
yshui: I just feel that's a bit inefficient? Maybe I can add a con_has_one_child() or con_is_only_child()?

src/xcb.c
yshui: Why is that? What disadvantage does doing this have?

yshui:
Fixed most of your concerns.

Also @stapelberg ?

include/con.h
```diff
@@ -200,6 +206,12 @@ Con *con_for_window(Con *con, i3Window *window, Match **store_match);
  */
 int con_num_children(Con *con);
 
+/*
```
Airblader: In the header file we need `**`.

src/con.c
```diff
@@ -316,6 +316,55 @@ bool con_is_hidden(Con *con) {
 }
 
 /*
+ * Check whether a con is maximized in the given orientation.
+ *
+ * A con is considered maximized, if it:
```
Airblader: +1 for the explanation, but please make sure the documentation of the header is the same

src/con.c
```diff
@@ -316,6 +316,55 @@ bool con_is_hidden(Con *con) {
...
+            assert(false);
+    }
+
+    /* Go through all parents, if anyone of them is in a split container
+     * with more than one child, return false. */
+    Con *current = con;
+    while (current != NULL && current->type != CT_WORKSPACE) {
+        Con *parent = current->parent;
+        if (parent && parent->layout == layout) {
+            if (!con_has_num_children(parent, 1))
```
Airblader: I don't think we have any performance concerns here, but I don't mind this either.

src/xcb.c
Airblader: Mostly two reasons:

* a boolean argument in particular is hard to read when encountering it because "true" and "false" don't really convey what they actually *mean*. This can be solved by using enums.
* functions whose behavior depends on a boolean flag is considered bad practice, e.g., see [this](http://programmers.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior).

Airblader:
> Also, ping @stapelberg ?

He seems to be quite busy lately, so it might take a bit. But he will definitely look at it. :-) Don't forget about them tests!

stapelberg:
I think the change should be okay, fundamentally. I have a couple of comments (see the diff), and of course it might be that we discover a compelling reason to revert the commit after it has gone out to users to test.

include/con.h
```diff
@@ -62,6 +62,20 @@ bool con_is_split(Con *con);
 bool con_is_hidden(Con *con);
 
 /**
+ * Check whether a con is maximized in the given orientation.
+ *
+ * A con is considered maximized, if it:
+ *
+ *   1) is a fullscreen container
```
stapelberg: Why are fullscreen containers considered maximized? Applications can already detect whether they’re fullscreened?

Is this part of the spec? Is this necessary for the use-case at hand?

include/con.h
```diff
@@ -62,6 +62,20 @@ bool con_is_split(Con *con);
...
+ *   2) doesn't share space with another container in the
+ *      given orientation (horizontal/verical). Which is to
+ *      say, this container is not placed within a splith or
+ *      a splitv conatiner with more than one child.
```
stapelberg: nit: s/conatiner/container/

src/x.c
```diff
@@ -43,6 +43,7 @@ typedef struct con_state {
     bool unmap_now;
     bool child_mapped;
     bool is_hidden;
+    bool is_maximized_horz, is_maximized_vert;
```
stapelberg: For consistency, please place each struct member definition on its own line.

src/x.c
```diff
@@ -663,6 +664,30 @@ static void set_hidden_state(Con *con) {
...
 /*
+ * Sets or remove _NET_WM_STATE_MAXIMIZE_{HORZ, VERT} on con
+ *
+ */
+static void set_maximize_state(Con *con) {
+    if (!con->window)
+        return;
+
+    con_state *state = state_for_frame(con->frame.id);
+    bool con_maximized = con_is_maximized(con, HORIZ);
```
stapelberg: nit: `const bool`

src/con.c
```diff
@@ -727,6 +776,20 @@ int con_num_children(Con *con) {
...
+ *
+ */
+bool con_has_num_children(Con *con, int num) {
+    Con *child;
+    TAILQ_FOREACH(child, &(con->nodes_head), nodes) {
+        num--;
+        if (num < 0)
+            return false;
+    }
+    return num == 0;
```
stapelberg: For consistency, please use `return (num == 0);`

src/x.c
yshui: Why? It's re-assigned later. Are you suggesting to use two variables?

stapelberg: Yeah, I think I’d prefer two separate const variables.

include/con.h
yshui: It is. Chrome won't use the maximized layout even when fullscreened (via i3, not via using F11 in Chrome). And I think this fits the spec: fullscreened window is indeed using the maximal amount of vertical/horizontal space.

Airblader: I also think we should include this for now. We could also partially revert this later on as well if need be.

stapelberg: > It is. Chrome won't use the maximized layout even when fullscreened (via i3, not via using F11 in Chrome). And I think this fits the spec: fullscreened window is indeed using the maximal amount of vertical/horizontal space.

That seems like a change in chrome. I remember that we’ve had feature requests from people who reported that chrome behaved precisely the same way regardless of whether one used i3’s fullscreen feature or whether one pressed F11.

Regardless, how do other window managers implement this? Do they consider fullscreen to be maximized?

yshui: > chrome behaved precisely the same way regardless of whether one used i3’s fullscreen feature or whether one pressed F11.

Is this the expected behavior? Should we consider this a bug in i3?

> Regardless, how do other window managers implement this? Do they consider fullscreen to be maximized?

I think most wm don't have a fullscreen state like i3 has. Fullscreen mode in other wm is generally initiated by the application, not the wm

Airblader: > Fullscreen mode in other wm is generally initiated by the application, not the wm

Same difference, it all goes through the window manager. The question is whether fullscreening an application in other systems causes the window manager to treat it as maximized or not. I think it makes sense to mirror that behavior and would assume that they do not count fullscreen as maximized.

yshui: @Airblader at least KWin does this

Airblader:
@yshui Please drop a comment when you incorporated comments since we won't be notified otherwise. The tests still need fixed and a rebase is in order. :-)

yshui:
@Airblader Oh, sorry. I didn't know GitHub doesn't have notification for new commits..

BTW, should I write new tests? Or do I just need to fix the old ones?

Airblader:
@yshui I think we should definitely have tests here that verify that containers have the correct combination of these atoms set depending on the layout.

Airblader:
@yshui Are you going to continue work on this?

yshui:
Sorry, but I'm not using i3wm anymore.

Airblader:
Alright, thanks for answering. :)

sethpollen:
I know it has been several years, but I think this work may now have increased importance. See https://crbug.com/1495853. Not only does the lack of MAXIMIZED hints cause Chrome to draw extra space at the top of the screen, but it also means that clicking in the top few pixels of the screen doesn't switch tabs.

sethpollen:
@Airblader are you generally supportive of this change? If you think it's the right direction, I can try to pick up the code and finish it.

orestisfl:
@sethpollen Sure, feel free to pick it up.

Looking at the code, we might be able to simplify by re-using `con_descend_direction` but someone needs to check if that's possible.

include/con.h
sethpollen: I'm hoping to pick up this pull request and push it to completion, so I'd like to close out these open threads.

As far as I can tell, muffin (from Cinnamon) does not set [MAXIMIZED](https://github.com/linuxmint/muffin/blob/bf234250143d2b4c7bad9f1ef3bccafbb6e6c21b/src/core/window-private.h#L234) when a window [is fullscreened](https://github.com/linuxmint/muffin/blob/bf234250143d2b4c7bad9f1ef3bccafbb6e6c21b/src/core/window.c#L3875). The muffin codebase seems to treat these as independent properties.

I had a bit of a hard time navigating the KWin code, but I also don't see evidence of maximize-on-fullscreen [here](https://github.com/KDE/kwin/blob/032891be60714ff4798fe25441b812d8084644ac/src/xdgshellwindow.cpp#L737). Maybe I am missing something.

So maybe we should change this to consider fullscreen windows as not maximized.

sethpollen:
I copied this pull request, touched up some commenting, tried to resolve open comments, and added test coverage. The result is https://github.com/i3/i3/pull/5840. Can a reviewer take a look, when you have time? I apologize in advance if I made any n00b mistakes; this is my first time creating a pull request on Github (and my first time using Perl).

PR #2386, opened 2016-06-22
Don't change border style if BS_NORMAL is requested in motif hints

### LINKED ISSUE #2385: new_window behaviour change

```
Binary i3 version:  4.12-75-g4501218 (2016-06-22, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12-75-g4501218 (2016-06-22, branch "next") (pid 2364)
Loaded i3 config: /home/username/.config/i3/config (Last modified: 13 seconds ago)

The i3 binary you are running: /usr/bin/i3
```

**What I did:**
Had `new_window pixel 1` (or 2 or whatever) in config and updated i3 (from b1d70f2) to this version ( aka this patch is at fault https://github.com/i3/i3/commit/45012189bcb67ba7652d2ea25a3e03cd74a796e7 ).

**What I saw:**
[Dolphin4](https://userbase.kde.org/Dolphin) now has a title bar (apparently this is the only application that now has one).
xprop log for Dolphin4: https://gist.github.com/licaon-kter/6ab3010d6e5763f72bbb954ecc51dd58

**What I expected instead:**
All newly opened windows have a 1 (or 2 or whatever) pixel border and no titlebar.

yshui:
Fixes #2385

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -1193,7 +1193,7 @@ static bool handle_motif_hints_change(void *data, xcb_connection_t *conn, uint8_
     border_style_t motif_border_style;
     window_update_motif_hints(con->window, prop, &motif_border_style);
 
-    if (motif_border_style != con->border_style) {
+    if (motif_border_style != con->border_style && motif_border_style != BS_NORMAL) {
         DLOG("Update border style of con %p to %d\n", con, motif_border_style);
         con_set_border_style(con, motif_border_style, con->current_border_width);
 


acrisci:
I'm ok with this. We basically have no documentation on how these hints should be interpreted, so we are free to do what makes sense. We can interpret "normal" to mean the user default. I can't think of any situation where that would cause a problem.

yshui:
@acrisci And this seems to be the intention of the original motif hints patch. It's my mistake to not check for BS_NORMAL.

shachaf:
With this change, when a window sets itself borderless using motif hints, it can't set itself back to bordered. Is that intended? It's not compatible with Metacity, which the comment says is the goal, and it doesn't seem right to me.

If I understand the rationale of the change correctly, maybe it would be better to interpret BS_NORMAL as "what the border style would have been with no hints", rather than "what the border currently is"?

Airblader:
It's tricky to decide what to do here; on the one hand, some users probably prefer this to be enitrely managed by i3, but others might want Motif hints to be respected.

In any case, let's not discuss it in a merged PR. Can you please open a new issue referencing this issue/PR and outlining the case? Thanks!

Issue #2383, opened 2016-06-21
workspace next / prev commands behave unexpectedly from first / last workspace


jplatte:
Output of `i3 --moreversion 2>&- || i3 --version`:

```
Binary i3 version:  4.12 (2016-03-06, branch "4.12") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12 (2016-03-06, branch "4.12") (pid 694)
Loaded i3 config: /home/jplatte/.i3/config (Last modified: Mo 20 Jun 2016 23:57:37 CEST, 13948 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

URL to a logfile as per http://i3wm.org/docs/debugging.html: Will create one if actually required.

**What I did:**

Execute (via dmenu or xbindkeys mouse button mapping) `i3-msg 'workspace next'` on the last workspace on the right, secondary monitor.

**What I saw:**

i3 changes to the first workspace on that monitor.

**What I expected instead:**

i3 should change to the first workspace, no matter the output.

---

I can also reproduce this with 'workspace prev' on the first workspace when that workspace is on my left, primary monitor. In that case i3 switches to the last workspace on that output, instead of the overall last workspace. Interestingly, I can't reproduce this if I switch the outputs the workspaces are on.

I can however reproduce it when both the first and the last workspace are on the same output. In that case, it again depends on which output the workspaces are on. On the left monitor, 'workspace next' behaves unexpectedly (goes to the first workspace on the right monitor) and on the right monitor, 'workspace prev' behaves unexpectedly (goes to the last workspace on the left monitor).

I guess it could be that 'workspace next' on the last workspace always only considers workspaces on the right monitor / secondary output, while 'workspace prev' on the first workspace always only considers workspaces on the left monitor / primary output?

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

Airblader:
I can't actually reproduce this, it works as expected for me. Please provide the log file as described above as well as your i3 config and the output of `xrandr`. Ideally, try reproducing this with the default config.

jplatte:
Alright, here's the log: http://logs.i3wm.org/logs/5656540782526464.bz2

I created it by launching i3 with only the following two commands in my `.xinitrc`:

```
xrandr --output HDMI-1 --primary
[...]

PR #2392, opened 2016-07-08
break test 528 by changing workspaces

bebehei:
FYI for bug #2383


bebehei:
Switching the workspaces with `workspace next` or `workspace prev` should not care about the actual display, where the workspace is located.

stapelberg:
Can you integrate this as a commit into pull request #2393 please? In general, we don’t want to merge pull requests which make the tests fail, and having both commits in the same PR helps us and future readers to understand the change you’re making.

bebehei:
Sry. I forgot to close this stub by myself. Thanks for doing it.

PR #2393, opened 2016-07-09
fix #2383 for numbered workspaces

### LINKED ISSUE #2383: workspace next / prev commands behave unexpectedly from first / last workspace

bebehei:


bebehei:
this fixes #2383 for numbered workspaces. It still switches for named workspaces in a different way as the documentation suggests.

For any extra details, consider reading #2383.

Travis Job 1 failed because of docker. Could you please restart it?

stapelberg:
I’ll wait for @Airblader’s assessment on whether this is a correct fix for issue #2383, but aside from that, could you please modify the commit description to describe what you’re changing instead of just referencing the issue? We’d like to have the git history self-contained.

Thanks!

bebehei:
- [x] tests added and enhanced (found an uncovered bug)
    - The test configuration may not see straightforward, but it covers most corner cases now.
- [x] rebased
- [ ] @Airblader's OK.

Airblader:
@stapelberg I'm having trouble verifying this because from my point of view we haven't really decided on the correct behavior in the ticket, have we? I'd actually sort of refer to your judgement of how these commands should operate, I don't have a strong opinion about it… :-)

bebehei:
> @stapelberg I'm having trouble verifying this because from my point of view we haven't really decided on the correct behavior in the ticket, have we? I'd actually sort of refer to your judgement of how these commands should operate, I don't have a strong opinion about it… :-)

Quoting @stapelberg #2383:
> It’s intentional that numbered workspaces are sorted, whereas named workspaces are kept in the order in which they are created, both in i3bar and when switching workspaces using workspace next.

Isn't this a clear enough definition? Or do you need to understand my changes? Shall I elaborate them?

Airblader:
Ah, I must've indeed missed that somehow. It looks good to me, mostly going off the tests here. :-)

Airblader:
@stapelberg Should we go ahead with this?

stapelberg:
Sure.

@bebehei Could you rebase please?

bebehei:
@stapelberg rebased.

PR #2431, opened 2016-08-22
Move cleanup code into atexit funtion (`src/main.c:i3_exit`).

nimaje:
* Cleanup should happen on all exits, not only on `i3-msg exit`.
  * e.g. I normally call `poweroff` in a running i3 -> i3 is killed by a signal -> ipc_socket is not unlinked.
* As this only moves some code from a function which calls `exit(0)` into
a function called from `atexit` this should work (NOT TESTED).

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1565,12 +1565,6 @@ void cmd_layout_toggle(I3_CMD, const char *toggle_mode) {
  */
 void cmd_exit(I3_CMD) {
     LOG("Exiting due to user command.\n");
-#ifdef I3_ASAN_ENABLED
-    __lsan_do_leak_check();
-#endif
-    ipc_shutdown();
-    unlink(config.ipc_socket_path);
-    xcb_disconnect(conn);
     exit(0);
 
     /* unreached */
--- a/src/main.c
+++ b/src/main.c
@@ -24,6 +24,10 @@
 
 #include "sd-daemon.h"
 
+#ifdef I3_ASAN_ENABLED
+#include <sanitizer/lsan_interface.h>
+#endif
+
 /* The original value of RLIMIT_CORE when i3 was started. We need to restore
  * this before starting any other process, since we set RLIMIT_CORE to
  * RLIM_INFINITY for i3 debugging versions. */
--- a/src/main.c
+++ b/src/main.c
@@ -166,6 +166,13 @@ static void i3_exit(void) {
     ev_loop_destroy(main_loop);
 #endif
 
+#ifdef I3_ASAN_ENABLED
+    __lsan_do_leak_check();
+#endif
+    ipc_shutdown();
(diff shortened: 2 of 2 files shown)


Airblader:
Please don't write "NOT TESTED" in the commit message. If it's not tested, it won't be merged.

nimaje:
Sorry I saw that there is a travis testing and wanted to discuss this change (there is a test for existing of the tmpdir after i3 has exited https://travis-ci.org/i3/i3/jobs/154155169#L3450, ok it tests if the old tmpdir gets cleaned at restart, but tests for the new after it calls exit_gracefully($pid))

This pr only moves some code I didn't expected that tests fail.
Don't konw wath causes this https://travis-ci.org/i3/i3/jobs/154155169#L4483

I can test locally but I would have to install some things first.

Airblader:
I didn't so much mean the automatic tests. If you author a change I would expect you run the changed binary locally for a bit to see if it works. A smoke / sanity test, if you will. Since you hadn't built the project I guess you didn't runt the binary either. 

Travis fails because a test is failing (see line 475)

nimaje:
Ok will do that soon.
I wanted to first discuss the change, but I understand your point.

Airblader:
It's fine if you want to discuss it first, but you hadn't mentioned that originally so I viewed it as a request for merge.

nimaje:
ok locally tested.
Seems like everything works, except it only removes the files in /tmp with `i3-msg exit` or `i3-msg restart` not if i3 is killed via a signal.
As this change breaks a travis test this cleanup is new although it is moved from where `i3-msg exit` should be defined.
Locally I have corrected the test to exit i3 after checking the tmpdirs not before.

I made this pr because i3 doesn't cleanup in /tmp with my useage and the code in `src/command.c:cmd_exit` looked like the right cleanup code (and like code which should be called at any exit).

Airblader:
@nimaje Are you going to continue work on this?

nimaje:
Yes

Airblader:
@nimaje Not to bother you, but it's been another 3.5 weeks. If you'll continue on it that's fine, but I'm just checking in because otherwise we can close the PR.

nimaje:
I give up on this for now, maybe I look later on this again.
I don't know why it changes the test result (this implies `exit_gracefully($pid)` always hits 'kill with signal case' in failed test and these changes work, but on my machine it doesn't call `i3_exit` at exit from signal).
You could try if it works for you if you want.
To try:
1. start a binary with this changes (suggestion in xephyr)
2. use `i3 --get-socketpath`
3. kill i3 with SIGINT (e.g. `kill -2 i3pid` where i3pid is the last number in output of step 2)
4. check whether the socketfile (output of step 2) exists.
It shouldn't exist anymore

Issue #2418, opened 2016-08-06
Some keybindings does not work when num lock is enabled


KabbAmine:
Output of `i3 --moreversion 2>&- || i3 --version`:

```
Binary i3 version:  4.12-118-g155f034 (2016-08-03, branch "gaps-next") © 2009 Michael Stapelberg and contributors
(Getting version from running i3, press ctrl-c to aborRunning i3 version: 4.12-118-g155f034 (2016-08-03, branch "gaps-next") (pid 2882)
Loaded i3 config: /home/k-bag/.i3/config (Last modified: sam. 06 août 2016 15:20:19 MSK, 289 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

The logfile: _http://logs.i3wm.org/logs/5719570367119360.bz2_

**What I did:**

- Use i3 with the default `/etc/i3/config` file
- Enable numpad
- Launch `resize` mode with `$mod+r`
- Press `escape` to quit the mode

**What I saw:**

Nothing, the `escape` key is not taken account.

**What I expected instead:**

Quit the `resize` mode when pressing `escape`.

**Notes**

Note that a lot of combinaisons does not work when num lock is enabled (All keybindings where `shift` is involved, e.g `$mod+shift+r` to restart i3). 
This behavior started after d77d40173aa

stapelberg:
Thanks, will take a look in the next few days. Which keyboard layout are you using (`setxkbmap print`)?

KabbAmine:
`fr` most of the time, but the issue occurs when using `us`, `ru` or `ar`

rosshadden:
I don't know how you narrowed this down to `Numlock`, but I have the same issue.  I never would have figured out that was even remotely related.

I use `dvorak`, though this seems to be a problem when using `us` for me as well.

It might be helpful to be more specific, though.  These are the `setxkbmap` options I use:

```sh
setxkbmap \
	-layout us,us \
	-variant dvorak, \
	-option \
	-option grp_led:caps \
	-option grp:shifts_toggle \
	-option keypad:pointerkeys \
	-option compose:sclk \
	-option ctrl:nocaps
```

KabbAmine:
> I don't know how you narrowed this down to Numlock, but I have the same issue. I never would have figured out that was even remotely related.

No magic here :smile:  The issue started after the commit d77d4017 which is related to numlock.

pjozog:
[...]

PR #2443, opened 2016-09-09
Fix NumLock changes when using Shift

nmschulte:
I don't think this fixes all of the issues, but this does seem to fix the issue related to `bindsym`s w/ `Shift` not working when the numeric lock is enabled.  I noticed that some of the `bindsym`s used in "modes" do not work w/ numeric lock still.

```
bindsym $mod1+b exec notify-send "$mod1+b"
bindsym $mod1+Shift+b exec notify-send "$mod1+Shift+b"
```

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -394,7 +394,7 @@ static void add_keycode_if_matches(struct xkb_keymap *keymap, xkb_keycode_t key,
             /* Also bind the key with active NumLock+CapsLock */
             ADD_TRANSLATED_KEY(bind->event_state_mask | xcb_numlock_mask | XCB_MOD_MASK_LOCK);
         } else {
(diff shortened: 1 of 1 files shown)


nmschulte:
I don't know much about the XKB/X keyboard stack, but from reviewing the code and a referencing the docs for the used API, along w/ the changes causing issue, this is what I came up with.  It seems like this is just a "thinko," and some of the prior cases were missed w/ the new logic.

stapelberg:
I think this is obsoleted by commit 0469716fd6cbe22916b1f7a0cc8193f6025e8405. Please let me know in case I missed something, ideally on issue #2418 itself.

stapelberg:
Merged this manually (with another fix applied on top). Thanks!

PR #2445, opened 2016-09-11
Hint to use $TERMINAL with i3-sesible-terminal

simonnagl:
I had the issue because of wrong configuration the preferred terminal was no longer in the $PATH.
I did not know this option of i3-sensible-terminal. I fried of mine helped me out.

I hope this can help some people in the future.

### DIFF
--- a/i3.config
+++ b/i3.config
@@ -36,6 +36,9 @@ floating_modifier Mod1
 
 # start a terminal
 bindsym Mod1+Return exec i3-sensible-terminal
+# If you have more than one terminal emulator enabled you can set $TERMINAL
+# to set your preferred terminal as default.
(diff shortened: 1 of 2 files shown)


Airblader:
I don't get it. Why use `i3-sensible-terminal` at all if you're gonna force which terminal to use anyway?

stapelberg:
As @Airblader mentioned, it’d be more efficient to use e.g. `bindsym $mod+Return exec exec /usr/bin/urxvtc` directly (note the double `exec` to avoid extraneous shell processes with e.g. `dash`).

`i3-sensible-terminal` only uses the `TERMINAL` environment variable in case you already have that in your environment for some reason (perhaps you don’t like to touch program-specific configuration files).

Airblader:
I'll close this for the aforementioned reasons.

PR #2450, opened 2016-09-12
Fix colormap handling for containers.

### LINKED ISSUE #2435: Colormap not installed for container windows / can not take screenshot with xwd
Version: 4.12 (2016-03-06, branch "4.12")

It is not possible to take screenshot with xwd. For other window managers xwd works (dwm, blackbox, just to name few). It is easy to reproduce by running the command:

**xwd -silent -root -out /tmp/screen_cap.xwd**

The received output is error:
X Error of failed request:  BadColor (invalid Colormap parameter)
  Major opcode of failed request:  91 (X_QueryColors)
  Resource id in failed request:  0x0
  Serial number of failed request:  71
  Current serial number in output stream:  71

Digging deeper into it, i found that for most of i3 containers, colormaps are not set, got by executing:
**> xwininfo -tree -root | grep i3**
xwininfo: Window id: 0x40 (the root window) "i3"
  Root window id: 0x40 (the root window) "i3"
     **0x600039** "[i3 con] container around 0x809c4f4c0": ("i3-frame" "i3-frame")  1200x15+0+685  +0+685
        **0xa00003** "i3bar for output xroot-0": ("i3bar" "i3bar")  1200x15+0+0  +0+685
   ...

**> xwininfo -stats -id 0x600039 | grep -i colormap**
Colormap: 0x0 (not installed)

The only window which has color-map set is "i3bar":
**> xwininfo -stats -id 0xa00003 | grep -i colormap**
  Colormap: 0x20 (installed)

Which basically is the same as for root window.

Searching into code reveals that in file i3bar/src/xcb.c:1360 the root_screen->default_colormap is used (colormap pointer is set in other function).

In file src/x.c:129 it can be noticed that different colormap is used, kind of defined in line 107, but then for some reasons and cases released on line 143, the same file.

I guess people already have had trouble with it, saw discussions:
http://www.freelists.org/post/i3-discuss/i3-full-screencapture-with-xwd-not-working-only-in-i3
and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=832954

The thing is that i don't understand much of how to correctly use colormaps with xcb, thus i can not fix this by my self and information on internet is not very verbose on it either. I believe that with the right knowledge this is not a hard bug fixed.

Airblader:
This commit correctly handles colormaps by

* Using the static default colormap we determine on startup if the
  con has the corresponding depth. This avoids creating pointless
  colormaps.
* Not freeing the default colormap to not have stray colormaps on
  containers. This fixes an issue with certain programs such as xwd.
* Creating a custom colormap when necessary and freeing it when the
  container is killed.

fixes #2435

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -105,11 +105,18 @@ void x_con_init(Con *con) {
     uint32_t mask = 0;
     uint32_t values[5];
 
-    /* For custom visuals, we need to create a colormap before creating
-     * this window. It will be freed directly after creating the window. */
     xcb_visualid_t visual = get_visualid_by_depth(con->depth);
-    xcb_colormap_t win_colormap = xcb_generate_id(conn);
-    xcb_create_colormap_checked(conn, XCB_COLORMAP_ALLOC_NONE, win_colormap, root, visual);
+    xcb_colormap_t win_colormap;
+    if (con->depth != root_depth) {
+        /* We need to create a custom colormap. */
+        win_colormap = xcb_generate_id(conn);
+        xcb_create_colormap(conn, XCB_COLORMAP_ALLOC_NONE, win_colormap, root, visual);
+        con->colormap = win_colormap;
+    } else {
+        /* Use the default colormap. */
+        win_colormap = colormap;
+        con->colormap = XCB_NONE;
+    }
 
     /* We explicitly set a background color and border color (even though we
      * don’t even have a border) because the X11 server requires us to when
@@ -143,9 +150,6 @@ void x_con_init(Con *con) {
                         8,
                         (strlen("i3-frame") + 1) * 2,
                         "i3-frame\0i3-frame\0");
-
-    if (win_colormap != XCB_NONE)
-        xcb_free_colormap(conn, win_colormap);
 
     struct con_state *state = scalloc(1, sizeof(struct con_state));
     state->id = con->frame.id;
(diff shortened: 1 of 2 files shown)


Airblader:
I think we also need to dump this property so we can restore it after an in-place restart.

Airblader:
Actually I can't quite explain why, but not doing so seems to be correct. At least when I attempt to dump & parse it, I get pretty bad glitching. Perhaps colormaps are freed by the server if the client disconnects? Either way, everything seems to be OK.

src/x.c
```diff
@@ -105,11 +105,18 @@ void x_con_init(Con *con) {
...
-    /* For custom visuals, we need to create a colormap before creating
-     * this window. It will be freed directly after creating the window. */
     xcb_visualid_t visual = get_visualid_by_depth(con->depth);
-    xcb_colormap_t win_colormap = xcb_generate_id(conn);
-    xcb_create_colormap_checked(conn, XCB_COLORMAP_ALLOC_NONE, win_colormap, root, visual);
+    xcb_colormap_t win_colormap;
+    if (con->depth != root_depth) {
+        /* We need to create a custom colormap. */
+        win_colormap = xcb_generate_id(conn);
+        xcb_create_colormap_checked(conn, XCB_COLORMAP_ALLOC_NONE, win_colormap, root, visual);
```
psychon: @Airblader made me look at this diff and now I am again complaining about an unrelated issue:

This line has a memory leak. When using a `_checked` request you should eventually pass the resulting `xcb_void_cookie_t*` to `xcb_request_check` to know about the result of the operation. If you don't care about it, then don't use the `_checked` variant. (Or call `xcb_discard_reply` on the cookie, but that would be quite pointless)

(The memory leak happens inside libxcb

Airblader: Yeah, it definitely is unnecessary. I'll change it, thanks. :-) And as always thanks for reviewing (and answering my questions.)

Airblader:
@stapelberg OK, should be good to go now.

PR #2453, opened 2016-09-14
travis/cleanup-bintray: use dpkg --compare-versions

stapelberg:
…instead of lexicographically sorting strings, which fails for the
following situation:

    4.12-96-g086276b
    4.12-97-g59c070b
    4.12-108-gb850cfb

This bug resulted in new packages being built and uploaded, then
immediately deleted.

Thanks to eeemsi for reporting the issue.


stapelberg:
@Airblader Could you take a look at this PR please? :)

Airblader:
Just for reference, assigning the review to somebody also triggers a notification, so an explicit comment isn't necessarily needed (but of course adds context) :-)

PR #2457, opened 2016-09-17
Assign the sticky value for _NET_WM_DESKTOP on scratchpad windows.

### LINKED ISSUE #2456: _NET_WM_DESKTOP on scratchpad windows
We currently don't modify `_NET_WM_DESKTOP` on a window when it's moved to the scratchpad. We should investigate the specs on how to handle this case. Perhaps removing it would be allowed, otherwise we need to see whether there is a better way to handle this case given that the scratchpad isn't a normal workspace.

Airblader:
fixes #2456

### DIFF
--- a/src/ewmh.c
+++ b/src/ewmh.c
@@ -124,11 +124,13 @@ void ewmh_update_desktop_viewport(void) {
 }
 
 static void ewmh_update_wm_desktop_recursively(Con *con, const uint32_t desktop) {
-    /* Recursively call this to descend through the entire subtree. */
     Con *child;
+
+    /* Recursively call this to descend through the entire subtree. */
     TAILQ_FOREACH(child, &(con->nodes_head), nodes) {
         ewmh_update_wm_desktop_recursively(child, desktop);
     }
+
     /* If con is a workspace, we also need to go through the floating windows on it. */
     if (con->type == CT_WORKSPACE) {
         TAILQ_FOREACH(child, &(con->floating_head), floating_windows) {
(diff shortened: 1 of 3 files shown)


Airblader:
I can't make sense of the failing tests. They pass reproducibly on my machine and executing what seems to be failing (moving a workspace to scratchpad) manually on my machine also works fine. I'll have to dig deeper on why it's failing.

Airblader:
OK, so this is starting to become super weird. Through some test commits I have identified that if the line

````
xcb_change_property(conn, XCB_PROP_MODE_REPLACE, window, A__NET_WM_DESKTOP, XCB_ATOM_CARDINAL, 32, 1, (uint32_t[]){wm_desktop});
```

is executed in the scratchpad case, the test fails. And since it times out I assume i3 crashes. Given that everything in that line is used before, I am currently at a complete loss of ideas as to what is failing.

Airblader:
I've put some extra `does_i3_live` in the failing test case. Oddly, this made the tests pass for gcc, but fail at the same point in clang. It also shows that the test fails before it even really started – the only i3 command executed at that point for this test case is `fresh_workspace`, not even the windows have been opened yet.

Airblader:
It seems like opening the fresh workspace is what causes the crash. Since I cannot reproduce this either through running tests locally nor doing it myself on my system I'm still having a hard time finding the root cause of the crash.

Airblader:
OK, I found an issue now which I believe to be the root cause of the crash, though I don't quite know where the crash happens yet (probably some assertion). When we restart, we see the `_NET_WM_DESKTOP` value and assume the window is sticky. Once that window is taken from the scratchpad it is now sticky – boom.

Airblader:
@stapelberg Finally done. Nothing like some good old foo debugging on a remote machine…

stapelberg:
Can you rebase please?

Airblader:
@stapelberg Rebased.

stapelberg:
Sorry, the other PR I merged earlier actually conflicts with this one. Could you rebase again please?

Airblader:
@stapelberg Fixed

PR #2458, opened 2016-09-17
Handle _NET_ACTIVE_WINDOW for scratchpad windows.

### LINKED ISSUE #2455: _NET_ACTIVE_WINDOW client messages for scratchpad windows should show the window
i3 currently rejects a `_NET_ACTIVE_WINDOW` client message if the window to be activated is on an internal workspace. I think we should change this such that it behaves like `scratchpad show`. This accomodates for the request in a consistent way.

This came up in an issue over at rofi, see [here](https://github.com/DaveDavenport/rofi/issues/464) after they removed i3-specific workarounds.

Airblader:
We call scratchpad_show() on _NET_ACTIVE_WINDOW requests if the request
came from a pager. This is consistent with the i3 »focus« command because
we assume the user requested to see the window, so we do the only
sensible thing.

fixes #2455

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -716,14 +716,13 @@ static void handle_client_message(xcb_client_message_event_t *event) {
         }
 
         Con *ws = con_get_workspace(con);
-
         if (ws == NULL) {
             DLOG("Window is not being managed, ignoring _NET_ACTIVE_WINDOW\n");
             return;
         }
 
-        if (con_is_internal(ws)) {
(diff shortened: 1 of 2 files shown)


stapelberg:
Could you add a testcase for this please?

Airblader:
There is a test for the old behavior which still seems to pass. I'll look into that.

Airblader:
@stapelberg I've added a test. The old test still passed because I forgot the change only applies to pagers. :-)

stapelberg (approved):

PR #2471, opened 2016-09-23
fix i3bar crashing when I3SOCK present

### LINKED ISSUE #2468: i3bar crashes if I3SOCK is set
This is due to socket_path being free'd in i3bar/src/main.c:144, which can not happen because socket_path is a pointer to an environment variable.
I'm thinking about a solution for this.

koebi:
When I3SOCK is present, socket_path might be a pointer to an
environment variable, which cannot be free'd in line 157. This
commit duplicates the string if I3SOCK is present, thus making
socket_path a free-able pointer again.
fixes #2468


stapelberg:
Can you rebase to the next branch please? I can’t merge this otherwise :).

koebi:
gh says this is merge-able, and I think it is based on next. I am not sure though, I had some trouble with git earlier on…

PR #2472, opened 2016-09-24
Unconditionally depend on libpcre >= 8.10

stapelberg:
Even Debian oldstable has 8.30, so let’s get rid of conditional
compilation where we don’t actually need it.


stapelberg:
The travis build is failing because of the Perl transition in Debian, so I’ll hold off a few days before going forward with this PR.

PR #2476, opened 2016-09-25
Add the option to toggle between stacked/tabbed layout

BenoitAverty:
This PR adds a parameter called "stack_tab" to the "layout toggle" command. This makes it possible to toggle between stacked and tabbed layouts, muck like the "layout toggle split" works for splitv/splith layouts.

I don't like the name of the parameter very much, but I didn't have a better idea, so suggestions are welcome on that point :)

A possible improvement would be to remember the last layout used between stacked and tabbed, so that running the layout toggle stack_tab command when the layout is neither of them sets the layout to the last one used (again, much like the layout toggle split works).

I added test cases and documentation, but if I forgot anything feel free to tell me and I'll add what's missing.

Thanks for the work on i3, I'm falling in love with it :)

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1811,8 +1811,8 @@ bindsym $mod+t split toggle
 
 === Manipulating layout
 
-Use +layout toggle split+, +layout stacking+, +layout tabbed+, +layout splitv+
-or +layout splith+ to change the current container layout to splith/splitv,
+Use +layout toggle split+, +layout toggle stack_tab+, +layout stacking+, +layout tabbed+, +layout splitv+
+or +layout splith+ to change the current container layout to splith/splitv, stacking/tabbed,
 stacking, tabbed layout, splitv or splith, respectively.
 
 To make the current window (!) fullscreen, use +fullscreen enable+ (or
@@ -1826,13 +1826,14 @@ enable+ respectively +floating disable+ (or +floating toggle+):
 *Syntax*:
 --------------------------------------------
 layout default|tabbed|stacking|splitv|splith
-layout toggle [split|all]
+layout toggle [split|stack_tab|all]
 --------------------------------------------
 
 *Examples*:
 --------------
 bindsym $mod+s layout stacking
 bindsym $mod+l layout toggle split
+bindsym $mod+l layout toggle stack_tab
 bindsym $mod+w layout tabbed
 
 # Toggle between stacking/tabbed/split:
--- a/src/con.c
+++ b/src/con.c
@@ -1699,8 +1699,8 @@ void con_set_layout(Con *con, layout_t layout) {
 /*
  * This function toggles the layout of a given container. toggle_mode can be
  * either 'default' (toggle only between stacked/tabbed/last_split_layout),
- * 'split' (toggle only between splitv/splith) or 'all' (toggle between all
- * layouts).
+ * 'split' (toggle only between splitv/splith), 'stack_tab' (toggle only between
+ * stacking/tabbed) or 'all' (toggle between all layouts).
  *
  */
 void con_toggle_layout(Con *con, const char *toggle_mode) {
@@ -1724,6 +1724,17 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
             else
                 con_set_layout(con, L_SPLITH);
         }
+    } else if (strcmp(toggle_mode, "stack_tab") == 0) {
+        /* Toggle between stacked / tabbed layouts. When the current layout is
+         * neither, set the layout to staked first. */
+         if (parent->layout != L_STACKED && parent->layout != L_TABBED)
+            con_set_layout(con, L_TABBED);
+         else {
+            if (parent->layout == L_STACKED)
+                con_set_layout(con, L_TABBED);
+            else
+                con_set_layout(con, L_STACKED);
+         }
     } else {
         if (parent->layout == L_STACKED)
             con_set_layout(con, L_TABBED);
--- a/src/con.c
+++ b/src/con.c
@@ -1727,14 +1727,8 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
     } else if (strcmp(toggle_mode, "stack_tab") == 0) {
         /* Toggle between stacked / tabbed layouts. When the current layout is
          * neither, set the layout to staked first. */
-         if (parent->layout != L_STACKED && parent->layout != L_TABBED)
-            con_set_layout(con, L_TABBED);
-         else {
-            if (parent->layout == L_STACKED)
-                con_set_layout(con, L_TABBED);
-            else
-                con_set_layout(con, L_STACKED);
-         }
+        layout_t new_layout = (parent->layout == L_TABBED) ? L_STACKED : L_TABBED;
+        con_set_layout(con, new_layout);
     } else {
         if (parent->layout == L_STACKED)
             con_set_layout(con, L_TABBED);
--- a/include/commands.h
+++ b/include/commands.h
@@ -223,7 +223,7 @@ void cmd_move_direction(I3_CMD, const char *direction, long move_px);
 void cmd_layout(I3_CMD, const char *layout_str);
 
 /**
- * Implementation of 'layout toggle [all|split]'.
+ * Implementation of 'layout toggle [all|split|stack_tab]'.
  *
  */
 void cmd_layout_toggle(I3_CMD, const char *toggle_mode);
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -104,11 +104,11 @@ state LAYOUT:
   'toggle'
       -> LAYOUT_TOGGLE
 
-# layout toggle [split|all]
+# layout toggle [split|stack_tab|all]
 state LAYOUT_TOGGLE:
   end
       -> call cmd_layout_toggle($toggle_mode)
-  toggle_mode = 'split', 'all'
+  toggle_mode = 'split', 'stack_tab', 'all'
       -> call cmd_layout_toggle($toggle_mode)
 
 # append_layout <path>
@@ -128,7 +128,7 @@ state WORKSPACE:
       -> call cmd_workspace_back_and_forth()
   'number'
       -> WORKSPACE_NUMBER
-  workspace = string 
+  workspace = string
       -> call cmd_workspace_name($workspace, $no_auto_back_and_forth)
 
 state WORKSPACE_NUMBER:
--- a/src/commands.c
+++ b/src/commands.c
@@ -1533,7 +1533,7 @@ void cmd_layout(I3_CMD, const char *layout_str) {
 }
 
 /*
- * Implementation of 'layout toggle [all|split]'.
+ * Implementation of 'layout toggle [all|split|stack_tab]'.
  *
  */
 void cmd_layout_toggle(I3_CMD, const char *toggle_mode) {
--- a/testcases/t/192-layout.t
+++ b/testcases/t/192-layout.t
@@ -95,4 +95,16 @@ cmd 'layout toggle all';
 ($nodes, $focus) = get_ws_content($tmp);
 is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
 
+cmd 'layout toggle stack_tab';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'tabbed', 'layout now tabbed');
+
+cmd 'layout toggle stack_tab';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'stacked', 'layout now stacked');
+
+cmd 'layout toggle stack_tab';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'tabbed', 'layout now tabbed');
+
 done_testing;


src/con.c
```diff
@@ -1724,6 +1724,17 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
             else
                 con_set_layout(con, L_SPLITH);
         }
+    } else if (strcmp(toggle_mode, "stack_tab") == 0) {
+        /* Toggle between stacked / tabbed layouts. When the current layout is
+         * neither, set the layout to staked first. */
+         if (parent->layout != L_STACKED && parent->layout != L_TABBED)
```
Airblader: Please use braces, this looks really odd without them.

Airblader: This entire if/else could be rewritten in a clearer way as

````
if (layout == tabbed) {
    …
} else if (layout == stacked) {
    …
} else {
    …
}
```

In fact, you could even shorten it to this:

````
layout_t new_layout = (parent->layout == L_TABBED) ? L_STACKED : L_TABBED;
con_set_layout(new_layout);
```

src/con.c
```diff
@@ -1724,6 +1724,17 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+            con_set_layout(con, L_TABBED);
+         else {
+            if (parent->layout == L_STACKED)
```
Airblader: Please use braces, we are trying to get this consistent.

docs/userguide
```diff
@@ -1811,8 +1811,8 @@ bindsym $mod+t split toggle
 
 === Manipulating layout
 
-Use +layout toggle split+, +layout stacking+, +layout tabbed+, +layout splitv+
-or +layout splith+ to change the current container layout to splith/splitv,
+Use +layout toggle split+, +layout toggle stack_tab+, +layout stacking+, +layout tabbed+, +layout splitv+
```
Airblader: I agree that this name is not good. I don't think we currently have a good terminology to describe this, we only have »split« meaning the opposite.

Airblader:
Travis is failing due to wrong formatting.

BenoitAverty:
Hi Airblader, thank you for the review.

I apologize for the lack of feature request and the formatting, I was in a hurry and forgot to check the contributing guide :-/

I fixed the formatting.

Do you want me to open an issue corresponding to the PR or do we continue the discussion here ?

Airblader:
No, we can deal with it here for now. We mostly recommend opening an issue to avoid doing work that ends up not merged because we reject the feature. You've written the code already now, so it's too late for that (but of course we might still end up not merging it.)

BenoitAverty:
Hi,

I may have some time in the coming days to work on this PR. I'd like to implement the memorization of the last layout between stacked/tabbed, so that the last one used is set when using the toggle command from a split layout.

Do you think this PR has a good change of being merged? Otherwise I'll probably implement this a bit later.

For the name of the parameter, I had several ideas, none of which are completely satisfactory.
 - Just `stacking` or `stack`, because the windows are stacking instead of tiling. Problem : confusion with existing term in i3.
 - `overlap`, for a new term but not completely appropriate because it suggests a partial overlap (maybe that's just me)
 - `stack_mode` or `tab_mode`

another completely different option would be to allow complete choice on the layouts that are included when toggling. For example, we'd allow this in the config file:
```
#useless because almost identical to layout toggle split
bindsym $mod+w layout toggle splith,splitv

#This would make the function I want
bindsym $mod+w layout toggle stacking,tabbed

# More customization would be possible
bindsym $mod+w layout toggle stacking,splitv,splith
```

That seems a bit more complicated though, and not necessarily very useful if we find a good name for the current PR.

What do you think ?

s3rb31:
This is good. Let's expand on this ;)

I'd really like the idea of the `layout toggle` option being capable of cycling through all of the layouts based on user definition. We already have the keywords for `layout`. Why not make `layout toggle` accept a (comma?) seperated list. For example: `layout toggle split,tabbed` or `layout stacking,tabbed,splith`. 

I think this would allow a much more flexible way of cycling through the layouts, it wouldn't even introduce new terminology and it would adress the original PR "on-the-fly". 

I'd like to hear what you think. Greetings

BenoitAverty:
Yes that was my last suggestion in my previous message. I like the Idea, but the code for the PR becomes more complex, and I'm not sure the added value justifies it.

s3rb31:
Let me look at the code in detail today and see if I can do something.

Airblader:
What's the status here? Anybody still working on it?

s3rb31:
Yes, sorry for the late response. This slipped out of my field of vision over the new year, thanks for the heads up. Well, I have some observations:

First, I don't think it is possible to use comma (`,`) as a seperator for this feature without some (read: heavy) modification to the config parser. This is because the comma seperates commands in the i3 config file. If it is possible I'd be glad about some advice how to do it. If it is not possible, my question would be what other single characters are acceptable as a delimiter? For testing I used slash (`/`) ..

Further, I am not sure on how to define the the parser specs for the new possible arguments to `layout toggle`. One possibility would be something along these lines:  
```
# layout default|stacked|stacking|tabbed|splitv|splith
state LAYOUT:
  layout_mode = 'default', 'stacked', 'stacking', 'tabbed', 'splitv', 'splith'
      -> call cmd_layout($layout_mode)
  'toggle'
      -> LAYOUT_TOGGLE

# layout toggle [split|all]
state LAYOUT_TOGGLE:
  end
      -> call cmd_layout_toggle($toggle_mode)
  toggle_mode = 'split', 'all'
      -> call cmd_layout_toggle($toggle_mode)
  # new entry
  toggle_custom = string
      -> call cmd_layout_toggle($toggle_custom)
```
This would simply pass everything as one argument, but then it would not be possible to detect any wrong arguments for the `layout toggle` command (and the entry above our new one would be redundant).

Another option would be to pass the modes via N different arguments wich could be checked one by one, then. But I find this looks quite hackish as you have the same possible options for every argument.

So it would be nice if someone who knows the parser quite good could give some thoughts about how this could be implemented the most elegant way. 

Another solution I just came up with right now would be to use the solution from above and introduce another level in the parser spec, so that the `split` and all `all` options for `layout toggle` remain checkable. Something like this:
```
# layout default|stacked|stacking|tabbed|splitv|splith
state LAYOUT:
  layout_mode = 'default', 'stacked', 'stacking', 'tabbed', 'splitv', 'splith'
      -> call cmd_layout($layout_mode)
  'toggle'
      -> LAYOUT_TOGGLE

# layout toggle [split|all]
state LAYOUT_TOGGLE:
  end
      -> call cmd_layout_toggle($toggle_mode)
  toggle_mode = 'split', 'all'
      -> call cmd_layout_toggle($toggle_mode)
  'custom'
      -> LAYOUT_TOGGLE_CUSTOM

# layout toggle custom <args>
state LAYOUT_TOGGLE_CUSTOM:
  toggle_mode = string
      -> call cmd_layout_toggle($toggle_mode)
```

If it is an option to pass one single argument, seperated by a specific delimiter, then I already have some code based on strtok. 

It will split the passed argument via a specified delimiter, build an array of `layout_t`s (int) from the passed argument, and set the layout to the one following the current active layout within the array, or, if the current layout is not inside the array, the first entry.

But the post is already long so I'll wait for people to join the discussion on the above observations and deduce on how to tackle this thing and if what I posted is the right direction.

Greetings

Airblader:
> First, I don't think it is possible to use comma (,) as a seperator for this feature without some (read: heavy) modification to the config parser.

You're right and IMHO we shouldn't change this behavior. I think a good alternative would be `+`. Alternatively, and I like this the most, I think, we separate them by space and just read in the entire rest of the line and split it in the handler function. Losing the ability to report the parser error directly is fine with me. So it'd look like this:

````
bindsym $mod+x layout toggle splith tabbed stacked
````

I think it should be possible to change the parser to deal in a better way with this (introduce passing arrays to the handler function) and to get back the parser error reporting, but I wouldn't develop that in this PR and instead do a followup PR for it.

> and the entry above our new one would be redundant

Just as a small note: no, they wouldn't be, for the reason you stated yourself – better parser errors. :-)

**Comment by s3rb31** [comment]
> > and the entry above our new one would be redundant

> Just as a small note: no, they wouldn't be, for the reason you stated yourself – better parser errors. :-)

I think that dont understand this. If we have this:

```
state LAYOUT_TOGGLE:
  end
      -> call cmd_layout_toggle($toggle_mode)
  toggle_mode = 'split', 'all'
      -> call cmd_layout_toggle($toggle_mode)
  toggle_custom = string
      -> call cmd_layout_toggle($toggle_custom)
```

How would the parser then differentiate between an arbitrary string (which would cause no error) and a disallowed word? `split` and `all` would also be captured by the string which would make the explicit definition superfluous. That was what I meant to say.

But if thats not a problem. I like your idea by simply staying with spaces but treating it as a single argument. This way we you keep a uniform

**Comment by Airblader** [comment]
@s3rb31 Did you delete a comment? Or is Github just slow with database replication? :-)

**Comment by s3rb31** [comment]
> > and the entry above our new one would be redundant

> Just as a small note: no, they wouldn't be, for the reason you stated yourself – better parser errors. :-)

I think that dont understand this. If we have this:

```
state LAYOUT_TOGGLE:
  end
      -> call cmd_layout_toggle($toggle_mode)
  toggle_mode = 'split', 'all'
      -> call cmd_layout_toggle($toggle_mode)
  toggle_custom = string
      -> call cmd_layout_toggle($toggle_custom)
```

How would the parser then differentiate between an arbitrary string (which would cause no error) and a disallowed word? `split` and `all` would also be captured by the string which would make the explicit definition superfluous. That was what I meant to say.

But if thats not a problem, everything is fine. I like your idea by simply staying with spaces but treating it as a single argument. This way we keep a uniform look in the config file and handle the special sauce within the implementation.

I'll implement what we've got so far, and post the code I have during this weekend!

_PS. Touching the parser is something I would leave to someone other since I don't code in Perl. But I absoluteley agree. It would be very nice if the parser would support arrays._

**Comment by Airblader** [comment]
> How would the parser then differentiate between an arbitrary string (which would cause no error) and a disallowed word? split and all would also be captured by the string which would make the explicit definition superfluous.

Right, I didn't think about it enough. :-) I think I'd still prefer the explicit (redundant) lines in the spec because it documents as much of the allowed values as possible that way.

> I'll implement what we've got so far, and post the code I have during this weekend!

Great, thanks!

> PS. Touching the parser is something I would leave to someone other since I don't code in Perl. But I absoluteley agree. It would be very nice if the parser would support arrays.

No problem whatsoever. Like I said, cleaning this up in a separate PR is better, anyway.

> Edit: Yes I deleted this post before because I accidentally hit enter. I hate it when Enter is Send ... xD

:-)

**Comment by s3rb31** [comment]
Hello everyone,

> I think I'd still prefer the explicit (redundant) lines in the spec because it documents as much of the allowed values as possible that way.

I did not know it either, but the parser seems to detect even partial words if they appear in the beginning. If it is not commented out, it will execute the partial detection and output an error:

```
15.01.2017 21:21:51 - IPC: received: *layout toggle splith splitv*
15.01.2017 21:21:51 - commands_parser.c:parse_command:267 - COMMAND: *layout toggle splith splitv*
15.01.2017 21:21:51 - commands.c:cmd_criteria_init:173 - Initializing criteria, current_match = 0x677a20
15.01.2017 21:21:51 - commands.c:cmd_layout_toggle:1545 - toggling layout (mode = split)
15.01.2017 21:21:51 - con.c:con_toggle_layout:1713 - con_toggle_layout(0x202db10, split), parent = 0x1ffde00
15.01.2017 21:21:51 - con.c:con_set_layout:1617 - con_set_layout(0x202db10, 5), con->type = 2
15.01.2017 21:21:51 - ERROR: Expected one of these tokens: <end>, '[', 'move', 'exec', 'exit', 'restart', 'reload', 'shmlog', 'debuglog', 'border', 'layout', 'append_layout', 'workspace', 'focus', 'kill', 'open', 'fullscreen', 'sticky', 'split', 'floating', 'mark', 'unmark', 'resize', 'rename', 'nop', 'scratchpad', 'title_format', 'mode', 'bar'
15.01.2017 21:21:51 - ERROR: Your command: layout toggle splith splitv
15.01.2017 21:21:51 - ERROR:                                  ^^^^^^^^
```

But having them commented out would still document the possible values and since there is no other purpose for it I think thats okay? (Maybe this behaviour is something you want to evaluate further?)

Currently I am working with [this](https://gist.github.com/s3rb31/c3d187844b27590785cf442a6f3561f5) code (applied on current next). I am testing it with the modified testcase. If there is anything I missed in the test I'd be glad about some hint.

Of course this solution will led into the else-block starting at `src/con.c:1767` for any arguments to `layout toggle` not containing the delimiter (apart from `split` of course) resulting in 'layout toggle default'. 

I think this could be improved by toggling between the given layout and the previous one for anything else than `all` and `default`. But that would require another field like `last_split_layout` to save the previous layout. I havent really thought about this much but it sounds nice to me. Would like to hear what others think about this.

PS. I am not sure on how to contribute to this PR? Should I create an additional PR on @BenoitAverty's fork?

Airblader:
> But having them commented out would still document the possible values and since there is no other purpose for it I think thats okay?

Ah, okay, I didn't think about partial matches either. No, in that case I'd prefer not to have commented-out code.

I'm currently thinking that we should get the array-argument stuff in either with or prior to this PR. I don't like the manual parsing much. But that'll be up to @stapelberg; perhaps he'd be OK with it.

> PS. I am not sure on how to contribute to this PR? Should I create an additional PR on @BenoitAverty's fork?

Yes, please create your own PR for this.

s3rb31:
> Yes, please create your own PR for this.

Just did it. Have a look at #2649 :v:

Airblader:
Closing in favor of the new PR

Issue #323, opened 2011-02-23
i3 crashes when moving away from dock clients


i3bot:
[Originally reported by mist <i3@…>]
(Open a client as dock and focus downwards, i3 will crash

Issue #321, opened 2011-02-21
Dock area is moved


i3bot:
[Originally reported by mseed]
(1. Focus dock area (i3bar, trayer, ...)
2. Run terminal

Dock is moved.

PR #2477, opened 2016-09-25
Check output crossing on ENTER_NOTIFY to dockarea.

### LINKED ISSUE #2081: Mouse cursor jumping back to previously used workspace
I am facing a weird bug right now. I am sorry for a vague discription, but I wasn't able to fully identify and reproduce the issue yet.

Scenario: i3 4.11 on Ubuntu 14.04. Two monitors, same resolution.
Behavior: Mouse coursor jumping back to previously used workspace.

Detailed description: When moving/clicking with the mouse on a "broken" workspace, the mouse jumps back to the previously used workspace. This jump might be to the workspace on the other monitor or change workspace on the current monitor (if the previously used workspace was on the same monitor. The "broken" workspace remains broken independent from the program running. Once I move all programs to another workspace (thereby removing the "broken" workspace), the problem is resolved.

```
i3 version 4.11 (2015-09-30, branch "4.11") © 2009 Michael Stapelberg and contributors
```

Airblader:
When receiving an ENTER_NOTIFY event on a dock client we returned as to not
focus the dock client (cf. #321 and #323). However, we still need to check
for crossing output boundaries and if that happened focus the new output.

Otherwise it can happen that the cursor is on a different output than the
focused output. When opening a window, this would open it on the old output
and then warp the mouse there. This effect will be even worse if the window
is immediately moved with 'move position mouse' as the window will end up
in its correct position on the new output and the cursor warped to the old
output.

relates to #2081

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -148,15 +148,12 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
         enter_child = true;
     }
 
-    /* If not, then the user moved their cursor to the root window. In that case, we adjust c_ws */
-    if (con == NULL) {
+    /* If we cannot find the container, the user moved their cursor to the root
+     * window. In this case and if they used it to a dock, we need to focus the
+     * workspace on the correct output. */
+    if (con == NULL || con->parent->type == CT_DOCKAREA) {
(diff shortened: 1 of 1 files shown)


stapelberg:
Can you rebase please?

Airblader:
@stapelberg Rebased.

PR #2488, opened 2016-10-02
Bugfix: NULL dereference when stickying _NET_WM_DESKTOP_ALL windows

Ferdi265:
This PR fixes a bug where `i3` will crash when starting some programs that set `_NET_WM_DESKTOP_ALL`.

In these cases, `ws` will be `NULL` here, triggering a null pointer dereference.

This bug was originally tested and found by me with `lemonbar` and was introduced in commit a15ce8cb8dddf118ed27deb79eb5d1b4103e8303: Assign the sticky value for _NET_WM_DESKTOP on scratchpad windows. (#2457)

Sorry if this PR is not perfectly up to standards, this is my first time contributing to i3.


Ferdi265:
Sure thing :)

PR #2489, opened 2016-10-03
Add "output" to IPC events referencing a container

### LINKED ISSUE #2478: Add output to information given in workspace event from ipc
It would be really useful to have the name of the output which a desktop is/was on in the workspace event information you get from ipc events.  It is available from the get_workspaces, so seems like it should be here too.

pstray:
Add output to all events in ipc that references a container.

Closes #2478

### DIFF
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -294,6 +294,11 @@ void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     ystr("focused");
     y(bool, (con == focused));
 
+    if (con->type != CT_ROOT && con->type != CT_OUTPUT) {
+        ystr("output");
+        ystr(con_get_output(con)->name);
+    }
+
     ystr("layout");
     switch (con->layout) {
         case L_DEFAULT:


src/ipc.c
```diff
@@ -294,6 +294,9 @@ void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     ystr("focused");
     y(bool, (con == focused));
 
+    ystr("output");
+    ystr(get_output_for_con(con)->name);
```
Airblader: How will this behave for, say, scratchpad containers?

pstray: I have no idea. What are scratchpad containers?  But if that means they don't have an output i guess it would be solved with a test on the return of get_output_for_con?

Airblader: Scratchpad is an i3 feature (see userguide). They do have an output, but it's an internal fake output (which makes your change to the PR unnecessary – also, just as a BTW, we want to use braces in all statements). I don't think we want to expose this output. We should probably not set the property in this case, but I'm not 100% on that decision.

@stapelberg Thoughts?

pstray: Another test then, to check if it is a real output or not? Any pointers on what to check?
Regarding braces, there are a whole lot of single-line if-elses without braces in that file, but, yes, I do agree that braces are nice :) I'll fix that tomorrow.

Airblader: > Another test then, to check if it is a real output or not? Any pointers on what to check?

Let's first decide on a solution and then implement it.

> there are a whole lot of single-line if-elses without braces in that file

Yeah, but we want to move towards having it consistent. :-)

stapelberg: I think we should include pseudo outputs for scratchpad containers.

The only other alternative which comes to my mind is not including the “output” property at all, and that case would be indistinguishable from talking with an older i3 version which just doesn’t include the “output” property yet.

I think users of the IPC interface should be expected to handle internal outputs.

Airblader: Alright. @pstray Can you make these changes then?

pstray: Now the patch causes 'output' to always be present, and null only if the name is null.

btw, when I do restart in the next-branch, all input is captured in a small black box in the middle of the screen... any idea what that is?

Airblader: > btw, when I do restart in the next-branch, all input is captured in a small black box in the middle of the screen... any idea what that is?

That's i3 crashing and showing the crash dialog, which for some reason isn't rendering correctly (that's a known bug.) This means that restarting crashes i3 and needs to be fixed.

pstray: ah, right... from an an assert in con_get_output, causing it to die instead of returning null it seems... Need to change my code a bit so that doesn't happen then...

Is it really that that smart to have the assert there in con.c, line 365?

Airblader: I think the issue is that you are trying to print the output for all containers, including output and those higher up than output containers. Trying to determine the output for such a container doesn't make sense and the assertion rightfully reminds you of that. :-) In fact I'd expect i3 to crash when calling `i3-msg -t get_tree` as well – have you ever tried that? What did you test your PR with?

Your PR should limit printing the output to containers that are below output containers in the hierarchy.

pstray: Yes, probably... not sure what event causes the crash, because it only happens on restart when usign con_get_output. And testing is of course done my running the patched version of i3 :)

I checked the code for i3, and there are a lot of calls to con_get_output that still tests for a NULL return value, which of course is smart programming-wise to prevent things from crashing if con_get_output does return NULL sometime... 

Wouldn't say its non-informative, but it's definitly duplicate information. So no output on Cons of type ROOT and OUTPUT then?

Airblader: > not sure what event causes the crash

During a restart we dump the tree to file so we can read that file and restore the tree after the restart.

> And testing is of course done my running the patched version of i3

So running `i3-msg -t get_tree` didn't crash? Does the change to restricting for which containers to print the output fix the crash on restart?

> So no output on Cons of type ROOT and OUTPUT then?

I don't feel strongly for CT_OUTPUT, but for CT_ROOT it definitely doesn't make sense. :-)

src/ipc.c
```diff
@@ -294,6 +294,12 @@ void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     ystr("focused");
     y(bool, (con == focused));
 
+    if (con->type != CT_ROOT && con->type != CT_OUTPUT) {
+	ystr("output");
+	Con *output_con = con_get_output(con);
+	ystr(output_con->name);
```
Airblader: You can just inline `con_get_output` here now.

src/ipc.c
pstray: > So running i3-msg -t get_tree didn't crash?  Does the change to restricting for which containers to print the output fix the crash on restart?

Neither of the manual traversal in de19427e6df9c6ad9dfbbea5c5052e732a5aa335, or the output for only childs in 324ea5646b25ba5d911e7dd4b6206574ec1e0261causes a crash on restart or get-tree.

src/ipc.c
pstray: That is true :) Will fix

src/ipc.c
Airblader: Yeah, but the original »just dump the output on every node calling `con_get_output`« which also caused the crash on restart should've also caused a crash when calling `GET_TREE` manually, so I was wondering why it didn't crash when you were testing. :-)

pstray:
So, good to go?

stapelberg:
@Airblader I’d appreciate it if you could handle the review of this PR, given that you’ve been involved with the review so far.

@pstray Can you please squash your changes into a single commit? We’d like to keep the git history clean.

Airblader:
@pstray If you push the commit you need to leave a comment because pushing doesn't trigger a notification. :-)

@stapelberg Should be good to go now.

pstray:
@Airblader good to know :)

PR #2491, opened 2016-10-03
Optionally set layout of container in focus rather than parent

strake:
I have been using this patch personally for years, so i thought i ought to finally send it upstream.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -105,6 +105,10 @@ struct Config {
     /** Default orientation for new containers */
     int default_orientation;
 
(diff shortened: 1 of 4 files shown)


Airblader:
This PR has a couple of severe issues, most importantly that it isn't actually configurable but just defaults to being activated. 

But before we address them let's decide on whether it should be merged. To decide so I think you should describe the use case and motivation for it.

Airblader:
Given the vastly incomplete state I'm closing the PR now. If you want this merged please open a feature request answering the questions above so we can first discuss it. Thanks!

Issue #1758, opened 2015-06-16
Implement sticky_groups correctly


Airblader:
@stapelberg I just came across the "sticky_group" implementation in i3. As far as I can tell there is no way to set it and it is only set when loading a layout (but not printed when dumping one).

Are there any plans for this? Can it be removed? I'm not sure I'm using it right, but it seems to be broken to me.

This also seems to be different from what #11 describes. Looks more like the idea of this is to have user-opened containers marked as sticky groups where the contents will travel with the user when switching workspace to another container that has a sticky group. Am I understanding this correctly?

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
I played around more with it and I'm convinced now that the feature is broken. It causes all kinds of problems.

stapelberg:
> @stapelberg I just came across the "sticky_group" implementation in i3. As far as I can tell there is no way to set it and it is only set when loading a layout (but not printed when dumping one).

If it’s missing in dumped layouts, that should be fixed eventually, but see below.

> Are there any plans for this? Can it be removed? I'm not sure I'm using it right, but it seems to be broken to me.

There is no concrete plan. I started implementing this years ago, but stopped because I couldn’t figure out what a decent user interface for that feature looks like. Hence, as you figured out, it’s currently not well-tested and likely somewhat broken.

> This also seems to be different from what #11 describes. Looks more like the idea of this is to have user-opened containers marked as sticky groups where the contents will travel with the user when switching workspace to another container that has a sticky group. Am I understanding this correctly?

Yes, I think you’re understanding this correctly. In my words, a sticky group contains multiple containers, all of which display the same window, but only one of which can be visible at a time (or at least the others need to show a placeholder, which is something that I didn’t think of back then).

With regards to removing it, I think we have two possibilities:

1. Actually finish the feature, which would be my preferred course of action. I think it covers the use-case for which people use tags in other window managers, i.e. having windows be displayed in multiple window arrangements on different workspaces.
2. Remove the feature. That, IMO, requires an explanation of why it cannot work in i3 at all, aside from “we didn’t feel like doing the work to get it done”. I’m not yet convinced such an explanation exists :).

Airblader:
[...]

PR #2496, opened 2016-10-10
Implement 'swap' command.

### LINKED ISSUE #917: Swap container functionality
[Originally reported by mariusm@…]
(Sometimes I would like to be able to do something like this:

```
[con_mark="..."] swap
```

which would swap the contents of the focused and the marked container.

Also something like:

```
swap left | right | up | down
```

that would work similarly to move, but swap windows instead of move them (in the same parent container swap would be the same thing as move, but between different parent containers it would be different).

Airblader:
This patch introduces a new command 'swap' that swaps two containers so
that they assume each other's position and geometry.

fixes #917

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1976,6 +1976,39 @@ bindsym $mod+c move absolute position center
 bindsym $mod+m move position mouse
 -------------------------------------------------------
 
+=== Swapping containers
+
+Two containers can be swapped (i.e., move to each other's position) by using
+the +swap+ command. They will assume the position and geometry of the container
+they are swapped with.
+
+The first container to participate in the swapping can be selected through the
+normal command criteria process with the focused window being the usual
+fallback if no criteria are specified. The second container can be selected
+using one of the following methods:
+
++id+:: The X11 window ID of a client window.
++con_id+:: The i3 container ID of a container.
++mark+:: A container with the specified mark, see <<vim_like_marks>>.
+
+Note that swapping does not work with all containers. Most notably, swapping
+floating containers or containers that have a parent-child relationship to one
+another does not work.
+
+*Syntax*:
+----------------------------------------
+swap container with id|con_id|mark <arg>
+----------------------------------------
+
+*Examples*:
+-----------------------------------------------------------------
+# Swaps the focused container with the container marked »swapee«.
+swap container with mark swapee
+
+# Swaps container marked »A« and »B«
+[con_mark="^A$"] swap container with mark B
+-----------------------------------------------------------------
+
 === Sticky floating windows
 
 If you want a window to stick to the glass, i.e., have it stay on screen even
--- a/src/commands.c
+++ b/src/commands.c
@@ -1813,6 +1813,65 @@ void cmd_scratchpad_show(I3_CMD) {
 }
 
 /*
+ * Implementation of 'swap [container] [with] id|con_id|mark <arg>'.
+ *
+ */
+void cmd_swap(I3_CMD, const char *mode, const char *arg) {
+    HANDLE_EMPTY_MATCH;
+
+    owindow *match = TAILQ_FIRST(&owindows);
+    if (match == NULL) {
+        DLOG("No match found for swapping.\n");
+        return;
+    }
+
+    Con *con;
+    if (strcmp(mode, "id") == 0) {
+        long target;
+        if (!parse_long(arg, &target, 0)) {
+            yerror("Failed to parse %s into a window id.\n", arg);
+            return;
+        }
+
+        con = con_by_window_id(target);
+    } else if (strcmp(mode, "con_id") == 0) {
+        long target;
+        if (!parse_long(arg, &target, 0)) {
+            yerror("Failed to parse %s into a container id.\n", arg);
+            return;
+        }
+
+        con = (Con *)target;
+    } else if (strcmp(mode, "mark") == 0) {
+        con = con_by_mark(arg);
+    } else {
+        yerror("Unhandled swap mode \"%s\". This is a bug.\n", mode);
+        return;
+    }
+
+    if (con == NULL) {
+        yerror("Could not find container for %s = %s\n", mode, arg);
+        return;
+    }
+
+    if (match == TAILQ_LAST(&owindows, owindows_head)) {
+        DLOG("More than one container matched the swap command, only using the first one.");
+    }
+
+    if (match->con == NULL) {
+        DLOG("Match %p has no container.\n", match);
+        ysuccess(false);
+        return;
+    }
+
+    DLOG("Swapping %p with %p.\n", match->con, con);
+    bool result = con_swap(match->con, con);
+
+    cmd_output->needs_tree_render = true;
+    ysuccess(result);
+}
+
+/*
  * Implementation of 'title_format <format>'
  *
  */
--- a/src/con.c
+++ b/src/con.c
@@ -22,9 +22,11 @@ static void con_on_remove_child(Con *con);
 void con_force_split_parents_redraw(Con *con) {
     Con *parent = con;
 
-    while (parent && parent->type != CT_WORKSPACE && parent->type != CT_DOCKAREA) {
-        if (!con_is_leaf(parent))
+    while (parent != NULL && parent->type != CT_WORKSPACE && parent->type != CT_DOCKAREA) {
+        if (!con_is_leaf(parent)) {
             FREE(parent->deco_render_params);
+        }
+
         parent = parent->parent;
     }
 }
@@ -145,7 +147,7 @@ static void _con_attach(Con *con, Con *parent, Con *previous, bool ignore_focus)
 
         /* Insert the container after the tiling container, if found.
          * When adding to a CT_OUTPUT, just append one after another. */
-        if (current && parent->type != CT_OUTPUT) {
+        if (current != NULL && parent->type != CT_OUTPUT) {
             DLOG("Inserting con = %p after con %p\n", con, current);
             TAILQ_INSERT_AFTER(nodes_head, current, con, nodes);
         } else
@@ -526,6 +528,23 @@ bool con_inside_focused(Con *con) {
 }
 
 /*
+ * Checks if the container has the given parent as an actual parent.
+ *
+ */
+bool con_has_parent(Con *con, Con *parent) {
+    Con *current = con->parent;
+    if (current == NULL) {
+        return false;
+    }
+
+    if (current == parent) {
+        return true;
+    }
+
+    return con_has_parent(current, parent);
+}
+
+/*
  * Returns the container with the given client window ID or NULL if no such
  * container exists.
  *
@@ -803,10 +822,11 @@ void con_fix_percent(Con *con) {
     if (children_with_percent != children) {
         TAILQ_FOREACH(child, &(con->nodes_head), nodes) {
             if (child->percent <= 0.0) {
-                if (children_with_percent == 0)
+                if (children_with_percent == 0) {
                     total += (child->percent = 1.0);
-                else
+                } else {
                     total += (child->percent = total / children_with_percent);
+                }
             }
         }
     }
@@ -814,11 +834,13 @@ void con_fix_percent(Con *con) {
     // if we got a zero, just distribute the space equally, otherwise
     // distribute according to the proportions we got
     if (total == 0.0) {
-        TAILQ_FOREACH(child, &(con->nodes_head), nodes)
-        child->percent = 1.0 / children;
+        TAILQ_FOREACH(child, &(con->nodes_head), nodes) {
+            child->percent = 1.0 / children;
+        }
     } else if (total != 1.0) {
-        TAILQ_FOREACH(child, &(con->nodes_head), nodes)
-        child->percent /= total;
+        TAILQ_FOREACH(child, &(con->nodes_head), nodes) {
+            child->percent /= total;
+        }
     }
 }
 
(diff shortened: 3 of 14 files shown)


Airblader:
I'll still have to write quite a few testcases.

Airblader:
I also published a proof of concept implementation for #1758 using this feature here: https://github.com/Airblader/i3-sticky

Airblader:
And here in action, just because I think it's pretty cool: https://www.youtube.com/watch?v=kPGXqFPnswc

src/con.c
```diff
@@ -2116,3 +2136,94 @@ i3String *con_parse_title_format(Con *con) {
...
+        con_focus(con_descend_focused(current_ws));
+    }
+
+    /* #2: Move second to where first has been originally. */
+    result &= _con_move_to_con(second, fake, false, false, false, true);
+    /* If the first container originally contained the focused window, it makes
+     * sense to give the swapped-in second container focus now that first can
+     * no longer be focused if we moved it to another workspace. */
+    if (first_ws != second_ws && focused_within_first) {
+        con_focus(con_descend_focused(second));
```
Airblader: So this behavior makes sense to me but we should discuss it. Let's assume one of the cons we swap is focused. If we swap within the same workspace, the con will maintain focus meaning focus will stay on the con but be in a different location. 

Now lets assume the focused container is swapped into another workspace. I think it'd be disruptive to focus another workspace plus tooling (which the swap feature likely requires in most use cases) can do that job. Otherwise scripts have to undo workspace switching which could cause a flicker. 

In essence the current behavior tries to make minimal changes to focus but does it in a meaningful way if we have to. Any deviating behavior should be handled by the tooling.

src/con.c
```diff
@@ -2116,3 +2136,94 @@ i3String *con_parse_title_format(Con *con) {
...
+    assert(first != NULL);
+    assert(second != NULL);
+    DLOG("Swapping containers %p / %p\n", first, second);
+
+    if (first->type != CT_CON || second->type != CT_CON) {
+        ELOG("Only regular containers can be swapped, but found %d / %d\n", first->type, second->type);
+        return false;
+    }
+
+    if (con_is_floating(first) || con_is_floating(second)) {
```
Airblader: From my sticky group example I've seen that I really want to make this work with floating windows but I think we should take this feature one step at a time. Both to keep the PR reasonably small and because traditionally these features are likely to come back with a bug or two.

src/con.c
```diff
@@ -2116,3 +2136,94 @@ i3String *con_parse_title_format(Con *con) {
 
     return formatted;
 }
+
+/*
+ * Swaps the two containers.
```
Airblader: Should describe the behavior in more detail, in particular focus handling.

src/commands.c
```diff
@@ -1821,6 +1821,59 @@ void cmd_scratchpad_show(I3_CMD) {
...
+        return;
+    }
+
+    if (con == NULL) {
+        yerror("Could not find container for %s = %s\n", mode, arg);
+        return;
+    }
+
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
```
Airblader: Actually selecting more than one con makes no sense here. I think we should check for that, print a message and only handle the first matched container.

src/commands.c
```diff
@@ -1821,6 +1821,59 @@ void cmd_scratchpad_show(I3_CMD) {
 }
 
 /*
+ * Implementation of 'swap [container] [with] id|con_id|mark <arg>'.
+ *
+ */
+void cmd_swap(I3_CMD, const char *mode, const char *arg) {
+    HANDLE_EMPTY_MATCH;
+    // TODO XXX Extract conversion to long
```
Airblader: Todo

src/con.c
Airblader: This will actually change again slightly.

Airblader:
@stapelberg OK, I'll try to explain the focus behavior I'm going for now. I'd love if you could give it some thought because if you disagree, it'd be a substantial change so knowing early on is better (albeit I'm far past early already.)

There's basically three scenarios:

1. One of the two swapees is currently focused (or a parent thereof, I'll omit this from here on out) but they're on the same workspace.
2. One of the two swapees is currently focused, but the other one is on another workspace.
3. None of the swapees is focused.

And here's the intended focus behavior after swapping:

1. The previously focused container stays focused (which is in another physical location on the same workspace now.)
2. The non-focused swapee is focused in place of the other one. We cannot keep focus unless we switch workspaces, which is a drastic change. Keeping focus in the same physical location and the other involved swapee seems most intuitive to me.
3. The previously focused container stays focused and the swapees are inserted into the parents' focus stacks in the place where their counterpart used to be.

src/con.c
```diff
@@ -1120,6 +1144,9 @@ static bool _con_move_to_con(Con *con, Con *target, bool behind_focused, bool fi
         con_set_urgency(con, true);
     }
 
+    /* Ensure the container will be redrawn. */
+    FREE(con->deco_render_params);
```
Airblader: This is actually a bug that I'm *pretty* sure affects existing usecases already, though I haven't really checked. We do call `force_split_blahblah` from `on_remove_child` which is called in the next line, but we do it _on the parent_. Also that function ignores leaf containers. This leads to a broken rendering of urgent containers after moving.

Airblader:
@stapelberg Well now I really hope you agree because I think this is ready for review now. I'll also do some more local testing, though.

Airblader:
I should add another test for swapping a fullscreen container.

src/con.c
```diff
@@ -937,7 +959,7 @@ void con_disable_fullscreen(Con *con) {
     con_set_fullscreen_mode(con, CF_NONE);
 }
 
-static bool _con_move_to_con(Con *con, Con *target, bool behind_focused, bool fix_coordinates, bool dont_warp, bool ignore_focus) {
+static bool _con_move_to_con(Con *con, Con *target, bool behind_focused, bool fix_coordinates, bool dont_warp, bool ignore_focus, bool fix_pecentage) {
```
stapelberg: I agree that this function needs refactoring in a followup PR. I don’t know yet which approach I like best, so I’ll leave that up to you and comment on it once I see actual code :).

src/con.c
stapelberg: Can you send a fix for that in a separate PR please?

stapelberg:
The focus behavior sounds good to me on first glance.

docs/userguide
```diff
@@ -1947,6 +1947,39 @@ bindsym $mod+c move absolute position center
...
 -------------------------------------------------------
 
+=== Swapping containers
+
+Two containers can be swapped (i.e., move to each other's position) by using
+the +swap+ command. They will assume the position and geometry of the container
+they are swapped with.
+
+The first container to participate in the swapping can be selected through the
+normal command criteria process with the focused window being the usuall
```
stapelberg: s/usuall/usual/

docs/userguide
```diff
@@ -1947,6 +1947,39 @@ bindsym $mod+c move absolute position center
...
+fallback if no criteria are specified. The second container can be selected
+using one of the following methids:
```
stapelberg: s/methids/methods/

src/con.c
```diff
@@ -1046,9 +1068,11 @@ static bool _con_move_to_con(Con *con, Con *target, bool behind_focused, bool fi
     _con_attach(con, target, behind_focused ? NULL : orig_target, !behind_focused);
 
     /* 5: fix the percentages */
-    con_fix_percent(parent);
-    con->percent = 0.0;
-    con_fix_percent(target);
+    if (fix_pecentage) {
```
stapelberg: s/fix_pecentage/fix_percentage/

src/con.c
```diff
@@ -2116,3 +2144,154 @@ i3String *con_parse_title_format(Con *con) {
...
+ * Swaps the two containers.
+ *
+ */
+bool con_swap(Con *first, Con *second) {
+    assert(first != NULL);
+    assert(second != NULL);
+    DLOG("Swapping containers %p / %p\n", first, second);
+
+    if (first->type != CT_CON || second->type != CT_CON) {
+        ELOG("Only regular containers can be swapped, but found %d / %d\n", first->type, second->type);
```
stapelberg: Please improve this error message:

1. Please check each case individually, so that we can tell the user precisely which container is the problem.
2. Please include some container identification in the error message (`%p` and the title come to mind).
3. Please change the wording so that it’s clear what the magic numbers are (e.g. `[…] but found containers of type %d / type %d`).

src/con.c
```diff
@@ -2116,3 +2144,154 @@ i3String *con_parse_title_format(Con *con) {
...
+        return false;
+    }
+
+    if (con_is_floating(first) || con_is_floating(second)) {
+        ELOG("Floating windows cannot be swapped.\n");
+        return false;
+    }
+
+    if (first == second || con_has_parent(first, second) || con_has_parent(second, first)) {
+        DLOG("Cannot swap container with itself or into one another.\n");
```
stapelberg: Why DLOG instead of ELOG?

Also, like in my earlier comment, please split this into two conditionals. Should we use the same wording as in the documentation, i.e. “parent-child relationship”?

src/con.c
```diff
@@ -2116,3 +2144,154 @@ i3String *con_parse_title_format(Con *con) {
...
+    Con *old_focus = focused;
+
+    Con *first_ws = con_get_workspace(first);
+    Con *second_ws = con_get_workspace(second);
+    Con *current_ws = con_get_workspace(old_focus);
+    const bool focused_within_first = (first == old_focus || con_has_parent(old_focus, first));
+    const bool focused_within_second = (second == old_focus || con_has_parent(old_focus, second));
+
+    if (!con_fullscreen_permits_focusing(first_ws) || !con_fullscreen_permits_focusing(second_ws)) {
+        DLOG("Cannot swap because one target workspace is obscured.\n");
```
stapelberg: See my earlier comments about error messages.

src/con.c
```diff
@@ -2116,3 +2144,154 @@ i3String *con_parse_title_format(Con *con) {
...
+
+    double first_percent = first->percent;
+    double second_percent = second->percent;
+
+    /* De- and reattaching the containers will insert them at the tail of the
+     * focus_heads. We will need to fix this. But we need to make sure first
+     * and second don't get in each other's way if they share the same parent,
+     * so we select the closest previous focus_head that's not involved. */
+    Con *first_prev_focus_head = first;
+    while (first_prev_focus_head == first || first_prev_focus_head == second) {
```
stapelberg: Can `first_prev_focus_head == second` ever be true, given that we check `con_has_parent(first, second)` above?

stapelberg (changes requested):

src/con.c
Airblader: `first == second` isn't an error scenario, IMHO (it's a no-op). The other cases deserve an ELOG, agreed.

Airblader:
@stapelberg I'm *extremely* busy at work at the moment. I'll revisit this PR when I get time which might be a couple of weeks.

src/con.c
Airblader: Yes. If first and second are both inside the same tabbed container T[second, first].

docs/userguide
Airblader: fixed

docs/userguide
Airblader: fixed

src/con.c
Airblader: fixed

src/con.c
Airblader: fixed

src/con.c
Airblader: fixed

src/con.c
Airblader: fixed

src/con.c
Airblader: done

Airblader:
@stapelberg I think this is good to go now, so perhaps you could review it again. Once merged I'll create issues for the followup improvements, but for now I'd like to get some experience with this feature before a next release.

src/con.c
```diff
@@ -2159,3 +2187,169 @@ i3String *con_parse_title_format(Con *con) {
...
+        return false;
+    }
+
+    if (con_is_floating(first) || con_is_floating(second)) {
+        ELOG("Floating windows cannot be swapped.\n");
+        return false;
+    }
+
+    if (first == second) {
+        DLOG("Swapping container %p with itself, nothind to do.\n", first);
```
stapelberg: s/nothind/nothing

src/con.c
```diff
@@ -2159,3 +2187,169 @@ i3String *con_parse_title_format(Con *con) {
...
+
+    Con *old_focus = focused;
+
+    Con *first_ws = con_get_workspace(first);
+    Con *second_ws = con_get_workspace(second);
+    Con *current_ws = con_get_workspace(old_focus);
+    const bool focused_within_first = (first == old_focus || con_has_parent(old_focus, first));
+    const bool focused_within_second = (second == old_focus || con_has_parent(old_focus, second));
+
+    if (!con_fullscreen_permits_focusing(first_ws)) {
```
stapelberg: Is it strictly necessary to prevent swapping from working in this scenario? Similar to the sometimes-requested “fullscreen layout”, I can think of use-cases where one would want to swap some containers with the currently fullscreened container.

It’s also okay to address this in a subsequent PR if you’d like, just wanted to bring it up now.

stapelberg (approved):
Largely looks good — I’m happy to merge once the nits are addressed.

src/con.c
Airblader: Yes, fullscreen is already on the list of followup PRs from last year. We should definitely allow it, but it seemed this PR is big enough as is. :-)

Airblader:
@stapelberg The nit is fixed.

stapelberg (approved):

PR #2503, opened 2016-10-13
Implement new window::mark IPC event.

### LINKED ISSUE #2501: [enhancement] New window::mark event
I would like to see `change:mark` for the IPC `window` event. My particular usecase right now would be implementing sticky tiling windows after we get #2496 merged. I would use marks to coordinate sticky groups and need to be informed when a mark is added on a window.

Airblader:
This introduces a new type of 'window' event sent wit change:mark whenever
a mark on a window changes.

fixes #2501

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -754,14 +754,15 @@ defines whether pango markup shall be used for displaying this mode.
 This event consists of a single serialized map containing a property
 +change (string)+ which indicates the type of the change
 
-* +new+ - the window has become managed by i3
-* +close+ - the window has closed
-* +focus+ - the window has received input focus
-* +title+ - the window's title has changed
-* +fullscreen_mode+ - the window has entered or exited fullscreen mode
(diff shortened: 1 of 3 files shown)


Airblader:
I chose to send a separate event for each change of a mark even if all marks on a con are removed to set a new mark (i.e., one action resulting in multiple events) for two reasons:

1. It's easier.
2. If we later on want to provide information *which* mark changed we'd only add information, but not change how the events are sent.

docs/ipc
```diff
@@ -754,14 +754,15 @@ defines whether pango markup shall be used for displaying this mode.
 This event consists of a single serialized map containing a property
 +change (string)+ which indicates the type of the change
 
-* +new+ - the window has become managed by i3
```
Airblader: (If you're wondering, I've replaced the dash with the typographically correct hyphen)

PR #2507, opened 2016-10-16
Switch to autotools (GNU build system)

stapelberg:
I suggest to review commit-by-commit.

Refer to the commit description of  https://github.com/i3/i3/commit/3f33edf31b4286bee4b604c5d27f95c13ae52727 (main commit of this pull request) for rationale.

The website will be updated accordingly once this is merged.


Airblader:
Yeah, it sure is a surprise. :-) Given the size of the PR it'll take me a bit to review, though I'll probably not review it too in-depth anyway. When it comes to compiling every build is the best test there is. Plus I don't really know much about autotools.

acrisci:
I never thought I would live to see the day.

acrisci:
I hear @sardemff7 knows a lot about autotools...

Airblader:
> “make install” installs the same files (plus documentation and manpages)

So this is a change to before, right? We didn't use to install man pages because, if I recall correctly, you said the path for man pages isn't consistent. Is this obsolete?

Also, a quick attempt to prepare my AUR packages for the merge causes countless warnings of the type

````
warning: #warning _FORTIFY_SOURCE requires compiling with optimization (-O)
```

I'm not sure whether this is an issue in my package, though. :-)

Airblader:
The LICENSE file isn't copied into the build folder. I think it should. Same for I3_VERSION (?)

configure.ac
```diff
@@ -0,0 +1,169 @@
...
+AC_CHECK_FUNCS([atexit dup2 ftruncate getcwd gettimeofday localtime_r memchr memset mkdir rmdir setlocale socket strcasecmp strchr strdup strerror strncasecmp strndup strrchr strspn strstr strtol strtoul], , [AC_MSG_FAILURE([cannot find the $ac_func function, which i3 requires])])
+
+# Checks for libraries.
+
+AC_SEARCH_LIBS([floor], [m], , [AC_MSG_FAILURE([cannot find the required floor() function despite trying to link with -lm])])
+
+# libev does not ship with a pkg-config file :(.
+AC_SEARCH_LIBS([ev_run], [ev], , [AC_MSG_FAILURE([cannot find the required ev_run() function despite trying to link with -lev])])
+
+AC_CHECK_FUNCS([shm_open], , [AC_CHECK_LIB([rt], [shm_open])])
```
sardemff7: Why not `AC_SEARCH_LIBS` too?

configure.ac
```diff
@@ -0,0 +1,169 @@
...
+
+AC_CHECK_LIB([pthread], [pthread_condattr_setpshared])
```
sardemff7: Since you use autoconf-archive, why not `AX_PTHREAD`?

configure.ac
```diff
@@ -0,0 +1,169 @@
+#                                               -*- Autoconf -*-
+# Run autoreconf -fi to generate a configure script from this file.
+
+AC_PREREQ([2.69])
+AC_INIT([i3], [4.12], [https://github.com/i3/i3/issues])
+# For AX_EXTEND_SRCDIR
+AX_ENABLE_BUILDDIR
+AM_INIT_AUTOMAKE([foreign subdir-objects -Wall])
```
sardemff7: I saw `make dist-bzip2` in your `release.sh` commit, why not `no-dist-gzip dist-bzip2` here directly? Why not `dist-xz` too? :-)

sardemff7:
@Airblader 
> I'm not sure whether this is an issue in my package, though. :-)

Yes, `-O` should not be upstream. Period. A really really smart default at best.

> The LICENSE file isn't copied into the build folder. I think it should. Same for I3_VERSION (?)

No, build folder is for build-generated files, if you need a disted file, just pick it from the source folder. :-)

Airblader:
@sardemff7 Thanks for the input. My build system is configured to use `-O2`, but somehow it isn't applied when building my changed package. I'll have to look into that, but that's my problem. :-)

Airblader:
@sardemff7 I've tried to resolve my problem with the warnings I mentioned above and came across [this](http://arch-dev-public.archlinux.narkive.com/7049fxEM/build-issues-due-to-d-fortify-source-2-in-cppflags) and similar posts. They suggest this warning is triggered if the configure scripts are extremely old. Is this the case here?

I'm virtually building the package as @stapelberg's commit message recommends and I somewhat refuse to believe that Arch's default makepkg configuration is broken such that it surfaces with the i3 package but not countless other packages built using autoconf. :-) So I'd like to understand why my package build should have to override compiler settings.

Airblader:
Is it possible that the `/etc` folder is put into the wrong folder? It ends up beneath `/usr` for me, which seems wrong.

configure.ac
```diff
@@ -0,0 +1,170 @@
...
+AM_PROG_AR
+
+AX_FLAGS_WARN_ALL
+AC_MSG_CHECKING(whether compiler understands -Wunused-value)
+old_CFLAGS="$CFLAGS"
+CFLAGS="$CFLAGS -Wunused-value"
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([],[])],
+  AC_MSG_RESULT(yes),
+  AC_MSG_RESULT(no)
+  CFLAGS="$old_CFLAGS")
```
sardemff7: Use `AM_CFLAGS`. You have to test using `CFLAGS`, but you should always restore the value and add yours in `AM_CFLAGS` so that users can override yours.

sardemff7:
`*FLAGS` are *not* upstream, they *never should be*, ever. These variables are for users. At package level, the user is the packager.
If you put `-D_FORTIFY_SOURCE` in `CPPFLAGS`, it is your responsibility alone to add `-O2` too. If Arch have `-D_FORTIFY_SOURCE` in a global `makepkg.conf`, they must have `-O2` too.
If this flag comes from the build system, then you should have a default in `AM_CPPFLAGS`, but not `CPPFLAGS`.

I don’t know for your `/etc` issue. If prefix is `/usr`, Autoconf should use `/etc` and not `${prefix}/etc`.
But again, if the packaging system use its own values, it should take care of all cases.

Airblader:
@sardemff7 Arch has `-D_FORTIFY_SOURCE=2` in `CPPFLAGS` and `-O2` in `CFLAGS`. If you say that this is an issue with Arch's default config I'll take this up with the Arch mailing list. Thanks for the input. :-)

> I don’t know for your /etc issue. If prefix is /usr, Autoconf should use /etc and not ${prefix}/etc.

Using `--sysconfdir=/etc` in my PKGBUILD works. But I'd still say the commit message is breaking its promise here. Perhaps it should be more clear on specifying the necessary parameters for `configure` to actually get the same file layout as with the old make-based system independent of what just happens to be the default value.

The old make-based system actually had logic for this:

````
ifndef SYSCONFDIR
  ifeq ($(PREFIX),/usr)
    SYSCONFDIR=/etc
  else
    SYSCONFDIR=$(PREFIX)/etc
  endif
endif
```

sardemff7:
Autoconf has the same kind of code, it surprises me that it doesn’t work.

acrisci:
I would like to see an `autogen.sh` file which can generate the project minimally by calling:

    autoreconf --force --install --verbose

and whatever else you need. This is good for people who are not familiar with the GNU build system.

Here is [an example](https://github.com/acrisci/i3ipc-glib/blob/master/autogen.sh) from one of my projects that emulates the Gnome way of doing it (although you won't need the gtkdoc stuff).

acrisci:
Also I think that the project is reaching the level of complexity that a separate task runner script could be a good idea for things like building, generating, releasing, running with special options, doing tests, etc. I've been doing doing this for all my projects recently and I find that it works much better than documentation for demonstrating how certain things are done in the project. I'll write a POC for that.

i3bar/src/child.c
```diff
@@ -105,7 +105,8 @@ __attribute__((format(printf, 1, 2))) static void set_statusline_error(const cha
     char *message;
     va_list args;
     va_start(args, format);
-    (void)vasprintf(&message, format, args);
+    if (vasprintf(&message, format, args) == -1)
+        return;
```
stapelberg: Done.

debian/rules
```diff
@@ -4,14 +4,14 @@
...
-	dh_auto_build -- all docs mans
-
 override_dh_installchangelogs:
 	dh_installchangelogs RELEASE-NOTES-*
 
 override_dh_strip:
 	dh_strip --dbg-package=i3-wm-dbg
 
+override_dh_auto_test:
+	# TODO: enable tests
```
stapelberg: Intentional. I’ll leave this here and give it a whirl when doing the next round of Debian package cleanups. Issues here should primarily be about i3, and we don’t track the other shortcomings of the Debian packaging either, so I’d just leave it as-is.

stapelberg:
Thanks everyone for the review and discussion. I’ll address @sardemff7’s review comments in a second. If I forgot to address something else that came up in the discussion (i.e. not in the review itself), please let me know.

>> “make install” installs the same files (plus documentation and manpages)

> So this is a change to before, right? We didn't use to install man pages because, if I recall correctly, you said the path for man pages isn't consistent. Is this obsolete?

@Airblader You recall correctly, and yes, I’d say this is obsolete: all distributions have tooling to invoke configure in such a way that the correct manpage path is passed. So, we still don’t know for sure where they’ll end up, but I have very high confidence that packagers/distribution will naturally end up doing the right thing.

>> [about --prefix=/usr not resulting in --sysconfdir=/etc]

> Autoconf has the same kind of code, it surprises me that it doesn’t work.

@sardemff7 Can you point me to that code? I can’t see it in autoconf 2.69. In fact, testing with GNU hello 2.10 (which, IIUC, poses a reference example), when running `./configure --prefix=/usr && make lib/configmake.h`, I see `echo '#define SYSCONFDIR "/usr/etc"';` in the output.

> I would like to see an autogen.sh file which can generate the project minimally by calling:
> 
>     autoreconf --force --install --verbose
>
> and whatever else you need. This is good for people who are not familiar with the GNU build system.

@acrisci I’m a bit on the fence about this. Here are a couple of thoughts:

* The file would likely literally contain `exec autoreconf -fi`, so it won’t save a lot of typing (caveat: of course I can’t guarantee that we’ll never need any other tools). The only advantage I can see is that some people might only be familiar with `autogen.sh`, not with running `autoreconf -fi` directly.
* I remember that when I didn’t know how the GNU build system worked, I didn’t realize that running the `autogen.sh` script is what I should have done.
* We don’t want novice users to deal with building i3 anyway, as we strongly encourage people to use their distribution’s packages.

> Also I think that the project is reaching the level of complexity that a separate task runner script could be a good idea for things like building, generating, releasing, running with special options, doing tests, etc.

@acrisci This sounds like the exact opposite of where I’d like to go. I want the GNU build system be the (hopefully familiar, or at least worthy of time/knowledge investment) interface we present to our developers. But let’s discuss this in a separate issue please, this PR is already complex enough as-is :).

configure.ac
stapelberg: > I saw make dist-bzip2 in your release.sh commit, why not no-dist-gzip dist-bzip2 here directly? 

I guess that’s slightly cleaner. Done. I’ll leave the dist-bzip2 call in release.sh so that it’s robust against changes in the default for now.

> Why not dist-xz too? :-)

If you’re suggesting to also publish xz tarballs on our website, let’s discuss that in a separate issue.

If not, I’m not sure what you’re asking? IIUC, the effect of specifying dist-xz is that we end up with a bzip2 and an xz file, of which we’d use only the bzip2 file…?

configure.ac
stapelberg: Done.

configure.ac
stapelberg: Done.

stapelberg:
Alright, pushed my changes — please take another look and let me know what needs to be addressed now (as opposed to incremental improvements after this is merged).

Airblader:
LGTM as far as I can tell

sardemff7:
My bad, the `/etc` thing is code [from the doc](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf.html#Site-Defaults) for distributions to put nice default values.

stapelberg:
Thanks for clarifying. I’ll let @Airblader figure out whether Arch should use site defaults or otherwise change their build system to make this common case easier.

Given that there are no new comments, I’ll merge this PR now and we’ll sort out any incremental improvements afterwards.

PR #2508, opened 2016-10-17
Respect minimum size hints for floating windows.

### LINKED ISSUE #2436: x11-ssh-askpass opens with wrong size (tiny floating window)

**Affected versions:**

```
$ i3 --moreversion
Binary i3 version:  4.11 (2015-09-30, branch "4.11") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.10.4 (2015-09-08, branch "4.10.4") (pid 31630)

```

```
$ i3 --moreversion
Running i3 version: 4.10.4 (2015-09-08, branch "4.10.4") (pid 31630)

The i3 binary you just called: /home/kev/sw/i3/4.12/bin/i3
```

```
$ i3 --moreversion
Binary i3 version:  4.12-31-g0e29101 (2016-03-06, branch "master") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.10.4 (2015-09-08, branch "4.10.4") (pid 31630)

The i3 binary you just called: /home/kev/sw/i3/4.12+2016-09-03/bin/i3
```

**Unaffected version:**

```
$ i3 --moreversion
Binary i3 version:  4.10.4 (2015-09-08, branch "4.10.4") © 2009-2014 Michael Stapelberg and contributors
Running i3 version: 4.10.4 (2015-09-08, branch "4.10.4") (pid 31630)

The i3 binary you just called: /home/kev/sw/i3/4.10.4/bin/i3
```

Let me know if this is actually needed.  Since the problem appears to be reproducible in current versions (but not in 4.10.4 or 4.9), I don't think it is.

**What I did:**

Started the `ssh-askpass` (aka `x11-ssh-askpass`) program.

**What I saw:**

The floating ssh-askpass window is very small, with a size of something like 0x0 or 1x1 or similar.  If it is inhibited from grabbing the keyboard, then it can be resized to reveal the actual window contents.

Problem first occurs in 4.11 (from Ubuntu 16.04 repos), but also observed in manually compiled 4.12 and master.

root-20160903-224759

root-20160903-224807

root-20160903-224823

root-20160903-224829

**What I expected instead:**

As observed in earlier versions, eg. 4.10.4 or 4.9.1/4.9, the window should appear with the correct size.

This suggests a problem with window hint handling, or similar, in the affected versions.  I've haven't yet tried to bisect the commits between 4.10 and 4.11, because my hope is that you will have a better idea of where in the codebase to go looking.

root-20160903-224542

Airblader:
This commit introduces proper support for the minimum size on floating
windows by ensuring that it is respected during mapping, later changes as
well as resizes.

Furthermore, this commit fixes minor issues with how the hints are handled
during calculations.

fixes #2436

### DIFF
--- a/src/floating.c
+++ b/src/floating.c
@@ -72,18 +72,29 @@ void floating_check_size(Con *floating_con) {
     Rect floating_sane_max_dimensions;
     Con *focused_con = con_descend_focused(floating_con);
 
-    /* obey size increments */
-    if (focused_con->window != NULL && (focused_con->window->height_increment || focused_con->window->width_increment)) {
-        Rect border_rect = con_border_style_rect(focused_con);
+    Rect border_rect = con_border_style_rect(focused_con);
+    /* We have to do the opposite calculations that render_con() do
+     * to get the exact size we want. */
+    border_rect.width = -border_rect.width;
+    border_rect.width += 2 * focused_con->border_width;
+    border_rect.height = -border_rect.height;
+    border_rect.height += 2 * focused_con->border_width;
+    if (con_border_style(focused_con) == BS_NORMAL) {
+        border_rect.height += render_deco_height();
+    }
 
-        /* We have to do the opposite calculations that render_con() do
-         * to get the exact size we want. */
-        border_rect.width = -border_rect.width;
-        border_rect.width += 2 * focused_con->border_width;
-        border_rect.height = -border_rect.height;
-        border_rect.height += 2 * focused_con->border_width;
-        if (con_border_style(focused_con) == BS_NORMAL)
-            border_rect.height += render_deco_height();
+    if (focused_con->window != NULL) {
+        if (focused_con->window->min_width) {
+            floating_con->rect.width -= border_rect.width;
+            floating_con->rect.width = max(floating_con->rect.width, focused_con->window->min_width);
+            floating_con->rect.width += border_rect.width;
+        }
+
+        if (focused_con->window->min_height) {
+            floating_con->rect.height -= border_rect.height;
+            floating_con->rect.height = max(floating_con->rect.height, focused_con->window->min_height);
+            floating_con->rect.height += border_rect.height;
+        }
 
         if (focused_con->window->height_increment &&
             floating_con->rect.height >= focused_con->window->base_height + border_rect.height) {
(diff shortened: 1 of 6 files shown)


Airblader:
I want to put a bit more thought into this first, but I'll open it for review already – also so I can leave a few explanatory comments.

src/floating.c
```diff
@@ -74,18 +74,29 @@ void floating_check_size(Con *floating_con) {
...
-        /* We have to do the opposite calculations that render_con() do
-         * to get the exact size we want. */
-        border_rect.width = -border_rect.width;
-        border_rect.width += 2 * focused_con->border_width;
-        border_rect.height = -border_rect.height;
-        border_rect.height += 2 * focused_con->border_width;
-        if (con_border_style(focused_con) == BS_NORMAL)
-            border_rect.height += render_deco_height();
+    Rect border_rect = con_border_style_rect(focused_con);
+    /* We have to do the opposite calculations that render_con() do
```
Airblader: I think it's pretty ugly that we need to do this, although awesome seems to do it the same way. Either way I didn't introduce it, so for the sake of the PR I'll leave the reverse calculation in here.

src/floating.c
```diff
@@ -226,13 +252,16 @@ void floating_enable(Con *con, bool automatic) {
...
-    if (con_border_style(con) == BS_NORMAL)
+    if (con_border_style(con) == BS_NORMAL) {
         nc->rect.height += deco_height;
+    }
 
     /* Honor the X11 border */
     nc->rect.height += con->border_width * 2;
     nc->rect.width += con->border_width * 2;
 
+    floating_check_size(nc);
```
Airblader: This needed to be moved down. First we need to decide how big everything is, *then* we check the size.

testcases/t/221-floating-type-hints.t
```diff
@@ -62,10 +62,10 @@ sub open_with_fixed_size {
 
     my $flags = $XCB_ICCCM_SIZE_HINT_P_MIN_SIZE | $XCB_ICCCM_SIZE_HINT_P_MAX_SIZE;
 
-    my $min_width = 55;
-    my $max_width = 55;
-    my $min_height = 77;
-    my $max_height = 77;
+    my $min_width = 150;
```
Airblader: Just to get above the default minimum window size which will be applied in the config used for tests.

mosajjal:
I've patched the code and it doesn't seem to work. Here's a screenshot of my test environmnt. 
screenshot

Airblader:
Once again, this is about minimum size and not about VirtualBox. Please stop mixing unrelated issues.

Airblader:
@stapelberg Given that we just had a release, I think I'd prefer reviewing and getting this merged now. The PR had worked fine for me, but I think we'll need the community to test it on a broader basis.

goofle:
Hi,
Can you guide to create a path that won't respect size hints?
I don't need this behaviour.

Airblader:
Revert the commit locally and compile from source.

goofle:
@Airblader I'm sorry I gotta ask, why this behaviour isn't configurable?

By the way, am I good to go by just `make && sudo make install` after I squashed this commit?
If not please be kind and help me further with building and installing from source.

Airblader:
How to compile i3 is described in the hacking howto you can find on the website, however it may vary slightly on different distros. I'd recommend looking at how the package of your distro gets built and using those instructions.

> I'm sorry I gotta ask, why this behaviour isn't configurable?

It's a specification, and the specification says this is a must. We don't want to make every part of the entire specification configurable, especially if doing so violates it.

goofle:
@Airblader Thanks man. I wasn't able to find the hacking howto.
Maybe I will look at how it gets installed on my machine. But since I'm a total ignorant about the whole thing and I'm using ubuntu, think I'll be good by just following the hacking howto.

> It's a specification, and the specification says this is a must. We don't want to make every part of the entire specification configurable, especially if doing so violates it.

I understand, thanks again.

Issue #1849, opened 2015-08-17
Allow reload to take an optional configuration file path


kneitinger:
I implemented something that helps me out a lot, but running it by everyone before updating the documentation and submitting a pull request:

I modified the `reload` command to take an optional argument which is the configuration file to load.

For example in ~/.i3/config:

    bindsym $mod+x reload ~/.i3/another_config
and in ~/.i3/another_config

    bindsym $mod+x reload ~/.i3/config

The reason I did this was to have drastically different i3 experiences for different use-cases, beyond the scope of modes.  For instance, my general configuration is fine most of the time, but when playing music, I do not need 80% of my defined keybindings and they could better be used for different functions.

I have two questions.  One, of course, is this something people are interested in?  It has been very helpful to me, but I understand if not.  Second, is this better implemented as a modification to `reload` or should it be a new command just called `load`?

stapelberg:
> For instance, my general configuration is fine most of the time, but when playing music, I do not need 80% of my defined keybindings and they could better be used for different functions.

Are i3’s modes (see the last paragraphs of http://i3wm.org/docs/userguide.html#resizingconfig) not sufficient to cover this use-case?

> I have two questions. One, of course, is this something people are interested in? It has been very helpful to me, but I understand if not. Second, is this better implemented as a modification to reload or should it be a new command just called load?

Why can you not just replace your config file and use reload as-is?

kneitinger:
I really dig modes, and they have helped me create many solutions to tasks I wanted, however, I do think that modes are not sufficient for this use-case, mainly because of the problems that occur during nesting of modes.  I want *mostly* the same experience from one config to another, but with some altered keybindings.

If I were to enter a "music performance" mode, but then go into "resize" mode, I would have to place an additional keybind in the music mode for resize.  Then once I enter resize from music, there is no general way for a mode to essentially "pop" off a modes stack, so one must define keybinds in resize to return to both default and music modes in this example.

Regarding replacing the config file: that method is certainly a way to accomplish the same task, so it comes down to the question of which is more preferable.  It would not be difficult to write a little script to facilitate the task of copying a particular config to the proper location, and if this is a very fringe use-case, I get that.  However, if this ability was included, the cleanness of built-in functionality may influence people to experiment with this and come up with interesting uses.

Airblader:
Side note, but instead of copying files around I'd just use a symbolic link.

stapelberg:
[...]

PR #2510, opened 2016-10-18
Allow specifying a different config file when reloading

regnat:
This pr adds an optional "config_file" parameter to the "reload"
command, to reload i3 with a different config file than the one it was
started with.


Airblader:
Please see #1849 where this has been discussed and rejected.

PR #2524, opened 2016-10-31
Fix name of mcedit

jwilk:
It's `mcedit`, not `mc-edit`.

### DIFF
--- a/i3-sensible-editor
+++ b/i3-sensible-editor
@@ -9,7 +9,7 @@
 # mechanism to find the preferred editor
 
 # Hopefully one of these is installed (no flamewars about preference please!):
-for editor in "$VISUAL" "$EDITOR" nano nvim vim vi emacs pico qe mg jed gedit mc-edit; do
+for editor in "$VISUAL" "$EDITOR" nano nvim vim vi emacs pico qe mg jed gedit mcedit; do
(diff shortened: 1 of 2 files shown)


stapelberg:
Could it be that some distributions ship it as mc-edit? Should we have both here?

jwilk:
I doubt `mc-edit` is a thing. There's exactly one hit on [DCS](https://codesearch.debian.net/search?q=%5Cbmc-edit%5Cb) (i3 itself), no hits on [RPM search](http://rpm.pbone.net/); I couldn't find anything relevant on Google either.

@hboetes, it was apparently you who added `mc-edit` to the list of editors (ac3541c28ce44fb9ce992c605d1f6ca00557fd8e).
Was it a typo for `mcedit` or is there actually a distro that ships it under this name?

hboetes:
I can't look it up right now, but the intention was the editor that is
provided by the midnight commander package.

Am 31.10.2016 12:06 schrieb "Jakub Wilk" <notifications@github.com>:

Airblader:
@jwilk @hboetes What's the status here?

hboetes:
Right, the proper name is mcedit indeed.

Airblader:
@stapelberg Then this should be good to go.

PR #2526, opened 2016-10-31
Added build/ to .gitignore.

Airblader:


stapelberg:
This seems like a straight-forward addition at first glance, but consider that the name of the build directory can be user-defined, and actually defaults to the architecture triplet (e.g. `x86_64-pc-linux-gnu` in my case). “build” is merely what I recommended in the commit description, because it seems like a convention.

One could argue that adding the most common build directory names to `.gitignore` is strictly better than not having them, but I’d like to take a second to discuss this in more detail and see if we can’t find a better solution.

In fact, I’m wondering whether we could have `configure` automatically add the name of the build directory to `.git/info/exclude` (only in case there is a `.git` directory next to `configure`, of course)? That way, any build directory name would work, and `.git/info/exclude` is local to the checkout, not to the repository.

Airblader:
I really mostly only care about this for local builds of us developers. For package builds it doesn't matter at all since the checkout is ususally a throw-away checkout the user never gets to see anyway. I just don't want my tree to dirty because of this directory when I work on i3. I don't think there's much benefit in putting a lot of effort into this.

stapelberg:
I’m not talking about package builds, I also had just developers in mind. i3 developers (which is more people than just us core contributors) might just use `./configure` (getting the triplet build directory name) or a custom build directory name entirely.

Airblader:
Yeah, but I don't really see an issue with that. If they don't care about the dirty tree, that's fine. If they do, they can use `build` as a name. I don't think that's too much to ask of an developer. I still don't really think jumping through hoops does anybody any good here. All we need is one (empty) directory we can use for compiling that git won't track.

Airblader:
… of course if somebody later wants to improve this with such a mechanism, that's cool with me. Right now I just want `build` to be excluded until somebody gives us a reason for why that simple solution isn't sufficient in their case. :-)

stapelberg:
Here’s my suggestion (to be added just before the final `echo` in `configure.ac`):

```bash
git_dir=`git rev-parse --git-dir 2>/dev/null`
if test -n "$git_dir"; then
       srcdir=`dirname "$git_dir"`
       exclude_dir=`pwd | sed "s,^$srcdir,,g"`
       if ! grep -q "^$exclude_dir" "$git_dir/info/exclude"; then
               echo "$exclude_dir" >> "$git_dir/info/exclude"
       fi
fi
```

Do you want to incorporate that into your pull request or shall I file a new one?

Airblader:
It's your code entirely, so I think the commit should be authored by you. :-) That said, can we get build/ in there anyway? As a generic meaningful name that I can easily memorize and type across any machine I work with? :-)

stapelberg:
> It's your code entirely, so I think the commit should be authored by you. :-)

done: https://github.com/i3/i3/pull/2543

> That said, can we get build/ in there anyway? As a generic meaningful name that I can easily memorize and type across any machine I work with? :-)

I don’t follow — why is that necessary? With my PR, build directories will already be git-ignored…?

Airblader:
> With my PR, build directories will already be git-ignored…?

Yes, but the name is relatively hard to remember and can differ across machines. As someone using multiple machines I'd be happier to have one consistent name I can rely on (and type easily). I think having `build` in gitignore really doesn't hurt anyone and hence it's a low-cost nice-to-have. The autogenerated autotools ignores have similar things as well.

If you, however, disagree then I don't think it's worth talking about it much more and I'll just hack it into my local checkouts. :-)

stapelberg:
I still don’t follow.

If you want to name your build directory “build”, then go ahead. As soon as you run `cd build && ../configure` (which you need to do anyway in order to build), the directory will be git-ignored.

Am I misunderstanding something?

Airblader:
> Am I misunderstanding something?

No, apparently I was. If that's how it works that is totally fine with me. Thanks!

PR #2530, opened 2016-11-02
Fix two memory leaks.

Airblader:
This still leaves

```
==27513==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 35 byte(s) in 1 object(s) allocated from:
    #0 0x7fbe4b50fe60 in __interceptor_malloc /build/gcc-multilib/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:62
    #1 0x7fbe47db5c37 in _IO_vasprintf (/usr/lib/libc.so.6+0x6fc37)
```

but that doesn't seem to come from us.

### DIFF
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -206,7 +206,7 @@ int main(int argc, char *argv[]) {
     }
 
     if (!payload)
-        payload = "";
+        payload = sstrdup("");
 
     int sockfd = socket(AF_LOCAL, SOCK_STREAM, 0);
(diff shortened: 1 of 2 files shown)


Airblader:
Actually I do think that comes from us, but I'm not sure yet where from exactly. The only relevant `sasprintf` call in `src/main.c` looks fine to me.

Airblader:
OK, I found the last one. It was the unfree'd socket path.

PR #2537, opened 2016-11-04
Update config.keycodes link to proper location in user guide

JoshRosso:
Links to i3's config.keycodes in the User's guide navigates to a 404.
This commit updates the reference to the `/etc/config.keycodes`
file in GitHub.


stapelberg:
Thanks!

PR #2553, opened 2016-11-10
release.sh: add reminder to announce on reddit

stapelberg:


Airblader:
Can we also add a reminder to update the Github milestones? That's what i3bot will rely on.

stapelberg:
> Can we also add a reminder to update the Github milestones? That's what i3bot will rely on.

Sure. Can you send a PR to that effect, describing precisely which steps are necessary?

PR #2556, opened 2016-11-11
Use the DPI setting within the i3bar

mihaicmn:


Airblader:
I like the change, but please make sure the formatting is correct. :-)

mihaicmn:
Is there any procedure that has to be followed to format the code?

Airblader:
You just run `clang-format` on the code, i.e.,

```
clang-format -i $(find . -name "*.[ch]" | tr '\n' ' ')
```

Airblader:
The changes look good to me, but can you please squash them into a single commit so we can keep the git log clean?

Issue #2565, opened 2016-11-13
i3-msg/i3-ipc incorrectly replying to command


koebi:
Output of `i3 --moreversion 2>&- || i3 --version`:

Binary i3 version:  4.12-129-g3a359a0 (2016-10-03, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12-129-g3a359a0 (2016-10-03, branch "next") (pid 2893)
Loaded i3 config: /home/koebi/.i3/config (Last modified: Sun 13 Nov 2016 09:38:41 PM CET, 434 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: /usr/bin/i3

URL to a logfile as per debugging guide:
http://logs.i3wm.org/logs/5641332169113600.bz2

**What I did:**
I tried to switch workspaces using for example something like "i3-msg workspace 1" or the i3-ipc interface issuing the command "workspace 1".

**What I saw:**
i3-msg returned [], if the workspace i tried to switch to was the workspace I was currently on.
i3-msg returned [{"success":true}] otherwise.

For the i3-ipc-interface, I am not too certain regarding what it replies, but seeing i3-msg behave like this and therefore giving an explanation as to why the script using i3-ipc doesn't work as intended.
I therefore assume, that it replies the same as i3-msg.

**What I expected instead:**
As the [i3-ipc-documentation](http://i3wm.org/docs/ipc.html#_command_reply) implies, I expected the same repy in both cases, which I believe should be [{"success":true}].

i3bot:
Sorry, we can only support the latest major version. Please upgrade from 4.12 to 4.13, verify the bug still exists, and re-open this issue.

Airblader:
This is still valid in 4.13, reopening.

mihaicmn:
What should be the return status in case of having the "workspace_auto_back_and_forth" option enabled?

koebi:
I tried this both with the "workspace_auto_back_and_forth"-option enabled and disabled. In both cases, the command does the expected thing, so I think in both cases, the reply should be [{"success": true}].

Airblader:
Closing since the PR has been merged.

PR #2567, opened 2016-11-14
Fix incorrect reply to ipc command

mihaicmn:
Fixes issue #2565


mihaicmn:
There are some failing tests but they seem unrelated to my changes. Could someone please take a look?

mihaicmn:
Thank you Ingo!

Issue #1799, opened 2015-07-17
enhancement: Implement RandR 1.5 support


stapelberg:
See http://cgit.freedesktop.org/xorg/proto/randrproto/commit/?id=4109f29861a7066970a2afd079d0fc2c12828bcb for the spec changes.

We should evaluate whether i3 needs to understand the new concept of “monitors” — it is my understanding that we have to add support for it, otherwise we will misdetect tiled 4K displays (using Multistream Transport = MST), such as the Dell UP2414Q, as being multiple monitors instead of just one monitor.

Unfortunately, AFAICT, not a single graphics driver supports RandR 1.5 yet, so ideally we’d find a different way of testing such a scenario. Maybe contacting Dave Airlie or Keith Packard and asking for test data (in which format?) would help.

acrisci:
Do we need to support both versions?

Airblader:
> therwise we will misdetect tiled 4K displays […] as being multiple monitors instead of just one monitor.

Are you sure? From what I understand it'd be quite the opposite: the user *wants* to split one physical output into multiple monitors but since we don't support the monitors concept, we keep treating it as one (big) screen. Or am I reading the spec wrong?

Either way, I'm not sure why we should support a concept that has no graphics driver support yet. If we build that feature, we have to maintain code that we probably can't verify. We also keep around dead code that no one can use yet for an unknown amount of time. IMHO, we should build this feature once drivers support it at the earliest, ideally when we see people actually using the concept.

stapelberg:
> Do we need to support both versions?

Not entirely sure if that’s what you mean, but we do need to support RandR < 1.5 and RandR ≥ 1.5 :).

> Are you sure? From what I understand it'd be quite the opposite: the user wants to split one physical output into multiple monitors but since we don't support the monitors concept, we keep treating it as one (big) screen. Or am I reading the spec wrong?

I think you’re indeed reading it wrong. Check out http://keithp.com/blogs/MST-monitors/ for more background.

> Either way, I'm not sure why we should support a concept that has no graphics driver support yet. If we build that feature, we have to maintain code that we probably can't verify.

That’s why I mentioned that there might be other ways to verify our understanding of the spec.

> We also keep around dead code that no one can use yet for an unknown amount of time. IMHO, we should build this feature once drivers support it at the earliest, ideally when we see people actually using the concept.

[...]

PR #2573, opened 2016-11-20
tests: add inject_randr15

stapelberg:
This tool is similar to xtrace in usage in that it intercepts traffic to
the X server. The motivating feature for writing the tool is its ability
to inject prepared reply messages instead of the server’s reply. In
this particular case, we’ll inject a RRGetMonitors reply to test i3’s
RandR 1.5 code paths.

The added testcase is a noop for now, but with the code that’s lingering
in the randr15 branch, i3 does actually detect monitors as per the
injected reply:

    2016-11-20 21:10:05 - randr.c:__randr_query_outputs:618 -
      RandR 1.5 available, querying monitors
    2016-11-20 21:10:05 - randr.c:__randr_query_outputs:628 -
      1 RandR monitors found (timestamp 0)
    2016-11-20 21:10:05 - randr.c:__randr_query_outputs:646 -
      name DP3, x 0, y 0, width 3840 px, height 2160 px, width 520 mm,
      height 290 mm, primary 1, automatic 1

This is preparation work for issue #1799

### DIFF
--- a/testcases/inject_randr1.5.c
+++ b/testcases/inject_randr1.5.c
@@ -1,0 +1,441 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3 - an improved dynamic tiling window manager
+ * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
+ *
+ * inject_randr1.5.c: An X11 proxy which interprets RandR 1.5 GetMonitors
+ * requests and overwrites their reply with a custom reply.
+ *
+ * This tool can be refactored as necessary in order to perform the same
+ * purpose for other request types. The RandR 1.5 specific portions of the code
+ * have been marked as such to make such a refactoring easier.
+ *
+ */
+#include "all.h"
+
+#include <ev.h>
+#include <fcntl.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <sys/time.h>
+#include <sys/resource.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <libgen.h>
+
+static void uds_connection_cb(EV_P_ ev_io *w, int revents);
+static void read_client_setup_request_cb(EV_P_ ev_io *w, int revents);
+static void read_server_setup_reply_cb(EV_P_ ev_io *w, int revents);
+static void read_client_x11_packet_cb(EV_P_ ev_io *w, int revents);
+static void read_server_x11_packet_cb(EV_P_ ev_io *w, int revents);
+
+static char *sun_path = NULL;
+
+void cleanup_socket(void) {
+    if (sun_path != NULL) {
+        unlink(sun_path);
+        free(sun_path);
+        sun_path = NULL;
+    }
+}
+
+/* BEGIN RandR 1.5 specific */
+static void *injected_reply = NULL;
+static off_t injected_reply_len = 0;
+/* END RandR 1.5 specific */
+
+#define XCB_PAD(i) (-(i)&3)
+
+struct connstate {
+    /* clientw is a libev watcher for the connection which we accept()ed. */
+    ev_io *clientw;
+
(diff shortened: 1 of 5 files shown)


testcases/inject_randr1.5.c
```diff
@@ -0,0 +1,439 @@
...
+ * Exits the program with an error if the write failed.
+ *
+ */
+static void must_write(int n) {
+    if (n == -1) {
+        err(EXIT_FAILURE, "write()");
+    }
+}
+
+static void uds_connection_cb(EV_P_ ev_io *w, int revents);
```
Airblader: I guess we never talked about this before, specifically, but shouldn't forward declarations generally be at the top of the file?

testcases/inject_randr1.5.c
```diff
@@ -0,0 +1,439 @@
...
+    int displayp;
+    if (xcb_parse_display(getenv("DISPLAY"), &host, &displayp, NULL) == 0) {
+        errx(EXIT_FAILURE, "Could not parse DISPLAY=%s", getenv("DISPLAY"));
+    }
+    free(host);
+
+    struct sockaddr_un addr;
+    memset(&addr, 0, sizeof(addr));
+    addr.sun_family = AF_LOCAL;
+    snprintf(addr.sun_path, sizeof(addr.sun_path), "/tmp/.X11-unix/X%d", displayp);
```
Airblader: Is it guaranteed that it will always be placed under `/tmp`?

testcases/inject_randr1.5.c
```diff
@@ -0,0 +1,439 @@
...
+}
+
+static void read_server_setup_reply_cb(EV_P_ ev_io *w, int revents) {
+    struct connstate *connstate = (struct connstate *)w->data;
+    xcb_setup_failed_t setup_failed;
+    must_read(readall_into(&setup_failed, sizeof(setup_failed), w->fd));
+
+    switch (setup_failed.status) {
+        case 0:
+            errx(EXIT_FAILURE, "error authenticating at the X11 server");
```
Airblader: Do both (all) compilers see `errx` as a terminating statement or will they see it as a fall-through case?

testcases/inject_randr1.5.c
```diff
@@ -0,0 +1,439 @@
...
+    if (fcntl(fd, F_SETFD, FD_CLOEXEC)) {
+        warn("Could not set FD_CLOEXEC");
+    }
+
+    struct sockaddr_un addr;
+    memset(&addr, 0, sizeof(struct sockaddr_un));
+    addr.sun_family = AF_UNIX;
+    int i;
+    for (i = 0; i < 100; i++) {
+        /* XXX: The path to X11 sockets differs on some platforms (e.g. Trusted
```
Airblader: I guess that answers my question above. This makes me wonder whether this injection should be explicitly enabled when running tests so that someone building for one of these platforms (which admittedly would surprise me) doesn't run into errors?

testcases/inject_randr1.5.c
```diff
@@ -0,0 +1,439 @@
...
+        } else {
+            /* Let the user know bind() was successful, so that they know the
+             * error messages can be disregarded. */
+            fprintf(stderr, "Successfuly bound to %s\n", addr.sun_path);
+            sun_path = sstrdup(addr.sun_path);
+            break;
+        }
+    }
+
+    if (i == 99) {
```
Airblader: Relying on magic numbers that differ by one sounds suboptimal to me. Couldn't we use a boolean success flag instead?

testcases/inject_randr1.5.c
stapelberg: See below.

testcases/inject_randr1.5.c
stapelberg: Done.

testcases/inject_randr1.5.c
stapelberg: Done.

PR #2580, opened 2016-11-21
Implement RandR 1.5 support

### LINKED ISSUE #1799: enhancement: Implement RandR 1.5 support

stapelberg:
This comes with the intentionally undocumented --disable-randr15 command
line flag and disable-randr15 configuration directive. We will add
documentation before the release if and only if it turns out that users
actually need to use this flag in their setups. Ideally, nobody would
need to use the flag and everything would just keep working, but it’s
better to be safe than sorry.

fixes #1799

### DIFF
--- a/generate-command-parser.pl
+++ b/generate-command-parser.pl
@@ -65,7 +65,7 @@ for my $line (@raw_lines) {
 my $current_state;
 
 for my $line (@lines) {
-    if (my ($state) = ($line =~ /^state ([A-Z_]+):$/)) {
+    if (my ($state) = ($line =~ /^state ([A-Z0-9_]+):$/)) {
         #say "got a new state: $state";
         $current_state = $state;
     } else {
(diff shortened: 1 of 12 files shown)


stapelberg:
Tested this with my MST monitor at work.

When changing displays, i3bar doesn’t update correctly for some reason which I haven’t had time to debug yet. The symptom is that the i3bar instances for DP1-9 and DP1-9 end up on eDP1. When restarting i3 (and hence i3bar), the problem disappears.

I think we can fix that later, and gather some experience with RandR 1.5 now.

PR #2585, opened 2016-11-21
Use the DPI setting within the i3-config-wizard and i3-nagbar

### LINKED ISSUE #2461: i3-config-wizard should allow different/bigger fonts
When trying to setup i3 on a retina display (Dell XPS 13), the 8pt font used in the wizard is dinky and unreadable. It looks like options already exist for alternate fonts, they're just not implemented.

mihaicmn:
Also fixes #2461


Airblader:
I'm both happy to see that reading `Xft.dpi` makes a couple of things that much easier and sad to see that I missed all these opportunities initially (and thus for 4.13). :-)

PR #2586, opened 2016-11-24
Building on macos needs libiconv, add a check for it.

cresh:


Airblader:
What does this library contain that i3 uses?

cresh:
Without `-liconv`, linking on macos fails with
```
Undefined symbols for architecture x86_64:
  "_iconv", referenced from:
      _convert_ucs2_to_utf8 in libi3.a(libi3_a-ucs2_conversion.o)
      _convert_utf8_to_ucs2 in libi3.a(libi3_a-ucs2_conversion.o)
  "_iconv_open", referenced from:
      _convert_ucs2_to_utf8 in libi3.a(libi3_a-ucs2_conversion.o)
      _convert_utf8_to_ucs2 in libi3.a(libi3_a-ucs2_conversion.o)
ld: symbol(s) not found for architecture x86_64
```

Airblader:
Right, thanks. We actually include that header for all systems, so I'm not sure whether doing this only for  MacOS is correct or whether all need it. @stapelberg

stapelberg:
Not all operating systems need it — on Linux, glibc provides iconv. The preceding `AC_SEARCH_LIBS([shm_open], [rt])` call is actually very similar: on some operating systems, linking with `-lrt` is not necessary.

We should try not to add operating system-specific code to configure.ac, as the whole point of the GNU build system is to be portable :).

@cresh Could you remove the conditional please?

Airblader:
@stapelberg The change has been done.
@cresh When you make the changes please drop a comment as we don't get notified by Github for commits, unfortunately.

PR #2591, opened 2016-11-29
`move` syntax clarification

jolange:
Both x and y position need to be specified. The notation `[<px> px] [<px> px]` is misleading, because neither the x- nor the y-coordinate are optional.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1921,7 +1921,8 @@ move <left|right|down|up> [<px> px]
 # Moves the container either to a specific location
 # or to the center of the screen. If 'absolute' is
 # used, it is moved to the center of all outputs.
-move [absolute] position [[<px> px] [<px> px]|center]
+move [absolute] position <pos_x> [px] <pos_y> [px]
+move [absolute] position center
 
 # Moves the container to the current position of the
 # mouse cursor. Only affects floating containers.


docs/userguide
```diff
@@ -1921,7 +1921,7 @@ move <left|right|down|up> [<px> px]
 # Moves the container either to a specific location
 # or to the center of the screen. If 'absolute' is
 # used, it is moved to the center of all outputs.
-move [absolute] position [[<px> px] [<px> px]|center]
+move [absolute] position [<posx> px <posy> px|center]
```
Airblader: That's still not correct, though. For one, this currently suggests that anything after »position« is optional (which is untrue). On the other hand, the »px« arguments very much are optional. It gets a little tricky now since we have no other established way to group things.

I think to keep this correct yet understandable this section should be split to document

    move [absolute] position <pos_x> [px] <pos_y> [px]
    move [absolute] position center

separately. What do you think?

jolange: You are right:
- I think a common notation for this case is using parentheses: `(first option|second option)`. But since this is used in no place within the i3 documentation, making it two lines is probably cleaner.
-  I actually just saw that the `px` are optional when looking at the parser specs to verify that my understanding of the command was correct. I chose to omit that to avoid even more confusion. But your suggested notation is also used e.g. in `tray_padding <px> [px]`, so I think that would be consistent and also well-readable with the two-line notation.
I'll update the PR immediately (squashing once everybody is satisfied).

Airblader:
The change looks good to me, except for squashing.

jolange:
@Airblader done.

PR #2596, opened 2016-12-01
Fix read of uninitialized memory

ybznek:
Previous code was reading whole array, it was "slower" and it
read uninitialized memory


ybznek:
Hi, is this fail really caused by my change?

Issue #2572, opened 2016-11-20
Programmatical switch of floating window focus doesn't move active window to the top


mgoral:
Binary i3 version:  4.13 (2016-11-08) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.13 (2016-11-08) (pid 3741)o abort…)
Loaded i3 config: /home/virgo/.config/i3/config (Last modified: pon, 10 paź 2016, 17:49:45, 3545707 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: x-window/i3

URL to a logfile as per http://i3wm.org/docs/debugging.html:

http://logs.i3wm.org/logs/5729327995944960.bz2

A little explanation:

Main window WM_NAME: Subconvert
First dialog (window A) WM_NAME: Donnie.Darko.LIMITED.DVDRip.DivX.ViTE.srt - Synchronize Subtitles
Second dialog (window B) WM_NAME: dupa.srt - Synchronize Subtitles

**What I did:**

Several programs (e.g. Gimp) can have opened several floating dialog boxes at once and allow switching them e.g. by selecting some position from the menu. That's exactly what I do: open one non-modal dialog window "A", then the second one "B" (which is placed over window "A") and then switch to window "A" by selecting some option in their parent window.

**What I saw:**

This changes active window focus to "A", but window "B" is still on top of floating windows stack.

**What I expected instead:**

I expected to move window "A" over the window "B" as it would be the same behavior as when manually switching their focus (via a keyboard or clicking on them).

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

Airblader:
As I said on reddit, this makes sense to me and we should fix it, IMHO.

PR #2616, opened 2016-12-26
Raise floating window to top when it gets focus

sandsmark:
fixes https://github.com/i3/i3/issues/2572

### DIFF
--- a/src/con.c
+++ b/src/con.c
@@ -206,6 +206,15 @@ void con_focus(Con *con) {
         con_focus(con->parent);
 
     focused = con;
+
+    /* Raise floating window unless it is fullscreen */
+    Con *ws = con_get_workspace(con);
+    Con *fs = (ws ? con_get_fullscreen_con(ws, CF_OUTPUT) : NULL);
+    Con *floatingcon = con_inside_floating(con);
+    if (floatingcon != NULL && fs != con) {
+      floating_raise_con(floatingcon);
+    }
+
     /* We can't blindly reset non-leaf containers since they might have
      * other urgent children. Therefore we only reset leafs and propagate
      * the changes upwards via con_update_parents_urgency() which does proper
@@ -500,7 +509,9 @@ bool con_is_docked(Con *con) {
  *
  */
 Con *con_inside_floating(Con *con) {
-    assert(con != NULL);
+    if (con == NULL)
+        return NULL;
+
     if (con->type == CT_FLOATING_CON)
         return con;
 
--- a/src/click.c
+++ b/src/click.c
@@ -251,15 +251,12 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
         goto done;
     }
 
-    /* 2: focus this con. */
+    /* 2: focus this con, raises if necessary. */
     con_focus(con);
 
-    /* 3: For floating containers, we also want to raise them on click.
-     * We will skip handling events on floating cons in fullscreen mode */
+    /* 3: We will skip handling events on floating cons in fullscreen mode */
     Con *fs = (ws ? con_get_fullscreen_con(ws, CF_OUTPUT) : NULL);
     if (floatingcon != NULL && fs != con) {
-        floating_raise_con(floatingcon);
-
         /* 4: floating_modifier plus left mouse button drags */
         if (mod_pressed && event->detail == XCB_BUTTON_INDEX_1) {
             floating_drag_window(floatingcon, event);
--- a/src/floating.c
+++ b/src/floating.c
@@ -196,7 +196,11 @@ void floating_enable(Con *con, bool automatic) {
     /* We insert nc already, even though its rect is not yet calculated. This
      * is necessary because otherwise the workspace might be empty (and get
      * closed in tree_close_internal()) even though it’s not. */
-    TAILQ_INSERT_TAIL(&(ws->floating_head), nc, floating_windows);
+    if (set_focus) {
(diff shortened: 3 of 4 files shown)


src/con.c
```diff
@@ -217,6 +217,10 @@ void con_focus(Con *con) {
         workspace_update_urgent_flag(con_get_workspace(con));
         ipc_send_window_event("urgent", con);
     }
+
+    if (con->type == CT_FLOATING_CON) {
```
Airblader: I think we should use `con_inside_floating(con)` instead and then call `floating_raise_con` on the container returned from that function (if `!= NULL`). Compare with the call to it in `src/click.c`.

src/con.c
```diff
@@ -217,6 +217,10 @@ void con_focus(Con *con) {
...
+        floating_raise_con(con);
```
Airblader: The click handling code calls `con_focus` (almost) unconditionally, but `floating_raise_con` only based on some fullscreen behavior. With this change we would bypass these checks and now unconditionally call `floating_raise_con` upon clicks, ignoring fullscreen behavior. Have you investigated this? What's your conclusion?

At the very least this now produces duplicated calls.

src/con.c
Airblader: I'd prefer a test for this. Given the question about fullscreen windows I'd also like a test for that specific case.

Airblader:
Thanks for your contribution. See my review comments. As a final remark, please include in the commit message itself that and which issue the commit fixes.

sandsmark:
I'll have to brush up on my perl before writing a test if that is necessary, I haven't written perl in a couple of years.

sandsmark:
The failing test fails even without my patch, but sometimes work if I add in some sleep()s before and after creating the fixed size windows.

sandsmark:
Test was broken by a3013969976c0f2572e6137f8a6b0ed306287b5e

Found it by bisecting.

Airblader:
Thanks for the test fix, but I'd prefer this in a separate PR given that it's unrelated. On a side note, any idea why this is necessary? We do only send 5+7 values which is why I changed it originally. Mhm…

sandsmark:
I was a bit puzzled myself, I'm asking around for people that know perl better than I do.

And I'd rather just keep it in this PR as it is so small and blocks this change, and creating a new branch etc. for a new PR is more work. :-P

sandsmark:
well, the pack() shouldn't take more than what is available, but it seems like it puts in a 0 at the end (changing it to 'C5N*' and adding a 0 works as well). I assume this is the missing win_gravity, but the ordering of the parameters makes no sense to me. Re-ordering to match this makes it completely broken: https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.3

and I agree about moving the checking code to the right place, I'll update my patch.

sandsmark:
ah, never mind about the fullscreen checking. in the click handler it also needs the fullscreen check for moving, resizing, etc., so I'm not sure how it could be consolidated without making the code worse.

Airblader:
Ah, right. I forgot about that. Never mind then. :-)

sandsmark:
Found another issue, where newly created floating windows that don't get focus get stacked on top. I attempted to solve it, but I couldn't get the TAILQ_INSERT_BEFORE() to not crash when trying to put in the new window underneath the `focused` con...

src/con.c
```diff
@@ -500,7 +509,9 @@ bool con_is_docked(Con *con) {
  *
  */
 Con *con_inside_floating(Con *con) {
-    assert(con != NULL);
```
Airblader: This can't be necessary. `con_focus` also asserts `con != NULL`, so the new code path shouldn't be able to call this with `NULL`.

sandsmark: The issue was the recursion, in one of the tests (t/229-cleanup-tmpdir.t) this assert was triggered, apparently one of the parents was NULL.

sandsmark:
The set_focus seems to correctly imply whether the newly created floating window is/gets focused, so the only issue is to insert the new con in the right place in floating_windows (right below the currently focused floating con). So this new patch actually works, the only issue is that unfocused new floating cons always gets pushed to the bottom, not just beneath the currently focused con. Unless I'm misunderstanding you.

FWIW, what I'm trying to fix here is https://git.reviewboard.kde.org/r/129706/. So my first patch here fixed the first version there, but my last patch there doesn't work with this first patch. :-)

Airblader:
@sandsmark I'm sorry this has been hanging for a while now. Let's get back to it! What's the current status on the PR?

sandsmark:
Sorry, been pretty busy with dayjob recently, haven't really had time to work on this. I'll try to pick it up soonish.

Airblader:
And please don't forget to just drop a comment when you've changed it, otherwise I don't know when to look in here! Thanks again!

src/con.c
```diff
@@ -206,6 +206,15 @@ void con_focus(Con *con) {
         con_focus(con->parent);
 
     focused = con;
+
+    /* Raise floating window unless it is fullscreen */
+    Con *ws = con_get_workspace(con);
```
stapelberg: Please move these two lines so that they’re only ran for floating windows. con_focus is invoked quite often, so we shouldn’t make it slower than it needs to be.

src/con.c
stapelberg: It’s worthwhile to fix this in a separate commit. We shouldn’t call con_focus on a container which isn’t underneath a workspace in the hierarchy.

Airblader:
I'm going to close this for now. If you do pick up on the PR again please let us know and we can reopen (or send a new one).

PR #2620, opened 2016-12-31
Free allocated X resource value.

Airblader:
This fixes a little memory leak.

### DIFF
--- a/libi3/dpi.c
+++ b/libi3/dpi.c
@@ -53,6 +53,10 @@ void init_dpi(void) {
     DLOG("Found Xft.dpi = %ld.\n", dpi);
 
 init_dpi_end:
+    if (resource != NULL) {
+        free(resource);
+    }
+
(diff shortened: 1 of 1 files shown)


Airblader:
There seems to be a really bad memory leak in i3bar. I'm leaking dozens of MB in a few minutes and it doesn't seem to end. I'm not convinced it's an i3bar issue, but I'm still looking into it. If somebody could monitor i3bar on their end to see if it's just me that'd be pretty cool.

Airblader:
Not using a pango font I get a constant and low memory usage, opposed to the 0.5 GB I already accumulated in the past twenty minutes or so when using a pango font.

Airblader:
The leak seems to have come to a halt at 444 MB now. Which is still way beyond what i3bar should need. I'm fairly convinced this is a bug in pango or fontconfig, though.

PR #2624, opened 2017-01-04
Rewrite the signal handler dialogs.

### LINKED ISSUE #2422: Crash dialog is absolutely black

```
Binary i3 version:  4.12-85-g012de8c (2016-08-03, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12-85-g012de8c (2016-08-03, branch "next") (pid 21815)
Loaded i3 config: /home/v/.config/i3/config (Last modified: Wed 10 Aug 2016 11:39:53 PM MSK, 44948 seconds ago)

```

No.

**What I did:**

`pkill -11 -x i3`

**What I saw:**

2016-08-11-120829_1600x900_scrot

**What I expected instead:**
Crash dialog is drawn normally.

Airblader:
This commit is a rewrite of the popup dialogs used when i3 crashes. We now
use our draw_util suite and both properly react to EXPOSE events and clean
up the windows when the handler exits.

As a side-effect, this fixes #2422

### DIFF
--- a/src/sighandler.c
+++ b/src/sighandler.c
@@ -3,10 +3,6 @@
  *
  * i3 - an improved dynamic tiling window manager
  * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
- * © 2009 Jan-Erik Rediger
- *
- * sighandler.c: Interactive crash dialog upon SIGSEGV/SIGABRT/SIGFPE (offers
- *               to restart inplace).
  *
  */
 #include "all.h"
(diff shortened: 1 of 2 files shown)


Airblader:
Screenshot:

screenshot-2017-01-04_10-44-28

Airblader:
I've also noticed that with the conditional compilation gone from the draw utils we can remove some function arguments there now. I'll take care of that once this is merged.

src/sighandler.c
```diff
@@ -294,22 +308,30 @@ void handle_signal(int sig, siginfo_t *info, void *data) {
...
         /* Strip off the highest bit (set if the event is generated) */
         int type = (event->response_type & 0x7F);
-        if (type == XCB_KEY_PRESS) {
-            sig_handle_key_press(NULL, conn, (xcb_key_press_event_t *)event);
+        switch (type) {
+            case XCB_KEY_PRESS:
+                sighandler_handle_key_press((xcb_key_press_event_t *)event);
+                break;
+            case XCB_EXPOSE:
+                sighandler_handle_expose();
```
psychon: Minor (unimportant) optimisation: Replace this with
```C
case XCB_EXPOSE:
    if (((xcb_expose_event_t *) event)->count == 0)
        sighandler_handle_expose();
```
(Only handle the last one in a series of expose events; quite obviously this does not make much of a difference)

Airblader: Thanks, I'll update the PR. :-)

stapelberg:
Travis fails because you forgot to clang-format after your latest change :)

psychon:
Sorry for breaking things! :-)

Airblader:
I don't usually clang-format it because I just type it correctly. This is what I get for being hasty about making the change. :-) I'll fix it later.

Airblader:
@stapelberg Done and rebased

PR #2625, opened 2017-01-07
Report errors raised during error log creation

Drahflow:
This PR fixes a null pointer dereference after `fopen` on the errorlog failed (in my case because of wrong permissions).


stapelberg:
Thanks!

PR #2626, opened 2017-01-09
linking vim_like_marks from show_marks documentation

jolange:
Someone on IRC asked what "marks" mean here. Since it's the first occurrence of marks in the userguide, this should be linked IMO.


Airblader:
Thank you!

PR #2628, opened 2017-01-10
Fix build outside repo

nmschulte:
The Autoconf script's logic wasn't accounting for building outside of the working tree w/ respect to version detection; this led to a build error.

`nmschulte@desmas-l:~/builds/i3$ ~/src/i3/configure`
2017_01_09-19_58_40-721x351

`nmschulte@desmas-l:~/builds/i3$ make`
2017_01_09-19_58_48-721x351

This also updates the auto build-dir exclusion logic to only apply if within the git working tree, and uses those semantic CLI options from `git rev-parse`, to improve legibility.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,8 +31,8 @@ AX_EXTEND_SRCDIR
 
 AS_IF([test -d ${srcdir}/.git],
       [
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks!

PR #2629, opened 2017-01-11
fix conditional in configure script

nmschulte:
oops.
2017_01_10-20_48_33-703x161

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,7 +155,7 @@ else
 fi
 
 in_git_worktree=`git rev-parse --is-inside-work-tree 2>/dev/null`
-if "$in_git_worktree" == "true"; then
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -155,7 +155,7 @@ else
 fi
 
 in_git_worktree=`git rev-parse --is-inside-work-tree 2>/dev/null`
-if "$in_git_worktree" == "true"; then
+if [[ "$in_git_worktree" = "true" ]]; then
```
stapelberg: I think we should use `[` and `]` here to retain portability: http://serverfault.com/a/52050

nmschulte: When Autoconf processes this, the configure script ends up with single brackets.

stapelberg: Ah, that makes sense.

PR #2633, opened 2017-01-13
Validate that a binding mode is not defined more than once.

### LINKED ISSUE #2615: Config validation should fail if a mode is defined multiple times
When the same binding mode is defined multiple times i3 will happily load the config. We shoud error out in this case. Just spent a few minutes debugging an issue with a coworker… :-)

Airblader:
While defining the same mode usually wouldn't hurt and, in fact, the old behavior
allows to split the definition of a binding mode into several blocks, this
can lead to user errors where they accidentally define a mode twice and don't understand
why the mode behaves a certain way (this has been observed in real life :-)).

There's no good usecase for splitting a single binding mode into multiple blocks, thus
the new behavior is better.

fixes #2615

### DIFF
--- a/src/config_directives.c
+++ b/src/config_directives.c
@@ -126,6 +126,15 @@ CFGFUN(enter_mode, const char *pango_markup, const char *modename) {
         ELOG("You cannot use the name %s for your mode\n", DEFAULT_BINDING_MODE);
         exit(1);
     }
+
+    struct Mode *mode;
+    SLIST_FOREACH(mode, &modes, modes) {
+        if (strcmp(mode->name, modename) == 0) {
+            ELOG("The binding mode with name \"%s\" is defined at least twice.\n", modename);
+            exit(1);
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Airblader:
@aaschmid :-)

src/config_directives.c
```diff
@@ -126,6 +126,15 @@ CFGFUN(enter_mode, const char *pango_markup, const char *modename) {
         ELOG("You cannot use the name %s for your mode\n", DEFAULT_BINDING_MODE);
         exit(1);
     }
+
+    struct Mode *mode;
+    SLIST_FOREACH(mode, &modes, modes) {
+        if (strcmp(mode->name, modename) == 0) {
+            ELOG("The binding mode with name \"%s\" is defined at least twice.\n", modename);
+            exit(1);
```
stapelberg: `exit` seems a bit harsh, but seeing that we’ve been using `exit` in this code branch for 5 years, I won’t insist on changing it.

stapelberg (approved):

PR #2636, opened 2017-01-13
Only react on the last Expose event in a series of events.

Airblader:
Thanks to @psychon for pointing this out during the review of PR #2624.
This commit extends this change to all other occurences of Expose events
within i3.


psychon (approved):
:-)

stapelberg (approved):

PR #2638, opened 2017-01-13
Don't exit() on redefined binding mode.

Airblader:
Doing a hard exit() is a rather harsh action for something i3 can handle
perfectly fine and is only meant to be a check to make debugging easier
for users in certain situations.


Airblader:
@stapelberg Since I just killed my session when merging it all together in i3-gaps, compiling and restarting i3 (from having multiple modes for testing earlier), I'll agree and softened up the check.

stapelberg:
Could we also remove the `exit` in line 127 while we’re at it?

Airblader:
@stapelberg Done. We should at least `return`, though, so the user cannot overwrite the default mode.

Issue #2642, opened 2017-01-15
Migrate all tools to draw_util


Airblader:
We should switch all tools to use `libi3/draw_util.c` so we have all drawing abstracted.

- [x] i3-config-wizard
- [x] i3-input
- [x] i3-nagbar
- [x] placeholder windows

i3 4.13

mihaicmn:
@Airblader are you planning to implement this enhancement? Otherwise I would be glad to work on it.

Airblader:
@mihaicmn Go ahead! :-) But I'd propose doing one PR for each of the bullet points above separately.

Airblader:
Closing as all PRs have been merged.

@mihaicmn Thanks!

PR #2645, opened 2017-01-17
Migrate i3-input to draw_util

mihaicmn:
See #2642 for details.


stapelberg:
@Airblader Could you take a look at this please? I think you’re more familiar with draw_util. Thanks!

PR #2649, opened 2017-01-20
layout toggle: take any combination of layouts as arguments (continuation of #2476)

s3rb31:
Okay so here is the new PR on the `layout toggle` thing.

I have added two commits. The first one condenses the code and makes it much less redundant as the  cycle algorithm can handle the `split` part without much modification. The second one updates the documentation.

As @Airblader mentioned in the foregoing discussion (#2476) all of this might stay proof of concept, until there is a proper way to pass arrays of arguments to functions and validate them via the parser specification.

So now, some feedback! :P I am curious to see what others think about the code and about some ideas on how to implement something that makes it possible to define variadic functions in the parser specification.

Greetings

Edit: I just forgot to delete the obsolete specification as suggested by Airblader. Fixed now.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1802,7 +1802,8 @@ The +toggle+ option will toggle the orientation of the split container if it
 contains a single window. Otherwise it makes the current window a split
 container with opposite orientation compared to the parent container.
 Use +layout toggle split+ to change the layout of any split container from
-splitv to splith or vice-versa.
+splitv to splith or vice-versa. You can also define a custom sequence of layouts
+to cycle through with +layout toggle+, see <<manipulating_layout>>.
 
 *Syntax*:
 --------------------------------
@@ -1822,6 +1823,11 @@ Use +layout toggle split+, +layout stacking+, +layout tabbed+, +layout splitv+
 or +layout splith+ to change the current container layout to splith/splitv,
 stacking, tabbed layout, splitv or splith, respectively.
 
+Specify up to four layouts after +layout toggle+ to cycle through them. Every
+time the command is executed, the layout specified after the currently active
+one will be applied. If the currently active layout is not in the list, the
+first layout in the list will be activated.
+
 To make the current window (!) fullscreen, use +fullscreen enable+ (or
 +fullscreen enable global+ for the global mode), to leave either fullscreen
 mode use +fullscreen disable+, and to toggle between these two states use
@@ -1834,6 +1840,7 @@ enable+ respectively +floating disable+ (or +floating toggle+):
 --------------------------------------------
 layout default|tabbed|stacking|splitv|splith
 layout toggle [split|all]
+layout toggle [split|tabbed|stacking|splitv|splith] [split|tabbed|stacking|splitv|splith]…
 --------------------------------------------
 
 *Examples*:
@@ -1847,6 +1854,15 @@ bindsym $mod+x layout toggle
 
 # Toggle between stacking/tabbed/splith/splitv:
 bindsym $mod+x layout toggle all
+
+# Toggle between stacking/tabbed/splith:
+bindsym $mod+x layout toggle stacking tabbed splith
+
+# Toggle between splitv/tabbed
+bindsym $mod+x layout toggle splitv tabbed
+
+# Toggle between last split layout/tabbed/stacking
+bindsym $mod+x layout toggle split tabbed stacking
 
 # Toggle fullscreen
 bindsym $mod+f fullscreen toggle
--- a/src/con.c
+++ b/src/con.c
@@ -1719,28 +1719,64 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
         parent = con->parent;
     DLOG("con_toggle_layout(%p, %s), parent = %p\n", con, toggle_mode, parent);
 
-    if (strcmp(toggle_mode, "split") == 0) {
-        /* Toggle between splits. When the current layout is not a split
-         * layout, we just switch back to last_split_layout. Otherwise, we
-         * change to the opposite split layout. */
-        if (parent->layout != L_SPLITH && parent->layout != L_SPLITV)
-            con_set_layout(con, parent->last_split_layout);
-        else {
-            if (parent->layout == L_SPLITH)
-                con_set_layout(con, L_SPLITV);
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";
+
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        /* L_DEFAULT is used as a placeholder value to distinguish if
+         * the first layout has already been saved. (it can never be L_DEFAULT) */
+        layout_t new_layout = L_DEFAULT;
+        bool current_layout_found = false;
+        char *tm_dup = sstrdup(toggle_mode);
+        char *cur_tok = strtok(tm_dup, delim);
+
+        for (layout_t layout; cur_tok != NULL; cur_tok = strtok(NULL, delim)) {
+            if (strcasecmp(cur_tok, "split") == 0) {
+                /* Toggle between splits. When the current layout is not a split
+                 * layout, we just switch back to last_split_layout. Otherwise, we
+                 * change to the opposite split layout. */
+                if (parent->layout != L_SPLITH && parent->layout != L_SPLITV) {
+                    layout = parent->last_split_layout;
+                } else {
+                    layout = (parent->layout == L_SPLITH) ? L_SPLITV : L_SPLITH;
+                }
+            } else {
+                bool success = layout_from_name(cur_tok, &layout);
+                if (!success || layout == L_DEFAULT) {
+                    ELOG("The token '%s' was not recognized and has been skipped.\n", cur_tok);
+                    continue;
+                }
+            }
+
+            /* If none of the specified layouts match the current,
+             * fall back to the first layout in the list */
+            if (new_layout == L_DEFAULT) {
+                new_layout = layout;
+            }
+
+            /* We found the active layout in the last iteration, so
+             * now let's activate the current layout (next in list) */
+            if (current_layout_found) {
+                new_layout = layout;
+                free(tm_dup);
+                break;
+            }
+
+            if (parent->layout == layout) {
+                current_layout_found = true;
+            }
         }
-    } else {
+
+        con_set_layout(con, new_layout);
+    } else if (strcasecmp(toggle_mode, "all") == 0 || strcasecmp(toggle_mode, "default") == 0) {
         if (parent->layout == L_STACKED)
             con_set_layout(con, L_TABBED);
         else if (parent->layout == L_TABBED) {
-            if (strcmp(toggle_mode, "all") == 0)
+            if (strcasecmp(toggle_mode, "all") == 0)
                 con_set_layout(con, L_SPLITH);
             else
                 con_set_layout(con, parent->last_split_layout);
         } else if (parent->layout == L_SPLITH || parent->layout == L_SPLITV) {
-            if (strcmp(toggle_mode, "all") == 0) {
+            if (strcasecmp(toggle_mode, "all") == 0) {
                 /* When toggling through all modes, we toggle between
                  * splith/splitv, whereas normally we just directly jump to
                  * stacked. */
--- a/src/util.c
+++ b/src/util.c
@@ -66,7 +66,7 @@ __attribute__((pure)) bool name_is_digits(const char *name) {
     return true;
 }
 
-/**
+/*
  * Set 'out' to the layout_t value for the given layout. The function
  * returns true on success or false if the passed string is not a valid
  * layout name.
@@ -75,20 +75,23 @@ __attribute__((pure)) bool name_is_digits(const char *name) {
 bool layout_from_name(const char *layout_str, layout_t *out) {
     if (strcmp(layout_str, "default") == 0) {
         *out = L_DEFAULT;
+        return true;
     } else if (strcasecmp(layout_str, "stacked") == 0 ||
                strcasecmp(layout_str, "stacking") == 0) {
         *out = L_STACKED;
+        return true;
     } else if (strcasecmp(layout_str, "tabbed") == 0) {
         *out = L_TABBED;
+        return true;
     } else if (strcasecmp(layout_str, "splitv") == 0) {
         *out = L_SPLITV;
+        return true;
     } else if (strcasecmp(layout_str, "splith") == 0) {
         *out = L_SPLITH;
-    } else {
-        return false;
+        return true;
     }
 
-    return true;
+    return false;
 }
 
 /*
--- a/src/util.c
+++ b/src/util.c
@@ -66,6 +66,31 @@ __attribute__((pure)) bool name_is_digits(const char *name) {
     return true;
 }
 
+/**
+ * Set 'out' to the layout_t value for the given layout. The function
+ * returns true on success or false if the passed string is not a valid
+ * layout name.
+ *
+ */
+bool layout_from_name(const char *layout_str, layout_t *out) {
+    if (strcmp(layout_str, "default") == 0) {
+        *out = L_DEFAULT;
+    } else if (strcasecmp(layout_str, "stacked") == 0 ||
+               strcasecmp(layout_str, "stacking") == 0) {
+        *out = L_STACKED;
+    } else if (strcasecmp(layout_str, "tabbed") == 0) {
+        *out = L_TABBED;
+    } else if (strcasecmp(layout_str, "splitv") == 0) {
+        *out = L_SPLITV;
+    } else if (strcasecmp(layout_str, "splith") == 0) {
+        *out = L_SPLITH;
+    } else {
+        return false;
+    }
+
+    return true;
+}
+
 /*
  * Parses the workspace name as a number. Returns -1 if the workspace should be
  * interpreted as a "named workspace".
--- a/testcases/t/192-layout.t
+++ b/testcases/t/192-layout.t
@@ -95,4 +95,72 @@ cmd 'layout toggle all';
 ($nodes, $focus) = get_ws_content($tmp);
 is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
 
+cmd 'layout toggle splith splitv';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splith', 'layout now splith');
+
+cmd 'layout toggle splith splitv';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
+
+cmd 'layout toggle stacked splitv tabbed';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'tabbed', 'layout now tabbed');
+
+cmd 'layout toggle stacking splitv tabbed';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'stacked', 'layout now stacked');
+
+cmd 'layout toggle stacking splitv tabbed';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
+
+cmd 'layout toggle splitv i stacking tabbed';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'stacked', 'layout now stacked');
+
+cmd 'layout toggle stacked';
+($nodes, $focus) = get_ws_content($tmp);
+# this is correct if it does nothing
+is($nodes->[1]->{layout}, 'stacked', 'layout now tabbed');
+
+cmd 'layout toggle tabbed stacked';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'tabbed', 'layout now stacked');
+
+# obsoletes 'split' ;)
+cmd 'layout toggle splith splitv';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splith', 'layout now splith');
+
+# nonsense but works expectedly
+cmd 'layout toggle split split';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
+
+cmd 'layout toggle split split';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splith', 'layout now splith');
+
+# testing with arbitrary length and garbage
+cmd 'layout toggle stacking splith tabbed splitv stacking';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'tabbed', 'layout now tabbed');
+
+cmd 'layout toggle stacking splith garbage tabbed splitv stacking';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
+
+cmd 'layout toggle stacking splith garbage tabbed splitv stacking';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'stacked', 'layout now stacked');
+
+cmd 'layout toggle splitv splith garbage splitv tabbed stacking splitv';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
+
+cmd 'layout toggle splitv garbage    tabbed';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'tabbed', 'layout now tabbed');
+
 done_testing;
--- a/include/util.h
+++ b/include/util.h
@@ -70,6 +70,14 @@ Rect rect_sub(Rect a, Rect b);
 __attribute__((pure)) bool name_is_digits(const char *name);
 
 /**
+ * Set 'out' to the layout_t value for the given layout. The function
+ * returns true on success or false if the passed string is not a valid
+ * layout name.
+ *
+ */
+bool layout_from_name(const char *layout_str, layout_t *out);
+
+/**
  * Parses the workspace name as a number. Returns -1 if the workspace should be
  * interpreted as a "named workspace".
  *
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -110,7 +110,7 @@ state LAYOUT:
 state LAYOUT_TOGGLE:
   end
       -> call cmd_layout_toggle($toggle_mode)
-  toggle_mode = 'split', 'all'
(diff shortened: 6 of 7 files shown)


docs/userguide
```diff
@@ -1833,7 +1838,7 @@ enable+ respectively +floating disable+ (or +floating toggle+):
 *Syntax*:
 --------------------------------------------
 layout default|tabbed|stacking|splitv|splith
-layout toggle [split|all]
+layout toggle [default|all] OR [split|tabbed|stacking|splitv|splith] [ up to 4 times ... ]
```
Airblader: My proposal would be `[split|tabbed|stacking|splitv|splith]+` which is how arguments occuring once or more are usually handled. Any reason we're limited to 4? That seems like an arbitrary restriction we should get rid of.

src/con.c
```diff
@@ -1719,19 +1719,54 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // space is used as delimiter
+
+    if (strcmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        char *tm_dup = sstrdup(toggle_mode);
+        char *cur_tok = strtok(tm_dup, delim);
+
+        int index = 0;
+        layout_t *layouts = (layout_t *)smalloc(sizeof(layout_t) * 4);
```
Airblader: Please don't cast allocations.

src/con.c
```diff
@@ -1719,19 +1719,54 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+        while (cur_tok != NULL) {
+            if (strcmp(cur_tok, "stacked") == 0 ||
```
Airblader: These two conditions can easily fit on the same line.

src/con.c
```diff
@@ -1719,19 +1719,54 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+                strcmp(cur_tok, "stacking") == 0)
+                layouts[index++] = L_STACKED;
+            else if (strcmp(cur_tok, "tabbed") == 0)
```
Airblader: Please use braces (`{}`) everywhere.

src/con.c
```diff
@@ -1719,19 +1719,54 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+                /* Toggle between splits. When the current layout is not a split
+                 * layout, we just switch back to last_split_layout. Otherwise, we
+                 * change to the opposite split layout. */
+                if (parent->layout != L_SPLITH && parent->layout != L_SPLITV)
+                    layouts[index++] = parent->last_split_layout;
+                else
+                    layouts[index++] = (parent->layout == L_SPLITH) ? L_SPLITV : L_SPLITH;
+            }
+
+            if (index == 3)
```
Airblader: It seems to me like currently the counting is off here. If index is 3 at this point we haven't actually written the fourth element yet, thus allowing one value too few. But as I said above, let's get rid of this restriction anyway.

docs/userguide
s3rb31: It is the only thing that makes sense at all. We do not have more than four different layouts to cycle through. Or am I missing something?!

src/con.c
s3rb31: True, thats off by one, I will fix this. I just increased the number to four because that would make more sense (read above) and obviously forgot it here.

docs/userguide
Airblader: Yes, but we still shouldn't have hardcoded restrictions like this in certain code paths.

src/con.c
s3rb31: Will do it, but why? There are many places which do not use braces in the code. Is it just for safety because here there are many `if`s below each other?

Airblader: We want to get it to a consistent style over time.

Airblader:
Also we should log an error if the string contains a token we don't recognize.

Airblader:
@s3rb31 I see you pushed more commits. When you made changes please leave a comment. Unfortunately we don't get notifications for pushed commits on PRs. :-)

Airblader:
@stapelberg I think you can review this now.
@s3rb31 One thing left to do is to squash the commits into a nice feature commit with a proper commit message. You may want to wait for @stapelberg's review, though.

s3rb31:
> @s3rb31 I see you pushed more commits. When you made changes please leave a comment. Unfortunately we don't get notifications for pushed commits on PRs. :-)

Sure. I didn't know there're no notifications for new commits ...

> @s3rb31 One thing left to do is to squash the commits into a nice feature commit with a proper commit message. You may want to wait for @stapelberg's review, though.

No problem, do you have an example commit, so that I can see what a proper commit message looks like? I am not very experienced with git, so to squash them together I'd do a soft reset with the last N commits and the recommit and push them? Will that remove all the other commits?

Airblader:
> do you have an example commit, so that I can see what a proper commit message looks like?

It doesn't need to be too fancy. Just a quick explanation of what it's about and what issue it fixes. Here's an example:

````
commit 04dcf42397a7cf0e1284a2c6b16fb40c15af1f3e
Author: Tony Crisci <tony@dubstepdish.com>
Date:   Sun Jan 22 17:08:32 2017 -0500

    Add the ipc shutdown event (#2652)
    
    This event is triggered when the connection to the ipc is about to
    shutdown because of a user action such as with a `restart` or `exit`
    command. The `change` field indicates why the ipc is shutting down. It
    can be either "restart" or "exit".
    
    fixes #2318
````

> so to squash them together I'd do a soft reset with the last N commits and then recommit and push them?

You can just do `git rebase -i HEAD~7` (since you have 7 commits). That'll let you edit the commits which are listed one per line (they'll be in »chronological« order). You can replace the »pick« with a »squash« (or just »s«) for a commit to squash it into the previous commit. 

So you'll want to do that for all of your PR's commits except the first one (you don't want to squash that into someone else's commit). Just save that file (depending on what editor you use, typically it's nano so you save via Ctrl+X) and git will do the squashing. 

After that you need to force-push the branch (but make sure with `git log -p` first, just to be sure you did it right). Here's also a short tutorial on it: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

Of course a soft reset and committing again is also an option, but might as well improve your git-fu. :-)

**Comment by s3rb31** [comment]
Okay, I think I figured it out. I'll be sitting tight for a review and then recommit. ;)

File: docs/userguide [hunk 6]
```diff
@@ -1833,7 +1838,7 @@ enable+ respectively +floating disable+ (or +floating toggle+):
 *Syntax*:
 --------------------------------------------
 layout default|tabbed|stacking|splitv|splith
-layout toggle [split|all]
+layout toggle [all] OR [split|tabbed|stacking|splitv|splith]+
```
stapelberg: It seems like this would be clearer without the “OR” in the middle. Also, we should denote the repetition in some way. How about this?

```
layout toggle [split|all]
layout toggle [split|tabbed|stacking|splitv|splith] [split|tabbed|stacking|splitv|splith]…
```

(Filed https://github.com/i3/i3/issues/2673 about us not using a defined notation technique)

docs/userguide
```diff
@@ -1822,6 +1823,10 @@ Use +layout toggle split+, +layout stacking+, +layout tabbed+, +layout splitv+
 or +layout splith+ to change the current container layout to splith/splitv,
 stacking, tabbed layout, splitv or splith, respectively.
 
+Specify up to four arguments after +layout toggle+ to cycle through them. Every
```
stapelberg: s/arguments/layouts/

docs/userguide
```diff
@@ -1822,6 +1823,10 @@ Use +layout toggle split+, +layout stacking+, +layout tabbed+, +layout splitv+
...
+time the command is executed, the layout specified after the currently active one
```
stapelberg: What happens in case the currently active one is not in the list?

src/con.c
```diff
@@ -1719,19 +1719,52 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-        if (parent->layout != L_SPLITH && parent->layout != L_SPLITV)
-            con_set_layout(con, parent->last_split_layout);
-        else {
-            if (parent->layout == L_SPLITH)
-                con_set_layout(con, L_SPLITV);
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // space is used as delimiter
+
+    if (strcmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
```
stapelberg: Since you changed the parser state `LAYOUT_TOGGLE` from a list of constant tokens to a string, please use `strcasecmp` (here and below) to be forgiving with regards to lower/upper-case spelling.

s3rb31:
Okay, I adressed the review. Should I wait some more time with squashing or is it okay like this?

src/con.c
```diff
@@ -1719,28 +1719,61 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // space is used as delimiter
+
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        char *tm_dup = sstrdup(toggle_mode);
+        char *cur_tok = strtok(tm_dup, delim);
+
+        int index = 0, size = 4;
+        layout_t *layouts = smalloc(sizeof(layout_t) * size);
```
stapelberg: Why do you need to allocate memory for holding the layouts at all? You could also do the comparison with `parent->layout` while you iterate over the layout tokens and use a boolean for whether the next value should be applied or not, right?

stapelberg:
> Should I wait some more time with squashing or is it okay like this?

The docs look good. It makes sense to start squashing now, even if we’re still changing the code.

Thanks!

src/con.c
s3rb31: Yes, that seems to be right, very good thoughts. Let me think a little bit about how to do it the most elegant way.

s3rb31:
Hey everyone,

I've redone the layout comparision without memory allocation, please have a look! I think the code looks much cleaner now and should be less error prone.

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-         * layout, we just switch back to last_split_layout. Otherwise, we
-         * change to the opposite split layout. */
-        if (parent->layout != L_SPLITH && parent->layout != L_SPLITV)
-            con_set_layout(con, parent->last_split_layout);
-        else {
-            if (parent->layout == L_SPLITH)
-                con_set_layout(con, L_SPLITV);
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // space is used as delimiter
```
Airblader: This comment doesn't add any useful information.

docs/userguide
```diff
@@ -1834,6 +1840,7 @@ enable+ respectively +floating disable+ (or +floating toggle+):
 --------------------------------------------
 layout default|tabbed|stacking|splitv|splith
 layout toggle [split|all]
+layout toggle [split|tabbed|stacking|splitv|splith] [split|tabbed|stacking|splitv|splith]…
```
Airblader: Can we use this instead?

````
layout toggle [split|tabbed|stacking|splitv|splith]+
````

File: src/con.c [hunk 13]
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        layout_t new_layout = L_DEFAULT;
+        bool set_next_layout = false;
+        char *tm_dup = sstrdup(toggle_mode);
+        char *cur_tok = strtok(tm_dup, delim);
+
+        for (layout_t layout = L_DEFAULT; cur_tok != NULL; cur_tok = strtok(NULL, delim)) {
+            // parse the specific layouts from the passed string (toggle_mode)
+            if (strcasecmp(cur_tok, "stacked") == 0 || strcasecmp(cur_tok, "stacking") == 0) {
```
Airblader: The logic to map strings to layouts now exists twice in the code (the other one is in `src/commands.c`). I'd prefer extracting this into a function (e.g., `con_layout_from_name` in `src/con.c`) used by both places. This keeps it consistent, avoids duplication and makes this block of code more concise.

Only the `split` case will need to be handled only here.

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+                layout = L_TABBED;
+            } else if (strcasecmp(cur_tok, "splitv") == 0) {
+                layout = L_SPLITV;
+            } else if (strcasecmp(cur_tok, "splith") == 0) {
+                layout = L_SPLITH;
+            } else if (strcasecmp(cur_tok, "split") == 0) {
+                /* Toggle between splits. When the current layout is not a split
+                 * layout, we just switch back to last_split_layout. Otherwise, we
+                 * change to the opposite split layout. */
+                if (parent->layout != L_SPLITH && parent->layout != L_SPLITV)
```
Airblader: Please use braces everywhere.

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+                    layout = parent->last_split_layout;
+                else
+                    layout = (parent->layout == L_SPLITH) ? L_SPLITV : L_SPLITH;
+            } else {
+                ELOG("The token '%s' was not recognized and has been skipped.\n", cur_tok);
+                continue;
+            }
+
+            // save the leftmost layout
+            if (new_layout == L_DEFAULT)
```
Airblader: Braces here as well. :-)

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-            con_set_layout(con, parent->last_split_layout);
-        else {
-            if (parent->layout == L_SPLITH)
-                con_set_layout(con, L_SPLITV);
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // space is used as delimiter
+
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        layout_t new_layout = L_DEFAULT;
```
Airblader: I think this one deserves a comment explaining that `L_DEFAULT` is used as a placeholder since it cannot be actually used by the user.

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+            } else {
+                ELOG("The token '%s' was not recognized and has been skipped.\n", cur_tok);
+                continue;
+            }
+
+            // save the leftmost layout
+            if (new_layout == L_DEFAULT)
+                new_layout = layout;
+
+            // if told to, activate the parsed layout
```
Airblader: This comment sounds a little odd to me. We're telling this ourselves here. I'd prefer a comment explaining what's going on, e.g., »We found the active layout in the last iteration, so activate this one now.«

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+            if (set_next_layout) {
+                new_layout = layout;
+                break;
+            }
+
+            // current layout is active,
```
Airblader: Please use `/**/` for multi-line comments. In fact I think this code would speak for itself, especially given the comment right above for where we use the flag, so I don't think we'd need it at all.

And use braces. :-)

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-        else {
-            if (parent->layout == L_SPLITH)
-                con_set_layout(con, L_SPLITV);
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // space is used as delimiter
+
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        layout_t new_layout = L_DEFAULT;
+        bool set_next_layout = false;
```
Airblader: I'd prefer the name `found_current_layout`. What do you think?

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+
+            // current layout is active,
+            // remember to apply the next one
+            if (parent->layout == layout)
+                set_next_layout = true;
         }
-    } else {
+
+        con_set_layout(con, new_layout);
+        free(tm_dup);
```
Airblader: Can't we already free this much earlier? We should free memory as soon as we don't need it anymore.

Also, we have a `FREE` macro which I think we should try to use in all places.

testcases/t/192-layout.t
```diff
@@ -95,4 +95,68 @@ cmd 'layout toggle all';
...
+is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
+
+cmd 'layout toggle stacking splith garbage tabbed splitv stacking';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'stacked', 'layout now stacked');
+
+cmd 'layout toggle splitv splith garbage splitv tabbed stacking splitv';
+($nodes, $focus) = get_ws_content($tmp);
+is($nodes->[1]->{layout}, 'splitv', 'layout now splitv');
+
```
Airblader: Can we add a testcase for multiple consecutive spaces? Does this even work currently without throwing an error message?

docs/userguide
```diff
@@ -1822,6 +1823,11 @@ Use +layout toggle split+, +layout stacking+, +layout tabbed+, +layout splitv+
 or +layout splith+ to change the current container layout to splith/splitv,
 stacking, tabbed layout, splitv or splith, respectively.
 
+Specify up to four layouts after +layout toggle+ to cycle through them. Every
+time the command is executed, the layout specified after the currently active
+one will be applied. If the currently active layout is not in the list, the
+leftmost layout will be activated.
```
Airblader: s/leftmost layout/first layout in the list/

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+    const char delim[] = " ";  // space is used as delimiter
+
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        layout_t new_layout = L_DEFAULT;
+        bool set_next_layout = false;
+        char *tm_dup = sstrdup(toggle_mode);
+        char *cur_tok = strtok(tm_dup, delim);
+
+        for (layout_t layout = L_DEFAULT; cur_tok != NULL; cur_tok = strtok(NULL, delim)) {
+            // parse the specific layouts from the passed string (toggle_mode)
```
Airblader: Don't think we'll need this comment anymore with the changes below.

src/con.c
```diff
@@ -1719,28 +1719,65 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+                if (parent->layout != L_SPLITH && parent->layout != L_SPLITV)
+                    layout = parent->last_split_layout;
+                else
+                    layout = (parent->layout == L_SPLITH) ? L_SPLITV : L_SPLITH;
+            } else {
+                ELOG("The token '%s' was not recognized and has been skipped.\n", cur_tok);
+                continue;
+            }
+
+            // save the leftmost layout
```
Airblader: I'd prefer a comment like »None of the specified layouts match the current layout, so fall back to using the first layout specified in the list.«

docs/userguide
s3rb31: Well, I really don't care about this one, but I think the general discussion about what to use, should be held at #2673. @stapelberg told me to change it to what we have it now, so I really don't know what to do now ...

testcases/t/192-layout.t
s3rb31: I will add such testcases. But there are no further changes needed because `strtok` skips zero-length matches. So it currently doesn't throw an error. It will not even be logged by the ELOG.

docs/userguide
Airblader: Ah, okay. Let's leave it like that then.

docs/userguide
s3rb31: Why?! :P I think the wording is quite elgant! Also `leftmost` is common terminology within computing. ;)

Airblader: You generally don't refer to left(most) and right(most) elements in a list, but to first and last elements.

src/con.c
s3rb31: Well, we could do `free(tm_dup)` before `con_set_layout` or within the break-if (ll.1758) in the for-loop but not earlier. It is needed by the `strtok` call in the for-loop.

src/con.c
s3rb31: I'll extend the comment. I just want to make clear why we are defining a single character string constant here.

src/con.c
s3rb31: Yes, that might clarify things. I will add a comment.

src/con.c
s3rb31: I am okay with that.

src/con.c
s3rb31: Yes you're right. But wouldn't that be something for an additional PR?

s3rb31:
Okay, I fixed most of the things. I also added a testcase for multiple delimiters following each other. Please have a look again.

src/con.c
```diff
@@ -1719,28 +1719,69 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-            con_set_layout(con, parent->last_split_layout);
-        else {
-            if (parent->layout == L_SPLITH)
-                con_set_layout(con, L_SPLITV);
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // Space is used as delimiter to split toggle_mode
+
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        /* L_DEFAULT is used as a placeholder value to destinct if the
```
Airblader: That's not a proper sentence because »distinct« is not a verb. Much like the other places let's please use »first« instead of »leftmost«.

src/con.c
```diff
@@ -1719,28 +1719,69 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
-         * layout, we just switch back to last_split_layout. Otherwise, we
-         * change to the opposite split layout. */
-        if (parent->layout != L_SPLITH && parent->layout != L_SPLITV)
-            con_set_layout(con, parent->last_split_layout);
-        else {
-            if (parent->layout == L_SPLITH)
-                con_set_layout(con, L_SPLITV);
-            else
-                con_set_layout(con, L_SPLITH);
+    const char delim[] = " ";  // Space is used as delimiter to split toggle_mode
```
Airblader: I still prefer not having this comment here, but if you insist on it at least please use a consistent style. We don't really use same-line comments, but comments above the line. But again, my preference is to remove it because I think the meaning of this constant is perfectly clear.

src/con.c
```diff
@@ -1719,28 +1719,69 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
...
+    if (strcasecmp(toggle_mode, "split") == 0 || strstr(toggle_mode, delim)) {
+        /* L_DEFAULT is used as a placeholder value to destinct if the
+         * leftmost layout has already been saved. (it can never be L_DEFAULT) */
+        layout_t new_layout = L_DEFAULT;
+        bool current_layout_found = false;
+        char *tm_dup = sstrdup(toggle_mode);
+        char *cur_tok = strtok(tm_dup, delim);
+
+        for (layout_t layout = L_DEFAULT; cur_tok != NULL; cur_tok = strtok(NULL, delim)) {
+            if (strcasecmp(cur_tok, "stacked") == 0 || strcasecmp(cur_tok, "stacking") == 0) {
```
Airblader: My comment on refactoring the duplication still stands.

s3rb31:
Okay, I've implemented a function to map from the layout name to the corresponding layout_t value. I decided to put it in `util.c` because it seemed to fit in there better. If you want to have it in `src` please tell me and I will put it there.

Travis is failing because of checking `layout_t` against `-1` (unsigned vs signed). But it is necessary to differentiate further between `L_DEFAULT` and a really invalid value because `L_DEFAULT` is a possible value in `cmd_layout` in `src/commands.c`. 

I'd be glad if someone could tell me how to resolve this. We could simply cast the `layout_t`-value to `int`, but that does seem a little hacky. I'd rather define an error/max value inside the enum. For example `L_ERROR = -1` or `L_MAX = 7`. What do you think about this?

Airblader:
> decided to put it in src/util.c because it seemed to fit in there better.

My expectation of this file would be to contain functions that are not domain-specific, but it already does contain quite a bit of domain-specific code, so I think this is fine.

> I'd be glad if someone could tell me how to resolve this.

The »C way« to do this is to use out parameters and a return code:

````
bool layout_from_name(const char *layout_str, layout_t *layout) {
    // … code …
    if(something_that_finds_a_layout) {
        *layout = L_SPLITV;
        return true;
    }

    return false;
}

layout_t layout;
if(!layout_from_name(layout_str, &layout)) {
    // couldn't be found
}
````

File: src/util.c [hunk 28]
```diff
@@ -66,6 +66,31 @@ __attribute__((pure)) bool name_is_digits(const char *name) {
     return true;
 }
 
+/**
```
Airblader: nit: The actual source files only use `/* … */`, not `/** … */`

src/util.c
```diff
@@ -66,6 +66,31 @@ __attribute__((pure)) bool name_is_digits(const char *name) {
...
+               strcasecmp(layout_str, "stacking") == 0) {
+        *out = L_STACKED;
+    } else if (strcasecmp(layout_str, "tabbed") == 0) {
+        *out = L_TABBED;
+    } else if (strcasecmp(layout_str, "splitv") == 0) {
+        *out = L_SPLITV;
+    } else if (strcasecmp(layout_str, "splith") == 0) {
+        *out = L_SPLITH;
+    } else {
+        return false;
```
Airblader: For the mental understanding I would've preferred explicitly returning `true` in all branches, leave out all the `else` blocks and `return false` in the end, since that way each block of the function is its own unit and there's no mental dependency between the top and the bottom of the function. But I guess this is a matter of opinion, so this is fine.

Airblader:
And please don't forget to just drop a comment when you've changed it, otherwise I don't know when to look in here! Thanks again!

src/util.c
s3rb31: > (Just to make it clear, this is just an informational comment. No change requested here.)

Understood. But you're totally right on this one, so I'll change it.  ;)

Airblader: Alright! Just to make sure, I'd write it like this:

````
if (…) {
    *out = …
    return true;
}

if (…) {
    *out = …;
    return true;
}

return false
````

s3rb31:
> Thanks for your patience. :-) This is looking really good now. 

Thanks for all the very nice feedback. It's a pleasure to work on this project!

> There's one tiny little nit left. If you can take care of that one last thing I think we are ready to merge this!

I just fixed it! I think that should be it!

> And please don't forget to just drop a comment when you've changed it, otherwise I don't know when to look in here! Thanks again!

You're just too fast for me. ;) I was waiting for Travis and got distracted then ...

Airblader:
Thanks again!

Issue #2643, opened 2017-01-16
Allow dragging tiled windows with the mouse.


Airblader:
As suggested by @MForster in #2178:

> A container can be dragged by the title bar to one of the four sides of
another container. That container will then be split either horizontally
or vertically.

The PR there has open issues and has thus been closed until somebody is willing to continue working on it.

acrisci:
I'm going to do my best to continue this work in #2653

Airblader:
#2653 has now been abandoned. Anyone else feel free to pick up on @acrisci's continuation.

azadi:
I have gone through the code but I can't seem to find if there is a way to disable this feature short of compiling and removing this myself. Would someone be kind enough to share if they have found something on how to disable this? Thanks!

I realize this is an important feature for some people but I seem to be dragging the windows when clicking on them and it's affecting my workflow.

```
i3 --version
i3 version 4.21 (2022-09-21) © 2009 Michael Stapelberg and contributors
```

Airblader:
We will release a bugfix release very soon that will add a config switch to disable this.

azadi:
> We will release a bugfix release very soon that will add a config switch to disable this.

Very much appreciate the quick reply and support for disabling this. Thank you! I realize it's an important feature but I guess I am too used to not dragging windows anymore and hence. 

Many thanks.

joker234:
> > We will release a bugfix release very soon that will add a config switch to disable this.
> 
> Very much appreciate the quick reply and support for disabling this. Thank you! I realize it's an important feature but I guess I am too used to not dragging windows anymore and hence.
> 
> Many thanks.

I would really like that, too. Had to downgrade i3, because my workflow included clicking on the window titles and I moved them in 90% of the time :see_no_evil:

stapelberg:
i3 v4.21.1 was released a week ago with the fixes: https://i3wm.org/downloads/RELEASE-NOTES-4.21.1.txt

joker234:
> i3 v4.21.1 was released a week ago with the fixes: https://i3wm.org/downloads/RELEASE-NOTES-4.21.1.txt

Thanks for the hint. I only checked the Debian archives where it [doesn't seem to be released yet](https://tracker.debian.org/pkg/i3-wm).

stapelberg:
Yeah, I’m assuming @sur5r hasn’t had a chance to upload the bugfix release to Debian yet.

PR #2653, opened 2017-01-22
[wip] Tiling drag

acrisci:
A continuation of #2178 

TODO

- [ ] Support stacks and tabs
- [x] Don't always create new splits
- [ ] Drag onto the root window
- [x] Drop indicator for inner/outer regions
- [x] Implement outer region drop
- [ ] Initiate tiling drag from titlebar
- [ ] Emit move event properly on tiling drop
- [ ] Refactor `insert_con_into()` into con.c
- [ ] Configure color of drop indicator
- [ ] Bug: i3 crashes if a window closes while the dragging is indicating to drop it onto that window.
- [ ] Tests

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -499,6 +499,7 @@ i3_SOURCES = \
 	include/shmlog.h \
 	include/sighandler.h \
 	include/startup.h \
+	include/tiling_drag.h \
 	include/tree.h \
 	include/util.h \
 	include/window.h \
@@ -540,6 +541,7 @@ i3_SOURCES = \
 	src/sd-daemon.c \
 	src/sighandler.c \
 	src/startup.c \
+	src/tiling_drag.c \
 	src/tree.c \
 	src/util.c \
 	src/version.c \
--- a/include/all.h
+++ b/include/all.h
@@ -64,6 +64,7 @@
 #include "cmdparse.h"
 #include "xcursor.h"
 #include "resize.h"
+#include "tiling_drag.h"
 #include "sighandler.h"
 #include "move.h"
 #include "output.h"
--- a/include/con.h
+++ b/include/con.h
@@ -261,9 +261,9 @@ void con_enable_fullscreen(Con *con, fullscreen_mode_t fullscreen_mode);
 void con_disable_fullscreen(Con *con);
 
 /**
- * Moves a container to the side of a target container. This makes the the
- * container and its target the only two children of a split container with
- * appropriate layout.
+ * Moves a container to the side of a target container. This makes the container
+ * and its target the only two children of a split container with appropriate
+ * layout.
  *
  */
 void con_move_to_side_of_con(Con *con, Con *target, direction_t direction);
--- a/include/con.h
+++ b/include/con.h
@@ -261,6 +261,14 @@ void con_enable_fullscreen(Con *con, fullscreen_mode_t fullscreen_mode);
 void con_disable_fullscreen(Con *con);
 
 /**
+ * Moves a container to the side of a target container. This makes the the
+ * container and its target the only two children of a split container with
+ * appropriate layout.
+ *
+ */
+void con_move_to_side_of_con(Con *con, Con *target, direction_t direction);
+
+/**
  * Moves the given container to the currently focused container on the given
  * workspace.
  *
--- a/include/move.h
+++ b/include/move.h
@@ -17,3 +17,13 @@
  *
  */
 void tree_move(Con *con, int direction);
+
+typedef enum { BEFORE,
+               AFTER } position_t;
+
+/**
+ * This function detaches 'con' from its parent and inserts it either before or
+ * after 'target'.
+ *
+ */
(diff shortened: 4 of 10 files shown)


Airblader:
Glad to see this being continued. I think this is a really cool feature to have in i3.

FYI @MForster

MForster:
Awesome! 👍

acrisci:
There is definitely an interesting UX problem here. I'll share my initial thoughts. Any input is appreciated.

I think we should define two separate drop regions. There will be an "inner region" and an "outer region" on the edges.

Which region you drop in will determine whether it will do a split, or insert it as a sibling within the parent container.

These will then be divided into four quadrants. The quadrant will determine the direction of the move.

```
         +------------------------------------------------+
         |XX                                             X|
         |  X           SIBLING ABOVE                  XX |
         |   X+--------------------------------------+X   |
         |    |                                      |    |
         |    |               SPLIT ABOVE            |    |
         |    |                                      |    |
         |    |                  +                   |    |
         |    |                  |                   |    |
         |    |                  |                   |    |
         |    |                  |                   |    |
SIBLING LEFT  |SPLIT LEFT        |        SPLIT RIGHT|   SIBLING RIGHT
         |    |          +----------------+          |    |
         |    |                  |                   |    |
         |    |                  |                   |    |
         |    |                  |                   |    |
         |    |                  |                   |    |
         |    |                  +                   |    |
         |    |                                      |    |
         |    |               SPLIT BELOW            |    |
         |    |                                      |    |
         |    +--------------------------------------+    |
         |XX X         SIBLING BELOW                  X X |
         +-X----------------------------------------------+

```

This is just the first thing I thought of. Does that make sense?

Airblader:
Yeah, that makes sense to me. I wonder whether we could use regions to switch the container to stacked or tabbed as well. Might be too complex a mental model, though.

acrisci:
The outer/inner indicator is now implemented but not the actual movement.

acrisci:
I think mouse features are probably less useful when you need to press a key to initiate them. I'd like to initiate a drag with the mouse only.

What if dragging from the title bar initiated the tiling drag (in addition to the modifier)? That would be consistent with normal gui conventions where dragging a title bar is a movement action.

Airblader:
SGTM.

acrisci:
The outer region drop behavior is ready to test.

Airblader:
I think one general issue with the dragging feature is that percentages aren't kept. It's a pretty tough issue that I've tackled in my swap command PR (which is still open). I'm not sure we really do need to »fix«/support it for the dragging feature.

Also, i3 crashes if I move the mouse over i3bar while dragging a window.

Airblader:
Please disregard what I said about percentages. Not only don't we need to fix anything here, but it just doesn't even make sense since in general dragging windows changes the tree.

acrisci:
@Airblader can you think of what needs to be done besides my TODOs on the OP? Do we need tests?

Airblader:
@acrisci I'll have to take a look at the change itself (which I didn't yet). In general tests are always a good thing to have. As it stands this is probably a little tricky to test. I wonder whether we should also provide this feature as a command along the lines of

````
drag [window|container] [to] sibling|split left|right|top|bottom for <con_id>
````

which would have the added extra benefit of making it testable without simulating mouse behavior.

Apart from that I've played around with it for a few minutes now. Except for that one crash it seems to work fine even in the few edge cases I could immediately think of. But it did bring me back to my previous idea about drop locations to create tabbed and stacked containers (but perhaps let's just keep this as a future idea for now).

One minor thing coming to mind would be to allow separate customization of the drop indicator via its own `color.drop_indicator` directive (with the fallback to the focused window or something).

Airblader:
i3 crashes if a window closes while the dragging is indicating to drop it onto that window.

acrisci:
I think we can use XTEST (or xdotool) to simulate a mouse click then warp the pointer and release the mouse. That's kind of involved, but this feature is complicated enough that maybe that is worth it.

The `drag` command won't cover the case that you just mentioned here.

I think the outer region should be increased as well, but I'm not sure exactly how large it should be.

Airblader:
> The drag command won't cover the case that you just mentioned here or the other crash with the bar either.

I don't really strongly advocate implementing the command. Using XTEST is perfectly fine with me as well. :-)

> I think the outer region should be increased as well, but I'm not sure exactly how large it should be.

I agree that they seemed a little small. We don't have to get it perfect right away. Let's play with it a little.

Airblader:
@acrisci Do you plan on working on this still?

acrisci:
I think I just have too much going on right now to work on this, but I wish someone would continue this work because I think it has the potential to be a good feature.

Airblader:
Thanks. I do agree as well that this would be really cool and hope someone will continue. I also agree on not having enough time at the moment. :-) I'll close this for now since we have the open issue #2643 for it.

orestisf1993:
Watchers of this PR: I opened a continuation in #3085.

PR #2654, opened 2017-01-22
Migrate i3-config-wizard to draw_util

mihaicmn:
See #2642 for details.

### DIFF
--- a/i3-config-wizard/main.c
+++ b/i3-config-wizard/main.c
@@ -69,10 +69,16 @@
 #include "xcb.h"
 #include "libi3.h"
 
+#define TEXT_PADDING logical_px(4)
+#define WIN_POS_X logical_px(490)
+#define WIN_POS_Y logical_px(297)
+#define WIN_WIDTH logical_px(300)
+#define WIN_HEIGHT (15 * font.height + TEXT_PADDING)
+
+#define col_x(col) \
+    (((col)-1) * char_width + TEXT_PADDING)
 #define row_y(row) \
-    (((row)-1) * font.height + logical_px(4))
-#define window_height() \
-    (row_y(15) + font.height)
+    (((row)-1) * font.height + TEXT_PADDING)
 
 enum { STEP_WELCOME,
        STEP_GENERATE } current_step = STEP_WELCOME;
@@ -90,8 +96,7 @@ static i3Font bold_font;
 static int char_width;
 static char *socket_path;
 static xcb_window_t win;
-static xcb_pixmap_t pixmap;
-static xcb_gcontext_t pixmap_gc;
+static surface_t surface;
 static xcb_key_symbols_t *symbols;
 xcb_window_t root;
 static struct xkb_keymap *xkb_keymap;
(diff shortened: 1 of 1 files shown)


i3-config-wizard/main.c
```diff
@@ -867,10 +859,10 @@ int main(int argc, char *argv[]) {
     xcb_create_window(
         conn,
         XCB_COPY_FROM_PARENT,
-        win,                                                                /* the window id */
-        root,                                                               /* parent == root */
-        logical_px(490), logical_px(297), logical_px(300), window_height(), /* dimensions */
-        0,                                                                  /* X11 border = 0, we draw our own */
+        win,                                         /* the window id */
+        root,                                        /* parent == root */
+        WIN_POS_X, WIN_POS_Y, WIN_WIDTH, WIN_HEIGHT, /* dimensions */
```
Airblader: I don't like the hardcoded position, but it was like this before. We should fix this and while we're at it move mapping the window further down since we're currently violating the specifications. But that's for a separate PR, so this is just FYI.

i3-config-wizard/main.c
```diff
@@ -69,10 +69,14 @@
 #include "xcb.h"
 #include "libi3.h"
 
+#define TEXT_PADDING logical_px(4)
+#define WIN_POS_X logical_px(490)
+#define WIN_POS_Y logical_px(297)
+#define WIN_WIDTH logical_px(300)
+#define WIN_HEIGHT (15 * font.height + TEXT_PADDING)
```
Airblader: This is now `1 * font.height` more than it was before, no?

i3-config-wizard/main.c
```diff
@@ -463,82 +466,71 @@ void errorlog(char *fmt, ...) {
 void debuglog(char *fmt, ...) {
 }
 
+static void txt(int x, int row, char *text, color_t fg, color_t bg) {
```
Airblader: The calls to this use a lot of duplicated arguments for x. I think we should improve this as well. It's odd anyway to use pixels for x, but a row for y. We should switch the x argument to a column argument instead (and yes, this does mean using slightly different indentations). Do you want to do this in this PR or would you prefer it to be done separately?

Airblader:
Despite my comments, before merging I'm going to give this a shot on my machine and then let @stapelberg know.

i3-config-wizard/main.c
mihaicmn: No, previously it was a bit indirect: row_y(15) + font.height = 14 * font.height + logical_px(4) + font.height

i3-config-wizard/main.c
mihaicmn: It makes sense. I will include the change in this PR.

mihaicmn:
@Airblader I updated the code.

i3-config-wizard/main.c
Airblader: Right. :-)

Airblader:
I'm actually not sure when I can get around to setting up testing this on my machine. Do you mind posting a screenshot of what it looks like?

mihaicmn:
2017-01-23-225606_3840x2160_scrot
2017-01-23-225706_3840x2160_scrot

Issue #2664, opened 2017-01-30
Feature request: Ability to hide the tab bar


kilbith:
In tabbed mode, the tab bar is redundant since I already have `tint2` listing the windows.

I need to hide this i3 tab bar to the profit of other panel/taskbars.

Airblader:
We've discussed this in the past and decided not to allow this. Did your position on this change, @stapelberg ?

stapelberg:
No change, I still think title bars must be visible in any non-default layout.

kilbith:
Is there an explanation somewhere for this decision ? How does it hurt to have it as an option ?

Airblader:
https://faq.i3wm.org/question/778/why-is-patch-not-merged-and-made-optional/

stapelberg:
The rationale is that hiding the title bars makes it significantly harder to build up a mental model of the layout that is on your screen. I.e., you can no longer distinguish different layouts from each other.

amosbird:
How about adding an option for merging title bar and i3bar?

Airblader:
i3bar is a standalone tool, so that wouldn't work. Plus it's part of the i3bar philosophy to display workspaces, not individual windows.

Airblader:
If you want to disable the titlebars I'd suggest just running your patch locally.

blueforesticarus:
*refusing to accept patches for the ability to turn off a default feature.*

You're clowning.

Nadrieril:
If anyone is interested in patching it themselves, it's very easy:
```patch
diff --git a/src/render.c b/src/render.c
index 478a1cc4..4644005a 100644
--- a/src/render.c
+++ b/src/render.c
@@ -25,10 +25,7 @@ bool has_adjacent_container(Con *con, direction_t direction);
  * Returns the height for the decorations
  */
 int render_deco_height(void) {
-    int deco_height = config.font.height + 4;
-    if (config.font.height & 0x01)
-        ++deco_height;
-    return deco_height;
+    return 0;
 }
 
 /*
```

enesaltinkaya:
How very apple of you.

blueforesticarus:
![Image](https://github.com/user-attachments/assets/48a42f02-d1fe-4a84-bf5c-6938c587727c)

PR #2665, opened 2017-01-30
[WIP] Add new option: hide_tab_container

kilbith:
Addresses https://github.com/i3/i3/issues/2664

Not sure if it's the perfect approach to implement that. I'm willing to add more documentation if you're keen with the idea.


Airblader:
This approach is definitel not clean. However, before discussing a PR we should discuss the feature itself.

stapelberg:
Thanks for the pull request. I’ll have to decline merging it, however. See #2664 for rationale.

PR #2668, opened 2017-02-02
restart bar status command on config reload

### LINKED ISSUE #2651: [Enhancement] Make config-reload recognize bar status_command change
Currently (4.13), a reload of the config does not re-evaluate the `status_command` of a `bar` block. If you change it, you have to restart i3 to take effect. I think the `status_command` should be checked for a change at a reload. This would also be in line with the discussion in #1446.

jolange:
Closes #2651


jolange:
Yep. If the right infrastructure is there already, you just need to find out where things are happening ;-)

PR #2672, opened 2017-02-05
numslock no longer affects bindings, fixes #2523

### LINKED ISSUE #2523: Mouse bindings don't work when numlock is enabled
**i3 version:**
Binary i3 version:  4.12-154-g664d11c (2016-10-26, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12-154-g664d11c (2016-10-26, branch "next") (pid 1933)

**What I did:**
I have this line in my i3 config:

```
bindsym --whole-window button9 kill
```

Which binds the top "thumb button" on any Logitech-like mouse to kill the window. See the leftmost button on this image:

image

**What I saw:**

The mouse binding doesn't work if numlock is ON.  If numlock is OFF, it works as expected.  I have a standard US pc105 keyboard.  This _may_ be related to the faulty logic exhibited in #2418.

syrrim:


Airblader:
The change breaks a test and I also don't think it's correct. We can't just always dismiss the NumLock key.

syrrim:
As a note, we already ignore the capslock key. The numslock key doesn't seem like that far of a leap to me. It has the same property of being easy to enable by mistake. My keyboard in particular has no light indicating its state, so I never know if I have numslock on or off.

Airblader:
The change to not always discard the NumLock key was intentional, see d77d40173aae250d2e5ba86a8bce7a6e857cbc67

stapelberg:
Simply ignoring the key is not the correct solution, at least not without a lengthy explanation as to why that must be the only option. As @Airblader mentioned, our change to respect the key was intentional in order to enable a use-case.

Issue #2663, opened 2017-01-30
[Enhancement] Improve documentation for workspace commands


xkpd3:
Hello,

I would love if default names for workspaces could be set.
To understand what I mean, let's assume the following config:

```
set workspace default name  "1:[1]"
set workspace default name  "2:[2]"
set workspace default name  "3:[3]"
bindsym $mod+1 workspace number 1
bindsym $mod+2 workspace number 2
bindsym $mod+3 workspace number 3
bindsym $mod+Shift+exclam move container to workspace number 1
bindsym $mod+Shift+quotedbl move container to workspace number 2
bindsym $mod+Shift+section move container to workspace number 3
```

The first three lines are not implemented. I'm switching to workspaces using the "workspace number", because I rename workspaces while working.

When a new workspace is created, it is created with the number as name.

It would be nice to be able to set a default name, which is used when a workspace is created.

I worked around this with a [bunch of scripts](https://github.com/xkpd3/xkpd3conf-i3/tree/master/bin). It works okay, but I wonder if this would be something other people also would like to see in the core.

What do you think?

Best Regards,
Stefan

fantomas911:
You can do something like this:
`bindsym $mod+1 workspace myDevelopWorkSpace
bindsym $mod+Shift+exclam move container to myDevelopWorkSpace`

xkpd3:
Hi @fantomas911,

As I wrote above, I rename workspaces **while working**. This is why I'm naming my workspaces with the special `<WSNumber>:<WSName>` string, because this enables me to use
`bindsym $mod+1 workspace number <WSNumber>`
`bindsym $mod+Shift+exclam move container to workspace <WSNumber>`
to switch or move to a workspace regardless of `<WSName>`.

However, if I use "workspace number", I can not set a default workspace name. THIS is what I want.
If I do not use "workspace <WSName>" then I have a much bigger problem with not matching key bindings after the rename.

Airblader:
I think you can just use

````
set $ws1_default "1:Foo"
set $ws2_default "1:Bar"
bindsym $mod+1 workspace number $ws1_default
bindsym $mod+2 workspace number $ws2_default
````

This should both name it correctly initially and deal with renames, I think.

xkpd3:
@Airblader I'll try that. I did not know that `workspace number "1:Foo"` is a valid statement. I'll report back this evening.

xkpd3:
@Airblader, you're my hero! This works perfectly fine!

[...]

PR #2674, opened 2017-02-05
Trying to explain the workspace number behavior

xkpd3:
The documentation did not explain how workspace number `"1: www"` is working.
Related to #2663

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2090,6 +2090,23 @@ i3-msg 'rename workspace to "2: mail"'
 bindsym $mod+r exec i3-input -F 'rename workspace to "%s"' -P 'New name: '
 --------------------------------------------------------------------------
 
+If you want to rename workspaces on demand while keeping the navigation stable,
+you can use a setup like this:
+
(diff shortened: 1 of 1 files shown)


docs/userguide
```diff
@@ -2090,6 +2090,20 @@ i3-msg 'rename workspace to "2: mail"'
...
+
+*Example*:
+-------------------------
+bindsym $mod+1 workspace number "1: www"
+bindsym $mod+2 workspace number "2: mail"
+...
+-------------------------
+
+The command +workspace number "1: mail"+ will switch to workspace 1 and ignore the
+text part. So even when the workspace has been renamed to "1: web", the above
```
stapelberg: Can you clarify that the text part is only ignored when the workspace already exists please?

xkpd3:
I've added a paragraph describing the creation case.

stapelberg:
Could you squash the commits into 1 single commit before I merge please?

xkpd3:
Commits squashed.

PR #2679, opened 2017-02-12
Add troubleshooting for title attributes during layout restoring.

Airblader:


Airblader:
@nmschulte Since you're fairly active lately perhaps you can review this? :-)

Airblader:
For context, this came up (once again) here: https://www.reddit.com/r/i3wm/comments/5tk1sy/layout_restore_help/

nmschulte:
No problem.  It reads just fine, I think.

Interestingly, I've been playing with `append_layout` and `i3-save-tree` lately, and I ran up against this "want to start terminals with shells in specific working directories in a layout" problem.  Here's what I found:

1) You _can_ start rxvt-unicode with a specific title, use the `-T` flag.  (v9.22)
2) `append_layout` will match `title`, but not _only_ `title`.  I had to also include `class` for i3 to match `title`s; it would not match if `title` as the only `swallows` property. (`"class": "^URxvt$"`)

I think #2 is a bug, but I didn't dig.

Airblader:
Thanks. I agree that item 2 sounds like a bug, so please feel free and open an issue for it. I do believe the documentation is still helpful because it's a fairly common problem people run into. Most people won't use a `-T` flag and not all emulators have that anyway. :-)

We have an open issue to also swallow on window title changes; fixing that will make this documentation obsolete, so I've added a comment to the ticket.

stapelberg:
> @stapelberg Github won't let me merge without a review by someone with write access to the repository, so unfortunately @nmschulte's review doesn't count. How should we deal with this for changes I authored without having to involve you?

For now, I’d like to keep “all changes must be reviewed” as a policy. In practice, reviews of your contributions are usually very quick. I think my availability for such reviews will get better within the next few weeks. One thing that would help is if you commented “rubber-stamp approval, @stapelberg”, in this order — that should make it into my inbox and signal that only a cursory glance is required :).

Airblader:
I agree that the policy should be kept (code reviews are crucial). I'll try to get review help from people like @nmschulte (assuming he's OK with it :-) ) and then proceed as you requested.

nmschulte:
Certainly, please don't hesitate to request reviews from me: I'm happy and willing to contribute.

Issue #2276, opened 2016-03-30
Crash at startup with multiple monitors.


layus:
Output of `i3 --moreversion 2>&- || i3 --version`:

Binary i3 version:  4.12-17-g84c39a1 (2016-03-30, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.12-17-g84c39a1 (2016-03-30, branch "next") (pid 989)
Loaded i3 config: /home/gmaudoux/.i3/config (Last modified: mer 30 mar 2016 16:16:31 CEST, 648 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3

URL to a logfile as per http://i3wm.org/docs/debugging.html: 

http://logs.i3wm.org/logs/5718016125829120.bz2

**What I did:**

Boot and exec `startx`.

**What I saw:**

i3 crashed, the small "i3 crashed" popup appearing, but with no text, just a black box.

**What I expected instead:**

Hmm... i3 starting normally :-).

**Extra info:**

My laptop is docked an plugged to an external monitor.
To manage docking and external monitors/projectors, i use autorandr, which is called automatically when drm events are noticed. This is done by my monitormonitors script.

My feeling is that i3 crashes because my postswitch script (see below) tries to send messages to i3 just when i3 is comping with the new monitor layout notified by xrandr.

Note that autorandr is always called once to check the disposition at the begining:

```
#!/usr/bin/env python

import functools
import os.path
import pyudev
import subprocess
from pprint import pprint
from threading import Timer


def main():
    BASE_PATH = os.path.abspath(os.path.dirname(__file__))
    path = functools.partial(os.path.join, BASE_PATH)
    call = lambda x, *args: subprocess.call([x] + list(args))
    autorandr = lambda: call('/usr/bin/autorandr', '-c', '--default', 'presentation')

    context = pyudev.Context()
    monitor = pyudev.Monitor.from_netlink(context)
    monitor.filter_by(subsystem='drm')
    monitor.start()

    timer = Timer(0.5, autorandr)
    timer.start()

    for device in iter(monitor.poll, None):
        pprint(device)
        pprint(device.attributes)
        # I can add more logic here, to run only certain kinds of devices are plugged.

        # This ensures that we fire only one call to autorandr for multiple events occuring within some interval.
        timer.cancel()
        timer.join()
        timer = Timer(0.5, autorandr)
        timer.start()


if __name__ == '__main__':
    main()
```

[...]

PR #2684, opened 2017-02-16
Temporary fix for #2276

layus:
@Airblader  It would be really nice to have this merged. I know it is not a proper fix, but at least we would not experience repeated segfaults in the meantime. #2276 has been open for a full year. How long will it take to receive a proper fix ?

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -95,6 +95,8 @@ static bool definitelyGreaterThan(float a, float b, float epsilon) {
  */
 static bool maybe_back_and_forth(struct CommandResultIR *cmd_output, const char *name) {
     Con *ws = con_get_workspace(focused);
+    if (ws == NULL)
+        return false;
 
     /* If we switched to a different workspace, do nothing */
     if (strcmp(ws->name, name) != 0)


Airblader:
Thanks for your contribution.

I'm going to decline this because I'd like to avoid merging workarounds. They tend to make something better in the short run, but overall it hides an issue that isn't understood and might only lead to making it even harder to figure out the problem.

> How long will it take to receive a proper fix ?

As soon as somebody spent the time to figure out the root cause and comes up with the fix for it (which doesn't necessarily have to be the same person). Everybody is welcome to do so. :-)

That said, if the issue is a big pain for you and you don't want to invest the time, you can run your patch locally. I hope you understand why I don't want to merge it, though.

layus:
Yep, I understand. I can indeed patch locally, but I find this quite deceptive for other users who may encounter the same issue.

Airblader:
Once you have the workaround in upstream, nobody wants to spend time on the issue ever again, though. :-)

layus:
Unless the issue comes out again. Otherwise the workaround is a real fix, even if no-one knows why ;-). *Half kidding here*

Airblader:
Once you stop understanding the software you deliver, you're in big trouble. :-)

PR #2685, opened 2017-02-17
Copy the entire window content on Expose events.

### LINKED ISSUE #2683: Borders disapear for a few (long) seconds
*i3 version 4.13-60-gad9c879c*

**Optional Conditions**
Many (multiple) windows

**What I did:**
Ctrl-Alt-F1 and Ctrl-Alt-F2 (jump back and forth)

**What I saw:**
Borders disappear for a few seconds (very noticeable) then reappear.
It's easily noticeable if you have multiple windows (e.g. in stacking layout).
(The easiest way to notice this is to have one more WM running at the same time, then jumping back-and-forth between them.)

**What I expected instead:**
Borders shouldn't disappear (or, at least, should redraw *more quickly*).
Something is apparently different since it's not as nearly as noticeable as before.
My previous build (a301396) doesn't display this behaviour.

Airblader:
With commit d58dbc3 we started ignoring Expose events in a sequence
except for the last one. Since we only copied the affected part of
the window in the Expose event handler, this caused incorrectly
rendered window decorations.

Instead of reverting to the old behavior, we now copy the entire window
content on this single, last event with the following rationale:
- It's cheaper to copy a larger chunk once than multiple smaller
  chunks doing one server roundtrip each.
- That's how we do it when rendering out decoration on decoration
  changes as well.

fixes #2683

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -614,12 +614,9 @@ static void handle_expose_event(xcb_expose_event_t *event) {
     }
 
     /* Since we render to our surface on every change anyways, expose events
-     * only tell us that the X server lost (parts of) the window contents. We
-     * can handle that by copying the appropriate part from our surface to the
-     * window. */
+     * only tell us that the X server lost (parts of) the window contents. */
     draw_util_copy_surface(&(parent->frame_buffer), &(parent->frame),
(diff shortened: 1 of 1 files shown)


Airblader:
@nmschulte :-)

PR #2690, opened 2017-02-22
docs: explain adding containers into stcratchpad

sudo-nice:


Airblader:
Thanks for your contribution. I'm going to have to reject this because floating nested containers are not something we actually (want to) support, it just more or less happens to work. For this reason I don't want to document this behavior, because that would turn it into supported behavior.

PR #2705, opened 2017-03-12
i3bar: accept 'primary' for output config option

lebenlechzer:
This adds the previously adressed but unresolved feature to accept 'primary' as output option in the i3bar config.

fixes #2311 
@Airblader @acrisci

PR #2725, opened 2017-04-03
Feature/output primary

### LINKED ISSUE #2311: i3bar: Accept "primary" for output config option
This is a feature request to add the argument "primary" as special case to the `output` config option for i3bar to enable the bar to automatically be put on the primary monitor without having to configure an output explicitly (helpful with occasionally changing monitor setups), similar to `tray_output`.

My specific use case is to be able to have an i3bar with a different `status_command` on the primary output.

nmschulte:
#2705 with the `clang-format` fixes

fixes #2311 

@lebenlechzer

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1378,7 +1378,7 @@ directive multiple times.
 
 *Syntax*:
 ---------------
-output <output>
+output primary|<output>
 ---------------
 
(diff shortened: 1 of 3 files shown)


Airblader:
Thansk!

Issue #2734, opened 2017-04-10
[Request] 1 pixel title bars


squigglezworth:
While we can set borders to be 1 pixel wide, the smallest it's possible to get stacked/tabbed container title bars is ~5px with font size 0. It would be nice to be able to set the height of the title bar, or just a 1pixel title setting that just won't display the title

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
See #2664

squigglezworth:
Except this is not talking about hiding title bars.

Airblader:
See the discussion in the pull request. This is just not a configurability that we want.

squigglezworth:
There is no "discussion" past "we don't want to *hide* the title bar" which i completely agree with. 

However, I don't need nor want more than a 1pixel indicator where my window is. I don't need a title.

Airblader:
It's a design decision in i3 to force the display of window titles in tabbed or stacked layouts; a decision we are not looking to change, see #2664. Setting the font size to 0 is just a workaround you are trying to use, but it leaves you with the issue you described; and we don't want to enable all kinds of flows by supporting workarounds used because there's features that we rejected.

If you feel strongly about this then I suggest running a local patch. You already have one ready by @nmschulte above.

squigglezworth:
I was hoping to revisit this issue and hopefully get it implemented in i3. 

To clarify, I still don't wish to completely disable title bars, nor do I want to support any particular hacky workaround that I myself use.

I'd like to be able to disable the drawing of container title bar *text*. Logically following that, the ability to set the title *bar's* height should be possible.
Possible config directives could be
`draw_container_titles <true|false>`
and 
`container_title_bar_height <px>`

I know for certain that I'm not the only one who doesn't utilize the container titles, and are using some workarounds (like setting the `title_format` to a blank span for every window) to achieve similar behavior.


I hope this makes it more clear the configuration I'm looking for, and the reasoning.

I do believe such options have merit and should be considered.

orestisf1993:
It does get pretty hard to distinguish the layout after you remove the titles:

[...]

PR #2735, opened 2017-04-11
add title_vertical_margin configuration

nmschulte:
for #2734, allows to specify the additional vertical margin over the font height used when rendering titlebars

I'm sure this will need work before, if ever, it goes in.  beyond just the concept, these need consideration:
- [ ] userguide update
- [ ] parser tests
- [ ] review of config parser state modifications
- [ ] overflow/upper limit on titlebar height

### DIFF
--- a/include/config_directives.h
+++ b/include/config_directives.h
@@ -57,6 +57,7 @@ CFGFUN(force_display_urgency_hint, const long duration_ms);
 CFGFUN(focus_on_window_activation, const char *mode);
 CFGFUN(show_marks, const char *value);
 CFGFUN(hide_edge_borders, const char *borders);
+CFGFUN(title_vertical_margin, const long margin_height);
 CFGFUN(assign, const char *workspace);
 CFGFUN(no_focus);
 CFGFUN(ipc_socket, const char *path);
--- a/include/configuration.h
+++ b/include/configuration.h
@@ -130,6 +130,11 @@ struct Config {
      * By default, this is disabled. */
     hide_edge_borders_mode_t hide_edge_borders;
 
+    /** The amount of vertical buffer ("margin") to include when calculating the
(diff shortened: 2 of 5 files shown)


Airblader:
I don't really like the idea of making this pixel value configurable. I think the better option is to disable titles for stacked / tabbed containers (per configuration).

nmschulte:
> disable titles for stacked / tabbed containers

It seems wrong to limit this feature to stacking or tabbing containers; is there something about splitting containers that precludes this from being useful there?

Note, originally, I interpreted the issue to be about the window titles within a container, but it seems the request is to elide the container title, and retain its windows'.

I can effect this change no problem, but I think I've missed something related to testing of the config file syntax; the automated tests note these failures: -- the first is related to `hide_edge_borders` (and the test uses `hide_edge_border` (no trailing `s`), but I think maybe this is the intentional error for the test)

>not ok 22 - errors dont harm subsequent statements
>not ok 25 - error message (2+2 context) ok
>not ok 26 - error message (0+0 context) ok
>not ok 27 - error message (1+0 context) ok
>not ok 29 - error message (0+2 context) ok

Airblader:
> It seems wrong to limit this feature to stacking or tabbing containers; is there something about splitting containers that precludes this from being useful there?

Normal window titles can already be hidden. Only for tabbed and stacked containers with more than one child we currently force window titles no matter what the border configuration is.

nmschulte:
Closed in favor of a new PR.

jolange:
> I think the better option is to disable titles for stacked / tabbed containers (per configuration).

But that has been rejected in the past, right? (E.g.  #2664)

nmschulte:
Possible difference: this change is to hide the title part only, not the entire bar.  The change you cite is to remove the bar entirely.

PR #2739, opened 2017-04-13
i3-random-terminal

sqwishy:
This is important for when i3 needs a terminal but which one should be a surprise.


Airblader:
I'm not sure this is a joke or serious, but I don't see any reason to have such script I'm i3.

PR #2752, opened 2017-04-27
Fix changing of root workspace layout from stacked/tabbed

### LINKED ISSUE #2464: Can't change root workplace layout from stacking or tabbing

Running i3 version: 4.12 (2016-03-06, branch "4.12") (pid 1632)
Loaded i3 config: /home/atk/.i3/config (Last modified: Tue 20 Sep 2016 05:22:34 PM MSK, 1900 seconds ago)

**What I did:**
I'm pressing super+w (select tabbing) on workspace with no windows opened, press super-enter (open rxvt-unicode) a few times, then ctrl-d same amount of times (to end terminal sessions), then I press super-h (select horizontal tiling mode) and super-enter (open terminals) a few times again

**What I saw:**

The second time I pressed super-enter a few times terminals continued to tab.

**What I expected instead:**

Terminals should be tiled horizontally, since I pressed super-h before opening them.

The same thing happens with super-v (vertical tiling) and super-s (stacking). There seems it is no way back from situation when "root workplace layout" is tabbing or stacking.

mihaicmn:
Whenever the layout of an empty workspace is changed we need to update the [_workspace_layout_](https://github.com/i3/i3/blob/next/include/data.h#L681-L687) accordingly.
 
Fixes #2464.

### DIFF
--- a/src/con.c
+++ b/src/con.c
@@ -1639,12 +1639,14 @@ void con_set_layout(Con *con, layout_t layout) {
      * whole workspace into stacked/tabbed mode. To do this and still allow
      * intuitive operations (like level-up and then opening a new window), we
      * need to create a new split container. */
-    if (con->type == CT_WORKSPACE &&
-        (layout == L_STACKED || layout == L_TABBED)) {
+    if (con->type == CT_WORKSPACE) {
         if (con_num_children(con) == 0) {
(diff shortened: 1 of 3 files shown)


Airblader:
Thank you!

Issue #2744, opened 2017-04-18
Feature Request: Key Release IPC message for Mod keys


muhmud:
I'm using i3 version 4.13

I think it would be nice to have an ipc message sent out when any of the mod keys are released. This would allow a client script to be used to implement Alt+Tab style window stacking, which is really useful for some users. The release event is required to know when cycling is over and to allow the stack to be updated.

I have done this in my own fork by adding code to test `state->eventType` and `state->keycode` in the `handle_event()` function in handlers.c when processing a `XCB_XKB_STATE_NOTIFY` message, and then sending out a message as required.

I was then able to use a slightly hacked version of i3ipc.py to create client python scripts that now allow me to do Alt+Tab style switching for both windows and workspaces (woohoo!).

Would be nice if this could be pulled in.

Thanks for a great WM!

Airblader:
I think the feature itself sounds acceptable. Can you create a PR?

muhmud:
OK, done, it's [here](https://github.com/i3/i3/pull/2756)

Airblader:
As explained in the PR I've changed my mind on accepting this. Sorry again for the confusion!

muhmud:
No problem, but a bit of a shame. 

If bindsym --release worked for mod keys by themselves, that would also be OK, but it doesn't.

PR #2756, opened 2017-04-29
Feature Request: Key Release IPC message for Mod keys

muhmud:
As requested, pull request for this [issue](https://github.com/i3/i3/issues/2744) of the same name.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -1365,9 +1365,11 @@ static void property_notify(uint8_t state, xcb_window_t window, xcb_atom_t atom)
  *
  */
 void handle_event(int type, xcb_generic_event_t *event) {
+    DLOG("handle_event()\n");
+    
     if (type != XCB_MOTION_NOTIFY)
         DLOG("event type %d, xkb_base %d\n", type, xkb_base);
-
+    
     if (randr_base > -1 &&
         type == randr_base + XCB_RANDR_SCREEN_CHANGE_NOTIFY) {
         handle_screen_change(event);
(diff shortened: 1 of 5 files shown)


src/handlers.c
```diff
@@ -1367,7 +1367,7 @@ static void property_notify(uint8_t state, xcb_window_t window, xcb_atom_t atom)
 void handle_event(int type, xcb_generic_event_t *event) {
     if (type != XCB_MOTION_NOTIFY)
         DLOG("event type %d, xkb_base %d\n", type, xkb_base);
-
+
```
Airblader: Please remove all of these incorrectly introduced whitespaces.

src/handlers.c
```diff
@@ -1402,16 +1402,26 @@ void handle_event(int type, xcb_generic_event_t *event) {
             }
         } else if (state->xkbType == XCB_XKB_STATE_NOTIFY) {
             DLOG("xkb state group = %d\n", state->group);
+
+            /* If the Alt, Windows or any other mod key is released generate an ipc event */
```
Airblader: So in other words mentioning Alt and Windows is just misleading here, so please rephrase it. Also, why is this restricted to modifier keys?

Airblader:
* Please update the documentation.
* Please add tests to verify this behavior.
* Please rebase and squash the commits so the git history stays clean and the patch can be applied.
* Please remove the incorrectly introduced tabs.

Airblader:
I'm actually even somewhat second-guessing my decision. I – mistakenly – assumed we already had such an event for the keydown itself. But currently we only have the binding event. What if we have a binding using `--release` (possibly just mapped to `nop`), shouldn't the binding event in this scenario come at the correct time?

Airblader:
The more I think about it the less I think we should actually implement this. There's nothing i3-specific about this, so I feel like this would be suited better for an external script or tool to watch for these events. That way it's also agnostic to the window manager.

I'm sorry for the initial »Go« for this. I hope it didn't cause too much trouble since you already said you had the patch locally available.

Issue #2329, opened 2016-04-29
i3-dmenu-desktop inserts an additional space in quotation marks


SirCmpwn:
See https://github.com/SirCmpwn/sway/issues/521

I consider this a flaw in i3-dmenu-desktop, not in sway.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

SirCmpwn:
See also https://github.com/i3/i3/pull/2265, which fixes this, but was closed for vague reasons.

stapelberg:
I think the title doesn’t accurately capture the problem (or maybe I’m misunderstanding). https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html states:

> The Exec key must contain a command line. A command line consists of an executable program optionally followed by one or more arguments.

In the example you referred to, the desktop file specified `Exec=firefox %u`, which is evaluated to “firefox ” (with a trailing space). But, depending on the desktop file, it might as well be evaluated to e.g. “firefox --no-new-window”, in which case the whitespace would be perfectly intentional.

So, can we find a more accurate description of the issue at hand please?

SirCmpwn:
No, it's true. The important part of the title is *in quotation marks*. Why not just have it do 'i3-msg exec firefox '? 'i3-msg exec "firefox "' is what's causing problems. The quote handling on i3's exec implementation is a bit wonky here.

stapelberg:
> No, it's true. The important part of the title is in quotation marks. Why not just have it do 'i3-msg exec firefox '? 'i3-msg exec "firefox "' is what's causing problems.

The reason why i3-dmenu-desktop quotes the argument to i3’s exec command is so that commas and semicolons are not incorrectly interpreted by i3 as command separators.

> The quote handling on i3's exec implementation is a bit wonky here.

In which way? How precisely do you propose to change it?

Overall, I still don’t understand what the issue is that we’re trying to fix. Can you elaborate please?

stapelberg:
I think I understand the part where you’re outlining the parsing difference in sway vs. i3.

What I’m still unclear on is what you’d like to change/see changed. What is your proposal?

For completeness:

> For example, with i3, I couldn't actually run a process called "firefox " through the exec command.

You can:

```shell
[...]

PR #2757, opened 2017-04-30
Fix i3-dmenu-desktop quoted command name

Jereq:
According to the Desktop Entry Specification
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables
the executable name or path of the executable may be quoted. This is not
properly respected when i3-dmenu-desktop extracts the command name from
the Exec entry.

Examples of values that fail and what they currently result in:

- `"bar"` -> `"bar"`
- `"foo/bar"` -> `bar"`
- `"foo foobar/bar"` -> `"foo`
- `"foo\sbar"` -> `"foo\sbar"`
- `foo\sbar` -> `foo\sbar`
- `"foo\\\\bar"` -> `"foo\\\\bar"`

### DIFF
--- a/i3-dmenu-desktop
+++ b/i3-dmenu-desktop
@@ -282,7 +282,31 @@ for my $app (keys %apps) {
     }
 
     if ((scalar grep { $_ eq 'command' } @entry_types) > 0) {
-        my ($command) = split(' ', $apps{$app}->{Exec});
+        my $command = $apps{$app}->{Exec};
+
+        # Handle escape sequences (should be done for all string values, but does
+        # matter here).
+        my %escapes = (
+            '\\s' => ' ',
+            '\\n' => '\n',
+            '\\t' => '\t',
+            '\\r' => '\r',
+            '\\\\' => '\\',
+        );
+        $command =~ s/(\\[sntr\\])/$escapes{$1}/go;
+
+        # Extract executable
+        if ($command =~ m/^\s*([^\s\"]+)(?:\s|$)/) {
+            # No quotes
+            $command = $1;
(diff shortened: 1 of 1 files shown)


Airblader:
Also see #2329. As explained there, we want to avoid breaking user configs unless we have a *specific* bug that happens rather than a theoretical one (albeit not conforming to the standard is, indeed, not ideal). Do you have such a situation?

Jereq:
Hi,

Thanks for the response.

I believe #2329 is describing a different, but related issue. #2329 discusses how the command is forwarded to i3, while this is specifically about how the command is parsed to get the executable name out. As such, this should have less room for breaking existing configs.

I tried to come up with scenarios where this would cause problems, and the only one I can see is if someone relies on one of the faulty examples I listed above when using dmenu. It would be a bit weird, but it is possible.

This will not affect how the programs are launched, once found in dmenu. That was the main reason I only changed how the command to name translation works, rather than "fixing" the string value parsing more generally (which would be needed to fully conform to the spec).

This is also not a theoretical bug. Jetbrains creates .desktop files with quoted paths for at least two of their IDEs. The  following is an example:
```
[Desktop Entry]
Version=1.0
Type=Application
Name=IntelliJ IDEA Community Edition
Icon=/opt/idea-IC-163.7743.44/bin/idea.png
Exec="/opt/idea-IC-163.7743.44/bin/idea.sh" %f
Comment=The Drive to Develop
Categories=Development;IDE;
Terminal=false
StartupWMClass=jetbrains-idea-ce
```
This example triggers the second example above resulting in `idea.sh"` appearing in dmenu. Depending on the installation path, example 3 and 6 can be triggered as well.

BR
// Sebastian

Airblader:
I see, I misunderstood the PR when giving it a quick look. Thanks for the explanation!

Airblader (approved):

PR #2759, opened 2017-04-30
Added a hint about the required config v4 version hint.

### LINKED ISSUE #2751: `exec_always` doesn't seem to work without `# i3 config file (v4)`

https://github.com/i3/i3/files/956246/i3.txt, the relevant part only, ie: it not detecting it as a v4 script.

**What I did:**

Type my own config from scratch

**What I saw:**

`exec_always` doesn't seem to work without `# i3 config file (v4)`

**What I expected instead:**
The behavior being documented or [it being able to detect more v4 stuff](https://github.com/i3/i3/blob/next/src/config_parser.c#L637) (see additional note)
____

and now the useful bug report:

**description:**
`exec_always` (and possibility other function) doesn't seem to work without `# i3 config file (v4)` in the script

This isn't documented in the [userguide](https://i3wm.org/docs/userguide.html)\*, took me a long time to figure it out.

**reproducible test** (~/.config/i3/config)**:**
```
exec_always xterm
```

**test note:**
add `# i3 config file (v4)` anywhere and then it works
WARNING: the test case doesn't contain binding for exiting

**additional note:**
[The issue which added the feature](https://github.com/i3/i3/issues/435) seem to have been tag as 4.0, suggesting it is only added in v4

\* https://i3wm.org/docs/4.13/userguide.html for future viewer when it's available

Airblader:
fixes #2751

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -297,6 +297,15 @@ keyboard layout. To start the wizard, use the command +i3-config-wizard+.
 Please note that you must not have +~/.i3/config+, otherwise the wizard will
 exit.
 
+Since i3 4.0, a new configuration format is used. i3 will try to automatically
+detect the format version of a config file based on a few different keywords,
+but if you want to make sure that your config is read with the new format,
+include the following line in your config file:
+
+---------------------
+# i3 config file (v4)
+---------------------
+
 === Comments
 
 It is possible and recommended to use comments in your configuration file to


Airblader:
rubber-stamp approval, @stapelberg

docs/userguide
```diff
@@ -297,6 +297,13 @@ keyboard layout. To start the wizard, use the command +i3-config-wizard+.
 Please note that you must not have +~/.i3/config+, otherwise the wizard will
 exit.
 
+Please note that the config must contain the following line since version 4.0
```
stapelberg: This is actually not entirely true. The “# i3 config file (v4)” marker is one way to make i3 interpret the config file as a version 4 file. Others can be seen at https://github.com/i3/i3/blob/fff3f79da9a87a1f790c6328f6615422f2b69b47/src/config_parser.c#L637-L640

Could you rephrase the paragraph to make it clear that i3 tries to detect v4 files automatically, but you can make sure it does the right thing by adding the hint?

Airblader:
@stapelberg I've rephrased it now.

PR #2762, opened 2017-05-03
Fix invalid JSON

mdirkse:
I was using the example JSON as input for tests that I'm writing for https://github.com/mdirkse/i3ipc-go and found two errors.


Airblader:
Thanks!

PR #2769, opened 2017-05-12
Force container to be redrawn upon moving to another container.

Airblader:
This is necessary as otherwise urgent containers can be broken
after moving them.

This is a finding within #2496.


Airblader:
rubber-stamp approval, @stapelberg

Issue #2764, opened 2017-05-05
Feature: make `primary` output available on all places where an output can be used


coredump:
Currently the `primary` output (as defined by `xrandr`) can be used on `tray_output`. 

It would be useful to have it as a possible output on other commands like `assign` or `move` like we currently have left/right/up/down.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
Sounds good to me. PRs welcome.

hwangcc23:
I will work on this for sending the PR.

orestisf1993:
Is `assign <criteria> [→] [output] <output>` part of this issue?

Airblader:
Yes, I would say so.

orestisf1993:
Anything else missing for this?

Airblader:
I guess not. Your commit message didn't include a "fixes #2764", so Github didn't close the issue. :-)

PR #2771, opened 2017-05-14
Support to get the primary output

hwangcc23:
This makes `primary` output available for assign or move commands.
Fix the issue #2764(https://github.com/i3/i3/issues/2764).

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1922,7 +1922,7 @@ output::
 ----------------------------------------------
 focus left|right|down|up
 focus parent|child|floating|tiling|mode_toggle
-focus output left|right|up|down|<output>
+focus output left|right|up|down|primary|<output>
 ----------------------------------------------
 
 *Examples*:
@@ -1944,7 +1944,16 @@ bindsym $mod+x focus output right
 
 # Focus the big output
 bindsym $mod+x focus output HDMI-2
+
+# Focus the primary output
+bindsym $mod+x focus output primary
 -------------------------------------------------
+
+-------------------------------
+Note that you might not have a primary output configured yet. To do so, run:
+-------------------------
+xrandr --output <output> --primary
+-------------------------
 
 === Moving containers
 
@@ -2162,8 +2171,8 @@ To move a container to another RandR output (addressed by names like +LVDS1+ or
 
 *Syntax*:
 ------------------------------------------------------------
-move container to output left|right|down|up|current|<output>
-move workspace to output left|right|down|up|current|<output>
+move container to output left|right|down|up|current|primary|<output>
+move workspace to output left|right|down|up|current|primary|<output>
 ------------------------------------------------------------
 
 *Examples*:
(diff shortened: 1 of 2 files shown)


docs/userguide
```diff
@@ -1922,7 +1922,7 @@ output::
 ----------------------------------------------
 focus left|right|down|up
 focus parent|child|floating|tiling|mode_toggle
-focus output left|right|up|down|<output>
+focus output left|right|up|down|<output|primary>
```
Airblader: This should actually read

    focus output left|right|up|down|primary|<output>

docs/userguide
```diff
@@ -2174,8 +2183,17 @@ bindsym $mod+x move workspace to output right
 
 # Put this window on the presentation output.
 bindsym $mod+x move container to output VGA1
+
+# Put this window on the primary output.
+bindsym $mod+x move container to primary output
```
Airblader: Shouldn't this be »output primary« instead of »primary output«?

src/randr.c
```diff
@@ -45,10 +45,14 @@ static Output *get_output_by_id(xcb_randr_output_t id) {
  */
 Output *get_output_by_name(const char *name) {
     Output *output;
-    TAILQ_FOREACH(output, &outputs, outputs)
-    if (output->active &&
-        strcasecmp(output->name, name) == 0)
-        return output;
+    int get_primary = strcasecmp("primary", name) == 0 ? 1 : 0;
```
Airblader: Please use a bool instead of an int for this. We can also avoid the redundant ternary:

    bool get_primary = (strcasecmp(name, "primary") == 0);

src/randr.c
```diff
@@ -45,10 +45,14 @@ static Output *get_output_by_id(xcb_randr_output_t id) {
...
+    TAILQ_FOREACH(output, &outputs, outputs) {
+        if (output->primary && get_primary)
```
Airblader: Please us ` {}` braces for all conditionals.

src/randr.c
```diff
@@ -45,10 +45,14 @@ static Output *get_output_by_id(xcb_randr_output_t id) {
...
+            return output;
+        if (output->active &&
```
Airblader: These two conditions can be on the same line.

Airblader (changes requested):
Thanks for the PR. There are a couple of minor comments, see review. Also, can we please add a test for at least one of the commands just so this path is tested?

hwangcc23:
> There are a couple of minor comments, see review. 

@Airblader 
Thanks for the review. Code is updated.

> Also, can we please add a test for at least one of the commands just so this path is tested?

There is a problem: I need to use 'xrandr' to configure a primary output, but I cannot use 'xrandr' with Xephyr.
See https://github.com/i3/i3/pull/1638.

Airblader (approved):

Airblader:
Thanks!

Issue #2640, opened 2017-01-13
Testsuite docs don't reflect the autotools update


Airblader:
The testsuite documentation should be updated to reflect the move to autotools we made. Everything has to be executed from the build directory now, not the source directory.

i3 4.13

stapelberg:
Fixed with commit bbfdcfddfeffded9c9f38797d923e11e1278aa79

PR #2777, opened 2017-05-21
Update testsuite docs

hwangcc23:
After moved to autotools, the method to run the i3 testsuite is changed.
Update the testsuite document to reflect the change.


Airblader:
Thanks. This fixes #2640

Issue #2511, opened 2016-10-21
Restarting i3 results in crash


aksr:
i3 version 4.12-133-ge51a89e

http://sprunge.us/MTHY

**What I did:**
1. i3 config: http://sprunge.us/BZhG
2. run goldendict*
3. restart (via key-keybinding)

**What I saw:**
i3 crashes

**What I expected instead:**
For i3 to be restarted normally.

**A few things more...**
This doesn't happen with 4.12 (I mean i3 doesn't crash with above setting.).
I'm not sure if the same is true for freezes.

*_The above bug can be reproduced using `firefox` (no need for goldendict).
Just the config file should be properly changed.
For example: `for_window [instance="^Navigator$" class="^Firefox$"] mark f, border normal 1`_

_Also, I'm not sure if the freezes are somehow related to this bug. 
(I'm not yet able to locate when or why they happen.)
But, to get i3 to be responsive again—I have to send signal `SIGCONT` to i3._

If you need more info, or, if I'm not specific enough, _please_ do tell.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

Airblader:
Can you bisect this issue between 4.12 and the commit you're using?

aksr:
I'm not sure If I have enough time atm. I tried to get some input on IRC before making noise on the github.

Airblader:
Pulling a gdb backtrace would also help. :-)

If you're not faster I'll try to repro this on the weekend and try to get it fixed. @stapelberg We should wait with 4.13 for this.

aksr:
Fair enough, thank you.

aksr:
Even the version 4.12 freezes randomly(?) (i3 is made responsive by sending signal SIGCONT to it).
I'm not able to find out when or why it freezes (yet).
There are others with the same issue: https://marc.ttias.be/arch-general/2016-09/msg00067.php
This means, freezes and the crash (described above) are *two separate* bugs.

I should probably open another bug issue.

Airblader:
For the freeze a backtrace might be helpful as well because we'd perhaps see where it's stuck. But yeah, it sounds like a separate issue.

Airblader:
I cannot reproduce the issue, so please provide the backtrace and, ideally, bisect the issue. :-)

aksr:
Which issue, the first or the second?

[...]

PR #2779, opened 2017-05-22
Fix the i3 crash caused by mark + restart commands

hwangcc23:
This patch fixes the issue #2511(https://github.com/i3/i3/issues/2511).

1). Memorize the marks, but only call con_mark once the container has finished parsing. (Credit: This is @Airblader's patch.)

2). Add a test case 267-regress-mark-restart.t for regression test to check if mark and restart command crash i3.


Airblader:
@stapelberg Since this is effectively my patch I think it'd be better if you could review this.

PR #2781, opened 2017-05-22
Apply numlock fallback to bindcode where necessary

### LINKED ISSUE #2559: bindcode considers numlock state for non-numpad keys

```
Running i3 version: 4.13-7-gad7dec3 (2016-11-12, branch "next") (pid 22254)
Loaded i3 config: /home/klemens/.config/i3/config (Last modified: So 13 Nov 2016 15:15:45 CET, 100 seconds ago)

The i3 binary you are running: /repos/testle/i3/build/i3
```

The following commands are needed in version 4.13 when using bindcode while in 4.12 the first one was enough:

```
bindcode --release 133 exec rofi
bindcode --release Mod2+133 exec rofi
```

This was introduced in d77d40173aae250d2e5ba86a8bce7a6e857cbc67. If it is intended behaviour, I think it should be documented better, as the release notes state that this only affects numpad keys:
> only add numlock fallback for keybindings where necessary (**allows users to correctly bind keys on the numpad**)

stapelberg:
This was broken with commit d77d40173aae250d2e5ba86a8bce7a6e857cbc67

fixes #2559

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -224,8 +224,9 @@ static Binding *get_binding(i3_event_state_mask_t state_filtered, bool is_releas
     const uint32_t xkb_group_state = (state_filtered & 0xFFFF0000);
     const uint32_t modifiers_state = (state_filtered & 0x0000FFFF);
     TAILQ_FOREACH(bind, bindings, bindings) {
-        if (bind->input_type != input_type)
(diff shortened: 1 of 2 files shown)


stapelberg:
Alright, then, in it goes. Feel free to revert at any point in case it causes trouble.

PR #2795, opened 2017-05-31
Fix manpage on configuration lookup order

Adaephon-GH:
The lookup order stated in the i3 manpage was is contratiction to the
actual lookup order, which was introduced with commit
https://github.com/i3/i3/commit/bfa12a581915d6a3de182fa6025fce108cac8eab.
Since that commit (6.5 years ago) the "traditional" paths took precedence
over XDG_CONFIG paths.

### DIFF
--- a/man/i3.man
+++ b/man/i3.man
@@ -170,10 +170,10 @@ Exits i3.
 
 When starting, i3 looks for configuration files in the following order:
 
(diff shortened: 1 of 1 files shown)


Airblader:
Sorry for the delay on this, but thanks for the update. :-)

Issue #2802, opened 2017-06-03
Renaming current workspace to "to<something>" fails


jolange:
Output of `i3 --moreversion 2>&- || i3 --version`:
```
Binary i3 version:  4.13 (2016-11-08) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.13-64-g4a2d5da0 (2017-02-20, branch "next") (pid 28535)
Loaded i3 config: /home/jo/cfg/i3/devel.conf (Last modified: Thu 27 Apr 2017 06:36:26 CEST, 3243685 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: /home/jo/projects/i3/build/i3
```

URL to a logfile as per http://i3wm.org/docs/debugging.html:

The relevant part:
```
03/06/17 19:35:00 - ipc.c:ipc_new_client:1194 - IPC: new client connected on fd 9
03/06/17 19:35:00 - IPC: received: *rename workspace to totally_important_ws*
03/06/17 19:35:00 - commands_parser.c:parse_command:265 - COMMAND: *rename workspace to totally_important_ws*
03/06/17 19:35:00 - commands.c:cmd_criteria_init:163 - Initializing criteria, current_match = 0x73c420
03/06/17 19:35:00 - Renaming workspace "to" to "tally_important_ws"
03/06/17 19:35:00 - commands.c:cmd_criteria_init:163 - Initializing criteria, current_match = 0x73c420
03/06/17 19:35:00 - ipc.c:ipc_receive_message:1152 - IPC: client disconnected
```

**What I did:**

Try to rename the current workspace to something starting with "to".

**What I saw:**
The rename fails, because the parsing seems to fail (as the log indicates, too). The start of the `to<something>` is interpreted as the `to` belonging to the command, while the `to` is assumed to be the old workspace name:
```
$ i3-msg "rename workspace to totally_important_ws"
ERROR: Your command: (null)
ERROR:               (null)
ERROR: Old workspace "to" not found
[{"success":false,"error":"Old workspace \"to\" not found"}]
```

Having a "to" contained somewhere else in the new workspace name (not at the beginning) works just fine.

**What I expected instead:**
Proper parsing and renaming.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

Adaephon-GH:
[...]

PR #2799, opened 2017-06-02
Add compilation instructions to the hacking howto

hwangcc23:
Fix #2792.

PR #2808, opened 2017-06-11
Fix 'rename workspace to tosomething'

hwangcc23:
See the issue #2802.
Revise the state machine for the 'rename workspace' command.
These scenarios are considered:
    a). 'rename workspace to to bla'
        state transitions: RENAME -> RENAME_WORKSPACE -> RENAME_WORKSPACE_LIKELY_TO -> RENAME_WORKSPACE_LIKELY_TO_NEW_NAME
    b). 'rename workspace to tosomething'
        state transitions: RENAME -> RENAME_WORKSPACE -> RENAME_WORKSPACE_LIKELY_TO
    c). 'rename workspace to to'
        state transitions: RENAME -> RENAME_WORKSPACE -> RENAME_WORKSPACE_LIKELY_TO
    d). 'rename workspace to bla'
        state transitions: RENAME -> RENAME_WORKSPACE -> RENAME_WORKSPACE_LIKELY_TO
    e). 'rename workspace bla to foo'
        state transitions: RENAME -> RENAME_WORKSPACE -> RENAME_WORKSPACE_TO -> RENAME_WORKSPACE_TO_NEW_NAME

### DIFF
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -274,24 +274,28 @@ state RENAME:
       -> RENAME_WORKSPACE
 
 state RENAME_WORKSPACE:
-  old_name = 'to'
+  'to'
       -> RENAME_WORKSPACE_LIKELY_TO
   old_name = word
       -> RENAME_WORKSPACE_TO
 
 state RENAME_WORKSPACE_LIKELY_TO:
-  'to'
-      -> RENAME_WORKSPACE_NEW_NAME
+  'to '
+      -> RENAME_WORKSPACE_LIKELY_TO_NEW_NAME
   new_name = word
       -> call cmd_rename_workspace(NULL, $new_name)
 
-state RENAME_WORKSPACE_TO:
-  'to'
-      -> RENAME_WORKSPACE_NEW_NAME
-
-state RENAME_WORKSPACE_NEW_NAME:
+state RENAME_WORKSPACE_LIKELY_TO_NEW_NAME:
+  new_name = string
+      -> call cmd_rename_workspace("to", $new_name)
   end
       -> call cmd_rename_workspace(NULL, "to")
+
+state RENAME_WORKSPACE_TO:
+  'to'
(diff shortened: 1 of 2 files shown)


hwangcc23:
@Airblader The build job of Travis CI with the compiler gcc is always failed in these days. It failed in my PR #2799 and #2808. I guess it is not caused by my changes. Could you check it? Thanks.

Error message:
> The command "docker pull ${BASENAME} || ./travis/docker-build-and-push.sh ${BASENAME} travis/travis-base.Dockerfile" failed and exited with 100 during .

parser-specs/commands.spec
```diff
@@ -274,24 +274,28 @@ state RENAME:
...
-  old_name = 'to'
+  'to'
       -> RENAME_WORKSPACE_LIKELY_TO
   old_name = word
       -> RENAME_WORKSPACE_TO
 
 state RENAME_WORKSPACE_LIKELY_TO:
-  'to'
-      -> RENAME_WORKSPACE_NEW_NAME
+  'to '
```
Airblader: I don't think I like this, it feels like more of a workaround. The actual parser should just identify the tokens correctly, IMHO. @stapelberg What do you think?

Airblader:
Regarding the breaking build, I restarted the job for the other PR and it worked. It looked like there was just a network problem.

parser-specs/commands.spec
stapelberg: We use the same technique in `MOVE_WORKSPACE` already, so it would at least be consistent. If you have a cleaner technique, I’m listening :)

Airblader: Alright; if we use it already then I think we're not making the code worse than it already is.

Airblader:
@hwangcc23 Can you adapt the commit message so it actually states that this "fixes #2802"?

hwangcc23:
@Airblader done

PR #2813, opened 2017-06-14
[WIP/RFC] Add mouse_warping=always config option

tohojo:
This adds an option to set mouse_warping to 'always', which will warp the cursor
to the middle of the window whenever focus is changed, not just when changing
between different outputs.

I find this behaviour to work better with focus follows mouse (mouse pointer is always in the focused window).

There are still some rough edges that need to be ironed out; most notably, changing windows by scrolling the mouse wheel on a titlebar shouldn't warp the pointer. But before looking into fixing that I wanted to check whether this was a patch you will accept, assuming I fix those issues?


Airblader:
Just a first random thought: I can somewhat understand the desire for this when _not_ using focus follows mouse, but you say you _specifically_ prefer it using that? That surprises me; I'd think it would be extremely annoying to have the mouse jump around while using it.

tohojo:
Ingo Bürk <notifications@github.com> writes:

> Just a first random thought: I can somewhat understand the desire for
> this when not using focus follows mouse, but you say you specifically
> prefer it using that? That surprises me; I'd think it would be
> extremely annoying to have the mouse jump around while using it.

Yeah, I've been using such a config with xmonad for years, and seem to
have gotten into the habit of alternating between switch focus with the
mouse and the keyboard. So I'd find myself switching focus with the
keyboard, then later using the mouse to "point myself" back to another
window. But because the mouse pointer was already over that window, I'd
just end up twiddling the pointer a bit, and focus would not change, so
I'd be typing back in the other window.

This was annoying me enough that I ended up writing this patch,
basically :)

-Toke

tohojo:
I'm using an IBM style pointing stick, BTW, which means going back and forward between mouse and keyboard does not entail moving my hands (which I suppose is why I developed this habit).

Airblader:
@stapelberg What should we do with this?

stapelberg:
I don’t think the complexity and lingering intricacies of supporting this feature are worth it — sorry.

stefanpl:
@oliversturm wrote a nice little script using `xdotool` to achieve this functionality. Have a look at [#2971](https://github.com/i3/i3/issues/2971#issuecomment-365268904). Works great for me!

chrisduerr:
I just wrote a very similar patch (of course without first looking for existing PRs because why would I?) and while I'm *somewhat* fine with this kind of feature not being merged upstream, I'd like to present an interesting usecase: Scrolling with rotary encoders.

I've recently built a keyboard with a rotary encoder and struggled with the mouse pointer (which is where scrolling will occur) not following my keyboard focus. Thus I had to move the mouse manually just to go back to the keyboard to scroll my window.

PR #2827, opened 2017-07-06
Fix 'Mouse binding don't work'

akash-akya:
For mouse buttons we are not considering the masks. I changed to treat it as normal key.
Currently because of the default CapsLock mask, CapsLock specialized mouse click wont have effect. 
But do we need it?

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -474,8 +474,7 @@ void translate_keysyms(void) {
 
             xcb_keycode_t key = button;
             bind->keycode = key;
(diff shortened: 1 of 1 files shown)


stapelberg:
Sorry, I don’t have time to get into the headspace to review this in the next 2 weeks or so.

@Airblader If this looks good to you, feel free to merge it and see if it breaks anything — maybe send an email to i3-discuss inviting mouse binding users to test this change.

Airblader:
Since the OP said it fixes the issue for them and it sounds reasonable to me, let's go ahead and try it.

PR #2836, opened 2017-07-12
Move SYSCONFDIR definition to makefile

### LINKED ISSUE #2832: During build, SYSCONFDIR is set to NONE/etc
**TL;DR** Run `../configure --sysconfdir=/etc` before running `make`

```
$ git rev-parse HEAD
cea7d135a50d3b012a20a23f36cb5547ef9fa576
```
---

Following the steps from [`doc/hacking_howto`](https://github.com/i3/i3/blob/cea7d13/docs/hacking-howto#L998-L1001)

```
$ autoreconf -fi
$ mkdir -p build && cd build
$ ../configure
$ make -j8
```

generated `i3` might not start, as it will try to read `NONE/etc` as its configuration file.

To my understanding the problem is that the path of the system-wide configuration files directory is hard-coded into the source code during the `../configure` step via evaluating the following string

```
#define SYSCONFDIR "`eval echo $sysconfdir`"
```

When you run `../configure` though, the value of `sysconfdir` is `NONE/etc` as `autoconf` set `sysconfdir` to `'${prefix}/etc'` - while `prefix` is still `NONE` - such to enforce GNU coding standards:

```
... when the user runs:
`make'
    she can still specify a different prefix from the one specified to configure,
    in which case, if needed, the package should hard code dependencies corresponding
    to the make-specified prefix
`make install'
    she can specify a different installation location, in which case the package must still
    depend on the location which was compiled in (i.e., never recompile when
    `make install' is run). This is an extremely important feature, as many people may
    decide to install all the files of a package grouped together, and then install links from
    the final locations to there.
```
https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Installation-Directory-Variables.html

I see the following possible fixes.

1. Find a way to make `autoconf` expand those variables as the configure script is generated - not leaving them unexpanded. This should still be against the GNU coding standards but, should that work, it'd be an easy and quick fix.
2. Instead of relying on `config.h`, use a configuration file that's parsed by `i3` at runtime, moving from macro's to static variables.
3. Try to read an environment variable, fallback to `/etc`
4. As proposed [here](https://lists.gnu.org/archive/html/automake/2007-01/msg00091.html) use a script to sweep the problem under the carpet.

acrisci:
Official autoconf docs contraindicate the use of the $sysconfdir
variable in configure.ac. For reference:

https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Installation-Directory-Variables.html

The macro SYSCONFDIR (ordinarilly set to "/etc") is required to find
a system configuration file installed with i3. In some build setups,
this may not be defined properly in when the configure script is
compiled.

Instead, define this variable in AM_CPPFLAGS as the documentation
indicates.

fixes #2832

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -215,6 +215,7 @@ asciidoc_MANS =
 endif
 
 AM_CPPFLAGS = \
+	-DSYSCONFDIR="\"$(sysconfdir)\"" \
 	-I$(top_builddir)/parser \
 	-I$(top_srcdir)/include \
 	@AX_EXTEND_SRCDIR_CPPFLAGS@
--- a/configure.ac
+++ b/configure.ac
@@ -59,8 +59,6 @@ AX_CHECK_ENABLE_DEBUG([yes], , [UNUSED_NDEBUG], [$is_release])
 
 AC_PROG_CC_C99
 
-AC_DEFINE_UNQUOTED(SYSCONFDIR, "`eval echo $sysconfdir`", [Location of system configuration files])
-
 # For strnlen() and vasprintf().
 AC_USE_SYSTEM_EXTENSIONS
 


acrisci:
I tried to test that the behavior was the same for the install location of the configuration files to the sysconfdir but I might have missed some edgecases.

PR #2837, opened 2017-07-12
Improve git detection in configure.ac

### LINKED ISSUE #2833: git is required to build i3
I just ran into this issue building in a vm without git.

The build fails with

```
In file included from ../../i3/src/ipc.c:12:0:
../../i3/src/ipc.c: In function ‘handle_get_version’:
../../i3/include/yajl_utils.h:19:50: error: expected expression before ‘)’ token
 #define y(x, ...) yajl_gen_##x(gen, ##__VA_ARGS__)
                                                  ^
../../i3/src/ipc.c:908:5: note: in expansion of macro ‘y’
     y(integer, MAJOR_VERSION);
     ^
../../i3/include/yajl_utils.h:19:50: error: expected expression before ‘)’ token
 #define y(x, ...) yajl_gen_##x(gen, ##__VA_ARGS__)
                                                  ^
../../i3/src/ipc.c:911:5: note: in expansion of macro ‘y’
     y(integer, MINOR_VERSION);
```

I believe the problem is that those don't get defined properly. After installing git, it worked.

We should add a check in `configure.ac` to make sure git is installed.

acrisci:
During the configure step, if the git binary is not on the system, the
*_VERSION variables will not be set properly which will cause a build
error.

Fix this bug by falling back to the non-git version detection scheme
when git is not installed.

fixes #2833

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,10 @@ AS_VAR_IF([_cv_gnu_make_command], [""], [AC_MSG_ERROR([the i3 Makefile.am requir
 
 AX_EXTEND_SRCDIR
 
-AS_IF([test -d ${srcdir}/.git],
+AC_CHECK_PROG([git],[git],[yes],[no])
+AM_CONDITIONAL([FOUND_GIT], [test "x$git" = xyes && test -d ${srcdir}/.git])
+
+AS_IF([FOUND_GIT],
       [
         VERSION="$(git -C ${srcdir} describe --tags --abbrev=0)"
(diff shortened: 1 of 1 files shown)


acrisci:
Interesting...

```
#   Failed test 'human readable version contains branch name'
#   at /usr/src/i3/build/../../i3/testcases/t/193-ipc-version.t line 35.
#                   '4.13-non-git'
#
```

Airblader:
What's the status here?

PR #2839, opened 2017-07-15
[WIP/RFC] Handle a case of a soon to be gone window calling XSetInputFocus

errge:
This is a rare case happening mostly for emacs-helm users.  It
can be reproduced without any special emacs configuration using
the x-focus-frame function and then immediately quitting emacs.

The problem was that in this case emacs calls XSetInputFocus with
revert to parent setting, which afterwards at the exit of emacs causes
keyboard focus to be lost and not set to the ewmh support window.

About my proposed patch: I know it's big and disgusting, but the whole test
set runs without error and I also added a new testcase.

Actually, changing back the last_timestamp to XCB_CURRENT_TIME
around the ewmh_window focusing in x.c is enough to make the bug gone, but
according to the git log there are other problems if we use XCB_CURRENT_TIME
for ewmh_window focusing.

Therefore I have set out to fix the actual bug that is being triggered by emacs.

Why my patch grew this big, is because when I detect the unmapping event that
can trigger this bug, I have to get a new xserver timestamp.  And the only way to
do this (at least this is how openbox is doing it) is to set a property and wait for
the resulting event.  But while I'm waiting for the event, I have to queue up the
events that otherwise would be lost forever and then reapply them at the beginning
of the queue.

Another option would be to introduce a new global variable called something like
"should_we_use_current_time_for_focusing_after_unmap" and set this to true
before tree_render in handle_unmap_notify_event and set back to false afterwards.
That will be definitely a much more easy to review changeset, if you are not worried
about the additional global variable.

Another option would be, that I make the "pushbackable event queue" a separate
refactoring that implements a "look_for_specific_event_in_queue_without_popping"
function.  Once we have this refactoring, the actual bugfix changeset will be nice.

I'd like to ask for your kind input regarding this and a discussion around which way
should we go from here.

### DIFF
--- a/include/handlers.h
+++ b/include/handlers.h
@@ -38,7 +38,7 @@ bool event_is_ignored(const int sequence, const int response_type);
  * event type.
  *
  */
-void handle_event(int type, xcb_generic_event_t *event);
+void handle_event(int type, xcb_generic_event_t* event, bool need_new_timestamp);
 
 /**
  * Sets the appropriate atoms for the property handlers after the atoms were
@@ -46,6 +46,15 @@ void handle_event(int type, xcb_generic_event_t *event);
  *
  */
 void property_handlers_init(void);
+
+/**
+ * These two functions implement an event queue, that can be used when
+ * we synchronously have to wait for a specific event in the middle of
+ * business logic, but we don't want to drop the events that are in
+ * front of it in the event queue.
+ */
+void queue_event(xcb_generic_event_t*);
+xcb_generic_event_t* pop_event();
 
 #if 0
 /**
--- a/src/ewmh.c
+++ b/src/ewmh.c
@@ -311,8 +311,8 @@ void ewmh_setup_hints(void) {
         0,                           /* border */
         XCB_WINDOW_CLASS_INPUT_ONLY, /* window class */
         XCB_COPY_FROM_PARENT,        /* visual */
-        XCB_CW_OVERRIDE_REDIRECT,
-        (uint32_t[]){1});
+        XCB_CW_EVENT_MASK | XCB_CW_OVERRIDE_REDIRECT,
+        (uint32_t[]){1, XCB_EVENT_MASK_PROPERTY_CHANGE});
     xcb_change_property(conn, XCB_PROP_MODE_REPLACE, ewmh_window, A__NET_SUPPORTING_WM_CHECK, XCB_ATOM_WINDOW, 32, 1, &ewmh_window);
     xcb_change_property(conn, XCB_PROP_MODE_REPLACE, ewmh_window, A__NET_WM_NAME, A_UTF8_STRING, 8, strlen("i3"), "i3");
     xcb_change_property(conn, XCB_PROP_MODE_REPLACE, root, A__NET_SUPPORTING_WM_CHECK, XCB_ATOM_WINDOW, 32, 1, &ewmh_window);
--- a/src/floating.c
+++ b/src/floating.c
@@ -697,7 +697,7 @@ static void xcb_drag_check_cb(EV_P_ ev_check *w, int revents) {
             case XCB_KEY_PRESS:
                 DLOG("A key was pressed during drag, reverting changes.\n");
                 dragloop->result = DRAG_REVERT;
-                handle_event(type, event);
+                handle_event(type, event, false);
                 break;
 
             case XCB_UNMAP_NOTIFY: {
@@ -713,7 +713,7 @@ static void xcb_drag_check_cb(EV_P_ ev_check *w, int revents) {
                     }
                 }
 
-                handle_event(type, event);
+                handle_event(type, event, false);
                 break;
             }
 
@@ -725,7 +725,7 @@ static void xcb_drag_check_cb(EV_P_ ev_check *w, int revents) {
 
             default:
                 DLOG("Passing to original handler\n");
-                handle_event(type, event);
+                handle_event(type, event, false);
                 break;
         }
 
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -21,6 +21,35 @@ int randr_base = -1;
 int xkb_base = -1;
 int xkb_current_group;
 
+struct Queued_Event {
+    xcb_generic_event_t *ev;
+
+    TAILQ_ENTRY(Queued_Event)
+    queued_events;
+};
+
+static TAILQ_HEAD(queued_event_head, Queued_Event) queued_events = TAILQ_HEAD_INITIALIZER(queued_events);
+
+void init_queued_events() {
+    TAILQ_INIT(&queued_events);
+}
+
+void queue_event(xcb_generic_event_t *ev) {
+    struct Queued_Event *new = smalloc(sizeof(struct Queued_Event));
+    new->ev = ev;
+    TAILQ_INSERT_TAIL(&queued_events, new, queued_events);
+}
+
+xcb_generic_event_t *pop_event() {
+    if (TAILQ_EMPTY(&queued_events))
+        return NULL;
+    struct Queued_Event *next = TAILQ_FIRST(&queued_events);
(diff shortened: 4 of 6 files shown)


acrisci:
I can reproduce this with `emacs -Q --eval '(progn (x-focus-frame nil) (sleep-for 2) (kill-emacs))'`.

In `src/xfns.c`, is the function [`x_focus_frame()`](https://github.com/emacs-mirror/emacs/blob/master/src/xfns.c#L4100) with this comment:

<blockquote>
/* In certain situations, when the window manager follows a
   click-to-focus policy, there seems to be no way around calling
   XSetInputFocus to give another frame the input focus .

   In an ideal world, XSetInputFocus should generally be avoided so
   that applications don't interfere with the window manager's focus
   policy.  But I think it's okay to use when it's clearly done
   following a user-command.  */
</blockquote>

I believe that in emacs terminology, a "frame" refers to a top-level managed window.

I think you could make the case that any application that calls `XSetInputFocus` on a top-level window is broken. The client should be using the `_NET_ACTIVE_WINDOW` client request and following the [icccm spec for input focus](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7). If that resolves the issue, it is preferable to the solution presented here.

acrisci:
This patch to emacs resolves the issue:

```diff
diff --git a/src/xfns.c b/src/xfns.c
index d8bf974..5d11096 100644
--- a/src/xfns.c
+++ b/src/xfns.c
@@ -4113,10 +4113,7 @@ x_focus_frame (struct frame *f, bool noactivate)
     }
   else
     {
-      XSetInputFocus (FRAME_X_DISPLAY (f), FRAME_X_WINDOW (f),
-		      RevertToParent, CurrentTime);
-      if (!noactivate)
-	x_ewmh_activate_frame (f);
+        x_ewmh_activate_frame (f);
     }
 
   x_uncatch_errors ();
```

The `x_ewmh_activate_frame()` function uses `_NET_ACTIVE_WINDOW` to focus the window. This is the correct way to request the window manager focus your window. It should be used when the `_NET_ACTIVE_WINDOW` atom is present in `NET_SUPPORTED` property of the root window.

Please open an issue with emacs and link to it here and I will help you make the case for this change.

errge:
@acrisci You are 100% wrong here, sorry to say.

In general, it's a very very bad idea to blame the project that is having comments when they are not sure that they are doing the right thing and point at that comments saying "even they think it's bad, so it must be bad".  I think you should be much more quick to blame a project, which have exchanged last_timestamp and XCB_CURRENT_TIME forth and back at least 4 times in the last 2 years and there are no comments at all around that code.  No commiters left a comment why the change is being made or why do they think the same solution gonna work this time around if it didn't work previously.  Emacs committers left a clue, so now you are blaming them.  Look into the i3 source code history around the "Still no window focused, better set focus" in x.c and see how many times XCB_CURRENT_TIME and last_timestamp is being exchanged because of random bugs.  On odd years we have last_timestamp, on even years we have XCB_CURRENT_TIME.  Most probably I don't have to do anything and i3 will work in 2018.

Also, I think you didn't understand the things that i3 is doing wrong, because my comment was too long and you were immediately concentrating on what Emacs is doing wrong, so let me explain one more time with an example, imaginary timeline:

1. 1980-01-01: i3 started
2. 1980-01-02: emacs is started
3. 1980-01-10: user needs a terminal, uses it for 5 minuets, so the X event timestamp is recorded in i3 memory as last_timestamp
4. 1981: user is using emacs the whole time, maybe uses XSetInputFocus once on the already activated window (frame in emacs terminology)
5. 1982-2006: user is using emacs during the whole time, no X events for i3
6. 2007-01-01: user quits emacs, which leaves the workspace empty, that causes X to lose input focus
7. 2007-01-01: i3 realizes that there is an emergency and there are no windows handling keyboard focus and i3 is unusable in this state, so to fix the emergency i3 sends the following message to X: "please activate my EWMH window, my current clock is 1980-01-10, use this clock information to handle race conditions correctly".

Put another way, your reproduction command will also work with sit-for 86400, if you don't touch your computer for a day.  Don't you think, that's strange?  The fact that this bug is only happening with i3 and not with other tiling window managers is also a fact that you should consider.

Are you still sure that Emacs is doing the wrong thing here?  I mean EWMH and ICCCM standards are here for a couple of years already, but I'm pretty sure X11 timestamping is a more basic concept and in this corner case i3 completely shits the bed regarding that.

And if your solution to this problem is to remove XSetInputFocus from the client application, then you shouldn't patch Emacs, but the X server, to refuse those calls.  It was always clear to me, that I can change Emacs to not call XSetInputFocus, but I was thinking if I do all the work and ask for advice politely on how to proceed, then you guys are interested in contributions, but I was wrong...

errge:
"Explanations exist; they have existed for all time; there is always a well-known solution to every human problem — neat, plausible, and wrong."  Usually quoted as: For every complex problem there is an answer that is clear, simple, and wrong.

I will not go to the GNU Emacs mailing list and ask them to do something that even I know is WRONG, just because it's neat, plausible, short, quick and easy.  I provided a testcase, I explained the problem, I offered to do the free coding for you to change my patch to your liking (after already spending 2 days just on the debugging).  If you want to use your own reputation, feel free to go around and write to every mailing list that governs a project that uses XSetInputFocus, I don't think it will be a fun (or useful) thing to do.

acrisci:
I apologize if I have upset you. It's obvious that you put a lot of effort into this and I should have been more sensitive to that in my comments.

Airblader:
All that being said, I do agree that the client is broken here. A client that forcefully rips focus control from the window manager without any regard to other clients should be expected to at least properly clean up after itself. 

The specifications exist for good reasons and in these times clients should be expected to follow them — or considered to be broken.

errge:
OK, thanks, I moved to i3 because I don't want to touch my mouse.  If breaking with Emacs in a way, that you can only recover with your mouse and only if you have a backup window to clic to is considered normal in i3 and it's not a bug, then I will have to look around, good to know.

Also, if sending out requests to the X server with a day old timestamp is not considered a bug, then I don't know what will be.

Also, I feel like that you guys don't care about the bug, just hate the many added lines, if I provide you with a one-liner addition that has nothing to do with the bug, but in practice fixes this behavior for most of the users, will you at least consider looking at that?  Instead of bashing the programs that I have been using for 15 years?  1 line fix in i3 is less than 3-5 in Emacs, after all.

Willing to share it on a new PR, but only if discussion will be around how it is good or bad for i3, not if you come at me for the programs I use.

Airblader:
> If breaking with Emacs in a way, that you can only recover with your mouse and only if you have a backup window to click to is considered normal in i3 and it's not a bug

It certainly is a bug – in emacs.

> Also, if sending out requests to the X server with a day old timestamp is not considered a bug

The timestamp mechanism exists precisely to avoid race conditions. Disregarding the mechanism because the timestamp is too old voids its entire purpose.

There is no reason for emacs to behave this way. There is no reason why emacs can't go through the window manager to request focus instead of violently ripping focus away. And if emacs for whatever reason decides to ignore the window manager's authority, they should at least be expected to properly clean up the mess they left behind.

> if I provide you with a one-liner addition that has nothing to do with the bug, but in practice fixes this behavior for most of the users, will you at least consider looking at that?

We will look at every patch we get sent (and are happy to discuss changed before you go through the trouble of writing the patch), but we will not merge it just because it's simple or just because it works around an issue; we merge fixes for things that need to be fixed in i3. Even if it's a big patch.

> Instead of bashing the programs that I have been using for 15 years? 

No one has said anything against emacs other than the fact that we think it has a bug. For which @acrisci  even provided a patch already.

I don't care much for your aggressions and unfounded attacks that we don't care about bugs, bash clients or go into any kind of politics and for this reason I am going to lock this discussion for not being productive. If you want to discuss a potential change, please open an issue where we can discuss it and we will do so, provided you can keep out the ad-hominems.

PR #2841, opened 2017-07-17
nagbar: autodismiss after given seconds

jwdeitch:
Requires compiling with -lpthreads

### DIFF
--- a/i3-nagbar/main.c
+++ b/i3-nagbar/main.c
@@ -25,6 +25,8 @@
 #include <limits.h>
 #include <fcntl.h>
 #include <paths.h>
+#include <time.h>
+#include <pthread.h>
 
 #include <xcb/xcb.h>
 #include <xcb/xcb_aux.h>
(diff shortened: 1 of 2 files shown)


Airblader:
Thanks for the contribution. Can you explain what this change is needed for? The nagbar is supposed to do its title justice, so automatically closing it goes against its purpose.

jwdeitch:
In my particular use-case, I have configured a keybinding to take a screenshot of a region on my screen and upload to S3. I wanted a way to notify me that the upload was complete and that the link was in my clipboard, without requiring my input. I can also see this being useful with emails, irc pings, etc.

Airblader:
Out of curiosity, why not run an actual notification daemon like Dunst and using send-notify to do this? 

i3-nagbar isn't really meant for transient notifications, but rather for nagging the user that an action is required (for example, config errors).

acrisci:
I think I agree with @Airblader here. Nagbar has a pretty limited use case within the project, primarily to notify users of configuration errors. It's not intended to be a general purpose application for notifications.

It sounds like your problem is much better solved by a real notification daemon. Lots of people like dunst but xfce also has a very nice standalone notification app. See `man notify-send` for more information.

jwdeitch:
Makes sense. Thanks for the feedback!

PR #2847, opened 2017-07-30
Document missing case of workspace event

tmerr:
After https://github.com/i3/i3/pull/2656 it looks like the only workspace event that's missing from the docs is workspace:move.

```
$ grep -r ipc_send_workspace_event
src/commands.c:        ipc_send_workspace_event("restored", parent, NULL);
src/commands.c:    ipc_send_workspace_event("reload", NULL, NULL);
src/commands.c:    ipc_send_workspace_event("rename", workspace, NULL);
src/move.c:    ipc_send_workspace_event("focus", ws, old_ws);
src/workspace.c:        ipc_send_workspace_event("init", workspace, NULL);
src/workspace.c:    ipc_send_workspace_event("focus", workspace, current);
src/workspace.c:        ipc_send_workspace_event("urgent", ws, NULL);
src/workspace.c:        ipc_send_workspace_event("init", ws, NULL);
src/workspace.c:    ipc_send_workspace_event("move", ws, NULL);
src/ipc.c:void ipc_send_workspace_event(const char *change, Con *current, Con *old) {
include/ipc.h:void ipc_send_workspace_event(const char *change, Con *current, Con *old);
```

So this should probably be added to the docs yeah?


Airblader:
Thanks!

PR #2860, opened 2017-08-19
Merge the AnyEvent-I3 repository into i3/AnyEvent-I3

stapelberg:
That way, we can use newly introduced commands in the testsuite immediately, instead of having to wait until individual developers update their environment’s copy of AnyEvent::I3.

My next PR will introduce changes to `src/ipc.c` _and_ `AnyEvent-I3/` _and_ `testcases/`, and everyone will be able to immediately verify the changes :).

I’ll push the (prefixed) AnyEvent-I3 tags after this is merged.


stapelberg:
Yep, no changes in there — I literally merged the old commits into the repo. In the depths of the git CLI, there probably is a command to verify what I’m proposing here is bit-for-bit identical with the contents of the AnyEvent-I3 repository :).

PR #2864, opened 2017-08-20
Improve error messages for incorrect config lines starting with “set”

### LINKED ISSUE #2564: Config parser is ignoring set* wrong statement
Version:
```
Running i3 version: 4.13 (2016-11-08) (pid 23105) abort…)
Loaded i3 config: /home/ael/.config/i3/config (Last modified: Sun 13 Nov 2016 09:06:21 PM CET, 520 seconds ago)

The i3 binary you are running: i3-with-/i3
```
As far as I know also the `4.12` is affected too.

**Bug**
All the unsupported statement in the config that start with `set` are ignored.
The bug is very easy to reproduce, you just need to insert the following wrong statement into your config:
```set -> try_to_notice_this_statement <- :)```

The line is ignored and no error is raised.
If you make a typo like `set_from_sourcd` it could be very difficult to find out.

stapelberg:
fixes #2564

### DIFF
--- a/parser-specs/config.spec
+++ b/parser-specs/config.spec
@@ -17,7 +17,8 @@ state INITIAL:
   end ->
   error ->
   '#'                                      -> IGNORE_LINE
-  'set'                                    -> IGNORE_LINE
+  'set '                                   -> IGNORE_LINE
+  'set	'                                  -> IGNORE_LINE
   'set_from_resource'                      -> IGNORE_LINE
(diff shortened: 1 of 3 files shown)


Airblader:
I'm only on mobile right now, so apologies if this is trivially answered by looking at more code, but why do we introduce this new boolean rather than using has_errors?

stapelberg:
`context` is only defined after the whole variable expansion thing is done. I figured it’d be easier to introduce a new variable than to verify the change in lifespan of an existing one.

Issue #2895, opened 2017-09-04
i3-msg swap container with con_id <invalid_con_id> crashes i3


orestisf1993:
Output of `i3 --moreversion 2>&- || i3 --version`:
```
Binary i3 version:  4.14-3-g5df53f5 (2017-09-04, branch "next") © 2009 Michael Stapelberg and contributors
(Getting version from running i3, press ctrl-c to abort…)
Running i3 version: 4.14-3-g5df53f5 (2017-09-04, branch "next") (pid 844)
Loaded i3 config: /home/orestis/.config/i3/config (Last modified: Mon 04 Sep 2017 06:26:40 PM EEST, 2555 seconds ago)

The i3 binary you just called: /usr/local/bin/i3
The i3 binary you are running: i3
```

URL to a logfile as per http://i3wm.org/docs/debugging.html:

http://logs.i3wm.org/logs/5726488619909120.bz2

**What I did:**
Open a window and
```
i3-msg swap container with con_id 1
```

**What I saw:**
i3 crashes

**What I expected instead:**
Nothing really I was just trying out the new command. It shouldn't crash though.

Airblader:
I can reproduce this. The problem is that we just cast the argument to `(Con *)` which then fails. I did it this way because we do the same for match criteria parsing, but since there we first compare the result with a container that we know is actually a container, we never run into this issue.

@stapelberg Is there a better way of checking the sanity of the argument other than comparing it against all known containers?

stapelberg:
> @stapelberg Is there a better way of checking the sanity of the argument other than comparing it against all known containers?

We don’t have any “magic ints” at the beginning of our structs (I think GObject does that), so the answer is no — we’ll need to compare against all known containers for the time being.

PR #2898, opened 2017-09-04
Fix crash with swap and invalid con_id

orestisf1993:
Fixes issue #2895.

I don't know if there is a more elegant way but at least it fixes the crash and seems to generally work.

### DIFF
--- a/include/con.h
+++ b/include/con.h
@@ -153,6 +153,13 @@ bool con_has_parent(Con *con, Con *parent);
 Con *con_by_window_id(xcb_window_t window);
 
 /**
+ * Returns the container with the given container ID or NULL if no such
+ * container exists.
+ *
+ */
+Con *con_by_con_id(long target);
+
+/**
  * Returns the container with the given frame ID or NULL if no such container
  * exists.
  *
--- a/src/commands.c
+++ b/src/commands.c
@@ -1841,7 +1841,7 @@ void cmd_swap(I3_CMD, const char *mode, const char *arg) {
             return;
         }
 
-        con = (Con *)target;
+        con = con_by_con_id(target);
     } else if (strcmp(mode, "mark") == 0) {
         con = con_by_mark(arg);
     } else {
--- a/src/con.c
+++ b/src/con.c
@@ -558,6 +558,22 @@ Con *con_by_window_id(xcb_window_t window) {
 }
 
 /*
+ * Returns the container with the given container ID or NULL if no such
+ * container exists.
+ *
+ */
+Con *con_by_con_id(long target) {
+    Con *con;
+    TAILQ_FOREACH(con, &all_cons, all_cons) {
+        if (con == (Con *)target) {
+            return con;
+        }
+    }
+
+    return NULL;
+}
+
+/*
  * Returns the container with the given frame ID or NULL if no such container
  * exists.
  *
--- a/testcases/t/265-swap.t
+++ b/testcases/t/265-swap.t
@@ -33,6 +33,38 @@ my ($result);
 my @urgent;
 
 ###############################################################################
+# Invalid con_id should not crash i3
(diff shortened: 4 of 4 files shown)


include/con.h
```diff
@@ -152,6 +152,13 @@ bool con_has_parent(Con *con, Con *parent);
  */
 Con *con_by_window_id(xcb_window_t window);
 
+/*
```
Airblader: Please use `**` for the header file (as the other functions).

src/commands.c
```diff
@@ -1841,7 +1841,7 @@ void cmd_swap(I3_CMD, const char *mode, const char *arg) {
             return;
         }
 
-        con = (Con *)target;
+        con = con_by_con_id(target);
```
Airblader: I do like this approach; it's consistent with the others.

src/con.c
```diff
@@ -558,6 +558,19 @@ Con *con_by_window_id(xcb_window_t window) {
 }
 
 /*
+ * Returns the container with the given container ID or NULL if no such
+ * container exists.
+ *
+ */
+Con *con_by_con_id(long target) {
+    Con *con;
+    TAILQ_FOREACH(con, &all_cons, all_cons)
```
Airblader: Please use braces (`{}`) both for `TAILQ_FOREACH` and the `if`.

src/con.c
```diff
@@ -558,6 +558,19 @@ Con *con_by_window_id(xcb_window_t window) {
...
+    if (con->window != NULL && con == (Con *)target)
```
Airblader: The check for a window is not necessary here and, in fact, not correct in my opinion. This function doesn't claim to check for a window, we have other functions for that already.

testcases/t/265-swap.t
```diff
@@ -33,6 +33,39 @@ my ($result);
...
 ###############################################################################
+# Invalid con_id should not crash i3
+# See issue #2895.
+###############################################################################
+
+$pid = launch_with_config($config);
+$ws = fresh_workspace;
+
+open_window;
+my $id = get_focused($ws) != 1 ? 1 : 2;
```
Airblader: We don't need this check; the con_id will never be something like `1` (and if it was, the test would still pass.)

Airblader: Also, let's please not introduce new variables here. I've designed this file such that all variables have more meaningful names and all variables are declared on top. This should just use `$A`.

testcases/t/265-swap.t
```diff
@@ -33,6 +33,39 @@ my ($result);
...
+
+###############################################################################
+# Swap 2 windows in different workspaces using con_id
+###############################################################################
+
+$pid = launch_with_config($config);
+
+$ws = fresh_workspace;
+open_window;
+$id = get_focused($ws);
```
Airblader: Same here. You actually call it »A« in the assertion, so let's make the test code easier to understand by naming the variables accordingly. :-)

Airblader:
Also, after the changes have been addressed, we'll need to squash the commits together.

Airblader:
One last thing, please drop a comment when you made the changes so that we get notified!

src/commands.c
orestisf1993: My only concern is that that the name `con_by_con_id` is a bit misleading because the only possible returned values are `target` after casting it to `(Con *)` or `NULL`.

orestisf1993:
I think everything is addressed. Should i squash everything together?

Airblader:
LGTM now. Can you squash it? Thanks!

orestisf1993:
Done

stapelberg:
We always merge into next, and cherry-pick into master if necessary.

stapelberg:
This is now cherry-picked.

PR #2901, opened 2017-09-07
Set marks to NULL after freeing 

### LINKED ISSUE #2900: Restart with Mark causes Exit

Binary i3 version:  4.14 (2017-09-04) © 2009 Michael Stapelberg and contributors
(Getting version from running i3, press ctrl-c tRunning i3 version: 4.14 (2017-09-04) (pid 8801)
Loaded i3 config: /home/bud/.config/i3/config (Last modified: ons  6 sep 2017 20:41:47, 952 seconds ago)

http://logs.i3wm.org/logs/5663741160980480.bz2

**What I did:**

restarted i3.

**What I saw:**

When executing the restart command, i3 exits to my login screen (lightdm) just like if the exit command would be executed. This only seems to happen if there are any marks. And it doesn't matter if it is from a binding (Mod4+Mod1+R in my case) or i3-msg restart.

**What I expected instead:**

A restart

*******

I'm sorry I haven't reported this earlier, i had the same issue with 4.13 , but I think it didn't happen in 4.12, but it might be me who didn't use marks back then.

My solution while waiting for a fix: having awk getting the con_id and con_mark of all containers and storing the values in an array (tmp[con_id]=con_mark), then doing i3-msg unmark (without a target, unmarking all containers), restart and then apply the marks again. 

Thanks for an amazing window manager.

orestisf1993:
Fixes #2900.

### DIFF
--- a/src/load_layout.c
+++ b/src/load_layout.c
@@ -157,6 +157,7 @@ static int json_end_map(void *ctx) {
             }
 
             free(marks);
+            marks = NULL;
             num_marks = 0;
         }
 
--- a/testcases/t/267-regress-mark-restart.t
+++ b/testcases/t/267-regress-mark-restart.t
@@ -20,6 +20,8 @@ use i3test;
 
 cmd 'open';
 cmd 'mark foo';
+cmd 'open';
+cmd 'mark bar';
 
 cmd 'restart';
 


orestisf1993:
Ok, I just realized that:

1. My testcase wasn't really for `swap`.
2. You already had a testcase that didn't catch the issue.

stapelberg:
Agreed. Let’s see if travis eventually decides to test this PR :)

orestisf1993:
done

budRich:
Excellent.

stapelberg:
…and cherry-picked into master. Thanks!

Issue #2810, opened 2017-06-13
[enhancement] Make "swap" work with floating windows


Airblader:
The `swap` command implemented in #2496 does not yet support floating windows. This should be implemented.

PR #2902, opened 2017-09-07
Make swap work with fullscreen windows

### LINKED ISSUE #2811: [enhancement] Make "swap" work with fullscreen windows
The `swap` command implemented in #2496 does not yet support fullscreen windows. This should be implemented.

orestisf1993:
Fixes #2811

### DIFF
--- a/src/con.c
+++ b/src/con.c
@@ -2276,15 +2276,14 @@ bool con_swap(Con *first, Con *second) {
     Con *current_ws = con_get_workspace(old_focus);
     const bool focused_within_first = (first == old_focus || con_has_parent(old_focus, first));
     const bool focused_within_second = (second == old_focus || con_has_parent(old_focus, second));
+    fullscreen_mode_t first_fullscreen_mode = first->fullscreen_mode;
+    fullscreen_mode_t second_fullscreen_mode = second->fullscreen_mode;
 
-    if (!con_fullscreen_permits_focusing(first_ws)) {
-        DLOG("Cannot swap because target workspace \"%s\" is obscured.\n", first_ws->name);
-        return false;
+    if (first_fullscreen_mode != CF_NONE) {
+        con_disable_fullscreen(first);
     }
-
-    if (!con_fullscreen_permits_focusing(second_ws)) {
-        DLOG("Cannot swap because target workspace \"%s\" is obscured.\n", second_ws->name);
-        return false;
+    if (second_fullscreen_mode != CF_NONE) {
+        con_disable_fullscreen(second);
     }
 
     double first_percent = first->percent;
@@ -2385,7 +2384,17 @@ bool con_swap(Con *first, Con *second) {
     second->percent = first_percent;
     fake->percent = 0.0;
 
+    SWAP(first_fullscreen_mode, second_fullscreen_mode, fullscreen_mode_t);
+
 swap_end:
+    /* The two windows exchange their original fullscreen status */
+    if (first_fullscreen_mode != CF_NONE) {
+        con_enable_fullscreen(first, first_fullscreen_mode);
+    }
+    if (second_fullscreen_mode != CF_NONE) {
+        con_enable_fullscreen(second, second_fullscreen_mode);
+    }
+
     /* We don't actually need this since percentages-wise we haven't changed
      * anything, but we'll better be safe than sorry and just make sure as we'd
      * otherwise crash i3. */
--- a/include/util.h
+++ b/include/util.h
@@ -57,6 +57,13 @@
 
 #define CALL(obj, member, ...) obj->member(obj, ##__VA_ARGS__)
 
+#define SWAP(first, second, type) \
+    do {                          \
+        type tmp_SWAP = first;    \
+        first = second;           \
+        second = tmp_SWAP;        \
+    } while (0)
+
 int min(int a, int b);
 int max(int a, int b);
 bool rect_contains(Rect rect, uint32_t x, uint32_t y);
--- a/testcases/t/291-swap.t
+++ b/testcases/t/291-swap.t
@@ -25,9 +25,16 @@ for_window[class="mark_B"] mark B
 EOT
 
 my ($ws, $ws1, $ws2, $ws3);
-my ($nodes, $expected_focus, $A, $B, $F);
+my ($node, $nodes, $expected_focus, $A, $B, $F);
 my ($result);
+my @fullscreen_permutations = ([], ["A"], ["B"], ["A", "B"]);
 my @urgent;
+
+sub fullscreen_windows {
+    my $ws = shift if @_;
+
+    scalar grep { $_->{fullscreen_mode} != 0 } @{get_ws_content($ws)}
+}
 
 ###############################################################################
 # Invalid con_id should not crash i3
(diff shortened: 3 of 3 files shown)


src/con.c
```diff
@@ -2374,5 +2373,12 @@ bool con_swap(Con *first, Con *second) {
     con_force_split_parents_redraw(first);
     con_force_split_parents_redraw(second);
 
+    if (first_fullscreen_mode != CF_NONE) {
+        con_enable_fullscreen(first, first_fullscreen_mode);
```
Airblader: Let's say I have two workspaces A and B. On A there are two windows, X and F1, where F1 is fullscreen. On B I have one window, F2, which is also fullscreen. B is active and I swap F2 with X. This means that on A, the now fullscreened container will be F2 and no longer F1. Correct?

Airblader: Also, `con_enable_fullscreen` mutates focus. We have to be careful that it doesn't interfere with what we are doing here.

orestisf1993: 1. For the current implementation the answer is "yes" but maybe it should not? I just realised that `swap` should probably also swap fullscreeniness. So X becomes fullscreen F2 looses the fullscreen and F1 should remain focused. Which behaviour do you prefer?
2. I think that it only does that in the correct cases. When you handle `CF_OUTPUT` it will only change focus for windows in an active workspace and for `CF_GLOBAL` it always changes focus. So it just makes sure that things are in a reasonable state.

orestisf1993: See https://github.com/i3/i3/pull/2902/commits/82e32fd596dea92cab39f4a870028ac1f00ec80c for (1.)

Airblader:
Regarding geometry the behavior of swap is to exchange geometries, so I think it makes sense to do the same for fullscreen windows, yes.

Airblader:
Have you also (manually) tested this with more exotic layouts? I know this is a weird thing, but it technically works and can lead to crashes if not careful: what about nested fullscreen containers and swapping child containers in it?

For example: `open, open, split v, open, focus parent, fullscreen enable` will fullscreen a stack of two vertical containers. What happens if you now swap only one of those two for a window on another workspace? What if that other window was fullscreened on its workspace?

testcases/t/265-swap.t
```diff
@@ -171,26 +178,72 @@ exit_gracefully($pid);
...
+    my $B_fullscreen = "B" ~~ @$fullscreen || 0;
+    $A->fullscreen($A_fullscreen);
+    $B->fullscreen($B_fullscreen);
+    sync_with_i3;
+
+    cmd '[con_mark=B] swap container with mark A';
+
+    $nodes = get_ws_content($ws1);
+    is($nodes->[0]->{window}, $B->{id}, 'B is on ws1:left');
+    # is($B->{fullscreen}, $A_fullscreen, "B got A's fullscreen mode");
```
Airblader: Why the commented out lines?

orestisf1993:
I just tested with lots of windows (there is an unrelated visual bug there). No crashes, behavior is as expected imo.

[Before fullscreen](https://i.imgur.com/ii6DG12.png).
[After fullscreen](https://i.imgur.com/yhRsiJ3.png).
[After `workspace 2` and `i3-msg swap mark A`](https://i.imgur.com/Jdo4Ajc.png).

There is this thing where focus in image 2 and 3 is not the same but I think that it is not an issue with my changes because:
1. It behaves the same with and without fullscreen mode.
2. i3 doesn't maintain multiple selected windows (like in image 2) when you change workspace. I hope I haven't confused the terminology here. To say it in a simplest manner when multiple green windows are selected and then you change to another workspace and go back again only one green window is selected.

I also implemented your example in a testcase.

testcases/t/265-swap.t
orestisf1993: I don't know. For whatever reason this doesn't work for me or I am misusing it. Should the following test work? https://github.com/orestisf1993/i3/commit/93b034621276fddd3f01164e3c92a5bd20a87603

orestisf1993: `$window->{fullscreen}` doesn't get updated even though I can confirm it visually.

orestisf1993:
This obviously needs some squashing and minor code improvements but I think the base functionality is right.

I have also implemented something for issue #2810. Should I commit here, open a new PR or wait for this to be merged? If we work on both issues at the same time we can also test for cases of swapping fullscreen + floating windows.

Airblader:
Let's keep the PRs separate, please. You can open the other one already if you want, you'll just have to rebase after this one is merged. That's up to you. :-)

I'd like to review but also manually try this out, I'll just have to get around to it. Perhaps I can find the time tomorrow.

Airblader:
The change itself looks good to me, though I haven't gotten around to testing it myself. Can you already try to clean this PR up for a final review, i.e.,

* squash the commits
* rebase onto the current HEAD
* get rid of any TODOs etc. in the tests

Thank you!

orestisf1993:
Yes, I will do it tomorrow. In the meanwhile, do you have an opinion for https://github.com/i3/i3/pull/2902#pullrequestreview-61325863?

orestisf1993:
Cleaned

src/con.c
```diff
@@ -2361,6 +2360,14 @@ bool con_swap(Con *first, Con *second) {
     second->percent = first_percent;
     fake->percent = 0.0;
 
+    /* The two windows exchange their previous fullscreen status */
```
orestisf1993: A problem here is that if `_con_move_to_con` fails, the windows are left as non-fullscreen instead of reverting to their initial state.

testcases/t/265-swap.t
orestisf1993: @stapelberg is there a consistent way to check if a window is fullscreen in the testsuite?

stapelberg: Check the `fullscreen_mode` property of the node in the tree (e.g. using `get_ws_content`)

orestisf1993: 🎉 Thanks :)

src/con.c
```diff
@@ -2361,7 +2360,26 @@ bool con_swap(Con *first, Con *second) {
     second->percent = first_percent;
     fake->percent = 0.0;
 
+#define SWAP(x, y, T) \
```
Airblader: It seems unnecessary to define a macro for this since we use it exactly once. I actually liked it better before where we have

```
if (first_fullscreen_mode != CF_NONE) {
    con_enable_fullscreen(second, first_fullscreen_mode);
}
```

as this is more in line with how we do other things in this function.

orestisf1993: I like the old way too but I made the swapping to reset the original fullscreen mode in case of failure.
Macro because it can be used for floating containers too but I agree it's stupid right now.
If you think it's not important to reset to the original state I'll gladly revert these changes.

Airblader: I might be missing something – how does swapping the variables help with the case of failure?

orestisf1993: Since `goto swap_end` (https://github.com/orestisf1993/i3/blob/b2f5f8ee01155d9b8b8adcad2d26f4850837c45a/src/con.c#L2309) skips the swap the original fullscreen modes are used in https://github.com/orestisf1993/i3/blob/b2f5f8ee01155d9b8b8adcad2d26f4850837c45a/src/con.c#L2376-L2381

Airblader: I see. In that case swapping these variables makes sense, but I don't like the one-off macro definition mid-code here. We should either change that or at least promote it to a globally available macro in `include/util.h`.

orestisf1993: I think it's ok now.

PR #2908, opened 2017-09-08
Use OVER operator for drawing text

heftig:
For opaque text, SOURCE is not any different from OVER. However, when
drawing color glyphs (which consist of RGBA pixels instead of strokes)
SOURCE's handling of alpha is not what we want.

I stumbled across this because cairo 1.15.8 seems to clear the surface
before drawing color emoji if the operator is SOURCE, deleting every-
thing drawn before. Arguably, the area outside the glyph bounds should
not be touched, but even if this is a cairo bug the problem of alpha
within the glyph remains.

### DIFF
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -105,7 +105,7 @@ static void draw_text_pango(const char *text, size_t text_len,
         pango_layout_set_text(layout, text, text_len);
 
     /* Do the drawing */
-    cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
+    cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
     cairo_set_source_rgb(cr, pango_font_red, pango_font_green, pango_font_blue);
     pango_cairo_update_layout(cr, layout);
     pango_layout_get_pixel_size(layout, NULL, &height);


Airblader:
Can you provide example screenshots of what this change looks like?

heftig:
I used the base config and added an emoji into the status bar. It doesn't look any different except without this patch all the text to the left of the emoji is missing.

Airblader:
When you add "except..." to "it doesn't look any different", then it means it does look different. If it didn't, we wouldn't need a patch, right? :-) I'd like to see the visual change before and after, thus I think screenshots would make sense.

heftig:
Old Cairo:
cairo 1.14.10 + i3 4.14, unpatched

New Cairo:
cairo 1.15.8 + i3 4.14, unpatched

New Cairo, i3 patched:
cairo 1.15.8 + i3 4.14, patched

stapelberg:
Thanks for the fix!

Could you rebase this onto the next branch to pick up commit 3137064efc577e90b342319365b173276b13d236 please?

These steps should do the trick:

```bash
git pull --rebase
git push -f git@github.com:heftig/i3 next
```

heftig:
Done.

stapelberg:
Thank you!

PR #2920, opened 2017-09-09
Consider RandR 1.5's monitors' output names in addition to monitor names

### LINKED ISSUE #2909: Auto-generated XRandr 1.5 monitor names break some existing configurations
Background: 

I have one of those early 4K monitors with two display chips, so to a PC it appears as two separate monitors (it connects via either two HDMI cables, or one DisplayPort cable, in which case it pretends to be two monitors daisy-chained to each other). It is then the driver's job to stitch these two panels and present them as a single monitor object to the OS, which works pretty well on Windows, however (up until recently) the proprietary NVIDIA driver for Linux did little to nothing in this regard. Thus, all X11 software saw the physical monitor as two distinct displays, and e.g. would full-screen themselves to cover only one half of the physical monitor. (Though 4K monitors made today don't have this issue, some 5K/6K monitors do.)

The generally accepted way to deal with this problem was to use Xinerama, however that didn't work if you had more than one monitor (as it would then gobble it up as well). I've written [some software](https://github.com/CyberShadow/hax11) to help with this issue, which is how I've been successfully using i3 as well.

Now, XRandr 1.5 comes along, and it introduces monitor objects. Support for these is added to the NVIDIA driver and i3: the X server exposes information about which outputs are actually part of one monitor, and i3 uses it to correctly tile and full-screen windows across the entire monitor. Hacks like Xinerama or hax11 are no longer needed and everything seems to work great. Hooray!

Well, everything except one thing:

When using XRandr 1.5, i3 outputs (e.g. the `output` setting in `bar` blocks in the configuration) now need to refer to monitor names, and not output names. So, one problem is that should the monitor name not match the previously-configured output name, the i3 configuration is going to break upon upgrading to 4.14. Another issue is that when stitching multiple outputs into a monitor object, the X server seems to auto-generate the name of thus-created monitors. For example, the output of `xrandr --listmonitors` on my system is:

```
Monitors: 2                
 0: +Auto-Monitor-49434132a2000000 3840/349x2160/393+1920+0  DP-1.8 DP-1.1.8
 1: +DVI-D-0 1200/519x1920/324+0+0  DVI-D-0
```

(note that `--listmonitors` is currently undocumented in xrandr's man page)

The number after `Auto-Monitor-` seems to come from the kernel's DRI stack, and I have no idea how stable it is.

I already generate my i3 config, so as a workaround I did something like this:

```bash
monitor_name=$(xrandr --listmonitors | grep -F DP-1.8 | cut -d ' ' -f 3 | cut -c 2-)
sed -e "s/MONITOR_NAME/$monitor_name/g" < config.in > config
```

So, what to do about this? Some options would be:

- When looking up XRandr 1.5 monitor objects by name, check the names for the outputs connected to a monitor object as well as the name of the monitor object itself. This should ensure that existing configurations aren't broken when upgrading on systems where monitor names don't match output names.

- Add a note about this to i3's documentation.

- Do nothing - this GitHub issue as it currently is should be enough to document this uncommon problem.

Either way, I just wanted to publicly document my findings after spending a few hours hunting down the reason why my i3-bar disappeared after upgrading, in case it might be helpful to someone.

CyberShadow:
Fixes #2909. 

I suggest reviewing commit by commit.

TODO:
- [x] Replace `OUTPUT_NAME` macro with `output_primary_name` function
- [x] ~~Add IPC field for alternative output names~~
- [x] ~~Make i3bar use the new IPC field~~
- [x] Update documentation
- [x] Add test?

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -869,6 +869,18 @@ The 'output' is the name of the RandR output you attach your screen to. On a
 laptop, you might have VGA1 and LVDS1 as output names. You can see the
 available outputs by running +xrandr --current+.
 
+If your X server supports RandR 1.5 or newer, i3 will use RandR monitor objects
+instead of output objects. Run +xrandr --listmonitors+ to see a list. Usually,
+a monitor object contains exactly one output, and has the same name as the
+output; but should that not be the case, you may specify the name of either the
+monitor or the output in i3's configuration. For example, the Dell UP2414Q uses
+two scalers internally, so its output names might be “DP1” and “DP2”, but the
+monitor name is “Dell UP2414Q”.
+
+(Note that even if you specify the name of an output which doesn't span the
+entire monitor, i3 will still use the entire area of the containing monitor
+rather than that of just the output's.)
+
 If you use named workspaces, they must be quoted:
 
 *Examples*:
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -579,6 +579,11 @@ static void dump_bar_bindings(yajl_gen gen, Barconfig *config) {
     y(array_close);
 }
 
+static char *canonicalize_output_name(char *name) {
+    Output *output = get_output_by_name(name, false);
+    return output ? output_primary_name(output) : name;
+}
+
 static void dump_bar_config(yajl_gen gen, Barconfig *config) {
     y(map_open);
 
@@ -588,8 +593,13 @@ static void dump_bar_config(yajl_gen gen, Barconfig *config) {
     if (config->num_outputs > 0) {
         ystr("outputs");
         y(array_open);
-        for (int c = 0; c < config->num_outputs; c++)
-            ystr(config->outputs[c]);
+        for (int c = 0; c < config->num_outputs; c++) {
+            /* Convert monitor names (RandR ≥ 1.5) or output names
+             * (RandR < 1.5) into monitor names. This way, existing
+             * configs which use output names transparently keep
+             * working. */
+            ystr(canonicalize_output_name(config->outputs[c]));
+        }
         y(array_close);
     }
 
@@ -599,7 +609,7 @@ static void dump_bar_config(yajl_gen gen, Barconfig *config) {
 
         struct tray_output_t *tray_output;
         TAILQ_FOREACH(tray_output, &(config->tray_outputs), tray_outputs) {
-            ystr(tray_output->output);
+            ystr(canonicalize_output_name(tray_output->output));
         }
 
         y(array_close);
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -829,7 +829,7 @@ IPC_HANDLER(get_outputs) {
         y(map_open);
 
         ystr("name");
-        ystr(output->name);
+        ystr(output_primary_name(output));
 
         ystr("active");
         y(bool, output->active);
--- a/src/randr.c
+++ b/src/randr.c
@@ -266,7 +266,11 @@ Output *create_root_output(xcb_connection_t *conn) {
     s->rect.y = 0;
     s->rect.width = root_screen->width_in_pixels;
     s->rect.height = root_screen->height_in_pixels;
-    s->name = "xroot-0";
+
+    struct output_name *output_name = scalloc(1, sizeof(struct output_name));
+    output_name->name = "xroot-0";
+    SLIST_INIT(&s->names_head);
+    SLIST_INSERT_HEAD(&s->names_head, output_name, names);
 
     return s;
 }
@@ -594,7 +598,12 @@ static bool randr_query_outputs_15(void) {
         Output *new = get_output_by_name(name, false);
         if (new == NULL) {
             new = scalloc(1, sizeof(Output));
-            new->name = sstrdup(name);
+
+            struct output_name *output_name = scalloc(1, sizeof(struct output_name));
+            output_name->name = sstrdup(name);
+            SLIST_INIT(&new->names_head);
+            SLIST_INSERT_HEAD(&new->names_head, output_name, names);
+
             if (monitor_info->primary) {
                 TAILQ_INSERT_HEAD(&outputs, new, outputs);
             } else {
@@ -643,14 +652,23 @@ static void handle_output(xcb_connection_t *conn, xcb_randr_output_t id,
 
     Output *new = get_output_by_id(id);
     bool existing = (new != NULL);
-    if (!existing)
+    if (!existing) {
         new = scalloc(1, sizeof(Output));
+        SLIST_INIT(&new->names_head);
+    }
     new->id = id;
     new->primary = (primary && primary->output == id);
-    FREE(new->name);
-    sasprintf(&new->name, "%.*s",
+    while (!SLIST_EMPTY(&new->names_head)) {
+        FREE(SLIST_FIRST(&new->names_head)->name);
+        struct output_name *old_head = SLIST_FIRST(&new->names_head);
+        SLIST_REMOVE_HEAD(&new->names_head, names);
+        FREE(old_head);
+    }
+    struct output_name *output_name = scalloc(1, sizeof(struct output_name));
+    sasprintf(&output_name->name, "%.*s",
               xcb_randr_get_output_info_name_length(output),
               xcb_randr_get_output_info_name(output));
+    SLIST_INSERT_HEAD(&new->names_head, output_name, names);
 
     DLOG("found output with name %s\n", output_primary_name(new));
 
--- a/src/randr.c
+++ b/src/randr.c
@@ -599,9 +599,39 @@ static bool randr_query_outputs_15(void) {
         if (new == NULL) {
             new = scalloc(1, sizeof(Output));
 
+            SLIST_INIT(&new->names_head);
+
+            /* Register associated output names in addition to the monitor name */
+            xcb_randr_output_t *randr_outputs = xcb_randr_monitor_info_outputs(monitor_info);
+            int randr_output_len = xcb_randr_monitor_info_outputs_length(monitor_info);
+            for (int i = 0; i < randr_output_len; i++) {
+                xcb_randr_output_t randr_output = randr_outputs[i];
+
+                xcb_randr_get_output_info_reply_t *info =
+                    xcb_randr_get_output_info_reply(conn,
+                                                    xcb_randr_get_output_info(conn, randr_output, monitors->timestamp),
+                                                    NULL);
+
+                if (info != NULL && info->crtc != XCB_NONE) {
+                    char *oname;
+                    sasprintf(&oname, "%.*s",
+                              xcb_randr_get_output_info_name_length(info),
+                              xcb_randr_get_output_info_name(info));
+
+                    if (strcmp(name, oname) != 0) {
+                        struct output_name *output_name = scalloc(1, sizeof(struct output_name));
+                        output_name->name = sstrdup(oname);
+                        SLIST_INSERT_HEAD(&new->names_head, output_name, names);
+                    } else {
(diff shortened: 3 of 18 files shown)


src/randr.c
```diff
@@ -594,7 +606,41 @@ static bool randr_query_outputs_15(void) {
...
+                              xcb_randr_get_output_info_name(info));
+
+                    if (strcmp(name, oname) != 0) {
+                        struct output_name *output_name = scalloc(1, sizeof(struct output_name));
+                        output_name->name = sstrdup(oname);
+                        SLIST_INSERT_HEAD(&new->names_head, output_name, names);
+                    }
+
+                    DLOG("Adding additional output name: %s -> %s\n", name, oname);
+                }
```
stapelberg: Add `FREE(info);` here to avoid a memory leak.

src/randr.c
```diff
@@ -594,7 +606,41 @@ static bool randr_query_outputs_15(void) {
...
+                    char *oname;
+                    sasprintf(&oname, "%.*s",
+                              xcb_randr_get_output_info_name_length(info),
+                              xcb_randr_get_output_info_name(info));
+
+                    if (strcmp(name, oname) != 0) {
+                        struct output_name *output_name = scalloc(1, sizeof(struct output_name));
+                        output_name->name = sstrdup(oname);
+                        SLIST_INSERT_HEAD(&new->names_head, output_name, names);
+                    }
```
stapelberg: Add ` else {
free(oname);
}` here to avoid a memory leak.

stapelberg:
For a commit-by-commit review, it’d be helpful if you could squash the history accordingly. I.e., don’t add commits to address review changes, rather modify your history and force-push. See `git rebase -i`.

CyberShadow:
> For a commit-by-commit review, it’d be helpful if you could squash the history accordingly. I.e., don’t add commits to address review changes, rather modify your history and force-push. See git rebase -i.

That's how I would do it normally, however in this particular case it's not easily possible to directly replace `output->name` with a function call, as unlike a macro, a function's return value is not an lvalue. Instead, I could squash 071c48f76a0a8d0427d9489c090e30404aca7ce9, d0266a7200c1916c5fd94dcc543a2f4918b4cbd7, and 95254e4068dcddf8eff9e07f0c82196cd2fc9134 into a single commit.

src/randr.c
CyberShadow: Thanks for noticing these, fixed. I think there may be a similar memory leak in `main.c` (the `output` variable in `get_window_position`), as that was the reference I used for the `xcb_randr_get_output_info_reply` invocation.

CyberShadow:
@stapelberg OK, I think this is ready for a closer look now.

I decided to try a different approach: instead of changing the IPC and updating i3bar to match, I made i3 canonicalize output names in the `get_bar_config` IPC reply. This saves having to do the same kind of changes in i3bar to make it keep track of all output names. Let me know what you think.

Remaining matters:

- It Works On My Machine™, but any thoughts about writing a test for this? I had a look at the existing inject_randr15 machinery, but unless there's a way to ask the X server to create dummy outputs, looks like it would need to be extended to also handle RRGetOutputInfo. Perhaps we can just make it return the CRTC number (not totally sure how that works) of the X server's existing output, and make sure whatever its name is, it's still matched when it's in i3's configuration? Alternatively, perhaps instead of `inject_randr15`, ask the X server to create a monitor object using `RRSetMonitor` / `xrandr --setmonitor` (I guess that would require RandR 1.5 in Xephyr/Xvfb)?

- Does this need any documentation changes anywhere?

src/ipc.c
```diff
@@ -588,8 +593,13 @@ static void dump_bar_config(yajl_gen gen, Barconfig *config) {
     if (config->num_outputs > 0) {
         ystr("outputs");
         y(array_open);
-        for (int c = 0; c < config->num_outputs; c++)
-            ystr(config->outputs[c]);
+        for (int c = 0; c < config->num_outputs; c++) {
+            /* Convert configuration output names to canonical names.
+             * Perform this here (instead of while loading the config)
```
stapelberg: You say “instead of while loading the config”, which I take to refer to https://github.com/i3/i3/blob/0f2bce391686e8c4126338ee28d0e199a49f13b5/i3bar/src/outputs.c#L188-L198
right?

Since you don’t remove the canonicalization there (and I don’t want to suggest that, keeping it as-is increases compatibility), shouldn’t the comment read “additionally”, not “instead”?

Also, what’s the bit about the respective outputs not yet existing? Why not mention the actual issue of changing output names with RandR 1.5?

stapelberg:
> I decided to try a different approach: instead of changing the IPC and updating i3bar to match, I made i3 canonicalize output names in the get_bar_config IPC reply. This saves having to do the same kind of changes in i3bar to make it keep track of all output names. Let me know what you think.

This is a good idea indeed!

> I had a look at the existing inject_randr15 machinery, but unless there's a way to ask the X server to create dummy outputs, looks like it would need to be extended to also handle RRGetOutputInfo

Extending it is fine. We can add any request we need.

> ask the X server to create a monitor object using RRSetMonitor / xrandr --setmonitor (I guess that would require RandR 1.5 in Xephyr/Xvfb)?

We use that in `t/533-randr15.t`, and it’s a bit unfortunate in hindsight: it modifies the environment, i.e. other tests can’t be expected to function the same way they function if they were started before `t/533-randr.t`. I’d like to not repeat this mistake more often, so a way which doesn’t modify the environment would be better.

> Does this need any documentation changes anywhere?

Yeah, please update the userguide to clarify that output names can be the name of a monitor or one of the individual output names which constitute that monitor. (Is that clear? Improvements welcome.)

src/ipc.c
CyberShadow: > which I take to refer to

Sorry, I mean when i3 reads `~/.i3/config` and populates `barconfigs`, i.e. here:

https://github.com/i3/i3/blob/0f2bce391686e8c4126338ee28d0e199a49f13b5/src/config_directives.c#L437

> Since you don’t remove the canonicalization there 

Since that's in i3bar code, doing that there would require the IPC change anyway :)

> Since you don’t remove the canonicalization there (and I don’t want to suggest that, keeping it as-is increases compatibility), shouldn’t the comment read “additionally”, not “instead”?

Sorry, I don't understand your question, even minding the above. Are you saying that not reverting the canonicalization in i3bar counts as another implied instance of canonicalization?

> Also, what’s the bit about the respective outputs not yet existing?

Sorry, perhaps I made some bad assumptions there. Considering the case of the user starting i3, plugging in an additional monitor, and trying to make i3 show a configured bar on the new output, how would i3 handle that?

> Why not mention the actual issue of changing output names with RandR 1.5?

Sorry, what do you mean by that? Do you mean that RandR 1.5 monitor objects can be created and deleted by the user arbitrarily? I don't see anything about renaming outputs in 1.5:

https://cgit.freedesktop.org/xorg/proto/randrproto/diff/randrproto.txt?id=randrproto-1.5.0&id2=randrproto-1.4.1

CyberShadow:
> Extending it is fine.

I was hoping you wouldn't say that ;) I'll look into it.

src/ipc.c
stapelberg: >> which I take to refer to
>
> Sorry, I mean when i3 reads ~/.i3/config and populates barconfigs, i.e. here:

Ah. That was the other place I checked, but there’s no output name canonicalization going on there, so that doesn’t seem to fit to your comment either.

>> Since you don’t remove the canonicalization there (and I don’t want to suggest that, keeping it as-is increases compatibility), shouldn’t the comment read “additionally”, not “instead”?
>
> Sorry, I don't understand your question, even minding the above. Are you saying that not reverting the canonicalization in i3bar counts as another implied instance of canonicalization?

Yes, I’d say so — we’ll end up with two places which can canonicalize an output name.

>> Also, what’s the bit about the respective outputs not yet existing?
>
> Sorry, perhaps I made some bad assumptions there. Considering the case of the user starting i3, plugging in an additional monitor, and trying to make i3 show a configured bar on the new output, how would i3 handle that?

As soon as you enable a new output (e.g. via `xrandr --output HDMI-2 --auto --right-of LVDS1`), i3 will open a new workspace on that output and display an i3bar instance. From i3bar’s perspective, a new bar config is received, i3bar queries the outputs, and off we go.

>> Why not mention the actual issue of changing output names with RandR 1.5?
> Sorry, what do you mean by that? Do you mean that RandR 1.5 monitor objects can be created and deleted by the user arbitrarily? I don't see anything about renaming outputs in 1.5:

Sorry I was unclear: I meant that the names changed between i3 4.13 and i3 4.14 due to using RandR 1.5 — i.e. the very top-level issue you’re trying to address :).

CyberShadow:
> Extending it is fine.

Done.

> Yeah, please update the userguide to clarify that output names can be the name of a monitor or one of the individual output names which constitute that monitor. (Is that clear? Improvements welcome.)

Done.

src/ipc.c
CyberShadow: Sorry, I still don't understand; I think we may be talking past each other. Do you think you could explain your point in other words, in more detail?

Alternatively, perhaps you could provide the text for what you think this comment should say, so that we may move on?

stapelberg: How about this?

Convert monitor names (RandR ≥ 1.5) or output names (RandR < 1.5) into monitor names. This way, existing configs which use output names transparently keep working.

src/randr.c
```diff
@@ -48,9 +48,15 @@ Output *get_output_by_name(const char *name, const bool require_active) {
     Output *output;
     bool get_primary = (strcasecmp("primary", name) == 0);
     TAILQ_FOREACH(output, &outputs, outputs) {
-        if ((output->primary && get_primary) ||
-            ((!require_active || output->active) && strcasecmp(output->name, name) == 0)) {
+        if (output->primary && get_primary)
```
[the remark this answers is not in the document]
CyberShadow: Done.

src/ipc.c
CyberShadow: Done, thanks.

docs/userguide
```diff
@@ -869,6 +869,14 @@ The 'output' is the name of the RandR output you attach your screen to. On a
 laptop, you might have VGA1 and LVDS1 as output names. You can see the
 available outputs by running +xrandr --current+.
 
+If your X server supports RandR 1.5 or newer, i3 will use RandR monitor objects
+instead of output objects. Run +xrandr --listmonitors+ to see a list. Usually,
+a monitor object contains exactly one output, and has the same name as the
+output; but should that not be the case, you may specify the name of either the
+monitor or the output in i3's configuration. (Note that even if you specify the
```
stapelberg: Let’s add an example here. How about:

For example, the Dell UP2414Q uses two scalers internally, so its output names might be “DP1” and “DP2”, but the monitor name is “Dell UP2414Q”.

include/util.h
```diff
@@ -78,6 +78,12 @@ __attribute__((pure)) bool name_is_digits(const char *name);
 bool layout_from_name(const char *layout_str, layout_t *out);
 
 /**
+ * Retrieves the primary name of an output.
+ *
+ */
+char *output_primary_name(struct xoutput *output);
```
stapelberg: Sorry, I didn’t spot this before: this function should go into include/output.h and include/output.c, not util.

docs/userguide
CyberShadow: Done.

include/util.h
CyberShadow: No problem. Done.

stapelberg:
Thanks a lot for this contribution!

CyberShadow:
It's been a pleasure. Thanks for the review.

PR #2923, opened 2017-09-10
Testsuite: avoid starting new i3 instances where possible

stapelberg:
This significantly reduces the time required to run the testsuite.


stapelberg:
That’s a fair concern. I’ve added the `kill_all_windows` convenience function, which is now invoked between all the tests. Please take another look.

PR #2925, opened 2017-09-10
Revert "Use OVER operator for drawing text (#2908)"

Airblader:
This reverts commit 16160462a30f186f5b72bb551ba2188670d4e45c.

### DIFF
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -105,7 +105,7 @@ static void draw_text_pango(const char *text, size_t text_len,
         pango_layout_set_text(layout, text, text_len);
 
     /* Do the drawing */
-    cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
+    cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
     cairo_set_source_rgb(cr, pango_font_red, pango_font_green, pango_font_blue);
     pango_cairo_update_layout(cr, layout);
     pango_layout_get_pixel_size(layout, NULL, &height);


Airblader:
@heftig 

Your change in #2908 causes problems for me where for example the binding mode indicator no longer displays any text (same for status blocks with urgent property). Can you also reproduce this in your machine? My cairo version is 1.14.10.

stapelberg (approved):

heftig:
Can you give me instructions on how to reproduce? I'm not actually an user of i3.

heftig:
Urgent window and workspace button (old cairo, patched i3):
screenshot from 2017-09-10 18-55-55
So it seems to work for me.

Airblader:
OK, so I actually run a fork which allows RGBA, thus the code in question looks like this (pre-patch):

```
    /* Do the drawing */
    cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
    cairo_set_source_rgba(cr, pango_font_red, pango_font_green, pango_font_blue, pango_font_alpha);
```

I *did* reproduce this with upstream i3 earlier before submitting the commit revert, but at the moment I can only reproduce it on my fork (so I may have made a mistake earlier when reproducing it) where I use a fully transparent color for the text; I guess this explains why with OVER, the text is no longer visible, as now cairo (correctly) just doesn't draw anything anymore.

I at least would shine more light on cairo here as I think the behavior your PR fixed reeks a lot like a bug in cairo (why should cairo clear the entire surface when drawing a glyph in one specific spot?). 
@psychon Do you know why cairo clears the entire surface when drawing a color glyph as described in #2908 using the SOURCE operator? Is this a bug?

Now, as you pointed out, either way the problem would persist within the glyph's boundaries. In my fork using SOURCE was a very conscious decision because it makes choosing the RGBA colors more intuitive (you don't have to worry about blending effects) and more powerful (you can make an area completely transparent).

psychon:
Hm. Could one of you write to the cairo mailing list about "this glyph thing"? It would be better if I don't have to play man-in-the-middle here. If you want me to guess: The cairo code has a temporary surface for color glyphs and draws this to the target surface. With OVER this does just the right thing. With SOURCE (or any other unbounded operator), it does not. So if my guess is right, then this is a design bug.

@Airblader What alpha values are you using in your fork? Could you give me the values that are given to `cairo_set_source_rgba`? If your alpha is 0, well, drawing alpha 0 with OVER is a no-op (you are drawing something that is fully transparent, i.e. invisible).

heftig:
I was really confused because—as I mentioned—SOURCE and OVER are equal for opaque drawing. Since this concerns a fork that uses alpha text, can we revert the revert on the main branch? The fork can continue to use SOURCE and/or look for a better solution.

Airblader:
@psychon Yes, in my vase it's an alpha value of 0. I do get that with OVER this is expectedly a no-op, no argument here. :-) It's just unsatisfying that this also affects color glyphs. I specifically use SOURCE in my fork because it allows two things:

* The actual color will be exactly what you draw with unseasonably of some blend.
* It allows drawing transparent areas over colored ones such that they can be completely transparent. With OVER, there is no way to do this.

That's why I chose SOURCE in my fork, but for color glyphs this now causes issues. Within the bounds of the glyph this is probably how it should be expected to work, though, so I'm really unsure what we should do here. That's not a question I was asking to you, though. :-)

Airblader:
@heftig Although it currently affects the fork the decision is still an important one here. Switching to OVER effectively means we can never support alpha values on text colors in the future. That's quite a restriction to impose.

Airblader:
(At least not support them in the way I chose to do it in the fork for the reasons mentioned above)

psychon:
Okay, so I wrote that mail to the cairo mailing list for you: https://lists.cairographics.org/archives/cairo/2017-September/028328.html

Airblader:
Thank you!

heftig:
`cairo_set_source_rgb`/`_rgba` sets the source pattern for drawing to an unbounded plane of a single color.

For drawing normal glyphs, Cairo gets A8 images ("coverage maps") from FreeType and uses them as the mask for the operation.

For drawing color glyphs, Cairo gets RGBA8 images from FreeType and uses them as the source for the operation. The mask becomes unbounded.

Normally you shouldn't encounter any color glyphs when drawing emoji-less text. However, if the user has misconfigured fontconfig to prefer emoji fonts into `sans-serif` or similar, this may be broken, as suddenly normal characters without emoji presentation will get the color glyphs.

For emoji characters and sequences pango automatically switches to the `emoji` font family. Fontconfig's default config defines several emoji fonts to serve this family. Hence, you get color glyphs for emoji characters even if the fonts you selected do not include the emoji fonts.

However, there's still a way around this. This patch should set up fontconfig to reject all color fonts in all cases:

```diff
diff --git i/Makefile.am w/Makefile.am
index f1ba3f6375c151cb..0231a565a0963f50 100644
--- i/Makefile.am
+++ w/Makefile.am
@@ -266,24 +266,27 @@ i3_LDADD = \
 	$(XKBCOMMON_LIBS) \
 	$(YAJL_LIBS) \
 	$(LIBPCRE_LIBS) \
+	$(FONTCONFIG_LIBS) \
 	$(PANGOCAIRO_LIBS) \
 	$(PTHREAD_LIBS) \
 	$(CODE_COVERAGE_LDFLAGS)
 
 libi3_CFLAGS = \
 	$(AM_CFLAGS) \
 	$(XCB_CFLAGS) \
 	$(XCB_UTIL_CFLAGS) \
 	$(XCB_UTIL_XRM_CFLAGS) \
 	$(YAJL_CFLAGS) \
+	$(FONTCONFIG_CFLAGS) \
 	$(PANGOCAIRO_CFLAGS)
 
 libi3_LIBS = \
 	$(top_builddir)/libi3.a \
 	$(XCB_LIBS) \
 	$(XCB_UTIL_LIBS) \
 	$(XCB_UTIL_XRM_LIBS) \
 	$(YAJL_LIBS) \
+	$(FONTCONFIG_LIBS) \
 	$(PANGOCAIRO_LIBS)
 
 libi3_a_CFLAGS = \
@@ -372,14 +375,16 @@ i3bar_i3bar_CFLAGS = \
 	$(libi3_CFLAGS) \
 	$(XCB_CFLAGS) \
 	$(XKBCOMMON_CFLAGS) \
+	$(FONTCONFIG_CFLAGS) \
 	$(PANGOCAIRO_CFLAGS) \
 	$(YAJL_CFLAGS)
 
 i3bar_i3bar_LDADD = \
 	$(libi3_LIBS) \
 	$(XCB_LIBS) \
 	$(XCB_UTIL_CURSOR_LIBS) \
 	$(XKBCOMMON_LIBS) \
+	$(FONTCONFIG_LIBS) \
 	$(PANGOCAIRO_LIBS) \
 	$(YAJL_LIBS)
 
diff --git i/configure.ac w/configure.ac
index 7d274e36b209e4fe..163c5baa60371cd4 100644
--- i/configure.ac
+++ w/configure.ac
@@ -99,6 +99,7 @@ PKG_CHECK_MODULES([XCB_UTIL_XRM], [xcb-xrm])
 PKG_CHECK_MODULES([XKBCOMMON], [xkbcommon xkbcommon-x11])
 PKG_CHECK_MODULES([YAJL], [yajl])
 PKG_CHECK_MODULES([LIBPCRE], [libpcre >= 8.10])
+PKG_CHECK_MODULES([FONTCONFIG], [fontconfig])
 PKG_CHECK_MODULES([PANGOCAIRO], [cairo >= 1.14.4 pangocairo])
 
 # Checks for programs.
diff --git i/libi3/font.c w/libi3/font.c
index 81091ea74a31da11..b07a7f4163092827 100644
--- i/libi3/font.c
+++ w/libi3/font.c
@@ -16,7 +16,9 @@
 
 #include <cairo/cairo-xcb.h>
 #include <pango/pangocairo.h>
+#include <fontconfig/fontconfig.h>
 
+static bool initialized_fontconfig = false;
 static const i3Font *savedFont = NULL;
 
 static xcb_visualtype_t *root_visual_type;
@@ -152,25 +154,61 @@ static int predict_text_width_pango(const char *text, size_t text_len, bool pang
     return width;
 }
 
+static void initialize_fontconfig (void)
+{
+    FcConfig *config = FcConfigCreate ();
+    FcBool res;
+
+    if (!FcConfigParseAndLoad(config, FcConfigFilename(NULL), FcTrue)) {
+        ELOG("Failed to load default fontconfig\n");
+        goto bail;
+    }
+
+    res = FcConfigParseAndLoadFromMemory(config,
+        "<fontconfig>"
+        "  <selectfont><rejectfont>"
+        "    <pattern><patelt name=\"color\"><bool>True</bool></patelt></pattern>"
+        "  </rejectfont></selectfont>"
+        "</fontconfig>", FcTrue);
+    if (!res) {
+        ELOG("Failed to add color-rejecting fontconfig\n");
+        goto bail;
+    }
+
+    if (!FcConfigSetCurrent(config)) {
+        ELOG("Failed to activate custom fontconfig\n");
+        goto bail;
+    }
+
+bail:
+    /* Always clean up our reference (FcConfigSetCurrent takes its own) */
+    FcConfigDestroy(config);
+}
+
 /*
  * Loads a font for usage, also getting its metrics. If fallback is true,
  * the fonts 'fixed' or '-misc-*' will be loaded instead of exiting. If any
  * font was previously loaded, it will be freed.
  *
  */
 i3Font load_font(const char *pattern, const bool fallback) {
     /* if any font was previously loaded, free it now */
     free_font();
 
     i3Font font;
     font.type = FONT_TYPE_NONE;
 
     /* No XCB connction, return early because we're just validating the
      * configuration file. */
     if (conn == NULL) {
         return font;
     }
 
+    if (!initialized_fontconfig) {
+        initialize_fontconfig();
+        initialized_fontconfig = true;
+    }
+
     /* Try to load a pango font if specified */
     if (strlen(pattern) > strlen("pango:") && !strncmp(pattern, "pango:", strlen("pango:"))) {
         const char *font_pattern = pattern + strlen("pango:");
```

heftig:
Made a patch to cairo that might solve the problem if it's accepted (in some form):

https://bugs.freedesktop.org/show_bug.cgi?id=102661

psychon:
A different patch was accepted, so apparently this is now fixed in cairo (but still does not have any unit tests, I think :-( )

Jesin:
Should this be re-applied now that it's fixed in Cairo?

xzfc:
Can we re-revert it back?

Advantages of `OVER` over `SOURCE` (pun intended):
* Subpixel font rendering (i3/i3#3604)
* No transparent garbage around color emoji characters (see the titlebar renders below)

@Airblader, I see your reasoning in https://github.com/i3/i3/pull/2925#issuecomment-328367740. This affects only i3-gaps (and hypothetical future version of vanilla i3) and only if a user has configured RGBA font colors. Thus I propose to use `OVER` by default and fallback to `SOURCE` in the said case. Not an ideal solution, but it covers more cases.

Patch for i3-gaps:
```diff
-    cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
+    cairo_set_operator(cr, pango_font_alpha == 1.0 ? CAIRO_OPERATOR_OVER : CAIRO_OPERATOR_SOURCE);
```

### Renders

[gaps-before.png](https://user-images.githubusercontent.com/5121426/69278442-41076080-0bda-11ea-8fee-4d5c3b67300d.png)
[gaps-after.png](https://user-images.githubusercontent.com/5121426/69278444-419ff700-0bda-11ea-9d42-0af880f04446.png)
[render script](https://gist.github.com/xzfc/9556fd6a93af0e33a4003eff848a0d68)

Left/right columns: disabled/enabled subpixel antialiasing in `fonts.conf`.
Rows: various titlebar background/text colors.

stapelberg:
> Can we re-revert it back?

Can you send a PR for that please?

xzfc:
Done: #3947

PR #2928, opened 2017-09-11
docs/hacking-howto: Promote "Using git / sending patches" section

### LINKED ISSUE #2907: Please add README or INSTALL file
There is currently no documentation in the project tree root on how to build i3. Please add a file describing the correct / supported way to do so.

If the documentation is elsewhere, please add a README file linking or directing readers to said location.

A README (or README.md) or INSTALL file is an established convention among open-source projects; it would be good if i3 followed suit.

Additionally, a README.md (or README) file will be automatically rendered by GitHub [on the project's main GitHub page](https://github.com/i3/i3), so it would be a good place to link to the project website and online documentation there.

CyberShadow:
Fixes #2907.

### DIFF
--- a/docs/hacking-howto
+++ b/docs/hacking-howto
@@ -8,6 +8,86 @@ touching i3’s source code. It should contain all important information to help
 you understand why things are like they are. If it does not mention something
 you find necessary, please do not hesitate to contact me.
 
+== Using git / sending patches
+
+=== Introduction
+
+For a short introduction into using git, see
+http://web.archive.org/web/20121024222556/http://www.spheredev.org/wiki/Git_for_the_lazy
+or, for more documentation, see http://git-scm.com/documentation
+
+Please talk to us before working on new features to see whether they will be
+accepted. A good way for this is to open an issue and asking for opinions on it.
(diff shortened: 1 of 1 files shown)


Airblader:
Can you squash those commits, please?

CyberShadow:
Are you sure? I did this change as one commit first, then I redid it as separate commits so it's possible to actually read the diff.

You can also squash it while merging in GitHub's UI, no?

Airblader:
Squashing shouldn't affect the overall diff. Since you only moved a few sections we don't need more than one commit for that.

CyberShadow:
> Squashing shouldn't affect the overall diff.

That's why there are multiple commits. It's not possible to discern the smaller changes from the overall diff (or that of a single squashed commit).

> Since you only moved a few sections we don't need more than one commit for that.

Not really. Sorry, have you read the commit messages? Or do you mean that the changes that weren't just literally moving a single block of text around are too minor to be in their own commits?

Airblader:
Ah, I actually misses that you made changes other than moving the section. You're completely right in this case, of course.

PR #2941, opened 2017-09-13
Fix focus order in floating_disable & floating_enable for unfocused windows

### LINKED ISSUE #1975: Inconsistent focus fallback behavior
A user described [this incident](https://faq.i3wm.org/question/6767/automatically-move-focus-back-to-tiling-when-floating-window-is-closed/?comment=6777#comment-6777) on the FAQ page. I can reproduce that I think is taking place here as follows.

Set up containers 1 through 4 as shown in the picture below. Now focus, in this order, 3–1–4–2. If you now keep closing the terminals, the focus will jump to 4 and then to 3, the other floating container, and not 1, which a user may expect to happen.

Now set up the same situation again and focus in the same order, 3–1–4–2. Type `sleep 3 && exit` into terminal 2 and switch focus to terminal 4. Once 2 is closed, if you exit out of terminal 4, focus will jump to terminal 1 and not 3 as before, even though from the user point of view there's not much of a difference between the two situations.

screenshot-2015-10-02_21-30-49

i3 4.11

### LINKED ISSUE #2938: Enabling / disabling floating for containers from other workspace loses their focus
```
Binary i3 version:  4.14-8-g85eb0976 (2017-09-05, branch "makepkg") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.14-40-gbcaec964 (2017-09-11, branch "remotes/origin/HEAD") (pid 902)
Loaded i3 config: /tmp/i3.config (Last modified: Tue 12 Sep 2017 11:42:24 AM EEST, 27 seconds ago)

The i3 binary you are running: /home/orestis/Documents/programming/i3/build/i3
```

http://logs.i3wm.org/logs/5751832013963264.bz2

**What I did:**
1. open 2 windows on workspace 1.
2. run `i3-msg floating enable && i3-msg mark A` for the second one. It is now focused & floating.
3. switch to workspace 2
4. run `i3-msg '[con_mark=A]' floating disable`
5. switch back to workspace 1

**What I saw:**
Window marked as A is not floating & is not focused.

**What I expected instead:**
Window marked as A is not floating & is still focused.

orestisf1993:
Fixes #2938 
Fixes #1975

### DIFF
--- a/src/con.c
+++ b/src/con.c
@@ -1436,40 +1436,6 @@ orientation_t con_orientation(Con *con) {
  *
  */
 Con *con_next_focused(Con *con) {
-    Con *next;
-    /* floating containers are attached to a workspace, so we focus either the
-     * next floating container (if any) or the workspace itself. */
-    if (con->type == CT_FLOATING_CON) {
-        DLOG("selecting next for CT_FLOATING_CON\n");
-        next = TAILQ_NEXT(con, floating_windows);
-        DLOG("next = %p\n", next);
-        if (!next) {
-            next = TAILQ_PREV(con, floating_head, floating_windows);
-            DLOG("using prev, next = %p\n", next);
-        }
-        if (!next) {
-            Con *ws = con_get_workspace(con);
-            next = ws;
-            DLOG("no more floating containers for next = %p, restoring workspace focus\n", next);
-            while (next != TAILQ_END(&(ws->focus_head)) && !TAILQ_EMPTY(&(next->focus_head))) {
-                next = TAILQ_FIRST(&(next->focus_head));
-                if (next == con) {
-                    DLOG("skipping container itself, we want the next client\n");
-                    next = TAILQ_NEXT(next, focused);
-                }
-            }
-            if (next == TAILQ_END(&(ws->focus_head))) {
-                DLOG("Focus list empty, returning ws\n");
-                next = ws;
-            }
-        } else {
-            /* Instead of returning the next CT_FLOATING_CON, we descend it to
-             * get an actual window to focus. */
-            next = con_descend_focused(next);
-        }
-        return next;
-    }
-
     /* dock clients cannot be focused, so we focus the workspace instead */
     if (con->parent->type == CT_DOCKAREA) {
         DLOG("selecting workspace for dock client\n");
@@ -1478,10 +1444,9 @@ Con *con_next_focused(Con *con) {
 
     /* if 'con' is not the first entry in the focus stack, use the first one as
      * it’s currently focused already */
-    Con *first = TAILQ_FIRST(&(con->parent->focus_head));
-    if (first != con) {
-        DLOG("Using first entry %p\n", first);
-        next = first;
+    Con *next = TAILQ_FIRST(&(con->parent->focus_head));
+    if (next != con) {
+        DLOG("Using first entry %p\n", next);
     } else {
         /* try to focus the next container on the same level as this one or fall
          * back to its parent */
@@ -1494,6 +1459,10 @@ Con *con_next_focused(Con *con) {
      * possible, excluding the current container */
     while (!TAILQ_EMPTY(&(next->focus_head)) && TAILQ_FIRST(&(next->focus_head)) != con) {
         next = TAILQ_FIRST(&(next->focus_head));
+    }
+
+    if (con->type == CT_FLOATING_CON && next != con->parent) {
+        next = con_descend_focused(next);
     }
 
     return next;
--- a/src/floating.c
+++ b/src/floating.c
@@ -176,6 +176,33 @@ void floating_enable(Con *con, bool automatic) {
         return;
     }
 
+    Con *focus_head_placeholder = NULL;
+    bool focus_before_parent = true;
+    if (!set_focus) {
+        /* Find recursively the ancestor container which is a child of our workspace.
+         * We need to reuse its focus position later. */
+        Con *ancestor = con;
+        while (ancestor->parent->type != CT_WORKSPACE) {
+            focus_before_parent &= TAILQ_FIRST(&(ancestor->parent->focus_head)) == ancestor;
+            ancestor = ancestor->parent;
+        }
+        /* Consider the part of the focus stack of our current workspace:
+         * [ ... S_{i-1} S_{i} S_{i+1} ... ]
+         * Where S_{x} is a container tree and the container 'con' that is beeing switched to
+         * floating belongs in S_{i}. The new floating container, 'nc', will have the
+         * workspace as its parent so it needs to be placed in this stack. If C was focused
+         * we just need to call con_focus(). Otherwise, nc must be placed before or after S_{i}.
+         * We should avoid using the S_{i} container for our operations since it might get
+         * killed if it has no other children. So, the two possible positions are after S_{i-1}
+         * or before S_{i+1}.
+         */
+        if (focus_before_parent) {
+            focus_head_placeholder = TAILQ_PREV(ancestor, focus_head, focused);
+        } else {
+            focus_head_placeholder = TAILQ_NEXT(ancestor, focused);
+        }
+    }
+
     /* 1: detach the container from its parent */
     /* TODO: refactor this with tree_close_internal() */
     con_detach(con);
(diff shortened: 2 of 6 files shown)


Airblader:
I don't expect test cases for this, but what happens in the following situations now?

* One floating window on another workspace, but it is not the focused one within the workspace. What has focus in that workspace after unfloating it?
* Two floating windows on another workspace and the focused one is unfloated
* Two floating windows on another workspace and the one not focused is unfloated

My expectation would be that

* the unfloated window will never be focused (on that workspace)
* the unfloated window is focused (on that workspace)
* the still floating window is focused (on that workspace)

orestisf1993:
> I don't expect test cases for this

Why not? I was looking ideas for more testcases.

1. 👍 
2. 👍 
3. 👎 Will investigate. The exact configuration is: tiled, floating1 floating2 & focus stack: floating2 > floating1 > tiled. For focus stack: floating2 > tiled > floating1 we already have the expected behavior.

Airblader:
By all means, feel free to add testcases. I just wanted to make clear that I'm only asking about the behavior. :-)

orestisf1993:
The problem is that I wasn't checking for floating containers in my loop. Will push after I write the tests.

orestisf1993:
Up! I actually fixed it by removing code.

orestisf1993:
Squashed, added fix for `floating_enable()`. Probably looking for more test case ideas.

src/con.c
```diff
@@ -1328,6 +1328,20 @@ orientation_t con_orientation(Con *con) {
...
 
 /*
+ * Returns the next container in the focus head of con's parent.
+ *
+ */
+Con *con_next_focused_tail(Con *con) {
+    Con *current;
+    TAILQ_FOREACH(current, &(con->parent->focus_head), focused) {
+        if (current == con) {
+            return TAILQ_NEXT(current, focused);
```
Airblader: Doesn't this mean this entire function is effectively the same as just `TAILQ_NEXT(con, focused)`?

include/con.h
```diff
@@ -324,6 +324,12 @@ bool con_move_to_mark(Con *con, const char *mark);
 orientation_t con_orientation(Con *con);
 
 /**
+ * Returns the next container in the focus head of con's parent.
+ *
+ */
+Con *con_next_focused_tail(Con *con);
```
Airblader: What does the `tail` refer to in this function?

src/floating.c
```diff
@@ -346,34 +359,43 @@ void floating_disable(Con *con, bool automatic) {
...
     Con *parent = con->parent;
 
+    Con *focus_tail_position = NULL;
+    Con *tiling_focused = con_descend_tiling_focused(ws);
+    /* No need to find focus order if focus is going to be set latter or if we are
+     * attaching to a workspace. */
+    if (!set_focus && tiling_focused->type != CT_WORKSPACE) {
+        focus_tail_position = con_next_focused_tail(parent);
+    }
+    DLOG("move_to_focus_head=%p\n", focus_tail_position);
```
Airblader: This log statement is now lying. :-) Either way, instead of dumping the variable name it'd be more useful to log what we are going to do with this container.

src/floating.c
```diff
@@ -346,34 +359,43 @@ void floating_disable(Con *con, bool automatic) {
     Con *ws = con_get_workspace(con);
     Con *parent = con->parent;
 
+    Con *focus_tail_position = NULL;
+    Con *tiling_focused = con_descend_tiling_focused(ws);
+    /* No need to find focus order if focus is going to be set latter or if we are
```
Airblader: s/latter/later/

orestisf1993:
Updated

orestisf1993:
Still has problems. Don't merge. Will edit this comment.

orestisf1993:
Updated. The new testcases crash my previous implementation.

orestisf1993:
@Airblader please remove the `waiting` label.

orestisf1993:
After https://github.com/i3/i3/pull/2969#issuecomment-330668431 I modified the code of `insert_con_into` to fix the problems of `move` I described and reused the function in `floating_disable`. It is in a very early stage and I can't clean it up right now but it might be a better alternative:  
https://github.com/orestisf1993/i3/tree/issue-2938-insert_into

I'll see if I can reuse it for `floating_enable` too.

orestisf1993:
Just pushed a commit that also fixes #1975. Will write a test for this and I could move it on a separate PR after #2975.

Airblader:
Sorry, I'm currently a little lost on as to what your status for the PR ist (?) :-)

orestisf1993:
I'd like to use `insert_con_into` from #2975 instead of the current implementation.

Airblader:
Fair enough, let's continue there first.

orestisf1993:
I've incorporated `insert_con_into` into `floating_disable`. Will investigate if `floating_enable` can be tidied up too but I think the PR can be reviewed right now.

I'll also run this patch on my machine for a while since it's a pretty risky PR.

src/floating.c
```diff
@@ -196,12 +221,22 @@ void floating_enable(Con *con, bool automatic) {
     /* We insert nc already, even though its rect is not yet calculated. This
      * is necessary because otherwise the workspace might be empty (and get
      * closed in tree_close_internal()) even though it’s not. */
-    if (set_focus) {
```
orestisf1993: We actually didn't need this branch since we call `con_activate` if `set_focus`.

src/floating.c
```diff
@@ -196,12 +221,22 @@ void floating_enable(Con *con, bool automatic) {
...
-        TAILQ_INSERT_TAIL(&(ws->floating_head), nc, floating_windows);
+    TAILQ_INSERT_HEAD(&(ws->floating_head), nc, floating_windows);
```
orestisf1993: Placement in the `floating_windows` stack is not ideal here. Eg if the focus order is F1 > T1 > F2 and we make T1 a floating container it should be below F1 but above F2. But since focus order doesn't always follow the floating order (with `focus_follows_mouse`) it is not a huge deal IMO.

src/floating.c
```diff
@@ -176,11 +176,36 @@ void floating_enable(Con *con, bool automatic) {
...
+         * [ ... S_{i-1} S_{i} S_{i+1} ... ]
+         * Where S_{x} is a container tree and the container 'con' that is beeing switched to
+         * floating belongs in S_{i}. The new floating container, 'nc', will have the
+         * workspace as its parent so it needs to be placed in this stack. If C was focused
+         * we just need to call con_focus(). Otherwise, nc must be placed before or after S_{i}.
+         * We should avoid using the S_{i} container for our operations since it might get
+         * killed if it has no other children. So, the two possible positions are after S_{i-1}
+         * or before S_{i+1}.
+         */
+        focus_before_parent = TAILQ_FIRST(&(con->parent->focus_head)) == con;
```
orestisf1993: I'll have to recheck this since it appears to be wrong. `con` can be the focused child of its parent but not the focused descendant of its ancestor, `focus_head_placeholder`.

Could use `is_focused_descendant()`.

src/floating.c
orestisf1993: The failed tests are probably because I changed the default from `_TAIL` to `_HEAD`. If we are not following the "ideal" strategy I think non-focused newly floating windows should be placed below all other existing floating windows, especially since one of them could be the currently focused window.

orestisf1993:
Up. This will need some time to be reviewed but we can slowly work on it.

Airblader:
I have some vacation in a week and will try to take a look at this then.

Airblader:
Can you resolve the conflicts? :-)

src/floating.c
```diff
@@ -346,27 +385,59 @@ void floating_disable(Con *con, bool automatic) {
     Con *ws = con_get_workspace(con);
     Con *parent = con->parent;
 
+    bool modify_focus_head = !set_focus;
+    Con *prev_focus_head = NULL;
+    Con *new_parent = con_descend_tiling_focused(ws);
+    if (new_parent->type != CT_WORKSPACE)
```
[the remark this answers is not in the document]
orestisf1993: Wait, what did you review here?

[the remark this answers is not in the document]
Airblader: I'm confused. What? :-)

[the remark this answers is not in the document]
orestisf1993: I can't find where you placed the comment here, it shows as outdated, and I can't grep the code shown above either.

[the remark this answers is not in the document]
Airblader: I'm a ninja! :-) I can't find it either. It was not that important anyway, so let's just ignore it.

PR #2950, opened 2017-09-14
Fix erratic behavior with single container child jumping outputs

### LINKED ISSUE #2466: Window in a single-child tabbed container moves across displays

```
Running i3 version: 4.12 (2016-03-06, branch "4.12") (pid 896)
Loaded i3 config: /home/simon/.i3/config (Last modified: Wed 21 Sep 2016 10:41:51 AM PDT, 4076 seconds ago)

```

http://logs.i3wm.org/logs/5713050069893120.bz2

**What I did:**
1. open a new, empty workspace
2. mod+enter+enter to create two terminals, side by side
3. focus right-hand terminal, type something in it to make sure it is distinguishable from the other terminal
4. mod+v to create a new container
5. mod+w to switch it to tabbed mode
6. alt+shift+left to move the terminal out of the tabbed container

**What I saw:**

On a multi-monitor setup, the terminal moves to the right-hand side of the monitor to the left. On a single-monitor setup (or if the workspace is on the far-left monitor), the terminal does not move at all, and the tabbed container does not get destroyed. I have reproduced this on two machines, and with the default config.

**What I expected instead:**

I expected the tabbed container to be destroyed, leaving the terminal in the same place. I have done this in older versions of i3 to destroy single-child containers, and it does not seem to be working anymore.

CyberShadow:
This fixes a regression introduced in commit 4e88c10564ca5366c2578908f62ec56625a26718: when attempting to move the single child of a container in the direction of another output, i3 would move the window to the output, despite the window not being at the edge of its output, instead of moving it to its parent container.

The bug occurred because the check for moving containers across outputs with non-default workspace layouts (issue #1603) did not actually verify that the moved window lies at the edge of the workspace, despite what its comment said.

Fixes #2466.

### DIFF
--- a/testcases/t/537-move-single-to-output.t
+++ b/testcases/t/537-move-single-to-output.t
@@ -1,0 +1,63 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • http://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • http://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • http://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Tests that windows inside containers with a single child do not jump
+# over other containers and erratically move across outputs.
+# Ticket: #2466
+# Bug still in: 4.14-63-g75d11820
+use i3test i3_config => <<EOT;
+# i3 config file (v4)
+font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
+
+fake-outputs 640x480+0+0,640x480+640+0
+
+workspace left output fake-0
+workspace right output fake-1
+EOT
+
+cmd 'workspace right';
+
+# ┌───────────────────────────┐ ┌───────────────────────────┐
+# │         Output 1          │ │     Output 2 - splith     │
+# │                           │ │┌───────────┐┌────────────┐│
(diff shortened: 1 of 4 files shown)


Airblader:
Thanks for the PR! This looks mostly good to me.

@stapelberg We should probably rebase this to also have the changes to the tests you made, right? Also, we should perhaps put this in the bugfix.

testcases/t/535-move-single-to-output.t
```diff
@@ -0,0 +1,70 @@
...
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Tests that windows inside containers with a single child do not jump
+# over other containers and erratically move across outputs.
+# Ticket: #2466
+# Bug still in: 4.14-63-g75d11820
+use List::Util qw(first);
+use i3test i3_autostart => 0;
```
stapelberg: Please use `i3_config` here, see https://github.com/i3/i3/pull/2945

CyberShadow: Done.

testcases/t/537-move-single-to-output.t
```diff
@@ -0,0 +1,64 @@
...
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Tests that windows inside containers with a single child do not jump
+# over other containers and erratically move across outputs.
+# Ticket: #2466
+# Bug still in: 4.14-63-g75d11820
+use List::Util qw(first);
```
stapelberg: Remove this line, you’re not using `first` anywhere.

CyberShadow: Done.

CyberShadow: Also removed unused `open_window` arguments.

Issue #2352, opened 2016-05-15
[feature request] Disable focus wrapping


Praron:
I think, that i3 is just beautiful, but there's one small thing, that I need and I can't make it, it's a disabling of focus wrapping. That will allow half-blind control in some cases or when it's difficult to see current focus.

Related information: https://faq.i3wm.org/question/874/how-to-disable-focus-wrapping/

stapelberg:
It’s been requested by 3 people in the last 4 years. I don’t think the low demand justifies the additional mental complexity (see also https://faq.i3wm.org/question/778/why-is-patch-not-merged-and-made-optional/).

I recommend to patch the source code if you feel strongly about this feature.

CyberShadow:
I would very much like this too

hencjo:
For the record, I'd love this feature as well. Idempotency over toggling.

CyberShadow:
I don't think I've ever used focus wrapping intentionally, but I do remember times of momentary frustration when I vainly attempted so focus the left/right-most window by holding $mod + arrow key, or confusion when focus warped to a seemingly arbitrary window (because the wrapping is done against the container, not output). I must assume that the behaviour is useful to some users, but it's been nothing but an anti-feature to me.

@Airblader Would you be willing to carry a patch for an option to disable focus wrapping in i3-gaps?

Airblader:
@stapelberg Are we sure about this decision? I do think there's a bit more demand for it than that (see these comments, but also other places). This feature does indeed make sense to me.

@CyberShadow Let's see whether we don't want to add this in i3 afterall; if not, I will consider it for i3-gaps based on how much of a patch it is (my main focus with i3-gaps right now is not so much adding features, but keeping it maintainable with upstream)

Airblader:
I'll reopen for now and based on what @stapelberg thinks he can close again. :-)

stapelberg:
Okay, let’s move this forward one step: if someone can suggest a PR and the change doesn’t turn out overly complicated, I’ll consider merging it.

PR #2953, opened 2017-09-15
Add "focus_wrapping" option

CyberShadow:
This implements enhancement request #2352.

The actual change is trivial:

```diff
 void tree_next(char way, orientation_t orientation) {
-    _tree_next(focused, way, orientation, true);
+    _tree_next(focused, way, orientation, config.focus_wrapping);
 }
```

The rest is bureaucracy. :)

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1039,12 +1039,28 @@ popup_during_fullscreen smart
 
 === Focus wrapping
 
-When in a container with several windows or child containers, the opposite
-window will be focused when trying to move the focus over the edge of a
-container (and there are no other containers in that direction) -- the focus
-wraps. If however there is another window or container in that direction, focus
-will be set on that window or container. This is the default behavior so you
-can navigate to all your windows without having to use +focus parent+.
+By default, when in a container with several windows or child containers, the
+opposite window will be focused when trying to move the focus over the edge of
+a container (and there are no other containers in that direction) -- the focus
+wraps.
+
+If desired, you can disable this behavior using the +focus_wrapping+
+configuration directive:
+
+*Syntax*:
+---------------------
+focus_wrapping yes|no
+---------------------
+
+*Example*:
+-----------------
+focus_wrapping no
+-----------------
+
+By default, focus wrapping does not occur if there is another window or
+container in the specified direction, and focus will instead be set on that
+window or container. This is the default behavior so you can navigate to all
+your windows without having to use +focus parent+.
 
 If you want the focus to *always* wrap and you are aware of using +focus
 parent+ to switch to different containers, you can use the
--- a/docs/userguide
+++ b/docs/userguide
@@ -1044,38 +1044,34 @@ opposite window will be focused when trying to move the focus over the edge of
 a container (and there are no other containers in that direction) -- the focus
 wraps.
 
-If desired, you can disable this behavior using the +focus_wrapping+
-configuration directive:
+If desired, you can disable this behavior by setting the +focus_wrapping+
+configuration directive to the value +no+.
 
-*Syntax*:
----------------------
-focus_wrapping yes|no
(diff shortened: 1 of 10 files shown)


Airblader:
The change looks good to me. I'm not a *huge* fan of having both `focus_wrapping` and `force_focus_wrapping` and would prefer a single `forcus_wrapping yes|no|force` instead; but removing `force_focus_wrapping` would be a breaking change.

@stapelberg How do you feel about that? We could do what we did in other places and add `focus_wrapping force` as an alias to `force_focus_wrapping yes`, focus the documentation on the new syntax and only keep a small mention of the old syntax for trackability when someone searches for it.

testcases/t/538-disable_focus_wrapping.t
```diff
@@ -0,0 +1,52 @@
...
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Tests that focus does not wrap when focus_wrapping is disabled in
+# the configuration.
+# Ticket: #2352
+# Bug still in: 4.14-72-g6411130c
+use List::Util qw(first);
```
stapelberg: Remove this line, you’re not using `first`

docs/userguide
```diff
@@ -1033,12 +1033,28 @@ popup_during_fullscreen smart
...
 === Focus wrapping
 
-When in a container with several windows or child containers, the first
-container will be focused when you use +focus down+ / +focus left+ on the last
-container -- the focus wraps. If however there is another window or container
-in that direction, focus will be set on that window or container. This is the
-default behavior so you can navigate to all your windows without having to use
-+focus parent+.
+By default, when in a container with several windows or child containers, the
+first container will be focused when you use +focus down+ / +focus left+ on the
```
stapelberg: I realize you haven’t touched this, but I don’t quite see why we only mention “focus down / focus left”, when any directional focus command might trigger wrapping.

Maybe “try to move focus over the edge of a container” is a more accurate term?

stapelberg:
> @stapelberg How do you feel about that? We could do what we did in other places and add focus_wrapping force as an alias to force_focus_wrapping yes, focus the documentation on the new syntax and only keep a small mention of the old syntax for trackability when someone searches for it.

Sounds good!

CyberShadow:
Sorry for the delay - I got involved with another project which has been rather time-consuming. I hope to make time to get back to this soon.

> We could do what we did in other places

What's an example option which has been changed in this way that I can use as a reference?

Airblader:
The special value `1pixel` for `border` is an example.

CyberShadow:
Thanks, feedback addressed. I wasn't sure if I should add a test for testing the new option names - let me know if I should.

parser-specs/config.spec
```diff
@@ -203,6 +204,11 @@ state MOUSE_WARPING:
   value = 'none', 'output'
       -> call cfg_mouse_warping($value)
 
+# focus_wrapping
+state FOCUS_WRAPPING:
+  value = word
```
stapelberg: Instead of using `word`, could you enumerate the possible values here? I.e. 1, yes, true, on, enable, active, 0, no, false, off, disable, inactive, force. That way, i3 can produce more useful error messages. Also, that way we can keep using `strcmp` instead of `strcasecmp` in `CFGFUN(focus_wrapping)`.

CyberShadow: Done.

Speaking of this, I noticed there doesn't seem to be any validation for boolean options, so e.g. `enalbe` will be silently parsed to `false`. I think there is room for improvement there too.

stapelberg: Agreed! Pull requests welcome.

Issue #2958, opened 2017-09-16
enhancement: make the IPC interface support network byte order


stapelberg:
Currently (4.14), i3 specifies byte order to be native byte order.

When writing that, we thought it would be fastest and most convenient for downstream programmers to not have to do any conversions.

We didn’t realize at the time that some programming languages might allow to specify a byte order, but not have the concept of a native byte order. This [discussion of native endianness in Go](https://www.reddit.com/r/golang/comments/3a88ay/does_go_need_native_endianess/) is just one example.

Ideally, the IPC interface would be easy to use from any language, and having to resort to unsafe hacks in modern languages is not a good experience.

Hence, I propose we extend the IPC interface as follows:

If the first message contains the magic string `i3-bge` (for big endian) instead of `i3-ipc`, then all values henceforth are assumed to be in network byte order and i3 will send messages with the `i3-bge` magic string itself.

Then, newer IPC libraries can just use network byte order.

Libraries which want to maintain compatibility with older i3 versions can start by sending an `i3-bge` message (e.g. `GET_WORKSPACES`, which is available in all versions of i3) and falling back to `i3-ipc` messages if the reply to the message is a closed connection instead of a `WORKSPACES` reply.

Thoughts? cc @Airblader @acrisci

Merovius:
Bikeshedding: `nbo` is also the acronym for "native byte order". Why not `i3-be`, or somesuch? Personally, I find "Big Endian" to be less ambiguous than "Network Byte Order" (if nothing else, because "Big Endian" has a counterpart "Little Endian", whereas the opposite of "Network Byte Order" is… "Host Byte Order"? Which isn't necessarily different).

stapelberg:
I’d like to keep the length the same to keep changes to the existing parsing code at a minimum. Let’s use `i3-bge` for Big Endian to avoid the acronym issue you pointed out.

acrisci:
I think I'd like it better if we made an explicit break in the ipc protocol and deprecate the old version instead of giving the impression that either are viable. That way we're future proof in case we need to make another break. It would only apply to the big endian version.

So maybe the magic should be `i3-ipc-v2` or `i3ipc2` or something like that.

stapelberg:
I sent PR #2961 which documents a technique to auto-detect i3’s byte order in memory-safe languages.

[...]

PR #2961, opened 2017-09-17
ipc: document how to detect i3’s byte order in memory-safe languages

stapelberg:
related to issue #2958 

cc @Merovius @acrisci @Airblader

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -889,3 +889,63 @@ Rust::
 	* https://github.com/tmerr/i3ipc-rs
 OCaml::
 	* https://github.com/Armael/ocaml-i3ipc
+
+== Appendix A: Detecting byte order in memory-safe languages
+
+Some programming languages such as Go don’t offer a way to serialize data in the
+native byte order of the machine they’re running on without resorting to tricks
+involving the +unsafe+ package.
+
+The following technique can be used (and will not be broken by changes to i3) to
+detect the byte order i3 is using:
+
+1. The byte order dependent fields of an IPC message are message type and
+   payload length.
+
+   * The message type +RUN_COMMAND+ (0) is the same in big and little endian, so
+     we can use it in either byte order to elicit a reply from i3.
+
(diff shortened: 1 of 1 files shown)


docs/ipc
```diff
@@ -889,3 +889,63 @@ Rust::
...
+     footnote:[A small payload is important: that way, we circumvent dealing
+     with UNIX domain socket buffer sizes, whose size depends on the
+     implementation/operating system. Exhausting such a buffer results in an i3
+     deadlock unless you concurrently read and write, which — depending on the
+     programming language — makes the technique much more complicated.].
+
+   * If i3 is running in little endian, this message is read in its entirety due
+     to the byte order independent payload length, then
+     https://github.com/i3/i3/blob/d726d09d496577d1c337a4b97486f2c9fbc914f1/src/ipc.c#L1188[silently
+     discarded] due to the unknown message type.
```
acrisci: linking to line numbers in documentation is a little risky

Airblader: But this points to a specific commit, so that danger isn't there. I still don't like it, though, because we also shouldn't reference code that might soon be outdated.

stapelberg: Regarding outdated code: the new code (and I’m not positive that we will actually change the IPC code much) will behave the same way, though — that’s what this document assures. If the code doesn’t behave the same way, the link provides a convenient entry point to a git bisecting session.

stapelberg:
For the record: I tested this on linux/amd64 (little endian) and linux/mips (big endian). If there are any other platforms which would be interesting to test, please let me know.

acrisci:
I've never tested i3ipc-python on other architectures but I've never gotten any complaints about byte order.

stapelberg:
> I've never tested i3ipc-python on other architectures but I've never gotten any complaints about byte order.

You got one now :) https://github.com/acrisci/i3ipc-python/issues/67 (I also sent a PR to fix it)

PR #2964, opened 2017-09-18
291-swap.t: make tests fail less

orestisf1993:


testcases/t/291-swap.t
```diff
@@ -220,14 +220,16 @@ $ws1 = fresh_workspace;
 
 $A = open_window(wm_class => 'mark_A');
 $F = open_window();
-$F->fullscreen(1);
+cmd 'fullscreen';
```
stapelberg: Here and below, why the switch to `cmd 'fullscreen'`? That is an entirely different code path (and wouldn’t require the `sync_with_i3` afterwards).

testcases/t/291-swap.t
```diff
@@ -253,21 +255,20 @@ $ws1 = fresh_workspace;
 
 open_window;
 $A = open_window(wm_class => 'mark_A');
-cmd "split v";
+cmd 'split v';
 open_window;
-cmd "focus parent";
-cmd "fullscreen enable";
+cmd 'focus parent';
+cmd 'fullscreen';
```
stapelberg: Why the change from `fullscreen enable` to `fullscreen`?

testcases/t/291-swap.t
```diff
@@ -253,21 +255,20 @@ $ws1 = fresh_workspace;
...
 
 $ws2 = fresh_workspace;
 $B = open_window(wm_class => 'mark_B');
-$B->fullscreen(1);
+cmd 'fullscreen';
 
 cmd '[con_mark=B] swap container with mark A';
-
 sync_with_i3;
-does_i3_live;
```
stapelberg: Why did you remove this? If this check causes trouble, it’s because i3 is crashing. We should fix the crash instead.

testcases/t/291-swap.t
orestisf1993: unrelated change for consistency with the rest of the file.

testcases/t/291-swap.t
orestisf1993: unrelated change.

It doesn't cause trouble but it doesn't do anything important either. In fact, this test never caused i3 to break and even if it did we don't need a `does_i3_live` to detect that.

orestisf1993:
Removed the unrelated changes. @stapelberg is the new sync usage in reasonable places now?

stapelberg:
Looks much better, thank you very much!

PR #2965, opened 2017-09-18
Bugfix: don’t invalidate layout upon invalid 'layout toggle' params

### LINKED ISSUE #2903: Certain commands crash i3

Binary i3 version:  4.14 (2017-09-04, branch "4.14") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.14 (2017-09-04, branch "4.14") (pid 8312)
Loaded i3 config: /home/cornerman/.i3/config (Last modified: Thu 07 Sep 2017 07:04:26 PM CEST, 92 seconds ago)

http://logs.i3wm.org/logs/5698829030522880.bz2

**What I did:**

I am maintaining a script to generate a bash completion script for i3 (https://github.com/cornerman/i3-completion). It runs bogus command against the i3 ipc interface and tries to find valid commands from the hints of the command parser ("Expected one of these tokens: ..."). With version `4.14`, a certain sequence of commands crashes i3.

Reproduction, which runs i3 in Xephyr and then send some commands to that instance:
```sh
git clone https://github.com/cornerman/i3-completion
cd i3-completion
# depends on: Xephyr, perl, perl-anyevent-i3
./gen.sh
```

**What I saw:**

After receiving some commands, i3 crashes with:
```
i3: ../../i3/src/ipc.c:334: dump_node: Assertion `false' failed.
i3: ../../i3/src/ipc.c:334: dump_node: Assertion `false' failed.
```
Which is this line
```c
DLOG("About to dump layout=default, this is a bug in the code.\n");
assert(false);
```

The last received command itself does not seem to be the problem, when I just try that it works.

**What I expected instead:**

i3 should keep running and the `gen.sh` script should complete.

stapelberg:
fixes #2903

### DIFF
--- a/src/con.c
+++ b/src/con.c
@@ -1841,7 +1841,9 @@ void con_toggle_layout(Con *con, const char *toggle_mode) {
             }
         }
 
-        con_set_layout(con, new_layout);
+        if (new_layout != L_DEFAULT) {
+            con_set_layout(con, new_layout);
+        }
     } else if (strcasecmp(toggle_mode, "all") == 0 || strcasecmp(toggle_mode, "default") == 0) {
         if (parent->layout == L_STACKED)
             con_set_layout(con, L_TABBED);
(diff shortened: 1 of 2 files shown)


maxh-:
Much appreciated!!

PR #2967, opened 2017-09-18
docs/ipc: "urgent": complete the list of container types

stapelberg:
Thanks chressie!


maxh-:
Does the focus array mean that windows get added there as they are focused?

stapelberg:
> Does the focus array mean that windows get added there as they are focused?

That’s unrelated to this change, so let’s move this discussion elsewhere if you have further questions: yes, the focus array contains an up-to-date list of node IDs in focus order, i.e. it changes whenever you change focus.

Issue #1819, opened 2015-07-29
Focus via criteria behind fullscreen container


cornerman:
Currently, if I have a fullscreen container and focus any underlying window via criteria, the fullscreen container is not disabled.

Say, I have two windows on a workspace: window A and window B. Then, I mark window A with `mark a`, fullscreen window B and switch to another workspace. If i now issue `[ con_mark = a ] focus`, this would actually switch focus to window A but leave the fullscreened window B in place. So, I end up with having focus on window A while window B is still blocking my view.

It behaves differently if I do not switch to another workspace and keep window B focused. Then the focus via criteria would actually do nothing and keep focus on the fullscreened window B.

I would expect focus via criteria to always take precedence over any fullscreen container. Other than directional focus changes, the user explicitly states which window should be focused.

i3 version 4.10.2-238-gc8b4303 (2015-07-27, branch "next") © 2009 Michael Stapelberg and contributors

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

acrisci:
I think that if you focus a tiling window behind the fullscreen container, it should drop out of fullscreen.

stapelberg:
Agreed. @cornerman’s use-case should behave as @acrisci noted.

teto:
I can totally imagine though a mode where once you are in fullscreen, you remain in fullscreen while focusing different windows (the focused window goes fullscreen while the previously focused exits fullscreen).

cornerman:
@teto Then, you could probably script something like that by listening to the window focus event and enabling fullscreen on every focused window.

teto:
Of course. Just wanted to illustrate another point of view since Michael
and Tony seemed to think leaving fullscreen is the obvious choice. It's
difficult to do comparisons, in windows if you alt tab between fullscreen
windows (e.g. games) they should remain fullscreen but flash usually leaves
fullscreen when you alt+tab (sometimes not, maybe it depends on the player).

2015-07-30 21:54 GMT+02:00 cornerman <notifications@github.com>:

Airblader:
I can reproduce this, but although the wrong window is focused, the x11 input focus does seem to be on the fullscreen container correctly.

orestisf1993:
Does this apply to global fullscreen mode as well?

Airblader:
[...]

PR #2969, opened 2017-09-19
Make 'focus' disable blocking fullscreen windows

orestisf1993:
This is my take on the issue. Before I move on I'd like some feedback on my take and the comments in the last commit message. The reason I didn't touch `con_fullscreen_permits_focusing()` is that too many stuff depend on it.

If this looks ok then I will:
- [x] ~~Mention that `focus` disable fullscreen in the userguide.~~
- [x] Fix `focus floating` as well.

Fixes issue #1819.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1951,6 +1951,9 @@ bindsym $mod+t floating toggle
 To change focus, you can use the +focus+ command. The following options are
 available:
 
+<criteria>::
+    Sets focus to the container that matches the specified criteria.
+    See <<command_criteria>>.
 left|right|up|down::
         Sets focus to the nearest container in the given direction.
 parent::
@@ -1970,6 +1973,7 @@ output::
 
 *Syntax*:
 ----------------------------------------------
+<criteria> focus
 focus left|right|down|up
 focus parent|child|floating|tiling|mode_toggle
 focus output left|right|up|down|primary|<output>
@@ -1977,6 +1981,9 @@ focus output left|right|up|down|primary|<output>
 
 *Examples*:
 -------------------------------------------------
+# Focus firefox
+bindsym $mod+F1 [class="Firefox"] focus
+
 # Focus container on the left, bottom, top, right
 bindsym $mod+j focus left
 bindsym $mod+k focus down
--- a/src/commands.c
+++ b/src/commands.c
@@ -1251,6 +1251,20 @@ void cmd_focus_direction(I3_CMD, const char *direction) {
 }
 
 /*
+ * Focus a container and disable any other fullscreen container not permitting the focus.
+ *
+ */
+static void cmd_focus_force_focus(Con *con) {
+    /* Disable fullscreen container in workspace with container to be focused. */
+    Con *ws = con_get_workspace(con);
+    Con *fullscreen_on_ws = (focused && focused->fullscreen_mode == CF_GLOBAL) ? focused : con_get_fullscreen_con(ws, CF_OUTPUT);
+    if (fullscreen_on_ws && fullscreen_on_ws != con && !con_has_parent(con, fullscreen_on_ws)) {
+        con_disable_fullscreen(fullscreen_on_ws);
+    }
+    con_focus(con);
+}
+
+/*
  * Implementation of 'focus tiling|floating|mode_toggle'.
  *
  */
@@ -1274,7 +1288,7 @@ void cmd_focus_window_mode(I3_CMD, const char *window_mode) {
             (!to_floating && current->type == CT_FLOATING_CON))
             continue;
 
-        con_focus(con_descend_focused(current));
+        cmd_focus_force_focus(con_descend_focused(current));
         success = true;
         break;
     }
@@ -1341,13 +1355,6 @@ void cmd_focus(I3_CMD) {
         if (!ws)
             continue;
 
-        /* Check the fullscreen focus constraints. */
-        if (!con_fullscreen_permits_focusing(current->con)) {
-            LOG("Cannot change focus while in fullscreen mode (fullscreen rules).\n");
-            ysuccess(false);
-            return;
-        }
-
         /* In case this is a scratchpad window, call scratchpad_show(). */
         if (ws == __i3_scratch) {
             scratchpad_show(current->con);
@@ -1371,7 +1378,7 @@ void cmd_focus(I3_CMD) {
          * So we focus 'current' to make it the currently focused window of
          * the target workspace, then revert focus. */
         Con *currently_focused = focused;
-        con_focus(current->con);
+        cmd_focus_force_focus(current->con);
         con_focus(currently_focused);
 
         /* Now switch to the workspace, then focus */
--- a/src/commands.c
+++ b/src/commands.c
@@ -1257,28 +1257,34 @@ void cmd_focus_direction(I3_CMD, const char *direction) {
 void cmd_focus_window_mode(I3_CMD, const char *window_mode) {
     DLOG("window_mode = %s\n", window_mode);
 
-    Con *ws = con_get_workspace(focused);
-    if (ws != NULL) {
-        if (strcmp(window_mode, "mode_toggle") == 0) {
-            if (con_inside_floating(focused))
-                window_mode = "tiling";
-            else
-                window_mode = "floating";
-        }
-        Con *current;
-        TAILQ_FOREACH(current, &(ws->focus_head), focused) {
-            if ((strcmp(window_mode, "floating") == 0 && current->type != CT_FLOATING_CON) ||
-                (strcmp(window_mode, "tiling") == 0 && current->type == CT_FLOATING_CON))
-                continue;
-
-            con_focus(con_descend_focused(current));
-            break;
-        }
+    bool to_floating = false;
+    if (strcmp(window_mode, "mode_toggle") == 0) {
+        to_floating = !con_inside_floating(focused);
+    } else if (strcmp(window_mode, "floating") == 0) {
+        to_floating = true;
+    } else if (strcmp(window_mode, "tiling") == 0) {
+        to_floating = false;
     }
 
-    cmd_output->needs_tree_render = true;
-    // XXX: default reply for now, make this a better reply
-    ysuccess(true);
+    Con *ws = con_get_workspace(focused);
+    Con *current;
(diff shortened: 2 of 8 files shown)


orestisf1993:
Pushed changes for `focus floating | tiling | mode_toggle`.

Airblader:
I'm not sure I like the approach. Before going into other ideas, is there a reason why it cannot just be »whenever `con_focus` is called on a container that is currently not focused but lives on a workspace with a fullscreened window, unfullscreen that window«?

orestisf1993:
I don't think it's a great idea because it's a very core function and we will be adding more side effects there and cause lot's of headaches. The only side effect it currently has is urgency hints and it already is buggy: `H [ V [ A B ] C]` focus C, set the urgent flag on A without focusing, `i3-msg '[con_mark=A]' move right`: urgent flag is lost.

Anyway, I'll try to make it work with `con_focus` and see if I can pass the tests with a reasonable amount of work.

Airblader:
But I don't quite see (yet) why the behavior change only applies to commands. A static outside function runs the risk of not using it in the places where it has to be used.

In general when I focus a container I should expect the function to do everything that has to be done to do so properly.

Of course I'd try to stay away from complicated stuff to serve all kinds of edge cases, hence me asking whether such cases even exist — naively, the change would sound simple to me.

@stapelberg What do you think about the current approach of handling this in commands.c?

orestisf1993:
IMO the current focus stack thing has some problems because of `con_focus` abuse. `move_tree` is a function that misbehaves for sure but the same erroneous logic could have been applied to other places as well.

Consider the layout: `H [ V1 [ A B ] V2 [ C D E ] ]`. Mark A as A. Focus in order A->B->C->D->E. Run `i3-msg '[con_mark=A]' move right` (from E). Start closing terminals. Focus order should be E->D->C->A->B (will explain why A > B)*. Instead it is E->A->D->C->B. As an extra problem A loses it's urgency flag. Reason is that `move_tree` calls `con_focus` and bumps A's focus order to the top and then `cmd_move_direction` bumps E's order again to the top.

I encountered the same situation with floating windows in https://github.com/i3/i3/pull/2941. I tried to explain my logic in a [comment](https://github.com/orestisf1993/i3/blob/e571bc03e60f4ba346caa0e8fc83d3f7c1d27b15/src/floating.c#L188-L197) there.

When node `A` is moved from `V1` to `V2` we should compare the focus order of the parents (in this case `V2` > `V1`) and put `A` either in the tail or the head of the new parent.  
*: This is the reason why A > B, we maintain focus order relative to the parent. If `V2` is focused then all nodes of `V2` are to be focused before any of the nodes in `V1`.

Anyway, to do what you are proposing we should split `con_focus` to a function that is actually used to focus specific containers (our case) and a function that fixes the focus stack (eg the way `move_tree` uses `con_focus` currently). Without the split a problem more relevant to our issue here is that moving a normal container to a workspace with a fullscreen container (eg to a different output) will unfullscreen the latter container.

Airblader:
> Without the split a problem more relevant to our issue here is that moving a normal container to a workspace with a fullscreen container (eg to a different output) will unfullscreen the latter container.

Why is this an issue? Isn't that the behavior we should expect?

Anyway, I just think we should carefully review the best way to tackle these issues. I'd like to avoid getting into a monkey-patching situation. That's why I think @stapelberg should join in here. :-)

stapelberg:
IIUC, the tension here is between `con_focus` being used as a building block for other commands (e.g. `move`), and it being used directly for the `focus` command.

I agree that it would be cleaner if `con_focus` just did the right thing, but at this point, even I cannot keep in my head all the implicit assumptions we make when calling `con_focus` as a building block. Hence, I don’t think we should require cleaning up `con_focus` in the context of this change.

Adding a `cmd_con_focus` wrapper in `src/commands.c` (similar to what the PR currently does) actually seems like a good middle ground to me. We should document the nature of the two functions, though: explicitly call `cmd_con_focus` the helper which should be used by all `focus` commands, and explicitly call `con_focus` the i3-internal building block for other commands.

Does that sound reasonable?

Airblader:
Yeah, OK. It comes with a grain of salt, but it's fine with me. :-)

orestisf1993:
> ... explicitly call cmd_con_focus the helper which should be used by **all** focus commands

Do we want to have this logic for all focus commands after all?

stapelberg:
> Do we want to have this logic for all focus commands after all?

I’d say for all `focus` commands, yes. Is there any reason against it?

orestisf1993:
Actually I already covered all uses of `con_focus` for `cmd_focus_X` commands. `focus left` etc don't use `con_focus`.

src/commands.c
```diff
@@ -1258,33 +1258,54 @@ void cmd_focus_direction(I3_CMD, const char *direction) {
 }
 
 /*
+ * Focus a container and disable any other fullscreen container not permitting the focus.
+ *
+ */
+static void cmd_focus_force_focus(Con *con, Con *ws) {
```
orestisf1993: Can remove the `ws` argument for cleanness, minor extra cycles.

testcases/t/156-fullscreen-focus.t
```diff
@@ -323,6 +314,92 @@ verify_move(2, 'prevented move to workspace by name');
...
+    cmd 'fullscreen';
+    is($x->input_focus, $second->id, 'fullscreen window focused');
+    is_num_fullscreen($tmp, 1, '1 fullscreen window');
+
+    cmd_focus_helper($mode, $first);
+    is_num_fullscreen($tmp, 0, 'no fullscreen windows');
+}
+
+################################################################################
+# When the fullscreen window is in an other workspace it should maintain it's
```
orestisf1993: s/it's/its

💩

orestisf1993:
Updated userguide.

src/commands.c
Airblader: Yeah, I'd remove it. This only runs on user commands, not constantly, so I think a couple of cycles don't justify carrying over the information.

docs/userguide
```diff
@@ -1927,24 +1930,28 @@ child::
 	The opposite of +focus parent+, sets the focus to the last focused
 	child container.
 floating::
-	Sets focus to the last focused floating container.
+	Sets focus to the last focused floating container, disabling any fullscreen windows that would block the change.
```
Airblader: Props for considering documentation, but in this case this strikes me as a detail that is more distracting than helping. We also don't document that setting the focus clears urgency hints etc. etc. :-)

orestisf1993: Ok reverted the change.

testcases/t/156-fullscreen-focus.t
```diff
@@ -323,6 +314,92 @@ verify_move(2, 'prevented move to workspace by name');
...
+    cmd "$command";
+    sync_with_i3;
+
+    is($x->input_focus, $target->id, "target window got focus after '$command'");
+}
+
+# 0: first & second tiling, focus using id
+# 1: first floating, second tiling, focus using 'focus floating'
+# 2: first tiling, second floating, focus using 'focus tiling'
+for my $mode (0 .. 2){
```
stapelberg: I’m not a big fan of modal tests. Could we pass a function to cmd_focus_helper instead?

src/commands.c
```diff
@@ -1258,33 +1258,55 @@ void cmd_focus_direction(I3_CMD, const char *direction) {
...
-            else
-                window_mode = "floating";
-        }
-        Con *current;
-        TAILQ_FOREACH(current, &(ws->focus_head), focused) {
-            if ((strcmp(window_mode, "floating") == 0 && current->type != CT_FLOATING_CON) ||
-                (strcmp(window_mode, "tiling") == 0 && current->type == CT_FLOATING_CON))
-                continue;
+    if (ws == NULL) {
+        yerror("No workspace was previously active.");
```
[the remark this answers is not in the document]
orestisf1993: Actually, let's just delete the check. It doesn't make sense for the focused container to not have a workspace. I took the check from `cmd_move_con_to_workspace_back_and_forth` but it is not the same situation.

testcases/t/156-fullscreen-focus.t
orestisf1993: can you elaborate on this a bit please?

stapelberg: Actually, playing around with the code, I think it would be better to just be explicit and incur the repetition, i.e. change mode 0 to:

```perl
kill_all_windows;
$tmp = fresh_workspace;
my $first = open_window;
my $second = open_window;

cmd 'fullscreen';
is($x->input_focus, $second->id, 'fullscreen window focused');
is_num_fullscreen($tmp, 1, '1 fullscreen window');

cmd '[id="'. $target->id .'"] focus';
sync_with_i3;

is($x->input_focus, $target->id, "target window got focus after '$command'");
is_num_fullscreen($tmp, 0, 'no fullscreen windows');
```

…and so on. I think the helper functions don’t pay for themselves here — they make the code significantly harder to follow.

orestisf1993:
@stapelberg I've made the changes you requested.

I've also added a new error reply with 28c9fbccb2841e93a5e4f7ff75d3bec386da81b8, tell me if you don't like it.

stapelberg:
Thanks! I’ll take a closer look in the evening.

src/commands.c
```diff
@@ -1250,35 +1250,55 @@ void cmd_focus_direction(I3_CMD, const char *direction) {
...
+}
+
 /*
  * Implementation of 'focus tiling|floating|mode_toggle'.
  *
  */
 void cmd_focus_window_mode(I3_CMD, const char *window_mode) {
     DLOG("window_mode = %s\n", window_mode);
 
+    bool to_floating;
```
stapelberg: nit: can you initialize this to `false` please? I think clang-analyze will complain about it otherwise.

testcases/lib/i3test/Test.pm
```diff
@@ -59,6 +60,17 @@ sub is_num_children {
     $tb->is_num($got_num_children, $num_children, $name);
 }
 
+sub is_num_fullscreen {
```
stapelberg: Can you please add some perldoc for this function, now that it is in i3test::Test?

orestisf1993:
up

stapelberg:
Thanks!

PR #2970, opened 2017-09-19
Do not canonicalize special output names 

### LINKED ISSUE #2948: i3bar tray stays on the old primary output after switching primary output
**What I did:**
I've been using dualhead setup (eDP-1 and DP-2) with DP-2 set as a primary output. Then, I've turned off DP-2 and designated eDP-1 as a new primary by running:
`xrandr --output eDP-1 --primary --mode 1920x1080 --pos 0x0 --rotate normal --output DP-1 --off --output DP-2 --off --output DP-3 --off --output HDMI-1 --off --output HDMI-2 --off --output HDMI-3 --off`

My i3 configuration contains the following configuration for i3bar:
```
bar {
  status_command i3status
  position top
  tray_output primary
}
```

**What I saw:**
Tray stayed inside the i3bar on the disabled output (so it was not visible to the user). Only after I've restarted i3 it moved to the i3bar on eDP-1 (new primary).

**What I expected instead:**
I've expected the tray to move to the i3bar displayed on a new primary output (and be visible). That was the behaviour that it exhibited before recent changes.

**Alternative scenario:**
If I enabled DP-2 and designated it as a new primary, the tray would remain visible (and fully working) on eDP-1. Only after i3 restart it would move to the new primary.

**More info:**
This breakage was introduced in commit 6de9cdd96fe89c63d874b13a52ac2acfb3907dd1 :
```
commit 6de9cdd96fe89c63d874b13a52ac2acfb3907dd1
Author: Vladimir Panteleev <git@thecybershadow.net>
Date:   Mon Sep 11 11:15:56 2017 +0000

    ipc: Canonicalize output names in bar configuration
    
    Convert the output names specified in the "output" and "tray_output"
    fields in bar blocks in i3's configuration to the referred output's
    primary name. This allows specifying names other than the primary
    output's name in the given fields without changing the IPC protocol.

:040000 040000 345b4983d784d0e7766dd61853f704abfdbb9c7f a3754bfb32a5dd648e0b1a1b7e835fdb10ef5462 M	src
```
If I remove the  canonicalize_output_name call (leaving only the original output name) from line 612 of `src/ipc.c` then thy tray moves correctly to the new primary output. My guess is that we are not re-canonicalizng output names somewhere after the 'primary' designation is changed. Restarting i3 forces the canonicalization of all names so it fixes this issue.

**Logs**
I haven't found anything interesting in i3 or i3bar logs, but I can regenerate them if needed.

**i3 version used:**
`Running i3 version: 4.14-63-g75d11820 (2017-09-13, branch "next")`
`Running i3 version: 4.14-87-g607e97e6 (2017-09-17, branch "next")`

*Edit: Updated versions and moved explanation for missing logs.*

CyberShadow:
Fixes #2948.

### DIFF
--- a/src/fake_outputs.c
+++ b/src/fake_outputs.c
@@ -28,8 +28,9 @@ static Output *get_screen_at(unsigned int x, unsigned int y) {
 /*
  * Creates outputs according to the given specification.
  * The specification must be in the format wxh+x+y, for example 1024x768+0+0,
+ * optionally followed by 'P' to indicate a primary output,
  * with multiple outputs separated by commas:
- *   1900x1200+0+0,1280x1024+1900+0
+ *   1900x1200+0+0P,1280x1024+1900+0
  *
  */
 void fake_outputs_init(const char *output_spec) {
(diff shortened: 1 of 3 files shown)


stapelberg:
Thank you very much!

PR #2973, opened 2017-09-21
Fixed issue of crashing when opening some new workspace referred to by number.

ViktorWalter:
i3 has a bug, where if I refer to the workspaces by number (names change dynamically in my use-case), then sometimes on switching to a yet unopened workspace leads to crash and logout.
I traced this issue to the src/workspace.c source, where the function `_workspace_show(Con *workspace)` accesses the previous workspace, after it was destroyed in this function, leading to access violation. Changing `output_push_sticky_windows(old_focus)` to `output_push_sticky_windows(focused)` fixes the issue. Other use-cases than mine may theoretically be affected.
I have originally made this pull request on the fork i3-gaps, and was told to bring it upstream.

### DIFF
--- a/src/workspace.c
+++ b/src/workspace.c
@@ -479,7 +479,7 @@ static void _workspace_show(Con *workspace) {
     ewmh_update_current_desktop();
 
     /* Push any sticky windows to the now visible workspace. */
-    output_push_sticky_windows(old_focus);
+    output_push_sticky_windows(focused);
 }
 
 /*


Airblader:
While I do believe that there might be a bug here, I also think the fix just isn't correct. The call is now done for a different workspace, which is wrong.

ViktorWalter:
I don't know about the test, but I am pretty sure that the call should have been done for the new focus in the first place, unless I am misunderstanding the purpose of `ouput_push_sticky_windows`. According to the comment, it iterates over all outputs, and pushes sticky windows to the currently visible workspace on that output. Doesn't that mean, that if the argument was to be `old_focus`, it would attempt to push those stickies to the previous workspace, even if it was destroyed?

ViktorWalter:
Also, the steps to reproduce: in my case this caused the crash wihout fail: Set shortcuts to switch workspaces by referring to them by numbers, not names. Say you have three monitors, with workspaces 1 - 3 automatically assigned to them right after login. Now, before starting any programs, switch to workspace 5. If that doesn't bring it down, switching to ws. 2 almost definitely will.

orestisf1993:
I've tried reproducing this with no success. IIUC, what you are describing is a very common setup among i3 users so I'd be surprised if these simple steps could trigger such a bug. Can you please share a minimal working example config (which when used can reliably reproduce the issue without any unnecessary fluff). Preferably using fake outputs.

Airblader:
I'm going to close this because I think the approach is incorrect and more investigation has to be done.

PR #2975, opened 2017-09-22
Fix move's focus bugs

orestisf1993:
This will also be used for https://github.com/i3/i3/pull/2941 if accepted.

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -374,11 +374,6 @@ void tree_move(Con *con, int direction) {
     }
 
 end:
-    /* We need to call con_focus() to fix the focus stack "above" the container
-     * we just inserted the focused container into (otherwise, the parent
-     * container(s) would still point to the old container(s)). */
-    con_focus(con);
-
     /* force re-painting the indicators */
     FREE(con->deco_render_params);
 
--- a/src/move.c
+++ b/src/move.c
@@ -13,6 +13,54 @@ typedef enum { BEFORE,
                AFTER } position_t;
 
 /*
+ * Returns the lowest container in the tree that has both a and b as descendants.
+ *
+ */
+static Con *lowest_common_ancestor(Con *a, Con *b) {
+    Con *parent_a = a;
+    while (parent_a) {
+        Con *parent_b = b;
+        while (parent_b) {
+            if (parent_a == parent_b) {
+                return parent_a;
+            }
+            parent_b = parent_b->parent;
+        }
+        parent_a = parent_a->parent;
+    }
+    assert(false);
+}
+
+/*
+ * Returns the direct child of ancestor that contains con.
+ *
+ */
+static Con *child_containing_con_recursively(Con *ancestor, Con *con) {
+    Con *child = con;
+    while (child && child->parent != ancestor) {
+        child = child->parent;
+        assert(child->parent);
+    }
+    return child;
+}
+
+/*
+ * Returns true if the given container is the focused descendant of ancestor, recursively.
+ *
+ */
+static bool is_focused_descendant(Con *con, Con *ancestor) {
+    Con *current = con;
+    while (current != ancestor) {
+        if (TAILQ_FIRST(&(current->parent->focus_head)) != current) {
+            return false;
+        }
+        current = current->parent;
+        assert(current->parent);
+    }
+    return true;
+}
+
+/*
  * This function detaches 'con' from its parent and inserts it either before or
  * after 'target'.
  *
@@ -23,6 +71,52 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
      * insert the entry before/after it, we call the on_remove_child callback
      * afterwards which might then close the con if it is empty. */
     Con *old_parent = con->parent;
+
+    /* We compare the focus order of the children of the lowest common ancestor. If con or
+     * its ancestor is before target's ancestor then con should be placed before the target
+     * in the focus stack. */
+    Con *lca = lowest_common_ancestor(con, parent);
+    if (lca == con) {
+        ELOG("Container is being inserted into one of its descendants.\n");
+        return;
+    }
+
+    Con *con_ancestor = child_containing_con_recursively(lca, con);
+    Con *target_ancestor = child_containing_con_recursively(lca, target);
+    bool moves_focus_from_ancestor = is_focused_descendant(con, con_ancestor);
+    bool focus_before;
+
+    /* Determine if con is going to be placed before or after target in the parent's focus stack. */
+    if (con_ancestor == target_ancestor) {
+        /* Happens when the target is con's old parent. Eg with layout V [ A H [ B C ] ],
+         * if we move C up. Target will be H. */
+        focus_before = moves_focus_from_ancestor;
+    } else {
+        /* Look at the focus stack order of the children of the lowest common ancestor. */
+        Con *current;
+        TAILQ_FOREACH(current, &(lca->focus_head), focused) {
+            if (current == con_ancestor || current == target_ancestor) {
+                break;
+            }
+        }
+        focus_before = (current == con_ancestor);
+    }
+
+    /* If con is the focused container in our old ancestor we place the new ancestor
+     * before the old ancestor in the focus stack. Example:
+     * Consider the layout [ H [ V1 [ A* B ] V2 [ C ] ] ] where A is focused. We move to
+     * a second workspace and from there we move A to the right and switch back to the
+     * original workspace. Without the change focus would move to B instead of staying
+     * with A. */
+    if (moves_focus_from_ancestor && focus_before) {
+        Con *place = TAILQ_PREV(con_ancestor, focus_head, focused);
+        TAILQ_REMOVE(&(lca->focus_head), target_ancestor, focused);
+        if (place) {
+            TAILQ_INSERT_AFTER(&(lca->focus_head), place, target_ancestor, focused);
+        } else {
+            TAILQ_INSERT_HEAD(&(lca->focus_head), target_ancestor, focused);
+        }
+    }
 
     con_detach(con);
     con_fix_percent(con->parent);
@@ -46,12 +140,28 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
 
     con->parent = parent;
 
+    if (parent == lca) {
+        if (focus_before) {
+            /* Example layout: H [ A B* ], we move A up/down. 'target' will be H. */
+            TAILQ_INSERT_BEFORE(target, con, focused);
+        } else {
+            /* Example layout: H [ A B* ], we move A up/down. 'target' will be H. */
+            TAILQ_INSERT_AFTER(&(parent->focus_head), target, con, focused);
+        }
+    } else {
+        if (focus_before) {
+            /* Example layout: V [ H [ A B ] C* ], we move C up. 'target' will be A. */
+            TAILQ_INSERT_HEAD(&(parent->focus_head), con, focused);
+        } else {
+            /* Example layout: V [ H [ A* B ] C ], we move C up. 'target' will be A. */
+            TAILQ_INSERT_TAIL(&(parent->focus_head), con, focused);
+        }
+    }
+
     if (position == BEFORE) {
         TAILQ_INSERT_BEFORE(target, con, nodes);
-        TAILQ_INSERT_HEAD(&(parent->focus_head), con, focused);
     } else if (position == AFTER) {
         TAILQ_INSERT_AFTER(&(parent->nodes_head), target, con, nodes);
-        TAILQ_INSERT_HEAD(&(parent->focus_head), con, focused);
     }
 
     /* Pretend the con was just opened with regards to size percent values.
--- a/testcases/t/113-urgent.t
+++ b/testcases/t/113-urgent.t
@@ -333,6 +333,34 @@ for ($type = 1; $type <= 2; $type++) {
     is($target_ws->{urgent}, 1, 'Target workspace is now marked urgent');
 
 ##############################################################################
+# Test that moving an unfocused container doesn't reset its urgency hint.
+##############################################################################
+    $tmp = fresh_workspace;
+    $win1 = open_window;
+    $win2 = open_window;
+    cmd 'split v';
+    $win3 = open_window;
+    set_urgency($win1, 1, $type);
+    sync_with_i3;
+
+    my $win1_info;
+
+    @content = @{get_ws_content($tmp)};
+    $win1_info = first { $_->{window} == $win1->id } @content;
+    ok($win1_info->{urgent}, 'win1 window is marked urgent');
+
+    cmd '[id="' . $win1->id . '"] move right';
+    cmd '[id="' . $win1->id . '"] move right';
+    @content = @{get_ws_content($tmp)};
+    $win1_info = first { $_->{window} == $win1->id } @content;
+    ok($win1_info->{urgent}, 'win1 window is still marked urgent after moving');
+
+    cmd '[id="' . $win1->id . '"] focus';
+    @content = @{get_ws_content($tmp)};
+    $win1_info = first { $_->{window} == $win1->id } @content;
+    ok(!$win1_info->{urgent}, 'win1 window is not marked urgent after focusing');
+
+##############################################################################
 
     exit_gracefully($pid);
 }
--- a/testcases/t/294-focus-order.t
+++ b/testcases/t/294-focus-order.t
@@ -106,4 +106,20 @@ $windows[0] = open_window;
 cmd 'move left';
 confirm_focus('split-v + move');
 
+######################################################################
+# Test that moving an unfocused container maintains the correct focus
+# order.
+# Layout: H [ A V1 [ B C D ] ]
+######################################################################
+
+fresh_workspace;
+$windows[3] = open_window;
+$windows[2] = open_window;
+cmd 'split v';
+$windows[1] = open_window;
+$windows[0] = open_window;
+
+cmd '[id=' . $windows[3]->id . '] move right';
+confirm_focus('split-v + unfocused move');
+
 done_testing;


include/con.h
```diff
@@ -115,6 +115,12 @@ Con *con_parent_with_orientation(Con *con, orientation_t orientation);
 Con *con_get_fullscreen_con(Con *con, fullscreen_mode_t fullscreen_mode);
 
 /**
+ * Returns the lowest container in the tree that has both a and b as descendants.
+ *
+ */
+Con *lowest_common_ancestor(Con *a, Con *b);
```
Airblader: Please name the function `con_*` as all the others.

src/con.c
```diff
@@ -488,6 +488,25 @@ Con *con_get_fullscreen_con(Con *con, fullscreen_mode_t fullscreen_mode) {
...
+        Con *parent_b = b;
+        while (parent_b) {
+            if (parent_a == parent_b) {
+                return parent_a;
+            }
+            parent_b = parent_b->parent;
+        }
+        parent_a = parent_a->parent;
+    }
+    return NULL;
```
Airblader: This should never happen, right? So we should probably `assert(false)` here.

src/move.c
```diff
@@ -24,6 +24,49 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
      * afterwards which might then close the con if it is empty. */
     Con *old_parent = con->parent;
 
+    Con *lca = lowest_common_ancestor(con, parent);
```
Airblader: We should drop a comment here explaining what the following lines are achieving.

src/move.c
```diff
@@ -24,6 +24,49 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
...
+    Con *ancestor = con;
+    bool moves_focus_from_ancestor = true;
+    while (ancestor->parent != lca) {
```
Airblader: If `con` and `parent` were on different outputs (mot just workspaces) (let's put aside whether that makes sense for a second), then the LCA would be the root node. Which means `ancestor->parent == NULL`, no?

src/move.c
```diff
@@ -24,6 +24,49 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
...
+        moves_focus_from_ancestor &= TAILQ_FIRST(&(ancestor->parent->focus_head)) == ancestor;
+        ancestor = ancestor->parent;
+    }
+    Con *target_ancestor = target;
+    while (target_ancestor->parent != lca) {
+        target_ancestor = target_ancestor->parent;
+    }
+
+    /* To determine if con is going to be placed before or after target in the parent's
+     * focus stack, we look at the stack order of the children of the least common
```
Airblader: s/least/lowest?

src/con.c
orestisf1993: Ok. A hypothetical use could be if the caller wants to split the tree somewhere temporarily and then call the function and check for `NULL`. Will make the change.

src/move.c
orestisf1993: `ancestor` and `target_ancestor` should always be children of `lca`, so no. Maybe I should give better names (ideas?) and write a comment as you said in another comment.

src/move.c
```diff
@@ -24,6 +24,52 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
      * afterwards which might then close the con if it is empty. */
     Con *old_parent = con->parent;
 
+    /* We compare the focus order of the children of the least common ancestor. If con or
+     * it's ancestor is before target's ancestor then con should be placed before target
+     * in the focus stack. */
+    Con *lca = con_lowest_common_ancestor(con, parent);
+    Con *ancestor = con;
+    bool moves_focus_from_ancestor = true;
+    while (ancestor->parent != lca) {
```
orestisf1993: Here, if `con == lca` we will make i3 segfault. Should we assume that `con` can't be an ancestor of `target`? If yes, should I add an assertion?

src/move.c
```diff
@@ -24,6 +24,52 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
...
+        moves_focus_from_ancestor &= TAILQ_FIRST(&(ancestor->parent->focus_head)) == ancestor;
+        ancestor = ancestor->parent;
+    }
+    Con *target_ancestor = target;
+    while (target_ancestor->parent != lca) {
```
Airblader: Perhaps extracting this logic into a `con_child_containing` function would improve readability and make this a bit more concise. We could also use it above by defining another function to turn it into something like

```
Con *ancestor = con_child_containing(lca, con);
bool moves_focus_from_ancestor = con_focused_inside(ancestor, con);
```

whereas the latter function checks that `con` is the focus head of `ancestor`, descending down the tree as necessary. This would mean that we need a few more iterations, but I think those cycles are well spent improving the readability of the change (which currently is my main concern here).

src/move.c
```diff
@@ -24,6 +24,52 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
...
+        target_ancestor = target_ancestor->parent;
+    }
+
+    /* To determine if con is going to be placed before or after target in the parent's
+     * focus stack, we look at the stack order of the children of the lowest common
+     * ancestor. */
+    Con *current;
+    bool focus_before;
+    TAILQ_FOREACH(current, &(lca->focus_head), focused) {
```
Airblader: Perhaps this, too, would be worth refactoring into a `TAILQ_FIRST_OF` kind of function / macro used like

    TAILQ_FIRST_OF(&(lca->focus_head), ancestor, target_ancestor, focused)

and returning whichever of the two comes first (which is what this does). I know that @stapelberg generally prefers only extracting something used more than once, but I think the abstraction would help simplifying the code and improving readability. What do you think?

src/move.c
```diff
@@ -24,6 +24,52 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
...
+     * before the old ancestor in the focus stack. Example:
+     * Consider the layout [ H [ V1 [ A B ] V2 [ C ] ] ] where A is focused. We move to
+     * a second workspace and from there we move A to the right and switch back to the
+     * original workspace. Without the change focus would move to B instead of staying
+     * with A. */
+    if (moves_focus_from_ancestor && focus_before) {
+        Con *place = TAILQ_PREV(ancestor, focus_head, focused);
+        TAILQ_REMOVE(&(lca->focus_head), target_ancestor, focused);
+        if (place) {
+            TAILQ_INSERT_AFTER(&(lca->focus_head), place, target_ancestor, focused);
```
Airblader: I think we have this kind of logic more than once in the code already. Perhaps a `TAILQ_INSERT_AFTER_OR_HEAD` would be good to extract. To not blow up the PR we could keep this as an option for a followup PR to explicitly do this kind of refactoring, though. Similarly, we might be able to use a `TAILQ_INSERT_BEFORE_OR_TAIL` in other places.

src/move.c
```diff
@@ -24,6 +24,52 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
      * afterwards which might then close the con if it is empty. */
     Con *old_parent = con->parent;
 
+    /* We compare the focus order of the children of the least common ancestor. If con or
+     * it's ancestor is before target's ancestor then con should be placed before target
```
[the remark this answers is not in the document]
orestisf1993: I think I'll bind "it's" to nagbar or something, I make this mistake 30 times per day.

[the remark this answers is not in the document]
Airblader: Probably not a good idea – it's and its both exist, it just depends on what it stands for. :-)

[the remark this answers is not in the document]
orestisf1993: Just to be aware of what I just typed

src/move.c
orestisf1993: Is `con_focused_inside` a good name? It might look like it is meant to check for `focused` inside `ancestor`.

orestisf1993: > into a `con_child_containing` function

maybe `con_ancestor_with_parent(Con* parent, Con* con)`?

src/move.c
orestisf1993: Is there a good way to have a macro that declares something (`current` for `TAILQ_FOREACH`) and "return" something? I think it should be a function.

src/move.c
Airblader: I'm open to other names, but generally speaking "child" is both a well known graph theoretical term and an established term in our code base. But feel free to propose names as we make the changes and then we can still revisit them.

src/move.c
Airblader: `GREP_FIRST` would be an example.

orestisf1993: IMO an inline function will be cleaner here because the `GREP_FIRST` way requires an extra argument.

Airblader: Fair enough. In that case we could also just make it specific and turn it into something like `con_first_focused_of(Con *parent, Con *a, Con *b)` or something like that.

src/move.c
orestisf1993: > Is `con_focused_inside` a good name? It might look like it is meant to check for `focused` inside `ancestor`.

Especially since we already have `con_inside_focused(Con *con)` which checks for the `focused` container.

src/move.c
orestisf1993: Changed my mind again, a macro can use `field`, generalizing it for other uses.

It should be enough to modify `GREP_FIRST` to have a `field` argument.

orestisf1993: What I have right now is:

```
#define TAILQ_FIRST_OF(head, destelm, elm1, elm2, field) \
    TAILQ_FOREACH(destelm, head, field) {                \
        if (destelm == elm1 || destelm == elm2) {        \
            break;                                       \
        }                                                \
    }
```

Using it like this:
```
Con *current;
TAILQ_FIRST_OF(&(lca->focus_head), current, ancestor, target_ancestor, focused);
bool focus_before = current == ancestor;
```

Break condition could be generalized to `condition`. Otherwise, I think it's possible to have a variable number of `elmx` arguments.

Airblader: We don't need to over-generalize the function for now. Once we need it, we can refactor it some more.

src/move.c
Airblader: Fair enough. Just propose a name you think fits and we can see what it looks like. :-) Names are easy to change.

orestisf1993:
Added the functions.

testcases/t/539-move-focus.t
```diff
@@ -0,0 +1,55 @@
...
+$D = open_window;
+is($x->input_focus, $D->id, 'Window D focused');
+
+$ws2 = fresh_workspace;
+cmd '[id=' . $A->id . '] move right';
+cmd "workspace $ws1";
+sync_with_i3;
+
+is($x->input_focus, $D->id, 'Window D still focused');
+kill_and_confirm_focus($C->id, 'C focused net');
```
Airblader: net?

testcases/t/539-move-focus.t
```diff
@@ -0,0 +1,55 @@
...
+kill_and_confirm_focus($B->id, 'C focused net');
```
Airblader: Should be B here and A below, right?

testcases/t/113-urgent.t
```diff
@@ -333,6 +333,35 @@ for ($type = 1; $type <= 2; $type++) {
     is($target_ws->{urgent}, 1, 'Target workspace is now marked urgent');
 
 ##############################################################################
+# Test that moving an unfocused container doesn't reset it's urgency hint.
+# See issue:
```
Airblader: assert(issue) :-)

testcases/t/113-urgent.t
```diff
@@ -333,6 +333,35 @@ for ($type = 1; $type <= 2; $type++) {
     is($target_ws->{urgent}, 1, 'Target workspace is now marked urgent');
 
 ##############################################################################
+# Test that moving an unfocused container doesn't reset it's urgency hint.
```
Airblader: its :-)

src/con.c
```diff
@@ -817,6 +836,35 @@ int con_num_visible_children(Con *con) {
...
+/*
+ * Returns true if the given container is the most focused descendant of ancestor,
+ * recursively.
+ *
+ */
+bool con_is_most_focused_descendant(Con *con, Con *ancestor) {
+    Con *current = con;
+    bool result = true;
+    while (result && current != ancestor) {
+        result &= TAILQ_FIRST(&(current->parent->focus_head)) == current;
```
Airblader: I think we can rewrite this to

```
while (…) {
    if (!…) {
        return false;
    }

    …
}

return true;
```

This will break out of the loop as soon as we know the result will be false, rather than always going through the entire hierarchy.

src/con.c
```diff
@@ -817,6 +836,35 @@ int con_num_visible_children(Con *con) {
...
+Con *con_child_of_ancestor(Con *ancestor, Con *con) {
+    Con *result = con;
+    while (result && result->parent != ancestor) {
+        result = result->parent;
+    }
+    return result;
+}
+
+/*
+ * Returns true if the given container is the most focused descendant of ancestor,
```
Airblader: It doesn't seem intuitive to me what a »most focused« container is? Can we not just do without the »most«, i.e., `con_is_focused_descendant`?

src/con.c
```diff
@@ -817,6 +836,35 @@ int con_num_visible_children(Con *con) {
 }
 
 /*
+ * Given an ancestor of a container and the container return the ancestor container of
+ * con that is a child of the given ancestor.
+ *
+ */
+Con *con_child_of_ancestor(Con *ancestor, Con *con) {
```
Airblader: I'm still not too happy with this name. What would you think about `con_child_containing`? A child is by definition a direct descendant, so the terminology should be clear.

include/queue.h
```diff
@@ -354,6 +354,13 @@
          (var) != TAILQ_END(head);                        \
          (var) = TAILQ_PREV(var, headname, field))
 
+#define TAILQ_FIRST_OF(destelm, head, elm1, elm2, field) \
```
Airblader: Let's maybe add a small explanation along the lines of

> Returns elm1, elm2 or NULL in destelm when going through the tail queue, depending on which one comes first.

src/con.c
orestisf1993: I do break if result is false but I can use an if if you prefer it.

testcases/t/539-move-focus.t
orestisf1993: I don't know why these ended up here. I corrected them somewhere but probably checked out something else somewhere?

src/con.c
orestisf1993: It's the child of the ancestor of con. Not too happy with the name either.

Can you explain the `con_child_containing` name?

testcases/t/113-urgent.t
orestisf1993: 😭

include/queue.h
orestisf1993: Ok. The rest didn't have any explanation, that's why I didn't add any.

src/con.c
Airblader: Oops :-) But yeah, I do think the other way is easier to read and avoids the result variable, so let's change it.

src/con.c
Airblader: It's the container's child (== direct descendant) which contains the given container.

orestisf1993: Ok I'll use the name then.

Just one quick question if `A->parent->parent->...->parent == B` do we say that B **contains** A?

Airblader: OK, so the question is whether »containing« is defined as »is ancestor of« or »is parent of«. I don't think we have a clear definition of this right now; in some parts we make this clear by appending `_recursively`, which I wouldn't mind doing here if it helps clearing things up.

orestisf1993:
@Airblader I've been wondering if it gets harder to track changes to a bigger PR after lots of reviews. Would you prefer it if I pushed `--fixup=` and `--squash=` type of commits and then rebased before the final merge?

orestisf1993:
Up

src/move.c
orestisf1993: Implemented this, tell me if you want to push it here after all. I also think that a macro
`#define TAILQ_INSERT_POSITION(head, listelm, elm, position, field)`
 that does everything will be handy:

- Compares `position` to `BEFORE` and `AFTER` (`position_t`)
- Uses `TAILQ_INSERT_AFTER_OR_HEAD` or `TAILQ_INSERT_BEFORE_OR_TAIL` accordingly.

orestisf1993: hmm, the code of this PR uses BEFORE-HEAD and AFTER-TAIL.

orestisf1993: I put the changes here, tell me if you want to push them in this PR: https://github.com/orestisf1993/i3/tree/move-con_focus-tmp

Airblader:
LGTM now, but I'd like @stapelberg to at least glance over it as well.

src/move.c
```diff
@@ -24,6 +24,42 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
...
+    Con *con_ancestor = con_child_containing_recursively(lca, con);
+    Con *target_ancestor = con_child_containing_recursively(lca, target);
+    bool moves_focus_from_ancestor = con_is_focused_descendant(con, con_ancestor);
+
+    /* To determine if con is going to be placed before or after target in the parent's
+     * focus stack, we look at the stack order of the children of the lowest common
+     * ancestor. */
+    Con *current;
+    TAILQ_FIRST_OF(current, &(lca->focus_head), con_ancestor, target_ancestor, focused);
+    bool focus_before = current == con_ancestor;
```
stapelberg: nit: `bool focus_before = (current == con_ancestor);` for consistency

src/move.c
```diff
@@ -24,6 +24,42 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
      * afterwards which might then close the con if it is empty. */
     Con *old_parent = con->parent;
 
+    /* We compare the focus order of the children of the least common ancestor. If con or
```
stapelberg: nit: s/least/lowest/

src/con.c
```diff
@@ -817,6 +836,35 @@ int con_num_visible_children(Con *con) {
 }
 
 /*
+ * Given an ancestor of a container and the container return the child of ancestor that
+ * recursively contains con.
```
stapelberg: I find this description hard to follow. How about the following instead?

> Returns the direct child of ancestor that contains con.

src/con.c
```diff
@@ -817,6 +836,35 @@ int con_num_visible_children(Con *con) {
...
+ *
+ */
+Con *con_child_containing_recursively(Con *ancestor, Con *con) {
+    Con *child = con;
+    while (child && child->parent != ancestor) {
+        child = child->parent;
+    }
+    return child;
```
stapelberg: Add `assert(child->parent);` to catch issues early (like in `con_lowest_common_ancestor`).

src/move.c
```diff
@@ -24,6 +24,42 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
...
+
+    Con *con_ancestor = con_child_containing_recursively(lca, con);
+    Con *target_ancestor = con_child_containing_recursively(lca, target);
+    bool moves_focus_from_ancestor = con_is_focused_descendant(con, con_ancestor);
+
+    /* To determine if con is going to be placed before or after target in the parent's
+     * focus stack, we look at the stack order of the children of the lowest common
+     * ancestor. */
+    Con *current;
+    TAILQ_FIRST_OF(current, &(lca->focus_head), con_ancestor, target_ancestor, focused);
```
stapelberg: I think this should be expressed inline to achieve more clarity:
```c
Con *current, *first = NULL;
TAILQ_FOREACH(current, &(lca->focus_head), focused) {
  if (current == con_ancestor || current == target_ancestor) {
    first = current;
    break;
  }
}
```
That way, readers of the code don’t need to look up what `TAILQ_FIRST_OF` does (i.e. this approach reduces mental complexity of an already very complicated function).

include/con.h
```diff
@@ -115,6 +115,12 @@ Con *con_parent_with_orientation(Con *con, orientation_t orientation);
 Con *con_get_fullscreen_con(Con *con, fullscreen_mode_t fullscreen_mode);
 
 /**
+ * Returns the lowest container in the tree that has both a and b as descendants.
+ *
+ */
+Con *con_lowest_common_ancestor(Con *a, Con *b);
```
stapelberg: Regarding all the new functions in con.c, I think it’d make more sense to place them in move.c as `static` functions until we need them elsewhere in the code base.

src/move.c
```diff
@@ -46,12 +82,24 @@ static void insert_con_into(Con *con, Con *target, position_t position) {
 
     con->parent = parent;
 
+    if (parent == lca) {
```
stapelberg: Could you add examples to each of the branches here please? Ideally using unicode drawing characters, like in testcases/t/510-focus-across-output.t

src/move.c
Airblader: That's what we had at first; I actually see it entirely from the opposite side. Inlining just introduces more branches and more complexity. With a well-named function we'd be abstracting that logic and only if the reader needs to know specific details of the function (or macro), they'd be looking up the definition. This is already a really long function, so to me breaking it up into pieces makes a lot of sense.

stapelberg: Sure, breaking up functions into self-contained pieces is a good idea in general. Maybe we just haven’t identified the correct pieces yet.

If we absolutely want to stick to having this bit of code abstracted away, let’s at least scope the macro to move.c, not to the entire code base.

orestisf1993: IMO moving the scope to move.c is not a great idea:

1. The macro is not exclusive to move.c semantically. It could be relevant to any place that uses TAILQs.
Not sure if I phrased this correctly.
2. Breaking `TAILQ_` macros in many files makes it harder to be aware of all of them and could lead to code repetition.

src/move.c
orestisf1993: Ok. One alternative would be to draw tree structures which could make it easier to understand the code for focus comparison between ancestors. A drawing of a workspace is closer to real life though.

btw, are there any tools that can help me with this?

orestisf1993:
:| move.c's `attach_to_workspace` suffers from the same issues.

Airblader:
What's the status on this one? Anything open from your point of view or should we get into the review process to merge this?

orestisf1993:
Yes you can continue the review process. I have left some items open from @stapelberg's review because I am not so sure about them.

As I pointed out `attach_to_workspace` has issues as well but they don't need to be fixed here.

Airblader:
I'm sorry it's been almost two weeks again, I'm just really lacking time right now. Perhaps @stapelberg can review it since it's his comments that are open anyway. :-)

src/move.c
stapelberg: I’m not aware of any i3-specific tools. There are tools which allow you to draw with Unicode drawing characters, but I haven’t tested any of them.

src/move.c
stapelberg: That’s fair. Let’s remove the macro then, until we use it in multiple places.

orestisf1993:
Up. I've added a new branch regarding `focus_before`. I found the problem because I couldn't find a case where `parent == lca` and `focus_before == 1`.

I think I changed everything @stapelberg mentioned.

PR #2977, opened 2017-09-22
Check container existance during drag events

### LINKED ISSUE #1627: Crash when a floating window is tiled while being resized via mouse
Ran into this while testing yesterday and managed to reproduce it on current next.

I'm able to fairly reliably reproduce this by doing the following:
1. Open a floating window
2. Begin resizing the floating window (titlebar right-click seems most reliable)
3. Disable floating on the window

Note that this sometimes crashes X itself for me, so it's probably wise to test on a second X server.

I've [pastebinned a backtrace](http://pastie.org/private/ocp4p0juf0uagfovpqft4q) which seems to imply there's a race going on. 0x1de5ed0 (the floating container) is closed at line 1173 in the log, but then con_inside_floating() for the just-tiled window still returns it later, which leads to the segfault.

http://logs.i3wm.org/logs/5154321532452864.bz2
i3 version 4.10.1-47-g3b81e40 (2015-04-01, branch "next")

orestisf1993:
Fixes #1627

I'd love to write a testcase for this but I don't know how and didn't find anything similar in the current tests. Any pointers?

### DIFF
--- a/include/con.h
+++ b/include/con.h
@@ -166,6 +166,13 @@ Con *con_by_window_id(xcb_window_t window);
 Con *con_by_con_id(long target);
 
 /**
+ * Returns true if the given container (still) exists.
+ * This can be used, e.g., to make sure a container hasn't been closed in the meantime.
+ *
+ */
+bool con_exists(Con *con);
+
+/**
  * Returns the container with the given frame ID or NULL if no such container
  * exists.
  *
--- a/src/floating.c
+++ b/src/floating.c
@@ -531,6 +531,11 @@ void floating_drag_window(Con *con, const xcb_button_press_event_t *event) {
     /* Drag the window */
     drag_result_t drag_result = drag_pointer(con, event, XCB_NONE, BORDER_TOP /* irrelevant */, XCURSOR_CURSOR_MOVE, drag_window_callback, event);
 
+    if (!con_exists(con)) {
+        DLOG("The container has been closed in the meantime.\n");
+        return;
+    }
+
     /* If the user cancelled, undo the changes. */
     if (drag_result == DRAG_REVERT)
         floating_reposition(con, initial_rect);
@@ -642,6 +647,11 @@ void floating_resize_window(Con *con, const bool proportional,
 
     drag_result_t drag_result = drag_pointer(con, event, XCB_NONE, BORDER_TOP /* irrelevant */, cursor, resize_window_callback, &params);
 
+    if (!con_exists(con)) {
+        DLOG("The container has been closed in the meantime.\n");
+        return;
+    }
+
     /* If the user cancels, undo the resize */
     if (drag_result == DRAG_REVERT)
         floating_reposition(con, initial_rect);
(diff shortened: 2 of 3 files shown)


orestisf1993:
Check failed because of `clang-format` version difference. Dropped https://github.com/i3/i3/pull/2977/commits/e1929db9892b1a4cec0568853907f819a1a12cc6.

src/floating.c
```diff
@@ -647,7 +647,7 @@ void floating_resize_window(Con *con, const bool proportional,
         floating_reposition(con, initial_rect);
 
     /* If this is a scratchpad window, don't auto center it from now on. */
-    if (con->scratchpad_state == SCRATCHPAD_FRESH)
+    if (con_exists(con) && con->scratchpad_state == SCRATCHPAD_FRESH)
```
Airblader: Are the checks here and below actually necessary? And if they are, I think it'd better to do a »fail fast« approach:

```
if (!con_exists(con)) {
    DLOG("The container has been closed in the meantime.\n");
    return;
}
```

This would also help reintroducing the bug by referencing `con` somewhere else without remembering to add the check.

src/floating.c
```diff
@@ -647,7 +647,7 @@ void floating_resize_window(Con *con, const bool proportional,
         floating_reposition(con, initial_rect);
```
Airblader: Why can this code path succeed without the check?

include/con.h
```diff
@@ -166,6 +166,12 @@ Con *con_by_window_id(xcb_window_t window);
 Con *con_by_con_id(long target);
 
 /**
+ * Returns true if the given container exists in the all_cons tail queue.
```
Airblader: Instead of describing how it is (indirectly) implemented, I'd prefer a description of what it does. Maybe along the lines of

> Returns true if the given container (still) exists. This can be used, e.g., to make sure the container hasn't been closed in the meantime.

Airblader:
As for the test, I think we can live without one here. It's likely to be a very race condition prone test anyway. Solving that would probably mean jumping through a lot of hoops.

src/floating.c
orestisf1993: The checks are necessary since this is the code that gets executed right after the floating disable command.

> This would also help reintroducing the bug by referencing con somewhere else without remembering to add the check.

How else could this be handled?

orestisf1993:
... This breaks resizing tiling windows. Closing until I resolve this.

orestisf1993:
Ok:

     /* The container that is being dragged or resized, or NULL if this is a
     * drag of the resize handle. */

Btw, I found a similar bug (in `next`) for resizing tiling windows and enabling floating.

src/floating.c
```diff
@@ -739,12 +749,16 @@ static void xcb_drag_check_cb(EV_P_ ev_check *w, int revents) {
     if (last_motion_notify == NULL)
         return;
 
-    dragloop->callback(
-        dragloop->con,
-        &(dragloop->old_rect),
-        last_motion_notify->root_x,
-        last_motion_notify->root_y,
-        dragloop->extra);
+    /* If we are resizing a container we should be certain that it exists. If
```
Airblader: I would like to rephrase this as this sounds a bit confusing (why would we start resizing something that doesn't exist?). We should explicitly state the scenario in which this happens, so something like this:

> Ensure that we are either dragging the resize handle (con is NULL) or the container still exists. The latter might not be true, e.g., if the window closed for any reason while the user is dragging it.

orestisf1993:
Done

Airblader:
Thanks! Waiting for Travis to merge

PR #2978, opened 2017-09-22
Issue 2572

### LINKED ISSUE #2572: Programmatical switch of floating window focus doesn't move active window to the top

orestisf1993:
Fixes #2572.

Based on https://github.com/i3/i3/pull/2616 (@sandsmark).

Differences from the original PR:
- `con_focus` changes apply only to containers with `CT_FLOATING_CON` since `con_focus` is called recursively. This way the `t/143-regress-floating-restart.t` bug isn't reintroduced (https://github.com/i3/i3/pull/2616#discussion_r94002671).
- `handle_focus_in` changes needed for the original issue. Tested with `xdotool windowfocus ...`.
- A small testcase.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -1221,7 +1221,9 @@ static void handle_focus_in(xcb_focus_in_event_t *event) {
         return;
     }
 
-    if (focused_id == event->event) {
+    /* Floating windows should be refocused to ensure that they are on top of
+     * other windows. */
+    if (focused_id == event->event && !con_inside_floating(con)) {
         DLOG("focus matches the currently focused window, not doing anything\n");
         return;
     }
@@ -1232,7 +1234,7 @@ static void handle_focus_in(xcb_focus_in_event_t *event) {
         return;
     }
 
-    DLOG("focus is different, updating decorations\n");
+    DLOG("focus is different / refocusing floating window: updating decorations\n");
 
     /* Get the currently focused workspace to check if the focus change also
      * involves changing workspaces. If so, we need to call workspace_show() to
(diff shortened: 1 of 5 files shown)


orestisf1993:
I think the `handle_focus_in` can be limited to floating containers only. Will check latter if this is preferable.

src/handlers.c
```diff
@@ -1221,19 +1221,12 @@ static void handle_focus_in(xcb_focus_in_event_t *event) {
         return;
     }
 
-    if (focused_id == event->event) {
```
Airblader: We can still ignore it for tiling windows, no?

orestisf1993: Yes, this is what I meant with my last comment. Will refine this latter.

Airblader: Oh, yeah, didn't see the comment. :-)

orestisf1993:
Up.

PR #2980, opened 2017-09-24
Use con_descend_focused for workspaces in _tree_next

### LINKED ISSUE #1160: Inconsistency of the focus-command when switching between containers
[Originally reported by dpoi@…]
(On executing a focus-command, i3 selects the next window of the given direction. If there is a new container, i3 will focus to the ACTIVE window of this container - even if there is an unactive window between the focused window and the active window of the target container.
But if the target container is on another display, it happens, that the focus-command will focus the UNACTIVE window of the target container.

orestisf1993:
Fixes #1160.

This is an alternative to https://github.com/orestisf1993/i3/commit/4a920e47ec6816c053aa577e4e697695a2bf2421 that doesn't modify floating focus behavior.

I'll also write tests once the main approach is decided.

### DIFF
--- a/testcases/t/510-focus-across-outputs.t
+++ b/testcases/t/510-focus-across-outputs.t
@@ -92,7 +92,7 @@ reset_focus $s3_ws;
 
 cmd "workspace $s2_ws";
 cmd 'focus right';
-is($x->input_focus, $sixth->id, 'sixth window focused');
+is($x->input_focus, $seventh->id, 'seventh window focused');
 reset_focus $s3_ws;
 
 cmd "workspace $s2_ws";
@@ -110,6 +110,8 @@ is($x->input_focus, $third->id, 'third window focused');
 cmd 'focus parent';
 cmd 'focus parent';
 cmd 'split v';
+# Focus second or else $first gets to the top of the focus stack.
+cmd '[id=' . $second->id . '] focus';
 reset_focus $s0_ws;
 
 cmd "workspace $s3_ws";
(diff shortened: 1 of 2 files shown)


Airblader:
Although the changed testcase does look odd at first glance, this would be more in line with the behavior for `H[ term1* H[ term2 term3* ] ]` (where the asterisk denotes the focused window within its parent). `focus right` in this case (from term1) will currently focus term3. Hence I think the change in this PR makes sense.

@stapelberg What do you think?

stapelberg:
I agree that this change makes sense, for the same reason that @Airblader mentioned.

Airblader:
@orestisf1993 So the way I understood it you wanted to write more tests?

orestisf1993:
Yes I wanted your opinion on the other approach as well but I'll write the tests tomorrow.

testcases/t/510-focus-across-outputs.t
```diff
@@ -92,7 +92,7 @@ reset_focus $s3_ws;
 
 cmd "workspace $s2_ws";
 cmd 'focus right';
-is($x->input_focus, $sixth->id, 'sixth window focused');
+is($x->input_focus, $seventh->id, 'seventh window focused');
```
orestisf1993: > Although the changed testcase does look odd at first glance

This is needed because of the new behavior.

testcases/t/510-focus-across-outputs.t
```diff
@@ -110,6 +110,8 @@ is($x->input_focus, $third->id, 'third window focused');
 cmd 'focus parent';
 cmd 'focus parent';
 cmd 'split v';
+# Focus second or else $first gets to the top of the focus stack.
```
orestisf1993: > Although the changed testcase does look odd at first glance

I am not 100% sure about this one here. Am I correct here?

orestisf1993:
Added the tests.

testcases/t/510-focus-across-outputs.t
Airblader: This does look a bit weird. But I assume the test failed without this change?

orestisf1993: Yes.

In a new workspace open 3 windows (focus order third > second > first), run `i3-msg floating enable, mark a, focus tiling, focus parent, focus parent, split v, "[con_mark=a]" focus`. Now "first" is going to be focused before "second" after `kill` or `focus tiling`.

PR #2998, opened 2017-09-29
Don't raise floating windows when focused because of focus_follows_mouse

### LINKED ISSUE #2990: Don't move floating windows to the top with focus_follows_mouse
```
Running i3 version: 4.14.1 (2017-09-24) (pid 1411)abort…)
Loaded i3 config: /home/eyenseo/.config/i3/config (Last modified: Mon 25 Sep 2017 15:07:42 CEST, 114 seconds ago)
```

Not needed.

**What I did:**

[Video](https://vimeo.com/235356494)

Had a small floating window over a big floating window, moved the mouse to copy text and then moved the mouse back to the small window.

**What I saw:**
After the mouse was back over the small window I saw the big window as the focus was moved to it and thus it was moved to the top.

**What I expected instead:**
The small window as  *I* did not change the focus. 

This should not be the default behavior - I had similar issues where the mouse mealy touched the big window because I was overshooting the small one, it is inconvenient enough that I use `focus_follows_mouse no` at the moment.

orestisfl:
Fixes #2990.

I am not too happy with the `_con_focus` name, I'd love to hear better suggestions for the change. If the code logic is ok and once we conclude on a name I'll properly reformat con.{c,h} (spacing, comments etc).

### DIFF
--- a/include/con.h
+++ b/include/con.h
@@ -39,6 +39,12 @@ void con_free(Con *con);
 void con_focus(Con *con);
 
 /**
+ * Sets input focus to the given container and raises it to the top.
+ *
+ */
+void con_activate(Con *con);
+
+/**
  * Closes the given container.
  *
  */
--- a/src/click.c
+++ b/src/click.c
@@ -241,7 +241,7 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
          * The splitv container will be focused. */
         Con *focused = con->parent;
         focused = TAILQ_FIRST(&(focused->focus_head));
-        con_focus(focused);
+        con_activate(focused);
         /* To prevent scrolling from going outside the container (see ticket
          * #557), we first check if scrolling is possible at all. */
         bool scroll_prev_possible = (TAILQ_PREV(focused, nodes_head, nodes) != NULL);
@@ -256,7 +256,7 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
     }
 
     /* 2: focus this con. */
-    con_focus(con);
+    con_activate(con);
 
     /* 3: For floating containers, we also want to raise them on click.
      * We will skip handling events on floating cons in fullscreen mode */
--- a/src/commands.c
+++ b/src/commands.c
@@ -1261,7 +1261,7 @@ static void cmd_focus_force_focus(Con *con) {
     if (fullscreen_on_ws && fullscreen_on_ws != con && !con_has_parent(con, fullscreen_on_ws)) {
         con_disable_fullscreen(fullscreen_on_ws);
     }
-    con_focus(con);
+    con_activate(con);
 }
 
 /*
@@ -1379,12 +1379,12 @@ void cmd_focus(I3_CMD) {
          * the target workspace, then revert focus. */
         Con *currently_focused = focused;
         cmd_focus_force_focus(current->con);
-        con_focus(currently_focused);
+        con_activate(currently_focused);
 
         /* Now switch to the workspace, then focus */
         workspace_show(ws);
         LOG("focusing %p / %s\n", current->con, current->con->name);
-        con_focus(current->con);
+        con_activate(current->con);
         count++;
     }
 
@@ -1496,7 +1496,7 @@ void cmd_move_direction(I3_CMD, const char *direction, long move_px) {
 
     /* the move command should not disturb focus */
     if (focused != initially_focused)
-        con_focus(initially_focused);
+        con_activate(initially_focused);
 
     // XXX: default reply for now, make this a better reply
     ysuccess(true);
@@ -1621,7 +1621,7 @@ void cmd_open(I3_CMD) {
     LOG("opening new container\n");
     Con *con = tree_open_con(NULL, NULL);
     con->layout = L_SPLITH;
-    con_focus(con);
+    con_activate(con);
 
     y(map_open);
     ystr("success");
@@ -2010,7 +2010,7 @@ void cmd_rename_workspace(I3_CMD, const char *old_name, const char *new_name) {
     }
 
     /* Restore the previous focus since con_attach messes with the focus. */
-    con_focus(previously_focused);
+    con_activate(previously_focused);
 
     cmd_output->needs_tree_render = true;
     ysuccess(true);
--- a/src/con.c
+++ b/src/con.c
@@ -243,13 +243,27 @@ void con_focus(Con *con) {
         workspace_update_urgent_flag(con_get_workspace(con));
         ipc_send_window_event("urgent", con);
     }
+}
 
-    /* Focusing a container with a floating parent should raise it to the top. Since
-     * con_focus is called recursively for each parent we don't need to use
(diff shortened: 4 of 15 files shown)


stapelberg:
Sorry for letting this linger for so long.

I think what we should do instead of introducing an option to con_focus is to pull the raise logic into a separate function and call it where appropriate — i.e. whenever we programmatically change focus, not as a reaction to a FocusIn event.

orestisf1993:
Shouldn't we keep FocusIn events? This will reintroduce #2572.

stapelberg:
I thought #2572 is about reacting to a menu-triggered raise, e.g. via the Window menu in GIMP. That would trigger a programmatic change, not a FocusIn. Am I misunderstanding that issue?

orestisf1993:
Hmm, I am not sure actually. I just checked by making the `con_focus` call in `handle_focus_in` not raise floating windows. Now, with `xdotool windowfocus ...` floating windows are not raised to the top. Isn't this a case that we want them to be raised to the top?

stapelberg:
No: xdotool windowfocus sends SetInputFocus.

The way to programmatically request focus from the window manager is via a _NET_ACTIVE_WINDOW ClientMessage sent to the root window. xdotool implements that as xdotool windowactivate.

orestisf1993:
Yes you are correct. But I think FocusIn events are not the cause of #2990, MotionNotify and/or EnterNotify are.

stapelberg:
We’d definitely get EnterNotify, but I think FocusIn events are also generated. It could be that we don’t receive them because we temporarily change the event mask when sending SetInputFocus — I’d have to look that up.

orestisf1993:
I think this comment explains it:

https://github.com/i3/i3/blob/4f751610c2a51fb35755db87fafee8598d6d7bd3/src/handlers.c#L1202-L1208

jolange:
What's the status here? I'm very interested in a solution to #2990, because it makes the current i3 version unusable for me.

stapelberg:
> I think what we should do instead of introducing an option to con_focus is to pull the raise logic into a separate function and call it where appropriate — i.e. whenever we programmatically change focus, not as a reaction to a FocusIn event.

That comment is still not addressed AFAICT.

orestisf1993:
> I think what we should do instead of introducing an option to con_focus is to pull the raise logic into a separate function and call it where appropriate — i.e. whenever we programmatically change focus, not as a reaction to a FocusIn event.

Doesn't this mean that we would have to make changes in most parts of the code where `con_focus` is called?

stapelberg:
Yes, we’ll need to identify where the raise function should be called (as opposed to just the focus function).

orestisf1993:
My problem with this approach is that we'll have to review each call of `con_focus` (about 50 of them) and see if it makes sense to add extra code to raise a possible floating container. Except for the 2 or 3 calls that cause #2990 I think that the other `con_focus` calls that also raise are either harmless (exclusively for tiling windows) or reasonable defaults until they cause a problem.

I propose to:
1. Make `_con_focus` a static
2. Keep `con_focus` as is in this PR
3. Add `con_focus_background` (or whatever appropriate name) that just calls `_con_focus(con, false);`
Use that when we explicitly don't want to raise floating containers

orestisf1993:
I pushed my proposal from https://github.com/i3/i3/pull/2998#issuecomment-349166626

stapelberg:
Thanks for your continued work on this PR.

> My problem with this approach is that we'll have to review each call of con_focus (about 50 of them) and see if it makes sense to add extra code to raise a possible floating container. Except for the 2 or 3 calls that cause #2990 I think that the other con_focus calls that also raise are either harmless (exclusively for tiling windows) or reasonable defaults until they cause a problem.

I don’t follow. You already did the review of each call. You identified the ones which should not raise. Why don’t you use string-replace on all con_focus calls to make them con_focus; con_raise;, then remove the raise from the 3 calls where we don’t want to raise? :)

orestisf1993:
My arguments against your approach are:

It increases boilerplate code for almost all calls of `con_focus`.

It can be misleading: there might be some places where we know that we are focusing a tiling container (eg in `con_swap`). Combining this with a raise call could confuse the reader.

Your approach will also need to introduce a new function since `floating_raise_con` already assumes floating containers. This new function will also need to use recursion (`con_inside_floating`) to identify the correct floating container.

stapelberg:
> It increases boilerplate code for almost all calls of con_focus.

I’m willing to live with that if it means avoiding options in con_focus.

> It can be misleading: there might be some places where we know that we are focusing a tiling container (eg in con_swap). Combining this with a raise call could confuse the reader.

Let’s not add raise calls in places where we know we operate on tiling containers, then. Also, let’s add a comment to the raise function documenting that it’s a no-op on tiling containers and safe to remove in that case.

orestisf1993:
> I’m willing to live with that if it means avoiding options in con_focus.

What's so bad with the option in `_con_focus`?

stapelberg:
It makes a very core piece of the code much harder to follow.

Function parameters which change the function behavior are the most prevalent anti-pattern in the i3 code base by far, so I’m really against introducing any more. Past refactorings and bug fixes have shown that these are very tricky to deal with.

A better alternative is to have well-scoped functionality in small functions, which is what I’m advocating for here.

orestisf1993:
I think it's not such a big issue here since the caller doesn't have to figure out the parameters like with `        con_move_to_workspace(con, ws, true, false, false)`. I also don't agree with a solution that introduces more code duplication, another prevalent problem in the i3 code base.

Would you agree to replace `con_focus` with:
```
void con_activate(Con *con){
    con_focus(con);
    con_raise(con); /* the new raise function */
}
```
?

This way we are making "focusing" and "activating" a window separate concepts, like in xdotool. You can hover over a window to "focus" it but you need to click it in order to "activate" it.

We could also move some `workspace_show` and urgency code inside `con_activate`.

stapelberg:
Sounds good, let’s give that a shot.

orestisf1993:
An alternative to prevent the project-wide rename and the inconsistency between the `focus` command and the "activate" concept I talked about would be to:

```
void con_focus(Con *con){
    con_focus_background(con);   /* or whatever name */
    con_raise(con);
}
```

would you prefer that?

stapelberg:
> would you prefer that?

No, I think I prefer the activate concept actually :)

orestisf1993:
I've pushed the rename. Let's check the changes to see if it doesn't make sense in some places.

stapelberg (approved):

stapelberg:
Looks good to me. We can fix calls one-by-one whenever we discover one where the window should not be raised.

Sonique:
So how can I use it in my i3wm configuration? Pretty annoying issue :)

Airblader:
@Sonique There's no need for configuration. If you run a version compiled with this patch it should work.

languitar:
Is there any plan when a new release will be made including this fix? The current behavior drives me crazy...

Airblader:
We're aiming to release over the next few days.

PR #3005, opened 2017-09-30
Implement the tick event

stapelberg:
I included the `xtest_sync_with_i3` implementation in this PR because the tick implementation exposes flakiness more aggressively: previously, flakiness was somewhat mitigated by our timeout (except when the timeout was not sufficient, of course).

The changes to the tests themselves are fairly mechanical.


stapelberg:
The Travis build failed because of 525-i3bar-mouse-bindings.t’s flakiness: synchronizing with i3 is not sufficient, we need to synchronize with i3bar. I’ll try to address this.

Issue #2729, opened 2017-04-06
i3 says "could not create graphical context" with TigerVNC


sphaugh:
Output of `i3 --moreversion 2>&- || i3 --version`:

`Binary i3 version:  4.13 (2016-11-08) © 2009 Michael Stapelberg and contributors`

URL to a logfile as per http://i3wm.org/docs/debugging.html:

http://logs.i3wm.org/logs/5637337312657408.bz2

**What I did:**

Install `tigervnc`, then do
```
cat > ~/.vnc/xstartup <<EOF
#!/bin/sh
exec i3
EOF
chmod +x ~/.vnc/xstartup
vncserver
```

**What I saw:**

```
sean@loefah ~> cat .vnc/loefah:1.log                                                                      

Xvnc TigerVNC 1.7.80 - built Jan  1 1970 00:00:01
Copyright (C) 1999-2016 TigerVNC Team and many others (see README.txt)
See http://www.tigervnc.org for information on TigerVNC.
Underlying X server release 11902000, The X.Org Foundation


Thu Apr  6 16:20:19 2017
 vncext:      VNC extension running!
 vncext:      Listening for VNC connections on all interface(s), port 5901
 vncext:      created VNC server for screen 0
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
04/06/2017 04:20:22 PM - [libi3] ERROR: Could not create graphical context. Error code: 9. Please report this bug.
[...]

PR #3016, opened 2017-10-09
Use default colormap only with default visual

sphaugh:
Related to #2729. Tested with NixOS and it seems good.

I used section 18.2.5 here for reference: https://xcb.freedesktop.org/xlibtoxcbtranslationguide/


sphaugh:
Nevermind, this breaks everything else, hah...

Airblader:
I assume that means we should close this. :-)

PR #3021, opened 2017-10-15
Bugfix: avert endless loop on unexpected EOF at ipc messages

xzfc:
Fix freeze on invalid ipc commands like

    echo -n $'i3-ipc\0\0\0\xa\0\0\0\0focus left' | socat - `i3 --get-socketpath`

Also, treat incomplete headers as IPC violation. Example of incomplete header:

    echo -n i3-ip | socat - `i3 --get-socketpath`

### DIFF
--- a/libi3/ipc_recv_message.c
+++ b/libi3/ipc_recv_message.c
@@ -13,6 +13,7 @@
 #include <stdint.h>
 #include <unistd.h>
 #include <errno.h>
+#include <inttypes.h>
 
 #include <i3/ipc.h>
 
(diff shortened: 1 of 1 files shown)


libi3/ipc_recv_message.c
```diff
@@ -41,7 +41,12 @@ int ipc_recv_message(int sockfd, uint32_t *message_type,
         if (n == -1)
             return -1;
         if (n == 0) {
-            return -2;
+            if (read_bytes == 0) {
+                return -2;
+            } else {
+                ELOG("IPC: unexpected EOF\n");
```
stapelberg: Here and below, let’s make the error messages a bit more helpful. How about:

```c
ELOG("IPC: unexpected EOF after %d bytes, want %d bytes\n", read_bytes, to_read);
```

libi3/ipc_recv_message.c
```diff
@@ -61,13 +66,17 @@ int ipc_recv_message(int sockfd, uint32_t *message_type,
     *reply = smalloc(*reply_length);
 
     read_bytes = 0;
-    int n;
     while (read_bytes < *reply_length) {
-        if ((n = read(sockfd, *reply + read_bytes, *reply_length - read_bytes)) == -1) {
+        int n = read(sockfd, *reply + read_bytes, *reply_length - read_bytes);
```
stapelberg: Make this `const int n` please

xzfc:
Updated.

libi3/ipc_recv_message.c
xzfc: I've rephrased slightly to distinguish EOF in payload/header:

```
IPC: unexpected EOF while reading payload, got 10 bytes, want 167772160 bytes
IPC: unexpected EOF while reading header, got 5 bytes, want 14 bytes
IPC: invalid magic in header, got "i3-ppc", want "i3-ipc"
```

stapelberg:
Neat, thanks!

PR #3022, opened 2017-10-17
Fix i3bar leaks

orestisf1993:


i3bar/src/xcb.c
```diff
@@ -1518,7 +1518,7 @@ void init_tray_colors(void) {
  */
 void clean_xcb(void) {
     i3_output *o_walk;
-    free_workspaces();
+    free_outputs();
     SLIST_FOREACH(o_walk, outputs, slist) {
```
Airblader: Isn't this dead code now?

i3bar/src/ipc.c
```diff
@@ -68,7 +68,6 @@ void got_output_reply(char *reply) {
     SLIST_FOREACH(o_walk, outputs, slist) {
         destroy_window(o_walk);
     }
-    FREE_SLIST(outputs, i3_output);
```
Airblader: What happens to an output which has been removed in the reply? Won't it just stick around in `outputs` now?

orestisf1993: `parse_outputs_json()` calls `free_outputs()`

i3bar/src/xcb.c
orestisf1993: yes

orestisf1993:
up

Airblader:
@orestisf1993 Can you take a look at the failing test?

orestisf1993:
Should be ok now.

Airblader:
Thanks!

PR #3023, opened 2017-10-17
Add support to resize floating container in percentage

### LINKED ISSUE #2816: Add support to resize floating container in percentage
Use case
--------

``` i3
for_window [floating] resize set 30% 90%, move position center
```

orestisf1993:
`resize set` is modified to accept both 'px' and 'ppt' height and width.

Fixes #2816.


Airblader:
LGTM, so I'll merge it now. Thanks!

PR #3027, opened 2017-10-22
_NET_ACTIVE_WINDOW: invalidate focus to force SetInputFocus call

stapelberg:
The sender of the _NET_ACTIVE_WINDOW client message might know better when to
set focus than i3, as i3 does not know about unmanaged (override_redirect=1)
windows.

related to https://github.com/i3/i3lock/issues/35

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -774,6 +774,8 @@ static void handle_client_message(xcb_client_message_event_t *event) {
                 scratchpad_show(con);
             } else {
                 workspace_show(ws);
+                /* Re-set focus, even if unchanged from i3’s perspective. */
+                focused_id = XCB_NONE;
                 con_focus(con);
             }
         } else {


Airblader:
I don't understand this change. This path is never reached for unmanaged windows?

stapelberg:
Correct, and it doesn’t need to be. In i3lock, we’re using the _NET_ACTIVE_WINDOW ClientMessage on the managed window which had focus before i3lock took it. Without this change, though, i3 won’t send a SetInputFocus request, because it didn’t notice the focus changes by the unmanaged window.

Airblader:
IMHO we are supporting broken client behavior with this change, just that in this case the broken client is i3lock. The reason we need this change is because the client caused us to be out of sync. Wouldn't a correct fix to avoid running out of sync, e.g., by processing FocusOut events? Openbox seems to do this, but unfortunately it is rather involved since one has to coordinate with FocusIn events.

But I think I've already lost the argument that we are turning i3lock into a broken client, so let's move on. At the very least the one positive thing about this change is that it's a safety net for when the request originates not from the broken client, but somewhere else.

stapelberg:
IIUC, maintaining sync would result in 1 fewer force-focus at the expense of more complicated code paths. If it turns out that this 1 force-focus is a problem, we can always switch to the maintain-sync approach :).

Issue #2608, opened 2016-12-15
Improving resize by adding percentage for fixed tiles


cyphunk:
It would be great if i3 config had an option to set the default width percentage for a newly opened window. For example, with one terminal open in tiled mode, when a browser window is opened it is just slightly too small and I always wind up resizing manually. (The same is true for editors and email clients) It would be great to have a way to specifiy a percentage for width of a certain windowed+tiled application. This is similar to issue https://github.com/i3/i3/issues/1219 except that here we are discussing tiled windows not floating windows.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
What’s the configuration surface you have in mind?

How does this feature position itself compared to saved layouts? I.e., if you are frequently creating a layout where there’s a browser and a terminal in a certain size, why not just use saved layouts?

Nymerea:
I'm not sure we spoke about the same feature, but I also would like to have a shortcut for resizing my window.

I mean, I often use a browser at my first half screen and open somes terminales at second halfs.

For browsing (with firefox), 50% of screen is too tiny, having a way to set windows take like 70% of screen would be great.

FichteFoll:
I would like to open my steam friendlist with a width of 350px (or 14% for my monitor) next to the main steam browser, because it would never contain any more data than what could be displayed in these 350px. I tried fiddling around with the various resize commands until I eventually went straight to the source code to find out that this isn't supported currently.

I would like to be able to do something like this:
```
for_window [class="Steam" title="Friends"] resize set width 350 or 14, move right
```

Airblader:
@FichteFoll The friend list is something that's open all the time anyway, no? So for this case you could just use layout saving & restoring.

FichteFoll:
@Airblader yes and no. I'll look into layout saving & restoring now that I found the docs for it. I think it will help me regardless.

martonbalazs:
Just found this and was thinking something similar. Browser is a good example: I sometimes open one or two terminals next to it but always prefer smaller space for those than for the browser and thus need to resize.

[...]

PR #3036, opened 2017-11-05
Implement 'resize set width|height <ppt> ppt' command

livanh:
This patch allows resizing tiling containers to exactly the given size (in ppt). The sintax is as in the title. The patch includes only the code now, I will update documentation if needed before merging.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2320,8 +2320,11 @@ space from all the other containers. The optional pixel argument specifies by
 how many pixels a *floating container* should be grown or shrunk (the default
 is 10 pixels). The ppt argument means percentage points and specifies by how
 many percentage points a *tiling container* should be grown or shrunk (the
-default is 10 percentage points). Note that +resize set+ will only work for
-floating containers.
+default is 10 percentage points).
+
+Notes about +resize set+: a value of 0 for <width> or <height> means "do
+not resize in this direction", and resizing a tiling container by +px+ is not
+implemented.
 
 It is recommended to define bindings for resizing in a dedicated binding mode.
 See <<binding_modes>> and the example in the i3
--- a/include/commands.h
+++ b/include/commands.h
@@ -63,7 +63,7 @@ void cmd_move_con_to_workspace_name(I3_CMD, const char *name, const char *no_aut
 void cmd_move_con_to_workspace_number(I3_CMD, const char *which, const char *no_auto_back_and_forth);
 
 /**
- * Implementation of 'resize set <px> [px] <px> [px]'.
+ * Implementation of 'resize set <width> [px | ppt] <height> [px | ppt]'.
  *
  */
 void cmd_resize_set(I3_CMD, long cwidth, const char *mode_width, long cheight, const char *mode_height);
--- a/src/commands.c
+++ b/src/commands.c
@@ -659,7 +659,7 @@ void cmd_resize(I3_CMD, const char *way, const char *direction, long resize_px, 
  */
 void cmd_resize_set(I3_CMD, long cwidth, const char *mode_width, long cheight, const char *mode_height) {
     DLOG("resizing to %ld %s x %ld %s\n", cwidth, mode_width, cheight, mode_height);
-    if (cwidth <= 0 || cheight <= 0) {
+    if (cwidth < 0 || cheight < 0) {
         ELOG("Resize failed: dimensions cannot be negative (was %ld %s x %ld %s)\n", cwidth, mode_width, cheight, mode_height);
         return;
     }
@@ -667,25 +667,84 @@ void cmd_resize_set(I3_CMD, long cwidth, const char *mode_width, long cheight, c
     HANDLE_EMPTY_MATCH;
 
     owindow *current;
+    bool success = true;
     TAILQ_FOREACH(current, &owindows, owindows) {
         Con *floating_con;
         if ((floating_con = con_inside_floating(current->con))) {
             Con *output = con_get_output(floating_con);
-            if (mode_width && strcmp(mode_width, "ppt") == 0) {
+            if (cwidth == 0) {
+                cwidth = output->rect.width;
+            } else if (mode_width && strcmp(mode_width, "ppt") == 0) {
                 cwidth = output->rect.width * ((double)cwidth / 100.0);
             }
-            if (mode_height && strcmp(mode_height, "ppt") == 0) {
+            if (cheight == 0) {
+                cheight = output->rect.height;
+            } else if (mode_height && strcmp(mode_height, "ppt") == 0) {
                 cheight = output->rect.height * ((double)cheight / 100.0);
             }
             floating_resize(floating_con, cwidth, cheight);
         } else {
-            ELOG("Resize failed: %p not a floating container\n", current->con);
+            if (current->con->window && current->con->window->dock) {
+                DLOG("This is a dock window. Not resizing (con = %p)\n)", current->con);
+                continue;
+            }
+
+            if (cwidth > 0 && mode_width && strcmp(mode_width, "ppt") == 0) {
+                /* get the appropriate current container (skip stacked/tabbed cons) */
+                Con *target = current->con;
+                Con *dummy;
+                resize_find_tiling_participants(&target, &dummy, D_LEFT, true);
+
+                /* Calculate new size for the target container */
+                double current_percent = target->percent;
+                char *action_string;
+                long adjustment;
+
+                if (current_percent > cwidth) {
+                    action_string = "shrink";
+                    adjustment = (int)(current_percent * 100) - cwidth;
+                } else {
+                    action_string = "grow";
+                    adjustment = cwidth - (int)(current_percent * 100);
+                }
+
+                /* perform resizing and report failure if not possible */
+                if (!cmd_resize_tiling_width_height(current_match, cmd_output,
+                                                    target, action_string, "width", adjustment)) {
+                    success = false;
+                }
+            }
+
+            if (cheight > 0 && mode_width && strcmp(mode_width, "ppt") == 0) {
+                /* get the appropriate current container (skip stacked/tabbed cons) */
+                Con *target = current->con;
+                Con *dummy;
+                resize_find_tiling_participants(&target, &dummy, D_DOWN, true);
+
+                /* Calculate new size for the target container */
+                double current_percent = target->percent;
+                char *action_string;
+                long adjustment;
+
+                if (current_percent > cheight) {
+                    action_string = "shrink";
+                    adjustment = (int)(current_percent * 100) - cheight;
+                } else {
+                    action_string = "grow";
+                    adjustment = cheight - (int)(current_percent * 100);
+                }
+
+                /* perform resizing and report failure if not possible */
+                if (!cmd_resize_tiling_width_height(current_match, cmd_output,
+                                                    target, action_string, "height", adjustment)) {
+                    success = false;
+                }
+            }
         }
     }
 
     cmd_output->needs_tree_render = true;
-    // XXX: default reply for now, make this a better reply
-    ysuccess(true);
+    ysuccess(success);
 }
 
 /*
--- a/testcases/t/541-resize-set-tiling.t
+++ b/testcases/t/541-resize-set-tiling.t
@@ -1,0 +1,147 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • https://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • https://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • https://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Tests resizing tiling containers
+use i3test;
+
+############################################################
+# resize horizontally
+############################################################
+
+my $tmp = fresh_workspace;
+
+cmd 'split h';
+
+my $left = open_window;
+my $right = open_window;
+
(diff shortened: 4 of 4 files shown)


src/commands.c
```diff
@@ -633,6 +633,47 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
 }
 
 /*
+ * Implementation of 'resize set width|height <ppt> ppt'.
+ *
+ */
+void cmd_resize_tiling_set(I3_CMD, const char *direction, long ppt) {
+    DLOG("set size to %ld ppt in %s direction\n", ppt, direction);
+    if (ppt<5){
```
Airblader: Why this arbitrary restriction?

src/commands.c
```diff
@@ -633,6 +633,47 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+        ELOG("Resize failed: requested size is too small (%ld)\n", ppt);
```
Airblader: This (and all other) log messages containing a value in ppt should include the `%` sign to provide a better understanding.

src/commands.c
```diff
@@ -633,6 +633,47 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+            continue;
+        }
+        if (con_inside_floating(current->con)) {
+            DLOG("This is a floating window. Not resizing (con = %p)\n)", current->con);
+            continue;
+        }
+
+        double current_percent = current->con->percent;
+        if (current_percent > ppt) {
+             if (!cmd_resize_tiling_width_height(current_match, cmd_output,
```
Airblader: Since we're doing the same call in both cases, it'd IMHO be better to use the if/else to define the arguments and only have one path actually execute the call.

src/commands.c
```diff
@@ -633,6 +633,47 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+                                                    current->con, "shrink", direction, (int)(current_percent*100)-ppt))
+                    return;
```
Airblader: Shouldn't this be logged? Also, in this case we are not returning any response on the IPC interface, which is bad.

Airblader (commented):
I only have a few quick comments, but this should not be taken as a merge decision.

We do have https://github.com/i3/i3/issues/2608 for that discussion which hasn't really reached any decision yet.

src/commands.c
Airblader: In fact, I don't really like that we are calling the handler of another command here, and especially like this. That feels quite hacky. This logic should be handled correctly in `con_*` functions, but will require some refactoring, I think.

src/commands.c
livanh: I saw that cmd_resize_tiling_width_height() returns without resizing if any of the involved windows could become smaller than 5%. I assumed that means "we don't want any window to be smaller than 5%", but feel free to suggest a different value. At the very least, we should restrict ppt to be non-negative.

src/commands.c
livanh: Ok, I'll do so

src/commands.c
livanh: No problem, I'll do it this way.

src/commands.c
livanh: This is how it's done in cmd_resize(). In fact, cmd_resize_tiling_width_height() prints some logging messages on its own, and using this command with i3-msg returns [{"success":true}] (or false). Is this the right way to check the response?

livanh:
I did't notice the discussion in #2608. If I understand it correctly, this would be a partial solution to that issue. At this time I'm not interested in doing the remaining part (i.e. associating some weight to each window), but it is an interesting idea nonetheless.

Airblader:
For the record, I don't think we'd implement something like the proposed weights idea. If at all, I think we'd stick with something like this.

livanh:
I just tried and it works also with `for_window` matching, as asked in #2608. Let me know if you decide to merge and what changes are needed.

pevu:
Please merge, I'd like to have this as well. Right now I have to write hacky wrapper scripts to get the same functionality, as I've seen other people do too.

src/commands.c
```diff
@@ -632,6 +632,47 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
     return true;
 }
 
+/*
+ * Implementation of 'resize set width|height <ppt> ppt'.
+ *
+ */
+void cmd_resize_tiling_set(I3_CMD, const char *direction, long ppt) {
+    DLOG("set size to %ld ppt in %s direction\n", ppt, direction);
+    if (ppt < 5) {
```
Airblader: Let's please just remove this check here.

src/commands.c
```diff
@@ -632,6 +632,47 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+            continue;
+        }
+        if (con_inside_floating(current->con)) {
+            DLOG("This is a floating window. Not resizing (con = %p)\n)", current->con);
+            continue;
+        }
+
+        double current_percent = current->con->percent;
+        if (current_percent > ppt) {
+            if (!cmd_resize_tiling_width_height(current_match, cmd_output,
```
Airblader: Can we just slightly change this logic to instead execute the resize for all matched containers regardless if it failed for one of them? We just collect the accumulated result of all calls such that in the end we report a success if and only if all were successful and a failure if a single one failed, but that way the resize still worked for all matches where it could work.

src/commands.c
```diff
@@ -632,6 +632,47 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+                return;
+        } else {
+            if (!cmd_resize_tiling_width_height(current_match, cmd_output,
+                                                current->con, "grow", direction, ppt - (int)(current_percent * 100)))
+                return;
+        }
+    }
+
+    cmd_output->needs_tree_render = true;
+    // XXX: default reply for now, make this a better reply
```
Airblader: Please remove this comment (I know it's in other places, but we don't need it here)

Airblader:
I'm willing to merge this feature, however there's a few things:

1. I'd prefer the syntax `resize set 35 ppt 50 ppt` which is in line with the current `px` symbol (which is optional). Otherwise we have two different syntax variations for doing the same thing.
2. The userguide has to be updates (`docs/userguide`).
3. This is easily testable, so we should have some tests for this.
4. I've added a few review comments which should be addressed (as much as they make sense after the other points above).

Airblader (dismissed):

pevu:
But what if you only wanted to adjust the height of the container? I thought this is for tiling containers, where usually either the height or the width is changed, not both at the same time.

Airblader:
Point taken. I still feel it'd be good to have the option to change both in an atomic operation. Perhaps we could use a placeholder value representing a nop. @stapelberg What do you think?

livanh:
My intended use is the same as @pevu, i.e. only resize in one direction. I can change the code to use the syntax you proposed (with a nop placeholder), but I'd like to clarify two points first:

1.  What happens if the unit (px/ppt) is omitted? Should we assume px for backwards compatibility, or turn it into an error?
2. Is it ok to call `cmd_resize_tiling_width_height` internally, or is it better to make a dedicated function?

Airblader:
> with a nop placeholder

Please note that we should first discuss what that placeholder looks like and if that's really what we want to go for. Hence me tagging @stapelberg to discuss it. :-)

> What happens if the unit (px/ppt) is omitted? Should we assume px for backwards compatibility, or turn it into an error?

We don't like breaking things, so we would be falling back to px here.

> Is it ok to call cmd_resize_tiling_width_height internally, or is it better to make a dedicated function?

I actually think calling it is OK.

stapelberg:
I agree with @Airblader: setting width/height in the same command seems reasonable to me.

livanh:
I think I did everything, except for the testcase. That would require some study and experimentation, which I do not have time to do today.

 As for the placeholder, I simply assumed a value of zero or less to mean "nop" for the corresponding direction, as they would not make sense anyway.

However, I just noticed that there is a conflict with #3023 (already merged), where "set <ppt> ppt <ppt> ppt" means to resize a floating container in percentage points. A possible way out would be to move my logic into `cmd_resize_set`, and follow the path for floating or tiling containers depending on the currently focused container. Does it make sense?

docs/userguide
```diff
@@ -2311,8 +2312,11 @@ space from all the other containers. The optional pixel argument specifies by
 how many pixels a *floating container* should be grown or shrunk (the default
 is 10 pixels). The ppt argument means percentage points and specifies by how
 many percentage points a *tiling container* should be grown or shrunk (the
-default is 10 percentage points). Note that +resize set+ will only work for
-floating containers.
+default is 10 percentage points).
+
+Note that +resize set+ takes the first form for floating containers and the
+second one for tiling containers. In the latest case, a negative or zero value
```
stapelberg: s/latest/latter/

include/commands.h
```diff
@@ -74,6 +74,12 @@ void cmd_resize_set(I3_CMD, long cwidth, long cheight);
  */
 void cmd_resize(I3_CMD, const char *way, const char *direction, long resize_px, long resize_ppt);
 
+/*
```
stapelberg: nit: start function comments in header files with `/**`

src/commands.c
```diff
@@ -632,6 +632,99 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+                   target->parent->layout == L_TABBED)
+                target = target->parent;
+
+            /* Then further go up until we find one with the matching orientation. */
+            while (target->type != CT_WORKSPACE &&
+                   target->type != CT_FLOATING_CON &&
+                   (con_orientation(target->parent) != HORIZ || con_num_children(target->parent) == 1))
+                target = target->parent;
+
+            /* Calculate new size for th target container */
```
stapelberg: nit: s/th/the/

src/commands.c
```diff
@@ -632,6 +632,99 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+ *
+ */
+void cmd_resize_set_ppt(I3_CMD, long cwidth, long cheight) {
+    DLOG("set size to %ld%% x %ld%% \n", cwidth, cheight);
+
+    HANDLE_EMPTY_MATCH;
+
+    owindow *current;
+    bool success = true;
+    char *action_grow_string = "grow";
```
stapelberg: Why declare these as a variable? You can just use "grow" below.

src/commands.c
```diff
@@ -632,6 +632,99 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
+            DLOG("This is a dock window. Not resizing (con = %p)\n)", current->con);
+            continue;
+        }
+        if (con_inside_floating(current->con)) {
+            DLOG("This is a floating window. Not resizing (con = %p)\n)", current->con);
+            continue;
+        }
+
+        if (cwidth > 0) {
+            Con *target = current->con;
```
stapelberg: Can `resize_find_tiling_participants` not be used here? It looks very similar to this block of code.

livanh:
I rebased over the latest `next` branch, moved the code into `cmd_resize_set`, removed the old parts, and added a test case. I can also squash the commits for a cleaner history, if you want.

For the comment on `resize_find_tiling_participants`, I tried using it, but it's not exactly the same. Sometimes the command works as expected, and sometimes it does nothing.
By looking at the code, the main difference seems to be in the `con_num_children(target->parent) == 1` condition. I actually copied this code from somewhere else in the file. If we can decide that one variant is definitely better than the other, I can handle this as a separate refactoring task.

src/commands.c
```diff
@@ -681,33 +681,104 @@ void cmd_resize(I3_CMD, const char *way, const char *direction, long resize_px,
...
     HANDLE_EMPTY_MATCH;
 
     owindow *current;
+    bool success = true;
     TAILQ_FOREACH(current, &owindows, owindows) {
         Con *floating_con;
         if ((floating_con = con_inside_floating(current->con))) {
             Con *output = con_get_output(floating_con);
-            if (mode_width && strcmp(mode_width, "ppt") == 0) {
+            if (cwidth == 0)
```
Airblader: Please use `{}` braces everywhere.

src/commands.c
```diff
@@ -681,33 +681,104 @@ void cmd_resize(I3_CMD, const char *way, const char *direction, long resize_px,
...
-            ELOG("Resize failed: %p not a floating container\n", current->con);
+            if (current->con->window && current->con->window->dock) {
+                DLOG("This is a dock window. Not resizing (con = %p)\n)", current->con);
+                continue;
+            }
+
+            if (cwidth > 0 && mode_width && strcmp(mode_width, "ppt") == 0) {
+                Con *target = current->con;
+                /* get the appropriate current container (skip stacked/tabbed cons) */
+                while (target->parent->layout == L_STACKED ||
```
Airblader: Please use `{}` braces for loops as well. Also, don't these two conditions fit on one line?

stapelberg:
> For the comment on resize_find_tiling_participants, I tried using it, but it's not exactly the same. Sometimes the command works as expected, and sometimes it does nothing.
> By looking at the code, the main difference seems to be in the con_num_children(target->parent) == 1 condition. I actually copied this code from somewhere else in the file.

So there is another duplicate of this functionality? Can you point me to it please?

> If we can decide that one variant is definitely better than the other, I can handle this as a separate refactoring task.

I’m a bit worried about pushing it to a later point in time, because then we might easily forget about it :).

Naively, I would argue that a variant which works in all the cases is better than the others. Could you please look into sending a PR which reduces duplication, so that we can make use of the new function in this PR?

Thanks!

livanh:
Please see #3111

stapelberg:
Could you clean up the git history of this PR please? See git’s “interactive rebase” feature. In the end, we’d like to have one git commit per logical change. Changes based on the code review don’t need to be reflected in the history. Thanks!

livanh:
I made everything into one commit, since we don't really need the initial syntax anymore. Is that ok?

stapelberg (approved):

stapelberg:
Yep. Thanks!

PR #3039, opened 2017-11-09
Update README with build info

eplanet:
I think it's better to provide a little more essential information for people that fork/clone the repository so they can concentrate on contributing rather than installing the requirements.

I just tested that from scratch in an empty Debian Stretch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,28 @@
 [![Issue Stats](http://www.issuestats.com/github/i3/i3/badge/issue?style=flat)](http://www.issuestats.com/github/i3/i3)
 [![Pull Request Stats](http://www.issuestats.com/github/i3/i3/badge/pr?style=flat)](http://www.issuestats.com/github/i3/i3)
 
+## Description
+
 i3 is a tiling window manager for X11.
 
 For more information about i3, please see [the project's website](https://i3wm.org/) and [online documentation](https://i3wm.org/docs/).
 
 For information about contributing to i3, please see [CONTRIBUTING.md](.github/CONTRIBUTING.md).
+
+## Development
+
+On debian-based systems, the following line will install all requirements:
+```bash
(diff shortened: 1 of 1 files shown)


Airblader:
If this is aimed at contributors, isn't it a sensible assumption that they generally installed i3 as a package at some point already? Then they already have all the dependencies.

We recently discussed the Readme for another PR. The basic problem is that we have a single source of truth, the docs folder, and would like to avoid duplicating things. That's why the Readme now links to the docs place on der website, which does already explain how to compile. Why is that not sufficient?

eplanet:
Installing i3 doesn't install the required build dependencies. There's no obvious listing of the build requirements neither in the [hacking-howto](https://i3wm.org/docs/hacking-howto.html) nor the Build-Depends of the package (that doesn't exist). That's why I thought adding these informations would be worth, especially in the README since a good majority of OS projects on github do that (including `i3status` for that matter).

The website is great and I understand the interest of pointing to it, so maybe I could port that PR to [hacking-howto](https://github.com/i3/i3.github.io/blob/master/_docs/hacking-howto) directly?

Airblader:
I've given this some more thought; in general, I agree that the README is the preferable place these days. My issue is mostly with the duplication. It's a shame the documents in the docs folder cannot be rendered as markdown by Github as this would probably simplify this situation quite a bit.

But assuming we won't be migrating everything from the website into the README, perhaps moving it to the hacking-howto is the more consistent choice here. Listing build dependencies is probably a good idea. However, I'd like @stapelberg to weigh in on this.

stapelberg:
The hacking-howto already contains the build instructions, we should link to it.

The preferred command to install i3 build-dependencies is `apt build-dep i3-wm` (ideally when using our repositories, so that recently introduced build-deps are covered).

I don’t think that belongs into the README, though. People should just install i3 from their package manager in the general case, or from our repos if they need a newer version. For the development use-case, let’s just refer to the hacking-howto and/or the DEPENDS file, which already lists all build-deps.

Airblader:
@stapelberg So based on your answer I guess we'll close the PR (doing so now)

PR #3041, opened 2017-11-12
docs/userguide: Correct mark/goto i3-input commands

d-e-s-o:
The userguide still mentions an old 'goto' command which no longer
exists and will be ignored silently (when used in the i3 config) or
causes an error to be reported (when invoked from the command line).
This change updates the userguide to correct this problem. In addition
to that it also updates the i3-input command shown to no longer use the
deprecated -p flag but -F instead.


Airblader:
Thanks!

PR #3051, opened 2017-11-18
i3bar: make better use of compile time string length evaluation

d-e-s-o:
The i3bar program contains a couple of locations where we use strlen on
a string that clearly is known at compile time and contained in the
binary. There is no point in doing so when the same information can be
retrieved statically by means of the sizeof operator.
This change fixes those occurrences. It also replaces a couple of magic
lengths with their symbolic counterparts and makes use of the return
value provided by the *printf* family of functions to avoid duplicated
string length assessment.

### DIFF
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -218,13 +218,19 @@ static int stdin_string(void *context, const unsigned char *val, size_t len) {
         return 1;
     }
     if (strcasecmp(ctx->last_map_key, "markup") == 0) {
-        ctx->block.pango_markup = (len == strlen("pango") && !strncasecmp((const char *)val, "pango", strlen("pango")));
+        const char pango[] = "pango";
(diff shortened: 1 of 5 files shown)


Airblader:
AFAIK, any respectable compiler will replace the strlen on a constant string at compile time already, making this (part of the) change unnecessary; since it just adds boilerplate we don't need, I'm not inclined to merge that.

Replacing the magic numbers does sound like a good idea, but we should use strlen there as well.

d-e-s-o:
Fair enough, I sort of expected such a response ;) Use of raw string literals without any syntactic connection in different places is not much better than that of magic numbers in my opinion, but up to you.

Airblader:
> Use of raw string literals without any syntactic connection in different places is not much better than that of magic numbers in my opinion

It is better in the sense that it makes it more clear what the number means and is likely to be more noticeable if, say, you are changing the value of the string you are looking for.

PR #3052, opened 2017-11-18
i3bar: replace magic numbers with more meaningful constructs

d-e-s-o:
In i3bar/src/config.c we compare string lengths agains magic numbers.
This change replaces those numbers with the lengths of the strings they
represent.


d-e-s-o:
Hey Ingo, would you mind reviewing this change? @Airblader

PR #3058, opened 2017-11-24
do not check for NULL in FREE macro

d-e-s-o:
free(3) is safe to invoke on a NULL pointer, in which case no action is
taken. This change adjusts the FREE macros to omit this unnecessary
check.


d-e-s-o:
Anybody? Trivial and non-invasive.

PR #3065, opened 2017-12-01
Skip internal workspaces with 'move workspace to output'

### LINKED ISSUE #3064: "[class=".*"] move workspace to output" fails when a container exists in the scratchpad
```
Binary i3 version:  4.14-100-g8144a0f4 (2017-09-19, branch "4.14-100-g8144a0f4") © 2009 Michael Stapelberg and contributors
```

https://logs.i3wm.org/logs/5758531089203200.bz2

**What I did:**

1. Open 2 windows
2. Move one of them to the scratchpad: `i3-msg move to scratchpad `
3. Run `i3-msg '[class=".*"]' move workspace to output ...`

**What I saw:**

Step 3 fails:
```
12/01/2017 02:34:38 PM - Trying to move workspace 0x614000002a40 / __i3_scratch to output "default".
12/01/2017 02:34:38 PM - ERROR: Could not get output from name "__i3".
12/01/2017 02:34:38 PM - ERROR: Cannot get current output. This is a bug in i3.
12/01/2017 02:34:38 PM - ERROR: Failed to move workspace to output.
```

**What I expected instead:**
The scratchpad workspace is ignored, all others are moved according to the command.

orestisf1993:
Fixes #3064.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1121,6 +1121,10 @@ void cmd_move_workspace_to_output(I3_CMD, const char *name) {
     owindow *current;
     TAILQ_FOREACH(current, &owindows, owindows) {
         Con *ws = con_get_workspace(current->con);
+        if (con_is_internal(ws)) {
+            continue;
+        }
+
         bool success = workspace_move_to_output(ws, name);
         if (!success) {
             ELOG("Failed to move workspace to output.\n");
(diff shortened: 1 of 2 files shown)


src/commands.c
```diff
@@ -1121,6 +1121,10 @@ void cmd_move_workspace_to_output(I3_CMD, const char *name) {
     owindow *current;
     TAILQ_FOREACH(current, &owindows, owindows) {
         Con *ws = con_get_workspace(current->con);
+        if (con_is_internal(ws)) {
```
Airblader: I think it'd be better to handle this in `workspace_move_to_output` or is there any reason why that wouldn't work?

orestisf1993: `workspace_move_to_output` already returns `false`. The problem is that `cmd_move_workspace_to_output` quits on the first unsuccessful call to it.

Airblader: Ah, I see. Alright then.

PR #3080, opened 2017-12-10
 Reduce repetition in cmd_move_con_to_workspace* 

orestisf1993:


Airblader:
I get the reasoning for the change, but I'm not actually sure I favor it. It somewhat outsources the handling of matches in a less transparent way such that I'm not sure the couple of (repeated) lines saved is worth the indirection of the extra call here.

orestisf1993:
IMO It's not only about the saved lines, it is made easily clear for the reader (without having to compare each function) that in the end, all `cmd_move_con_to_workspace`* do the same thing after they've acquired the target workspace.

Airblader:
I think my argument at the moment would be that as a reader, I would see

    cmd_move_con_to_workspace(ws);

and wonder: *which con?* just to be even further surprised that it's (potentially) more than one container. Admittedly, giving it a different name might resolve that issue.

orestisf1993:
Yeah the name is bad in retrospect. I'll find a better name.

orestisf1993:
- `move_matches_to_workspace`
- `move_matching_windows_to_workspace`
- `move_current_matches_to_workspace`

?

Airblader:
The first one sounds reasonable to me.

orestisf1993:
Up

Airblader:
Thanks!

PR #3078, opened 2017-12-10
Fix sticky-related crash (#3075)

orestisf1993:
The first commit is unrelated, I can remove it. For an explanation of problem & solution check the second commit message.

Fixes #3075.

PR #3081, opened 2017-12-11
Introduce ipc_send_workspace_event_from_gen

orestisf1993:
(From #3078)


Airblader:
I'm actually going to reject this PR. I get the sentiment to avoid repetition, but the introduction of this function feels forced and not very natural. I think we introduce more indirection than we gain from this.

PR #3087, opened 2017-12-13
 Fix focus order issues when encapsulating workspaces

orestisf1993:
I found this issue while adding new tests for #2975. I think the testcases are enough to explain the problem.

I don't know if the solution can be more elegant.

### DIFF
--- a/src/con.c
+++ b/src/con.c
@@ -809,6 +809,62 @@ Con *con_for_window(Con *con, i3Window *window, Match **store_match) {
     return NULL;
 }
 
+static int num_focus_heads(Con *con) {
+    int focus_heads = 0;
+
+    Con *current;
+    TAILQ_FOREACH(current, &(con->focus_head), focused) {
+        focus_heads++;
+    }
+
+    return focus_heads;
+}
+
+/*
+ * Iterate over the container's focus stack and return an array with the
+ * containers inside it, ordered from higher focus order to lowest.
+ *
+ */
+Con **get_focus_order(Con *con) {
+    const int focus_heads = num_focus_heads(con);
+    Con **focus_order = smalloc(focus_heads * sizeof(Con *));
+    Con *current;
+    int idx = 0;
+    TAILQ_FOREACH(current, &(con->focus_head), focused) {
+        assert(idx < focus_heads);
+        focus_order[idx++] = current;
+    }
+
+    return focus_order;
+}
+
+/*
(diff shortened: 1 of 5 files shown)


orestisf1993:
Any opinions on this?

src/con.c
```diff
@@ -795,6 +795,60 @@ Con *con_for_window(Con *con, i3Window *window, Match **store_match) {
...
+ * Iterate over the container's focus stack and return an array with the
+ * containers inside it, ordered from higher focus order to lowest.
+ *
+ */
+Con **get_focus_order(Con *con) {
+    Con **focus_order = smalloc(num_focus_heads(con) * sizeof(Con *));
+    Con *current;
+    int idx = 0;
+    TAILQ_FOREACH(current, &(con->focus_head), focused) {
+        focus_order[idx++] = current;
```
Airblader: Can we store the number of focus heads and add a sanity check here that bails out if we try to write out of bounds?

orestisf1993: But I malloc it through `num_focus_heads`, what would the improvement be?

Airblader: Sorry, I wasn't clear enough (and sorry for the delayed response; not getting much free time at my laptop right now :-) )

What I mean is to do it like this:

```
int num_focus_heads = num_focus_heads(con);
Con **focus_order = smalloc(num_focus_heads * sizeof(Con *));

// …

TAILQ_FOREACH(current, &(con->focus_head), focused) {
    if (idx >= num_focus_heads) {
        ELOG("…");
        break;
    }

    focus_order[idx++] = current;
}
```

It's just more defensive since the calculation of the array size happens elsewhere. Of course it should never happen (and currently doesn't), but we just avoid a potential out of bounds issue down the road.

orestisf1993: `num_focus_heads` counts the number of iterations using `TAILQ_FOREACH`. This function uses `TAILQ_FOREACH` to iterate through the `focus_head`s, increasing `idx` in the exact same way `num_focus_heads` does. This is a very core operation and should never fail.

I think an `assert`ion is a better choice since this would signify a huge error somewhere in the core logic.

Airblader: Fair enough, an assertion is fine with me :-)

orestisf1993:
Up. I can also resolve the conflicts if you want.

Airblader:
Yeah, please resolve the conflicts. :-)

orestisf1993:
Up

Issue #2767, opened 2017-05-10
i3bar feature request: x/y co-ordinates relative to module output


soumya92:
Currently i3bar's click events provide `x` and `y` relative to the root window (https://i3wm.org/docs/i3bar-protocol.html#_click_events), which is useful if you need to open an absolutely-positioned menu/applet.

It would be useful to have `x_frac`, `y_frac` (for the lack of a better name) that are relative to the module's output, such that a click on the top-left is `{"x_frac": 0, "y_frac": 0}`, and a click on the bottom right of the output is `{"x_frac": 1, "y_frac": 1}`.

This would allow a new level of interactivity, e.g., a volume module setting the volume to `(1 - y_frac) * 100`, a media player seeking to `x_frac * length`, or a 5-day weather forecast opening directly to day `int(x_frac * 5)`

Airblader:
The idea is OK to me, but I would prefer if we instead exposed pixel values relative to the block (`relative_x`, `relative_y`) and width and height of the block (`width`, `height`).

soumya92:
That was my initial thought too, but I assumed that adding two keys would be preferable to adding four. I'm all in favour of providing more information in the click event 😃

PMunch:
Maybe even better would be the x, y of the current implementation, the x, y of the block (in reference to root), and the width and height. That way you can easily calculate relative_x (block_x-click_x) and relative_y, fractions, and you could even start applications that are aligned with the block. So if I clicked eg. a date/time block I could create a calendar window aligned with the block. To keep it aligned however the user would obviously need to set a monospace font lest the bar would change below the window.

PR #3095, opened 2017-12-20
Add support for relative coordinates in i3bar click events

pallavagarwal07:
The JSON in the stdin of the status_command would get 4 new parameters:
```
x_rel  - relative position of click w.r.t. block (x coord)
y_rel  - relative position of click w.r.t. block (y coord)
width  - width of block
height - height of block
```
This was discussed in #2767

### DIFF
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -596,7 +596,7 @@ void child_click_events_key(const char *key) {
  * Generates a click event, if enabled.
  *
  */
-void send_block_clicked(int button, const char *name, const char *instance, int x, int y) {
+void send_block_clicked(int button, const char *name, const char *instance, int x, int y, int x_rel, int y_rel, int width, int height) {
     if (!child.click_events) {
         return;
     }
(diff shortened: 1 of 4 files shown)


i3bar/src/child.c
```diff
@@ -624,6 +624,18 @@ void send_block_clicked(int button, const char *name, const char *instance, int
     child_click_events_key("y");
     yajl_gen_integer(gen, y);
 
+    child_click_events_key("x_rel");
```
Airblader: Can we rename these to `relative_x` and `relative_y` please?

i3bar/src/xcb.c
```diff
@@ -523,7 +523,8 @@ void handle_button(xcb_button_press_event_t *event) {
                 block_x += render->width + render->x_offset + render->x_append + get_sep_offset(block) + sep_offset_remainder;
 
                 if (statusline_x <= block_x && statusline_x >= last_block_x) {
-                    send_block_clicked(event->detail, block->name, block->instance, event->root_x, event->root_y);
+                    send_block_clicked(event->detail, block->name, block->instance,
+                                       event->root_x, event->root_y, statusline_x - last_block_x, event->event_y, block_x - last_block_x, bar_height);
```
Airblader: The y coordinate will be wrong if the bar is either on the bottom or if there are multiple bars on top.

Airblader: Actually scratch that, I just read this wrong.

i3bar/src/child.c
pallavagarwal07: Done

pallavagarwal07:
Since we are changing the JSON format, shouldn't the version number be updated? Right now the version number is always `"version": 1`.

Airblader:
Thanks. Can you start the list items uppercased, though? Also, please squash those commits into a single one that also mentions "fixes #2767" and we're good to go I think. :-) 

We don't need to increment the version since this is a backwards compatible change.

pallavagarwal07:
Done

Airblader:
Thank you!

Issue #2778, opened 2017-05-21
Focusing with mouse stops working. Focus sometimes gets stuck on i3 and i cannot click inside any window.


exi:
Output of `i3 --moreversion 2>&- || i3 --version`:

Binary i3 version:  4.13 (2016-11-08) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.13 (2016-11-08) (pid 14558) abort…)
Loaded i3 config: //home/exi/.config/i3/config (Last modified: So 24 Jul 2016 05:58:08 PDT, 26025342 seconds ago)

The i3 binary you just called: /nix/store/fyp2zyyc66v76qcwqxq1siqzpy80kdc0-i3-4.13/bin/i3
The i3 binary you are running: /nix/store/fyp2zyyc66v76qcwqxq1siqzpy80kdc0-i3-4.13/bin/i3

**What I did:**

1. Open Firefox with some tabs
1. Open Chrome with some tabs
1. Click around in Chrome a bit. 

**What I saw:**

I now cannot focus any i3 container with my mouse anymore.
Also i cannot click anything in any window anymore.
Restarting i3 via key bindings resolves the problem until I click around in Chrome/Firefox some more.
Sometimes the i3 focus will get stuck on i3 itself and scolling the mouse will switch through my workspaces although
I clicked on a container and expected it to have focus.

**What I expected instead:**
My windows should have focus and should be selectable/clickable via mouse.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

exi:
I will try to capture a log of this although I'm having trouble reproducing this reliably.

Airblader:
Please do. I think without logs we have no chance here.

exi:
Log: http://logs.i3wm.org/logs/5657230527430656.bz2

exi:
The symptoms are: The focus is stuck to a process, whichever that is at the moment. And even switching containers with the i3 key bindings will keep the mouse focus on the other container. So for example, if the mouse focus is stuck to i3bar but the keyboard focus is stuck to firefox, I can type into firefox but clicking into it does not get the mouse focus to firefox. Scrolling over the firefox window will not scroll firefox but i3bar, which causes me to cycle to my workspaces.

exi:
[...]

PR #3097, opened 2017-12-20
Refocus focused window for FOCUS_IN events on the root window.

### LINKED ISSUE #2722: Bug i3 with tkinter application. I3 seems to stop replying.
Running i3 version: 4.13 (2016-11-08) (pid 451)to abort…)
Loaded i3 config: /home/tau/.i3/config (Last modified: Mon 20 Mar 2017 10:54:15 PM -03, 545611 seconds ago)

_REPLACE: i3 version output_
03/27/2017 06:28:04 AM - Additional arguments passed. Sending them as a command to i3.
[{"success":false,"parse_error":true,"error":"Expected one of these tokens: <end>, '[', 'move', 'exec', 'exit', 'restart', 'reload', 'shmlog', 'debuglog', 'border', 'layout', 'append_layout', 'workspace', 'focus', 'kill', 'open', 'fullscreen', 'sticky', 'split', 'floating', 'mark', 'unmark', 'resize', 'rename', 'nop', 'scratchpad', 'title_format', 'mode', 'bar'","input":"version","errorposition":"^^^^^^^"}]

The application below after running and one pressing 'Return' 
it should quit the application and give back focus to the previous
active windows. it seems i3 stops replying to commands from the keyboard
and i have to click on some container to get things working again.
Replacing <Return> for all other kind of events it fails too. Notice, it fails if you 
replace Entry for Button too.

I have other application that depends on the correct behavior. I hope you guys
have how to fix it.

Obs: if i comment self.resizable... it fails too.

    from Tkinter import *
    
    class BugI3(Tk):
    
        def __init__(self, title, default_data='', *args, **kwargs):
            Tk.__init__(self, *args, **kwargs)
            # when one quits the application
            # from an event it just makes i3 misbehaves
            # it cant get focus back to the previous application.
            self.bind('<Return>', lambda e: self.quit())
    
            self.title(title)
            self.resizable(0, 0)
    
            self.entry = Entry(self)
            self.entry.pack(side='left', expand=True, fill=BOTH)
    
    
    if __name__ == '__main__':
        bugi3 = BugI3('Word Pass')
        bugi3.mainloop()

### LINKED ISSUE #3096: i3 stops responding to keyboard input when certain windows close

```
Binary i3 version:  4.14.1-227-g13ca7830 (2017-12-12, branch "gaps-next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.14.1-227-g13ca7830 (2017-12-12, branch "gaps-next") (pid 1580)
Loaded i3 config: /home/ash/.config/i3/config (Last modified: Wed 20 Dec 2017 08:09:32 AM EST, 2481 seconds ago)

```

https://ptpb.pw/C9X4

**What I did:**

I closed Steam on a workspace by itself

**What I saw:**

i3 stopped responding to keyboard input entirely

**What I expected instead:**

i3 should continue to respond to keyboard input normally

**Extra info:**

This occurs inconsistently - mainly when a window is on a workspace by itself, but only with certain windows. This happens with Steam as well as with my project [yotredash](https://github.com/ashkitten/yotredash).

Airblader:
This deals with (admittedly somewhat misbehaving) clients which
use XSetInputFocus to take focus, but then don't properly restore
focus. This has been observed with TK apps, but also, e.g., Steam.

fixes #2722
fixes #3096

### DIFF
--- a/include/xcb.h
+++ b/include/xcb.h
@@ -53,6 +53,7 @@
                                                                   ConfigureNotify */                  \
                          XCB_EVENT_MASK_POINTER_MOTION |                                              \
                          XCB_EVENT_MASK_PROPERTY_CHANGE |                                             \
+                         XCB_EVENT_MASK_FOCUS_CHANGE |                                                \
                          XCB_EVENT_MASK_ENTER_WINDOW)
 
 #define xmacro(atom) xcb_atom_t A_##atom;
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -1207,6 +1207,14 @@ static bool handle_clientleader_change(void *data, xcb_connection_t *conn, uint8
  */
 static void handle_focus_in(xcb_focus_in_event_t *event) {
     DLOG("focus change in, for window 0x%08x\n", event->event);
+
+    if (event->event == root) {
+        DLOG("Received focus in for root window, refocusing the focused window.\n");
+        con_focus(focused);
+        focused_id = XCB_NONE;
+        x_push_changes(croot);
+    }
+
     Con *con;
     if ((con = con_by_window_id(event->event)) == NULL || con->window == NULL)
         return;


Airblader:
Currently only to have the PR already, I'd like someone to run this for a few days first.

Airblader:
I'd mostly want someone to run the patch to make sure there's no side effects causing regressions. So not using those apps is totally fine. Thanks for volunteering!

orestisf1993:
I was stuck in a state where clicking on an empty workspace would not focus it. I restarted and the problem disappeared. Don't know how to reproduce or if it was caused by your patch.

Just something to keep in mind.

Airblader:
I think you might have a point here. What happens with this patch in a multi-monitor setup? Say you have two outputs O1 and O2. O1 has a window, which is focused. O2 shows an empty workspace. Now you move the mouse from O2 to O2. What happens here?

orestisf1993:
It doesn't happen reliably. I've only noticed it with mouse clicks, not mouse movement since I don't use `focus_follows_mouse`. I'll try running i3 without your patch for a few days to see if the issue is elsewhere.

ashkitten:
fwiw i'm not seeing any side effects on my end

orestisf1993:
Do you use a second monitor?

ashkitten:
I do use a second monitor

orestisf1993:
I can't really reproduce my problem right now.

Airblader:
What you saw might be related to #2778 or the issues related to that one. I think we should treat it as such for now and maybe go ahead with merging this. @stapelberg Agreed?

orestisf1993:
(for the record I am neither for or against merging this, just documenting my problem)

> What you saw might be related to #2778 or the issues related to that one

No, it's not like this. I can click other windows just not inside the non-focused, visible workspace.

Airblader:
OK. Actually we should look into that, it does sound like something that might be caused by this patch.

orestisf1993:
Ok it happened again. I still have no idea how to reproduce it. My observation this time is that when it happens, it stays. Eg I can change workspaces using keys etc but I can never click on an empty workspace.

Clicking on windows, i3bar works as expected.

Airblader:
Perhaps this is some kind of race condition where `focused` is a closing client and we end up trying to focus a window that no longer exists. Or something like that.

siikamiika:
Some information probably related to this bug. `i3 version 4.14.1`, dual monitor

When I bindsym a script that invokes dmenu, like `bindsym $mod+something exec script`, dmenu doesn't (usually) respond to input. I can regain control over i3 by clicking another workspace with mouse. The script works fine when it isn't called by i3, though.

script:
```bash
#!/bin/bash
INPUT=$(echo | dmenu)
...
```

Similar thing happens with a tkinter python script, but it's possible to work around it by calling `xdotool windowactivate $id` ($id is the output of `xdotool getactivewindow` before creating the tkinter widget), then sleeping 0.1 sec and then calling `root.destroy()`. I don't know if there is some other, preferred way to let go of focus.

Airblader:
@stapelberg The majority of the feedback was positive, so perhaps we should merge this for now. If we do find proof that @orestisf1993 saw something caused by this we can fix it, but for now it seems to help more people than it does bad. What do you think?

stapelberg:
Sounds good, thanks!

stapelberg:
…though, maybe we should cut a release before?

Airblader:
Sure, that's fine by me!

Airblader:
@stapelberg This can be merged now then.

PR #3101, opened 2017-12-25
Add workspace vars to support DRY when customizing

tpraxl:
Relabeling the workspaces required the user to update the labels at least in two places of the i3 config file (switch to workspace + move focused container to workspace) to keep a consistent state.

This got even worse when the user started to reference the workspaces elsewhere, e.g. for assigning containers to specific workspaces.

This change supports the DRY principle for users. Relabeling the workspaces is now merely a matter of changing the respective variable ($ws1, $ws2, … $ws10).

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -104,17 +104,8 @@ bindsym Mod1+Shift+minus move scratchpad
 # If there are multiple scratchpad windows, this command cycles through them.
 bindsym Mod1+minus scratchpad show
 
-# Define workspaces along with their labels
-#
-# Customize your workspace labels here.
-# Example:
-#
-# set $ws1 "Terminals" 
-#
-# or
-#
-# set $ws1 "1: Terminals"
-#
+# Define names for default workspaces for which we configure key bindings later on.
+# We use variables to avoid repeating the names in multiple places.
 set $ws1 "1"
 set $ws2 "2"
 set $ws3 "3"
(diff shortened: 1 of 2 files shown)


etc/config
```diff
@@ -104,29 +104,52 @@ bindsym Mod1+Shift+minus move scratchpad
 # If there are multiple scratchpad windows, this command cycles through them.
 bindsym Mod1+minus scratchpad show
 
+# Define workspaces along with their labels
+#
+# Customize your workspace labels here.
```
Airblader: What I don't like about this comment is that it almost implies that workspace names in i3 have to be defined upfront; but workspace names are entirely dynamic, even at runtime. We already make the config that much longer introducing the variables, let's keep commenting this to a minimum (we have a userguide for more details). Something like:

    # Define names for default workspaces for which we configure key bindings later on.
    # We use variables to avoid repeating the names in multiple places.

tpraxl: Agree. I replaced the old comment with your proposal.

tpraxl:
@Airblader Just in case you didn't notice: the requested changes have been made.

Airblader:
Yes, I will get to this soon. The holidays are keeping me a bit busy. It's on my list though, don't worry. :-)

tpraxl:
@Airblader Very well. Thanks. Wish you a Happy New Year.

Airblader:
Thanks!

PR #3105, opened 2017-12-27
Docs: state that pango markup in i3bar requires the use of a pango font.

clonejo:
I was using the default font and it took me some hours to figure this out. Did not see it mentioned anywhere.


Airblader:
Thank!

PR #3111, opened 2018-01-06
Avoid some code duplication

livanh:
Some code and checks in cmd_resize_tiling_width_height() can be replaced with a call to resize_find_tiling_participants(). To get the same behavior, resize_find_tiling_participants() is extended so it can also look for a resize target in two direction simultaneously: left+right or top+bottom.
The need for this change emerged in the discussion of #3036, it and can be reused in that PR as well.

### DIFF
--- a/include/resize.h
+++ b/include/resize.h
@@ -11,6 +11,6 @@
 
 #include <config.h>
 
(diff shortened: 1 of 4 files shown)


stapelberg:
Thanks!

Issue #1129, opened 2013-12-01
Support vertical dock clients


i3bot:
[Originally reported by pie3mn@…]
(In the documentation dock mode is being described as lacking support for vertical bars which would free up vertical space which is sparse on todays widescreen displays already.

apeyser:
Oh no -- it should be me much simpler, and would be terribly useful. Don't bother with docking --- in fact, I don't use any dock thingies in the i3bar because that is, in imho, silly on a large screen. Just write the text from top to bottom or bottom to top --- if you can't read sideways, you're not using i3. 

Everyone is starting from interactive notions pioneered in the 80s for 640x480 single black & white screens made for folks who had never seen a gui at all before. There's no reason that a project like i3 should cater to those sensibilities. Just give me the json text bar I have now, just rotated.

If I want docked thingies,  I put them in a seperate graphical box, like stalonetray -- and in fact, it would be better to separate the informational text bar from interactive thingies. A menu and an info bar are different kinds of thingies, and sticking them together lead to all kinds of code complexities, interaction complexities and limitations of use. Text thingies can be quite small; interactive thingies need to be big, for example.

And I'd love the tabs also to be on the left or right --- exactly like now, just rotated.

Airblader:
Well, such a big feature would be built for a large group of people and I don't think for the "general public", reading sideways is a a route we should go down.

apeyser:
Why not? It's just a question of tradition, and the large group of people that i3 appeals to should have no problem adapting.

Airblader:
But once this "tradition" is set, it is set in stone and can't be changed easily. And I'm not convinced a 90 degrees rotated bar is really beneficial (also, I'm not sure XCB offers an easy way to rotate an entire pixmap like that).

naelyn:
I would like the ability to use gkrellm as well.  This is the one thing keeping me from switching to i3 for my primary WM.  Until then I have to relegate it to throwaway VMs for testing only.

Airblader:
I've seen a lot of people using other docks than i3bar with i3 so I think supporting this is worth having even without i3bar having it. 

I think I will have a look at this in a while. Tony stated that the assumption of horizontal docks is made in a lot of places. I'll have to check whether the effort is manageable.

drasill:
Now that _NET_WM_STATE_STICKY is supported, having a vertical dock is a matter of having a "right" gap on all workspaces, and have the dock being a sticky floating window ?

Airblader:
[...]

PR #3123, opened 2018-01-24
[WIP] Left and right dockarea

vesim987:
This PR add support for vertical dock client.

TODO:
- `handle_strut_partial_change` seem's to doesn't work
- do something with dock overlapping
- refactor `render_output`
- write tests

fixes: [https://github.com/i3/i3/issues/1129](https://github.com/i3/i3/issues/1129)

### DIFF
--- a/src/render.c
+++ b/src/render.c
@@ -313,7 +313,7 @@ static void render_output(Con *con) {
                 assert(false);
             }
             content = child;
-        } else if (!child->is_docked) {
+        } else if (!child->is_docked ) {
             DLOG("Child %p of type %d is inside the OUTPUT con\n", child, child->type);
             assert(false);
         }
@@ -342,10 +342,10 @@ static void render_output(Con *con) {
 
     /* First pass: determine the height of all CT_DOCKAREAs (the sum of their
      * children) and figure out how many pixels we have left for the rest */
-    Con *left_dock = 0;
-    Con *right_dock = 0;
-    Con *bottom_dock = 0;
-    Con *top_dock = 0;
+    Con *left_dock = NULL;
+    Con *right_dock = NULL;
+    Con *bottom_dock = NULL;
+    Con *top_dock = NULL;
 
     TAILQ_FOREACH(child, &(con->nodes_head), nodes) {
         if (child->type == CT_HDOCKAREA) {
--- a/src/render.c
+++ b/src/render.c
@@ -17,7 +17,9 @@ static void render_output(Con *con);
 static void render_con_split(Con *con, Con *child, render_params *p, int i);
 static void render_con_stacked(Con *con, Con *child, render_params *p, int i);
 static void render_con_tabbed(Con *con, Con *child, render_params *p, int i);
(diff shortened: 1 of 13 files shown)


vesim987:
Oh, i broke commit history.

okraits:
I guess you need to rebase your branch against the next branch.

Airblader:
Thanks for taking a shot at this! I really think this would be a cool feature to have in i3. As @okraits mentioned you'll need to rebase.

Once that is sorted out, two very high-level comments (I haven't reviewed the PR in detail yet):

1. It'd be good if you could write, in text form, some explanation as to how vertical dock clients fit into the container structure of i3 now. For one it'll help us understand, but I also think we'll need to document this a bit in, say, the hacking howto, as this is a change to the fundamental container structure.
2. Please don't rename things on the IPC (I just saw a "dockarea" → "hdockarea"). This breaks compatibility which we'd like to avoid when possible. Having "dockarea" and "vdockarea" is not the ideal situation, but it's better than breaking IPC clients.

Let us know via comment when you rebased the PR!

vesim987:
@Airblader I've fixed this branch. There was ~170 commits difference... idk how I did this. I've just created new branch, and applied my latest commit, and renamed it.

Airblader:
> add left and right position to i3bar

Please don't do this in this PR. It's a completely separate topic and one that has to be decided separately as well. I currently don't think we want a vertical mode for i3bar because we have no good answer to what the statusline should look like.

src/render.c
```diff
@@ -338,41 +343,100 @@ static void render_output(Con *con) {
 
     /* First pass: determine the height of all CT_DOCKAREAs (the sum of their
      * children) and figure out how many pixels we have left for the rest */
+    Con *left_dock = 0;
```
Airblader: Please use NULL over 0

src/render.c
```diff
@@ -338,41 +343,100 @@ static void render_output(Con *con) {
...
-        child->rect.height = 0;
-        TAILQ_FOREACH(dockchild, &(child->nodes_head), nodes)
-        child->rect.height += dockchild->geometry.height;
+            width -= child->rect.width;
+            
+        }
+    }
+
+    
+    if(bottom_dock) {
```
Airblader: Again, these blocks are a lot of repetition that should be refactored.

vesim987:
@Airblader do you have idea how to rewrite this part: https://github.com/vesim987/i3/blob/vertical-dockarea/src/handlers.c#L1376-L1415

op8867555:
Hi there, any progress on this? I tried this PR and found some problems:
* Incorrect bar position when using multiple monitors
* Incorrect bar position when running multiple bars
* `is_docked` is not restore when loading json layout which would crash i3 when reloading

Here is some changes to fixes problems above:
https://github.com/vesim987/i3/compare/vertical-dockarea...op8867555:vertical-dockarea

I'm new to i3, please take a look and point me out if I didn't do it right. :)

Airblader:
@vesim987 Are you continuing your work on this?

PR #3135, opened 2018-02-11
Update EWMH atoms when closing inactive workspace

### LINKED ISSUE #3126: i3 does not correctly set *DESKTOP* atoms
i3 does not correctly set `_NET_DESKTOP_NAMES`, `_NET_NUMBER_OF_DESKTOPS` and `_NET_WM_DESKTOP` atoms if the last window of the inactive workspace is closed. Here are the steps to reproduce the error:

1. Go to the empty workspace `#1` and open a terminal on it.
2. Do the same thing with the empty workspace `#3`.
3. Then go to the empty workspace `#2` and open a terminal again. Type in it: `sleep 5; exit`
4. Switch to the workspace `#1` and wait until the workspace `#2` disappears from the status line.
5. And for now, atom values are:
   ```
   $ xprop -root
   <...>
   _NET_DESKTOP_NAMES(UTF8_STRING) = "1", "2", "3"
   _NET_NUMBER_OF_DESKTOPS(CARDINAL) = 3
   <...>
   ```
   while the status line shows only `#1` and `#3`. Expected values are:
   ```
   _NET_DESKTOP_NAMES(UTF8_STRING) = "1", "3"
   _NET_NUMBER_OF_DESKTOPS(CARDINAL) = 2
   ```
6. Now switch to the workspace `#3` and do `xprop | grep _NET_WM_DESKTOP` with the terminal window `#3` itself. It will log `2`, while we are in the *second existent* workspace (value `1` is expected in this case).

```
Running i3 version: 4.14.1 (2017-09-24) (pid 11740)bort…)
Loaded i3 config: /home/user/.config/i3/config (<...>)

```

op8867555:
This PR is an attempt to fix #3126 by simply update EWMH atoms when an UnmapNotify received.
If it is ok, I'm thinking about writing tests for this, which one is better: creating a separated testcase or modifying existing ones?

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -330,6 +330,13 @@ bool tree_close_internal(Con *con, kill_window_t kill_window, bool dont_kill_par
         DLOG("parent container killed\n");
     }
 
+    if (ws == con) {
+        DLOG("Closing a workspace container, updating EWMH atoms\n");
+        ewmh_update_number_of_desktops();
+        ewmh_update_desktop_names();
+        ewmh_update_wm_desktop();
+    }
+
     con_free(con);
 
     /* in the case of floating windows, we already focused another container
--- a/testcases/t/294-update-ewmh-atoms.t
+++ b/testcases/t/294-update-ewmh-atoms.t
@@ -1,0 +1,111 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • https://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • https://build.i3wm.org/docs/lib-i3test.html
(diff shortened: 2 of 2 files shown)


Airblader:
Thanks for taking a look at this! I think theoretically speaking the more appropriate place to do this would be somewhere in the logic that modifies the tree and actually closes the now empty and not active workspace, seeing how the EWMH atoms are regarding dekstops and not the individual window.

op8867555:
I moved the code to `tree_close_internal` which modifying the tree and it only updates when closing a workspace container now.

Airblader:
The change looks fine to me now. As far as tests go, a new test would be better here I think. Also, when you did that, please squash the commits together into a single commit. Thanks!

op8867555:
Add the test and commits are squashed now. :)

Airblader:
Thank you!

PR #3141, opened 2018-02-15
Fix compatibility issue with git<=1.8.5-rc0

giuscri:
-C <path> was introduced in 1.8.5-rc0.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,8 +31,10 @@ AX_EXTEND_SRCDIR
 
 AS_IF([test -d ${srcdir}/.git],
       [
-        VERSION="$(git -C ${srcdir} describe --tags --abbrev=0)"
-        I3_VERSION="$(git -C ${srcdir} describe --tags --always) ($(git -C ${srcdir} log --pretty=format:%cd --date=short -n1), branch \\\"$(git -C ${srcdir} describe --tags --always --all | sed s:heads/::)\\\")"
+        pushd ${srcdir}
+        VERSION="$(git describe --tags --abbrev=0)"
+        I3_VERSION="$(git describe --tags --always) ($(git log --pretty=format:%cd --date=short -n1), branch \\\"$(git describe --tags --always --all | sed s:heads/::)\\\")"
+        popd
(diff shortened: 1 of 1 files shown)


Airblader:
git 1.8.5-rc0 has been released [4.5 years ago](https://github.com/git/git/releases?after=v1.8.5). Given that i3 depends on libraries that haven't even *existed* until 1.5 years ago (xcb-util-xrm) I wonder why we need this compatibility. Why not update git? There have been many crucial security fixes in git since then as well.

giuscri:
It's possible that some distributions are released with such an old version. See https://rpmfind.net/linux/RPM/centos/updates/7.4.1708/x86_64/Packages/git-1.8.3.1-12.el7_4.x86_64.html

jolange:
The currently supported enterprise linux versions of course all got quite old git versions (e.g. git 1.7.1 on EL6). But as you said, there are other dependencies that cause trouble anyway...

Airblader:
So how are these problems addressed (by actual i3 users)? If supporting such old git versions actually makes sense for some users I'm not opposed, but if they only get one step further in the compilation and then get stuck on missing libraries anyway, there's no point. One can always compile an equally old version of i3… :-)

jolange:
> So how are these problems addressed (by actual i3 users)? [...] One can always compile an equally old version of i3… :-)

Exactly that way (at least for my part).

> If supporting such old git versions actually makes sense for some users

I don't recall many people complaining yet, so it does not seem to be a huge problem.

Airblader:
@jolange Thanks for the input. The change in this PR is so trivial that I'm inclined to just merge it, but at the same time it feels odd for the mentioned reasons. 

I'll wait a bit for @gfazio's answer in the issue to hear at least a second user opinion it as well to see whether there was a good reason to have such an old git.

Airblader:
Given that there's no reply, I'll close this for now.

jirib:
Thx, this fixes i3 on Centos 7 which as git-1.8.3.1-14.el7_5.x86_64.

PR #3147, opened 2018-02-21
dump-asy: add prerequisites check; fix empty container names warnings

walker0643:
Since the script isn't installed via package manager with the benefit of dependency checks, the first chunk makes sure that both asymptote and gv are installed before allowing the script to fail with less useful error output.
 
The second chunk squelches the warnings created by empty container names.

### DIFF
--- a/contrib/dump-asy.pl
+++ b/contrib/dump-asy.pl
@@ -15,6 +15,12 @@ use AnyEvent::I3;
 use File::Temp;
 use File::Basename;
 use v5.10;
+use IPC::Cmd qw[can_run];
+
+# prerequisites check so we can be specific about failures caused
+# by not having these tools in the path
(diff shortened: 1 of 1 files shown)


orestisf1993:
The script is useful without `gv` imo.

Airblader:
It may be useful without, but currently it hard-relies on it. We can always change it so it only uses gv under certain circumstances, of course.

orestisf1993:
It didn't rely on it since it could finish all other stuff before trying to call it. Anyway, I guess I can add a `--no-gv` flag if it bothers me.

Airblader:
Yes, it does something else first, but then executes a command that is bound to fail. Hence its a hard dependency.

Adding a flag and thus making the check optional is fine with me if you want to do that. :)

Issue #3164, opened 2018-03-09
[Bug/Enhancement] X Resource RGB notation rgb:hh/hh/hh does not work


klorax:
X.Org Foundation recommends the RGB notation `rgb:hh/hh/hh` (which does not work when using X resources in i3) over `#rrggbb`: 

> "For  backward  compatibility,  an older syntax for RGB Device is supported, but its continued use is not encouraged. The syntax is an initial sharp sign character followed by a numeric specification ..."

Output of `i3 --moreversion 2>&- || i3 --version`:

N/A

URL to a logfile as per https://i3wm.org/docs/debugging.html:

N/A

**What I did:**

Entry in `$HOME/.Xresources`: `*color0: rgb:00/00/00`.
Entry in `$HOME/.i3/config`: `set_from_resource $var color0 #000000`.

**What I saw:**

Incorrect (gray != #000000) color.

**What I expected instead:**
To see the color specified.

Airblader:
Thanks for the issue. i3 doesn't explicitly support colors through X resources, we just allow setting variables from it.

That means this would be about supporting this color format in the i3 configuration in general, for which I don't see a good reason. I think it's reasonable to expect users to use the format i3 expects.

klorax:
Fair enough, to be honest I don't know why they switched, interested to find out but haven't found anything yet.

However two things might be good to change in i3:

1. The fallback color did not seem to be used in this case, which might be excpected behaviour.
2. There could be a note in the documentation section "X resources" about the color format notation.

Thanks!

Airblader:
I think there's a misunderstanding. i3 doesn't support »setting colors from Xresources«. It supports defining variables from Xresources. Whether these are used as colors, other values or even entire blocks of the config itself is entirely up to user – variables have no notion of syntax for their value. Hence using a variable as the color in, say, i3bar, cannot fall back to the fallback. The parsing of the variable succeeded, you just passed an invalid argument to i3bar.

klorax:
Ok, thanks for clearing up the part about the fallback!

I still think the documentation (userguide) could benefit from being a bit clearer though, currently it reads:

> "[variables] can also be created using a value configured in the X resource database. This is useful, for example, to avoid configuring color values within the i3 configuration. Instead, **the values can be configured, once, in the X resource database** to achieve an easily maintainable, consistent color theme across many X applications." (emphasis added)

As someone new, it is difficult know that i3 is getting the value for the resource verbatim and therefore must be in i3's format.

[...]

PR #3172, opened 2018-03-11
Docs [#3164]: Clarification about X resource value

klorax:
Clarification about X resource value: they are loaded verbatim and must therefore be in the format that i3 uses.
Solves #3164.


Airblader:
Thanks!

PR #3179, opened 2018-03-16
Correctly handle bindings for the same key with and without --release 

### LINKED ISSUE #2733: bindcode --release does not work (but bindsym --release does)
Hello,

I'm trying to show an xclock window only while a key is pressed, as is the case for i3bar in hide mode. Technically, this would be possible if i3 allowed assigning different commands when pressing and releasing the same key, i.e. one could then write

keysym [KEY]  [COMMAND1]
keysym --release [KEY] [COMMAND2]

As far as I understand, the current behaviour is to only use the keybinding which appears last in the configuration file.

Does anything speak against allowing such a double-binding on press and release for both the keysym and keycode statement? Maybe, to avoid confusion, one could also implement some new option like

keysym --pressrelease [KEY] [COMMAND1] [COMMAND2]

Best,

Simon

orestisfl:
Actually #2733's title is different from what I am fixing. I'll reread the thread tomorrow but `bindcode --release` works fine currently.

I am fixing the issue described in https://github.com/i3/i3/issues/2733#issuecomment-360805836.
I also wrote a small explanation in the second commit message.

- [x] I have
    ```
    mode "resize" {
    ...
        bindsym r --release exec --no-startup-id ~/bin/draw-resize.sh $(xdotool getactivewindow), mode "default"
    }
    bindsym $mod+r mode "resize"
    ```
    in my config. The first time I $mod+r it works as intended. The second time I $mod+r the r `--release` bindsym gets triggered immediately. This does not happen w/ `next`. Need to investigate.
- [ ] Tests from https://github.com/i3/i3/pull/3179#issuecomment-374261240

Fixes #2733

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -649,6 +649,14 @@ void switch_mode(const char *new_mode) {
         translate_keysyms();
         grab_all_keys(conn);
 
+        /* Reset all B_UPON_KEYRELEASE_IGNORE_MODS bindings to avoid possibly
+         * activating one of them. */
+        Binding *bind;
+        TAILQ_FOREACH(bind, bindings, bindings) {
+            if (bind->release == B_UPON_KEYRELEASE_IGNORE_MODS)
+                bind->release = B_UPON_KEYRELEASE;
+        }
+
         char *event_msg;
         sasprintf(&event_msg, "{\"change\":\"%s\", \"pango_markup\":%s}",
                   mode->name, (mode->pango_markup ? "true" : "false"));
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -198,6 +198,7 @@ void regrab_all_buttons(xcb_connection_t *conn) {
  */
 static Binding *get_binding(i3_event_state_mask_t state_filtered, bool is_release, uint16_t input_code, input_type_t input_type) {
     Binding *bind;
+    Binding *result = NULL;
 
     if (!is_release) {
         /* On a press event, we first reset all B_UPON_KEYRELEASE_IGNORE_MODS
@@ -271,23 +272,26 @@ static Binding *get_binding(i3_event_state_mask_t state_filtered, bool is_releas
         if (bind->release == B_UPON_KEYRELEASE && !is_release) {
             bind->release = B_UPON_KEYRELEASE_IGNORE_MODS;
             DLOG("marked bind %p as B_UPON_KEYRELEASE_IGNORE_MODS\n", bind);
-            /* The correct binding has been found, so abort the search, but
-             * also don’t return this binding, since it should not be executed
-             * yet (only when the keys are released). */
-            bind = TAILQ_END(bindings);
-            break;
-        }
-
-        /* Check if the binding is for a press or a release event */
-        if ((bind->release == B_UPON_KEYPRESS && is_release) ||
-            (bind->release >= B_UPON_KEYRELEASE && !is_release)) {
+            if (result) {
+                break;
+            }
             continue;
         }
 
-        break;
+        /* Check if the binding is for a press or a release event */
+        if ((bind->release == B_UPON_KEYPRESS && is_release)) {
+            continue;
+        }
+
+        if (is_release) {
+            return bind;
+        } else if (!result) {
+            /* Continue looping to mark needed B_UPON_KEYRELEASE_IGNORE_MODS. */
+            result = bind;
+        }
     }
 
-    return (bind == TAILQ_END(bindings) ? NULL : bind);
+    return result;
 }
 
 /*
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -227,9 +227,9 @@ static Binding *get_binding(i3_event_state_mask_t state_filtered, bool is_releas
         /* For keyboard bindings where a symbol was specified by the user, we
          * need to look in the array of translated keycodes for the event’s
          * keycode */
+        bool found_keycode = false;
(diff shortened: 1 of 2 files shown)


src/bindings.c
```diff
@@ -270,23 +271,26 @@ static Binding *get_binding(i3_event_state_mask_t state_filtered, bool is_releas
...
+        if ((bind->release == B_UPON_KEYPRESS && is_release)) {
             continue;
         }
 
-        break;
+        if (is_release) {
+            return bind;
+        } else {
+            /* Continue looping to mark needed B_UPON_KEYRELEASE_IGNORE_MODS. */
+            result = bind;
```
orestisf1993: This will change the current behaviour: if multiple bindings for the same key are defined (which is an error) the last of them will be executed. Will fix that.

Airblader:
So if it doesn't fix the issue mentioned in the commit message, we should change the commit message (to a "relates to")? Otherwise it'll auto-close the issue.

orestisf1993:
What I mean is that `bindcode --release` works fine in `next`.

I reread the whole thread since my understanding of the issue was not 100% correct.

### `bindsym` + `bindsym --release`
https://github.com/i3/i3/issues/2733#issuecomment-294005259
```
bindsym $mod+b exec notify-send "press"
bindsym --release $mod+b exec notify-send "release"
```
works in `next`.

### `bindcode` + `bindcode --release` with non-mod keys
https://github.com/i3/i3/issues/2733#issuecomment-294077895
```
bindcode 66 exec notify-send "press"
bindcode --release 66 exec notify-send "release"
```
works in `next`.


### `bindcode` + `bindcode --release` with mod keys
https://github.com/i3/i3/issues/2733#issuecomment-360445055
```
bindcode 133 exec notify-send "press"
bindcode --release 133 exec notify-send "release"
```
doesn't work in `next`, fixed in this PR.
Reason: once 133 (Super_L) is pressed the first `bind` with `->release == B_UPON_KEYPRESS` is matched and the loop exits in
https://github.com/i3/i3/blob/a9512c6345ae66c658e215396be1278754471671/src/bindings.c#L290
the other `bind` with `->release == B_UPON_KEYRELEASE` is never changed to `B_UPON_KEYRELEASE_IGNORE_MODS`.
The mods don't match in
https://github.com/i3/i3/blob/a9512c6345ae66c658e215396be1278754471671/src/bindings.c#L256-L257
since it's a mod key bindcode (I think, this is pretty confusing).


### `bindsym` + `bindsym --release` with mod keys
```
bindsym Super_L exec notify-send "press"
bindsym --release Super_L exec notify-send "release"
```
doesn't work in `next` but didn't work in the first version of this PR either, didn't notice.
Reason is that in this line
https://github.com/i3/i3/blob/a9512c6345ae66c658e215396be1278754471671/src/bindings.c#L239
`B_UPON_KEYRELEASE_IGNORE_MODS` is not checked.
This caused another bug. From 548d74015c50d7fae14bfb8bb1989acde5fc22ae:
> 1. press $mod, press x, release x, release $mod
> 2. press $mod, press x, release $mod, release x

case (2.) didn't work, now it should be fixed. Is there an issue for this?

I don't feel really confident about merging this since bindings are not tested very well.

Airblader:
> I don't feel really confident about merging this since bindings are not tested very well, let's give it some time.

Thanks, I appreciate that.

Maybe @stapelberg can chime in here, the bindings are more his area. :-)

orestisf1993:
I could start adding some more binding tests. Not just for the changes, the whole thing.

Airblader:
That'd be wonderful. Perhaps we do it in a separate PR?

stapelberg:
More tests are always appreciated :). I haven’t looked at the code in question, but I’d be fine merging a bindcode-related change at the beginning of the development cycle (which is where we are right now). Most users use bindsym, so potential breakages are fairly limited.

orestisf1993:
Actually, `258-keypress-release.t` and `290-keypress-numlock.t` test a bunch of cases after all, don't know if there are many stuff missing unrelated to this PR.

Maybe:
1. Test different order from 548d74015c50d7fae14bfb8bb1989acde5fc22ae
2. Test duplicate bindsym/bindcode (+permutations). In all cases the first binding should be called and an error should be produced.
3. With
    ```
    set $mod Mod4 # super_l
    bindsym Super_L nop 1
    bindsym --release nop 2
    bindsym $mod + a nop 3
    bindsym --release $mod + a nop 4
    ```
   - $mod press => nop 1
   - $mod release => nop 2
   - $mod press => nop 1
   - a press => nop 3
   - a release => nop 4
   - $mod release => ???

orestisf1993:
> I have
>```
>mode "resize" {
>...
>    bindsym r --release exec --no-startup-id ~/bin/draw-resize.sh $(xdotool getactivewindow), mode "default"
>}
>bindsym $mod+r mode "resize"
>```
>in my config. The first time I $mod+r it works as intended. The second time I $mod+r the r `--release` bindsym gets triggered immediately. This does not happen w/ `next`. Need to investigate.

ha! This can happen in `next` too but with `bindcode --release`. Try:

```
# i3 config file (v4)
set $mod Mod4
mode "resize" {
    bindcode --release 27 mode "default"
}
bindsym $mod+r mode "resize"
```

Airblader:
@orestisf1993 What's the status here?

orestisf1993:
I've been running this patch for a while, seems fine. I think we can review this.

Airblader (approved):

PR #3185, opened 2018-03-18
Remove 'method' from cmd_move_window_to_position

orestisf1993:


orestisf1993:
This could break some workflows. Unless if you are talking about something like `move to relative position …`.

I wonder if the original intention was to actually be able to move the window out of bounds.

orestisf1993:
> Unless if you are talking about something like `move to relative position …`.

Even though this can be executed with `move [direction] <pixels> px` for floating windows.

Airblader:
> This could break some workflows

While that's true and unfortunate, I do think we should not keep broken behavior broken for the sake of not breaking workflows. The `absolute` flag should dictate whether the given coordinates are relative to the workspace or the root window. I do think it makes sense to have both properly implemented, given that the command syntax for it is already there. What do you think?

orestisf1993:
I don't think that `move position` was ever meant to support relative movement. Especially since directional commands for floating containers exist since 2009 but the `move position px px` was implemented in 2012 and tests also checked for absolute coordinates (f78f137ed00f4e31f4752f3e26a8e133809b9f00).
It is also not mentioned in the userguide:

```
# Moves the container either to a specific location
# or to the center of the screen. If 'absolute' is
# used, it is moved to the center of all outputs.
move [absolute] position <pos_x> [px] <pos_y> [px]
move [absolute] position center
```

`[absolute]` is mentioned only for `center` and it exists for `position <pos_x> ...` because of the parser:
1. move absolute position center
2. move position center
3. move position <pos_x> [px] <pos_y> [px]
4. Now `absolute` must be accepted by `position <pos_x> [px] <pos_y> [px]` for `absolute position center` to work.

Airblader:
Yeah, I get all that. But the semantics of `absolute` are clear from the command where it works, so a user might well expect it to work in the `position` case as well.

At the very least we should clearly point out in the userguide that `absolute` has no effect for `position` commands.

orestisf1993:
Updated the userguide.

docs/userguide
```diff
@@ -2030,10 +2030,13 @@ Use the +move+ command to move a container.
 # defaults to 10 pixels.
 move <left|right|down|up> [<px> px]
 
-# Moves the container either to a specific location
-# or to the center of the screen. If 'absolute' is
-# used, it is moved to the center of all outputs.
+# Moves the container to the specified pos_x and pos_y
+# coordinates on the screen.
```
Airblader: I actually meant we should specifically point out that the absolute flag does *nothing* here. On second thought, I'd prefer that we instead just remove the `[absolute]` from the command syntax. Yes, the parser accepts (and ignores it), but we just don't document it since it doesn't do anything.

orestisf1993:
up

PR #3186, opened 2018-03-18
i3-msg: only print input + errorposition if they exist

orestisf1993:
Before:
```
$ i3-msg floating disable, move window to position 100 px 100 px
ERROR: Your command: (null)
ERROR:               (null)
ERROR: Cannot change position of a window/container because it is not floating.
[{"success":true},{"success":false,"error":"Cannot change position of a window/container because it is not floating."}]
```

After:
```
$ i3-msg floating disable, move window to position 100 px 100 px
ERROR: Cannot change position of a window/container because it is not floating.
[{"success":true},{"success":false,"error":"Cannot change position of a window/container because it is not floating."}]
```

### DIFF
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -95,8 +95,10 @@ static int reply_start_map_cb(void *params) {
 
 static int reply_end_map_cb(void *params) {
     if (!last_reply.success) {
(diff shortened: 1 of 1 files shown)


Airblader:
Thanks as always! :-)

PR #3192, opened 2018-03-20
Fix potential memory leak

Exagone313:
`xkb_state_new` uses `calloc` and may fail in a rare case, which would cause a memory leak (for `dummy_state_*` variables).
Note that `xkb_state_unref` checks if the parameter given is not null `(!state)` before freeing.
Calls to `xkb_state_new` have been grouped to remove code duplication.

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -431,31 +431,20 @@ static void add_keycode_if_matches(struct xkb_keymap *keymap, xkb_keycode_t key,
  *
  */
 void translate_keysyms(void) {
-    struct xkb_state *dummy_state = xkb_state_new(xkb_keymap);
-    if (dummy_state == NULL) {
-        ELOG("Could not create XKB state, cannot translate keysyms.\n");
-        return;
-    }
-
-    struct xkb_state *dummy_state_no_shift = xkb_state_new(xkb_keymap);
-    if (dummy_state_no_shift == NULL) {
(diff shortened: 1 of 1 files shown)


src/bindings.c
```diff
@@ -431,32 +431,21 @@ static void add_keycode_if_matches(struct xkb_keymap *keymap, xkb_keycode_t key,
...
+        (dummy_state_no_shift = xkb_state_new(xkb_keymap)) == NULL ||
+        (dummy_state_numlock = xkb_state_new(xkb_keymap)) == NULL ||
+        (dummy_state_numlock_no_shift = xkb_state_new(xkb_keymap)) == NULL) {
         ELOG("Could not create XKB state, cannot translate keysyms.\n");
-        return;
+        goto out;
     }
 
-    bool has_errors = false;
-    Binding *bind;
```
orestisf1993: No need to reposition the `*bind` here, right?

Exagone313: Yeah, but I felt like leaving this variable alone wasn't right.
I wonder if it would be appreciated a complete reformat of the code with all variables on the top of functions for example ;)

Airblader: Declaring everything on top is a really outdated practice. We definitely prefer declaring variables as close to where they are initialized as possible.

Exagone313:
I updated it.

PR #3193, opened 2018-03-21
xcb_drag_prepare_cb: drain events

orestisf1993:
As discussed in PR #3085, X11 events can appear while `dragloop->callback()` is running.

I can "verify" the results here with some new tests for floating dragging.

It might also be worth investigating why ev_run -> ev_loop makes a difference.

### DIFF
--- a/src/floating.c
+++ b/src/floating.c
@@ -687,8 +687,7 @@ struct drag_x11_cb {
     const void *extra;
 };
 
-static void xcb_drag_prepare_cb(EV_P_ ev_prepare *w, int revents) {
-    struct drag_x11_cb *dragloop = (struct drag_x11_cb *)w->data;
+static bool drain_drag_events(EV_P, struct drag_x11_cb *dragloop) {
     xcb_motion_notify_event_t *last_motion_notify = NULL;
     xcb_generic_event_t *event;
 
@@ -749,12 +748,14 @@ static void xcb_drag_prepare_cb(EV_P_ ev_prepare *w, int revents) {
 
         if (dragloop->result != DRAGGING) {
             free(last_motion_notify);
-            return;
+            ev_break(EV_A_ EVBREAK_ONE);
+            return true;
         }
     }
 
-    if (last_motion_notify == NULL)
-        return;
+    if (last_motion_notify == NULL) {
+        return true;
+    }
 
     /* Ensure that we are either dragging the resize handle (con is NULL) or that the
      * container still exists. The latter might not be true, e.g., if the window closed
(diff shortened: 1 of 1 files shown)


Airblader:
To clarify, you are not yet done with the PR because you want to add tests still?

orestisf1993:
The tests will be for floating dragging that will work because of the change here. If I write them, I will create some convenience functions to be used by #3085 also.

Whether this happens inside this PR is up to you.

Another reason I don't really like this change here is that it is more "trial and error" (the `ev_loop` part) rather than understanding why it works.

Airblader:
I don't understand it either, but since @stapelberg and you both confirmed that it works better, I'd be fine with merging it for now. I'm still not entirely sure whether getting this merged as-is is your goal, though. :-)

orestisf1993:
@stapelberg hasn't tested (maybe not even seen?) the `ev_loop` part. The `drain` part is understandable: while `dragloop` is running more events are produced (written in the file descriptor that ev_io is watching), we don't catch them (because ev_loop is not running?).

I'll play with some floating drag tests and maybe write the tiling drag ones concurrently to see if the issue is present in a different situation.

src/floating.c
```diff
@@ -767,7 +769,8 @@ static void xcb_drag_prepare_cb(EV_P_ ev_prepare *w, int revents) {
             last_motion_notify->root_y,
             dragloop->extra);
     }
-    free(last_motion_notify);
+    FREE(last_motion_notify);
+    goto drain;
```
stapelberg: I think we can easily avoid the goto here by moving the code between the label and goto into a separate function:

```c
/*
 * drain_drag_events returns true when it is certain there are no more
 * pending events for the specified drag loop, false otherwise.
 */
static bool drain_drag_events(struct drag_x11_cb *dragloop) {
     xcb_motion_notify_event_t *last_motion_notify = NULL;
     xcb_generic_event_t *event;
    while ((event = xcb_poll_for_event(conn)) != NULL) {
        // …
        if (last_motion_notify != (xcb_motion_notify_event_t *)event) {
             free(event);
        }
 
         if (dragloop->result != DRAGGING) {
             free(last_motion_notify);
             return true;
         }
     }
 
     if (last_motion_notify == NULL) {
         return true;
    }
 
     /* Ensure that we are either dragging the resize handle (con is NULL) or that the
      * container still exists. The latter might not be true, e.g., if the window closed
      * for any reason while the user was dragging it. */
     if (!dragloop->con || con_exists(dragloop->con)) {
         dragloop->callback(
             dragloop->con,
             &(dragloop->old_rect),
             last_motion_notify->root_x,
             last_motion_notify->root_y,
             dragloop->extra);
     }
    FREE(last_motion_notify);
    return false;
}

static void xcb_drag_prepare_cb(EV_P_ ev_prepare *w, int revents) {
     struct drag_x11_cb *dragloop = (struct drag_x11_cb *)w->data;
     xcb_motion_notify_event_t *last_motion_notify = NULL;
     xcb_generic_event_t *event;
 
    while (!drain_drag_events(dragloop)) {
        /* repeatedly drain events: draining might produce additional ones */
    }
}
```

stapelberg: I don’t think it’s less effort — I’m pretty pressed on time right now.

I’d appreciate it if you could continue the PR :). Thanks!

orestisf1993:
Up.

stapelberg:
Thanks!

Issue #3071, opened 2017-12-08
Feature Request: Lower, px-valued, or configurable minimum container height


pevu:
Output of `i3 --moreversion 2>&- || i3 --version`:
```
Binary i3 version:  4.14.1 (2017-09-24) © 2009 Michael Stapelberg and contributors
(Getting version from running i3, press ctrl-c to abort…)
Running i3 version: 4.14.1 (2017-09-24) (pid 1898)
Loaded i3 config: /home/pao/.config/i3/config (Last modified: Do 30 Nov 2017 20:07:16 CET, 666190 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

The minimum container height is too large for monitors in portrait mode (its approx. twice as large as in landscape mode).

There's a bug related to this:

    Steps to reproduce
    1. create a vertical tiling container
    2. open 2 terminals in it (I used urxvt)
    3. make the height of the bottom terminal as small as possible
    4. focus the top container and open new terminals until the bottom container
       has a height of approx. 16px or lower
    5. try to resize any of the terminals vertically (this will fail)

Making the minimum height a fixed px-value (like 16px as an approximate height of one line of text) would probably include fixing above bug.

Airblader:
IIRC the minimum size is percentage-based, not an absolute value. Can you explain why you need windows that are even smaller? What are they useful for?

I'm hesitant to add an option for this because I don't see what it'd be useful for (yet).

pevu:
For a 1080p screen, in portrait mode the absolute screen height is 1920px. 5% of 1920 are 96px or 6 lines of 16px each. 6 lines of text is too large for a minimum window size, people shouldn't reach that limit that easily. For reference, the Windows 10 taskbar has a height of 40px .

I put monitors in portrait mode, because I like the extra vertical space. It's a bit of a bummer to see i3 waste it.

Airblader:
I see. I still don't think this requires an option, but instead we shouldn't have such an arbitrary size restriction. Perhaps it'd make sense to just reduce the restriction to zero. Do you want to submit a PR?

pevu:
I agree about removing the restriction altogether. There won't be a pull request from me for this though, at least not until I'm able to finish some ongoing personal projects.

Airblader:
No worries; PRs are of course welcome by anyone. I'm fairly certain the fix is trivial and only involves changing all the `0.05` in `src/commands.c` to `0.0`s, but I haven't tested it.

[...]

PR #3194, opened 2018-03-21
Updated [resize grow] to harvest from other windows

moodboom:
Updated [resize grow] to be able to harvest pct from all other windows in the same container

### DIFF
--- a/i3bar/src/ipc.c
+++ b/i3bar/src/ipc.c
@@ -309,7 +309,7 @@ int i3_send_msg(uint32_t type, const char *payload) {
     char *buffer = smalloc(to_write);
     char *walk = buffer;
 
-    strncpy(buffer, I3_IPC_MAGIC, strlen(I3_IPC_MAGIC));
+    memcpy(buffer, I3_IPC_MAGIC, strlen(I3_IPC_MAGIC));
     walk += strlen(I3_IPC_MAGIC);
     memcpy(walk, &len, sizeof(uint32_t));
     walk += sizeof(uint32_t);
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -604,7 +604,7 @@ void handle_button(xcb_button_press_event_t *event) {
 
     const size_t len = namelen + strlen("workspace \"\"") + 1;
     char *buffer = scalloc(len + num_quotes, 1);
-    strncpy(buffer, "workspace \"", strlen("workspace \""));
+    memcpy(buffer, "workspace \"", strlen("workspace \""));
     size_t inpos, outpos;
     for (inpos = 0, outpos = strlen("workspace \"");
          inpos < namelen;


moodboom:
I have a question about the best way to deal with [min_pct].  Can we make that a configuration instead of a hardcoded value?

Airblader:
I haven't yet looked at the PR, but what is min_pct for, anyway? From my original understanding of the proposal I don't see why we'd need something like that. It has always been about resizing other windows when one of the windows has reached its minimum size already.

I'd definitely prefer not having to add any configuration for this.

moodboom:
It looked to me like the original code bailed out if the request pushed the window into negative size.  So it would bail out randomly at some point where you WOULD have crossed into negative.  To get the resizing math right, I needed to know the actual precise preferred minimum size, which I assume would be something greater than zero.  The previous code always checked against zero.  Correct me if I'm wrong, that's my first take on it.

moodboom:
I'm not sure why Travis reported failure, the log looked okay to me, maybe I missed something?

orestisf1993:
Run `clang-format -i src/commands.c`

moodboom:
Ah ok thank you @orestisf1993.

Airblader:
We explicitly got rid of some minimum width threshold not too long ago because there's just no way to please both small and large screen users, whether you use absolute or relative values (sorry, just remembered this right now). The idea there was that we don't have a good reason to have configuration for it since we can just use 0%.

With this change, I guess we would have a reason to re-introduce a minimum width. But of course we can't just reintroduce the problem in #3071. The problem with making this configurable is that some users will want it absolute, others percentage-based and soon enough people will want it to be a per-window setting etc.

I don't yet have an answer of what I'd prefer to do.

moodboom:
Absolutely understood.  That's why I was wondering about adding it, it will be different for different use cases I would think.  Everyone sets their parameters of [resize grow ...] differently.  We could set it as an additional parameter to [resize grow], but then it will get messy if it is ever needed elsewhere as well.

moodboom:
What about optional configurations, that can specify a pct_min or a px_min, and if neither are provided we default to zero?

One problem with using zero is that I've seen apps (gnome-terminal, nautilus) crash and misbehave if they get that small.

src/commands.c
```diff
@@ -554,28 +565,70 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
-        if (child == current)
-            continue;
-        if (child->percent - subtract_percent <= 0.0) {
-            LOG("Not resizing, already at minimum size (child %p would end up with a size of %.f\n", child, child->percent - subtract_percent);
+    LOG("current->percent before = %f\n", current->percent);
+
+    const double min_pct = 0.05;
+
+    // Grow
+    if (ppt > 0.0) {
```
orestisf1993: `ppt` is an `int`, compare it to `0`.

src/commands.c
```diff
@@ -554,28 +565,70 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
-    LOG("subtract_percent = %f\n", subtract_percent);
-    /* Ensure that the new percentages are positive. */
-    TAILQ_FOREACH(child, &(current->parent->nodes_head), nodes) {
-        if (child == current)
-            continue;
-        if (child->percent - subtract_percent <= 0.0) {
-            LOG("Not resizing, already at minimum size (child %p would end up with a size of %.f\n", child, child->percent - subtract_percent);
+    LOG("current->percent before = %f\n", current->percent);
+
+    const double min_pct = 0.05;
```
orestisf1993: Alternative to the `min_pct` configuration conversation: set it to `fabs(ppt / 100.0 / children)`. It should behave similarly to the current behaviour.

moodboom:
I tested further, and a min_pct of zero crashes i3.  A small value (eg 0.0001) works fine, but again, some windows misbehave and don't recover if they get that small.  Not i3's fault really, but seems best to defend against it.

After reading previous discussions, I understand the desire to keep things simple.  Any extra complexity has side effect.  To keep things as minimal as possible, I could add one config, min_pct, that would be optional, and default to something small (eg: .01).

orestisfl (changes requested):
I generally like the new behaviour, thanks for the idea and the implementation.

I haven't reviewed this extensively.

src/commands.c
```diff
@@ -528,6 +528,17 @@ static bool cmd_resize_tiling_direction(I3_CMD, Con *current, const char *way, c
     return true;
 }
 
+static int sort_cons_by_smallest_percent_cmp(const void *a, const void *b) {
```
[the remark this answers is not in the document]
moodboom: Sure thing, I changed the name.  Wasn't sure about the second comment...?

src/commands.c
```diff
@@ -528,6 +528,17 @@ static bool cmd_resize_tiling_direction(I3_CMD, Con *current, const char *way, c
...
+    Con *first = *((Con **)a);
+    Con *second = *((Con **)b);
+    if (first->percent < second->percent) {
```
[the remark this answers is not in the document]
moodboom: Done

src/commands.c
moodboom: Done

src/commands.c
```diff
@@ -554,28 +565,70 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
-    TAILQ_FOREACH(child, &(current->parent->nodes_head), nodes) {
-        if (child == current)
-            continue;
-        if (child->percent - subtract_percent <= 0.0) {
-            LOG("Not resizing, already at minimum size (child %p would end up with a size of %.f\n", child, child->percent - subtract_percent);
+    LOG("current->percent before = %f\n", current->percent);
+
+    const double min_pct = 0.05;
+
+    // Grow
```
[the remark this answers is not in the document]
moodboom: Done

src/commands.c
moodboom: Good one, done.  That means we can delay the decision about the configuration (although I vote yes fwiw :smile: ).

moodboom (commented):
Thanks for the improvements!

moodboom:
Note that with 

`    double min_pct = fabs(ppt / 100.0 / children);`

my windows get quite small and I see the misbehavior I've mentioned when windows get extremely small.

I'm still hoping we can add a min_pct configuration, but we can go with whatever you guys think best.

orestisf1993:
Does this happen on `next` too?

src/commands.c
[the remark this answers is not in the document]
orestisf1993: I'll leave a comment elsewhere.

src/commands.c
```diff
@@ -554,28 +559,71 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
-        if (child->percent - subtract_percent <= 0.0) {
-            LOG("Not resizing, already at minimum size (child %p would end up with a size of %.f\n", child, child->percent - subtract_percent);
+    LOG("current->percent before = %f\n", current->percent);
+
+    /* Consider making this configurable. */
+    double min_pct = fabs(ppt / 100.0 / children);
+
+    /* Grow */
+    if (ppt > 0) {
+        /* Sort cons by percent */
```
orestisf1993: You can mention here that the sort is in descending order since a reader would have to figure it out by reading the compare function and consulting the `qsort` manpage if they don't remember the `*compar` function details.

moodboom:
To be honest, I have been juggling between i3 next and i3 gaps gaps-next, and mostly testing with gaps-next.  Shall I try in next, if you think that may be subtly different?

src/commands.c
moodboom: Ah ok will do!

orestisf1993:
You can run i3 inside Xephyr, allowing you to change branches, make changes etc painlessly.
I have a script [here](https://github.com/orestisf1993/dotfiles/blob/b49a8f70c3578dff4f186bcf5d65adb9b97b6501/xorg/start-xephyr.sh) if you think it can help you.

I don't know the exact changes of i3-gaps but it would be nice to consider not breaking the i3-gaps users' workflow.

Anyway, do your windows get problematically small when you resize them on either `next` branch currently?

moodboom:
image
Yes it will shrink even smaller than shown, and apps like gnome-terminal do not recover and redraw correctly afterwards.

moodboom:
I should mention that I bind grow and shrink directly to key movements, and click them quickly to resize, so I use a very small growth percent of 3%.

orestisf1993:
Yeah, I am not sure about the current (`next`) behaviour either:

2018-03-22-012043_1366x729_escrotum

There are 3 terminals here, it gets buggy.

Actually, `double min_pct = fabs(ppt / 100.0 / children)` is not equivalent with the `next` behaviour after all. Containers can get 0 percent right now.

moodboom:
So I'm back to leaning towards a min_pct config.  What do you guys think?

orestisf1993:
Eg from emacs crash:

```
X protocol error: BadDrawable (invalid Pixmap or Window parameter) on protocol request 70
When compiled with GTK, Emacs cannot recover from X disconnects.
This is a GTK bug: https://bugzilla.gnome.org/show_bug.cgi?id=85715
For details, see etc/PROBLEMS.
Fatal error 6: Aborted
```

moodboom:
Yep it appears I need to prevent really small sizes.  I would think there are cases now where i3 allows a nearly-minimum-size window to be shrunken and those problems would occur, right?

moodboom:
What if I used a minimum pct min_pct equal to the resize amount?  Does that seem reasonable in all cases?

Airblader:
Just as an explanation, when we got rid of the minimum size the argument was that all we need to prevent are zero sizes (which may rightfully break clients), but anything above zero is up to the user: generally, they wouldn't want to resize a window down to, say, one pixel anyway. And a client breaking permanently after being resized to a small but positive size is something where I'd consider the client to be broken; we cannot possibly accommodate that in the window manager.

I do like the suggestion of using the resize percentage as the minimum width (in some way or another) because I'd still like to avoid adding configuration for this. Can we go with that for now?

I haven't yet had time to look at the PR itself.

moodboom:
Sure that sounds good.  That's how it is coded at the moment, and is working well for me for now, let me know what you guys think.

src/commands.c
```diff
@@ -554,28 +559,70 @@ static bool cmd_resize_tiling_width_height(I3_CMD, Con *current, const char *way
...
-    LOG("subtract_percent = %f\n", subtract_percent);
-    /* Ensure that the new percentages are positive. */
-    TAILQ_FOREACH(child, &(current->parent->nodes_head), nodes) {
-        if (child == current)
-            continue;
-        if (child->percent - subtract_percent <= 0.0) {
-            LOG("Not resizing, already at minimum size (child %p would end up with a size of %.f\n", child, child->percent - subtract_percent);
+    LOG("current->percent before = %f\n", current->percent);
+
+    double min_pct = fabs(ppt / 100.0);
```
orestisf1993: Now you can't resize for more than 50% right?

orestisf1993: And the tests fail. Because of it.

moodboom: Ah ok, thanks.  I'll take a look at that use case when I can.

moodboom:
If you guys still want an automatic min_size, how about using the resize ppt, with validation that clips low and high values?  No less than .005, no more than .03.

Or I can add an official min_pct config, with a reasonable default of 5%.  In my opinion, it seems to be a logical thing to make configurable in a window manager that supports so many possible screen layouts.  Is there a use case that would be negatively impacted by such a configuration?

moodboom:
Sorry, coding on phone is not a good idea.  :-)  I'll work on this later.

moodboom:
Are the test failures due to the tests expecting old behavior?  It seems like that may be the case, but maybe you all are more familiar with the test suite and can help me with that.

Resizing seem to be working well enough here considering the constraints.  I think I've gotten min_pct as "automatic" as I can.  A really large ppt grow will only bring the target to take up an equal share of the screen, not optimal but not horrible I guess.  It still seems inferior to adding configuration.  I'll wait for further feedback.  Thanks all.

orestisf1993:
No, the test should work. The failing test opens 4 windows and grows one of them to 50% with `resize grow width 10 px or 25 ppt`.

moodboom:
Correct me if I'm wrong, but I believe that the test will fail then with the new math.  min_pct will be set to fmax(.03,fmin(.25,fabs(.25))) = .25.  The "other" windows will not shrink as much as the test expects.  But it's doing what is expected with the new math.

As I mentioned, I'm not thrilled with the automatic behavior in all cases.

orestisf1993:
I mean, the use case the test represents should still be a valid one. I don't agree with the new formula, kinda restrictive don't you think?

moodboom:
Yes I agree it is.  I'd like to add the config.  What are your thoughts on that?

orestisf1993:
I'd like to give some time for Ingo or Michael to respond.

moodboom:
Sounds good, thanks @orestisf1993

moodboom:
@orestisf1993 maybe this auto min_pct is a little better, it's closer to your suggestion.

Airblader:
I'm still a bit undecided on it, to be honest. Out of curiosity, why is there the restriction to 0.03 in there? The other values can't ever reach zero, anyway, right?

Maybe a possible way forward would be getting some people to run this patch for a while to see how well it works out in everyday usage for different people; e.g., by asking on the mailing list and the subreddit. I'd like to rule out making things worse for some people here.

moodboom:
Thanks @Airblader will do.  I'll post a request to the subreddit and the mailing list.

The 0.03 is in there because if in general a window gets too small, things generally act up.  The previous approach of failing the grow if anything went negative was restrictive enough to prevent that in many cases.

I can remove it but it will more often result in very small sizes, if the grow ppt is small or there are a large number of small windows.  What would you prefer?

In my usage, I have found that a single user-modifiable min_pct config value is more precise than the auto behavior.

Airblader:
> I can remove it but it will more often result in very small sizes, if the grow ppt is small or there are a large number of small windows. What would you prefer?

I'd prefer removing it for now for the test and see how well it works without. The problem with a 0.03 is exactly the reason we removed the 5% limit we had previously: people have very differently sized monitors and there's just no single fixed number that fits everyone. And with the 0.03, we are practically re-introducing this limit since it'd be >= 0.03.

> In my usage, I have found that a single user-modifiable min_pct config value is more precise than the auto behavior.

Yes, and I really get understand that a configuration for this seems to make it easier. However, more configuration also means more complexity, which is why we try to avoid making things configurable that don't need to be configurable. If we find that there is no way around it, we can consider it; I'm not yet convinced that we can't find a good way to do it automatically. :-) Also, adding a configuration later on in a second step is easy, but removing one we didn't need to begin with is a breaking change.

moodboom:
@Airblader sounds great, thanks again!

moodboom:
Quick update: 

- The [request for feedback](https://redd.it/86nvzr) has 10 upvotes but no comments.
- I've been using it successfully here for three weeks with no issues, using a 6% minimum size.

moodboom:
Rebased onto latest i3 next commits.

moodboom:
UPDATE: This continues to perform well for me at 6% minimum size.  I have found that providing a specific min_pct configuration value is very helpful to make daily layout management more efficient, and prevent the current issues users are having with terminals crashing due to becoming (uselessly) small.

orestisf1993:
What if we ensured that each container will have at least 1 pixel? Kinda like `px_resize_to_percent`.

moodboom:
@orestisf1993 we could ensure 1 pix, that would help prevent i3-related crashes without adding any new configuration.  

But it would not address the following, which I feel are important:
* the window manager will easily allow windows to become "unuseably small", which makes the UI unfriendly
* many applications crash when made "unuseably small"; this problem will likely persist for the near future

I could set up the 1-pixel-minimum limit to get this pull request approved, and tackle these other issues later, if you like.

orestisf1993:
> I could set up the 1-pixel-minimum limit to get this pull request approved, and tackle these other issues later, if you like.

I think we could discuss this in a separate issue / PR and adjust the other resize functions as well.

Getting the final pixel size for split containers from `con->percent` isn't straightforward. I've added some code recently that dealt with this, see #3372.

For 1 pixel you'll need:
```
const orientation_t o = con_orientation(parent);
const int total = (o == HORIZ ? parent->rect.width : parent->rect.height);
const int target = (o == HORIZ ? 1 : 1 + con->deco_rect.height);
if (con->percent < (double)target / (double)total) {
    /* Less than 1px */
}
```

PR #3196, opened 2018-03-23
_con_move_to_con cleanup

orestisf1993:
See ~~review~~ commit by ~~review~~ commit

### DIFF
--- a/src/con.c
+++ b/src/con.c
@@ -1170,20 +1170,6 @@ static bool _con_move_to_con(Con *con, Con *target, bool behind_focused, bool fi
             floating_fix_coordinates(con, &(source_output->rect), &(dest_output->rect));
         } else
             DLOG("Not fixing coordinates, fix_coordinates flag = %d\n", fix_coordinates);
-
-        /* If moving to a visible workspace, call show so it can be considered
-         * focused. Must do before attaching because workspace_show checks to see
-         * if focused container is in its area. */
-        if (!ignore_focus && workspace_is_visible(target_ws)) {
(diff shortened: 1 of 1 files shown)


src/con.c
```diff
@@ -1170,20 +1170,6 @@ static bool _con_move_to_con(Con *con, Con *target, bool behind_focused, bool fi
             floating_fix_coordinates(con, &(source_output->rect), &(dest_output->rect));
         } else
             DLOG("Not fixing coordinates, fix_coordinates flag = %d\n", fix_coordinates);
-
-        /* If moving to a visible workspace, call show so it can be considered
-         * focused. Must do before attaching because workspace_show checks to see
```
Airblader: The commit message doesn't give an explanation for why the comment here can be ignored. The workspace_show later on happens after re-attaching.

orestisf1993: I've been trying but I don't understand the comment and how it justifies the _show call here.

It also causes a small bug:
- Open a window in an empty workspace
- Switch to another workspace
- seturgent to the first window
- Move another window to the first workspace
- Urgent flag is now reset

Airblader: Alright. I tried to make sense of this for a bit now as well and suspect it might just be a leftover from older times. We'll see if it causes issues.

orestisf1993: It was added in d8cf36ce83d2d9d6653ff08b1b2308c6f1bf1acd

Airblader: Yeah, that was a long time ago and the code looked quite a bit different back then, hence my assumption. :-)

PR #3200, opened 2018-03-25
_con_move_to_con: don't change focus when moving to active workspace

orestisf1993:


orestisf1993:
Ah this proved harder than it looked. I'll look into the failing tests.

orestisf1993:
Not a super pretty solution but _con_move_to_con is used / con_move_to_workspace is used in a bunch of places.

Airblader:
Looks like Travis had a hiccup there. Can you just push some change + push undoing that again so re-trigger Travis?

orestisf1993:
up

PR #3204, opened 2018-03-25
_con_move_to_con: focus_next isn't always con_next_focused(con)

orestisf1993:
`con_next_focused` uses con's parent. But since `con` can be inside an unfocused container this means that one of it's siblings could become focused in the current workspace.

There is another small issue with `focus_next`: since we descend when focusing later, moving an unrelated container while a parent container is selected will discard the user's selection.


orestisf1993:
(sorry if the requested review sends an extra notification, it was a missclick)

Airblader:
Don't worry about it :-) I usually instantly delete all but one email regarding the same issue anyway.

PR #3207, opened 2018-03-27
Introduce get_workspace_by_* functions

orestisf1993:
Could also be get_**existing**\_workspace_by_*


Airblader:
Yeah, let's rename them, please. Otherwise it's unclear whether they implicitly create the workspace.

PR #3210, opened 2018-03-29
add --modifier flag to i3-config-wizard (#3136)

### LINKED ISSUE #3136: Could i3-config-wizard be made scriptable?
TL;DR: would it be possible to add a `--meta=win` or `--meta=alt` option flag to **i3-config-wizard** and write a config without any user graphical interaction being required?

The use case of this enhancement is fairly specific: I'm using a provisioning tool (in my case Ansible) to bootstrap user workstations, and I want to deploy -- without user interaction --- a basic I3 config file, which will then be further modified by the provisioning tool to create the site-wide base configuration for I3.

I know the user's keymap choice and meta key preference before the Ansible playbook runs, but the actual provisioning is done headless, so I cannot and do not want to hand control over to the i3-config-wizard part way through the run. I do, however, want the benefits of a keymap-aware I3 config created as if the user had run i3-config-wizard and manually selected their preferred meta key. Right now my only choice seems to be build a template for every possible keymap and have Ansible select and copy over the right one, but this is really quite a painful choice.

Since the only actual interaction the i3-config-wizard provides is a boolean choice of whether or not to create the config, and if that is True a choice of meta key, it doesn't seem that it would be too difficult to allow something like:

`i3-config-wizard --meta win`

And have it assume that the user does indeed want to write a config appropriate for their meta key choice, and write that without going through the trouble of ever opening a window or grabbing the keyboard.

I'd write up a pull request, but frankly my C skills are so rusty I'm sure I'd just embarass myself.

joonatoona:
The --modifier flag accepts either alt or win, and will generate the
configuration file without opening a window

Fixes #3136.

### DIFF
--- a/i3-config-wizard/main.c
+++ b/i3-config-wizard/main.c
@@ -745,10 +745,12 @@ int main(int argc, char *argv[]) {
     char *pattern = "pango:monospace 8";
     char *patternbold = "pango:monospace bold 8";
     int o, option_index = 0;
+    bool headless_run = false;
 
     static struct option long_options[] = {
         {"socket", required_argument, 0, 's'},
         {"version", no_argument, 0, 'v'},
+        {"modifier", required_argument, 0, 'm'},
         {"limit", required_argument, 0, 'l'},
         {"prompt", required_argument, 0, 'P'},
         {"prefix", required_argument, 0, 'p'},
@@ -767,9 +769,18 @@ int main(int argc, char *argv[]) {
             case 'v':
                 printf("i3-config-wizard " I3_VERSION "\n");
                 return 0;
+            case 'm':
+                headless_run = true;
+                if (strcmp(optarg, "alt") == 0)
+                    modifier = MOD_Mod1;
+                else if (strcmp(optarg, "win") == 0)
+                    modifier = MOD_Mod4;
+                else
+                    err(EXIT_FAILURE, "Invalid modifier key %s", optarg);
+                break;
             case 'h':
                 printf("i3-config-wizard " I3_VERSION "\n");
-                printf("i3-config-wizard [-s <socket>] [-v]\n");
+                printf("i3-config-wizard [-s <socket>] [-v] [--modifier <modifier key>]\n");
                 return 0;
         }
     }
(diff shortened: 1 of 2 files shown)


i3-config-wizard/main.c
```diff
@@ -767,9 +769,18 @@ int main(int argc, char *argv[]) {
...
+                    modifier = MOD_Mod1;
+                else if (strcmp(optarg, "win") == 0)
+                    modifier = MOD_Mod4;
+                else
+                    err(EXIT_FAILURE, "Invalid modifier key %s", optarg);
+                break;
             case 'h':
                 printf("i3-config-wizard " I3_VERSION "\n");
-                printf("i3-config-wizard [-s <socket>] [-v]\n");
+                printf("i3-config-wizard [-s <socket>] [-v] [--modifier <modifier key>]\n");
```
orestisf1993: I think this should be `-m` for consistency.

Also, @Airblader:
- Should we write about this in the man page?
- Aren't some of the `long_options` above unused?

Airblader: > Should we write about this in the man page?

Yes, absolutely. :-)

> Aren't some of the long_options above unused?

Yeah, seems like it.

i3-config-wizard/main.c
```diff
@@ -767,9 +769,18 @@ int main(int argc, char *argv[]) {
...
+                    modifier = MOD_Mod1;
+                else if (strcmp(optarg, "win") == 0)
+                    modifier = MOD_Mod4;
+                else
+                    err(EXIT_FAILURE, "Invalid modifier key %s", optarg);
+                break;
             case 'h':
                 printf("i3-config-wizard " I3_VERSION "\n");
-                printf("i3-config-wizard [-s <socket>] [-v]\n");
+                printf("i3-config-wizard [-s <socket>] [-v] [-m <modifier key>]\n");
```
orestisf1993: `printf("i3-config-wizard [-s <socket>] [-v] [-m win|alt]\n");`, sorry didn't mention it.

See if you can also update the man page: https://github.com/i3/i3/blob/next/man/i3-config-wizard.man eg for reference: https://github.com/i3/i3/blob/next/man/i3bar.man

joonatoona: Not entirely sure if I got what `-s` does correct?

orestisf1993:
@Airblader can I use the github "Squash and Merge" feature or should the author squash the commits manually?

Airblader:
We typically ask the author to do this, but I don't see a reason for why you can't use Github's feature if you want to do so. The other day on i3status I even resolved merge conflicts myself. :-)

PR #3215, opened 2018-03-31
Small window decoration scrolling bugfix

orestisf1993:


Airblader:
Change looks good and I really appreciate your commit messages. Should we add a test for this?

orestisf1993:
Added the tests

Issue #3208, opened 2018-03-27
workspace <n> output <direction> can crash i3


orestisf1993:
Output of `i3 --moreversion 2>&- || i3 --version`:
```
Binary i3 version:  4.15-71-g45be56be (2018-03-27, branch "next") © 2009 Michael Stapelberg and contributors
```

URL to a logfile as per https://i3wm.org/docs/debugging.html:

https://logs.i3wm.org/logs/5678957424803840.bz2

**What I did:**
```
# i3 config file (v4)

workspace 5 output left
# I guess it should work with real outputs as well
fake-outputs 800x500+0+0,800x500+800+0,800x500+800+500,800x500+0+500
```
Run `i3-msg workspace 2, rename workspace 1 to 5`

**What I saw:**
```
==5487==ERROR: AddressSanitizer: heap-use-after-free on address 0x614000001e9c at pc 0x55556eb52de7 bp 0x7ffe742200c0 sp 0x7ffe742200b0
READ of size 4 at 0x614000001e9c thread T0
    #0 0x55556eb52de6 in con_get_workspace ../../i3/src/con.c:421
    #1 0x55556eb4380a in cmd_rename_workspace ../../i3/src/commands.c:2012
    #2 0x55556eb46957 in GENERATED_call parser/GENERATED_command_call.h:22
    #3 0x55556eb4a3aa in next_state ../../i3/src/commands_parser.c:185
    #4 0x55556eb4c4c4 in parse_command ../../i3/src/commands_parser.c:346
    #5 0x55556eba0ebf in handle_run_command ../../i3/src/ipc.c:124
    #6 0x55556ebac906 in ipc_receive_message ../../i3/src/ipc.c:1248
    #7 0x7f73d9fa44e2 in ev_invoke_pending (/usr/lib/libev.so.4+0x44e2)
    #8 0x7f73d9fa7ad5 in ev_run (/usr/lib/libev.so.4+0x7ad5)
    #9 0x55556ebb7a3f in ev_loop /usr/local/include/ev.h:835
    #10 0x55556ebc3495 in main ../../i3/src/main.c:964
    #11 0x7f73d8978f49 in __libc_start_main (/usr/lib/libc.so.6+0x20f49)
    #12 0x55556eb01749 in _start (/home/orestis/Documents/programming/i3/build/i3+0x189749)

0x614000001e9c is located 92 bytes inside of 448-byte region [0x614000001e40,0x614000002000)
freed by thread T0 here:
[...]

PR #3221, opened 2018-04-04
Issue 3208

ograff:
These two commits fix and test the issue of interpreting output assignments as directions when renaming a workspace.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1128,7 +1128,21 @@ void cmd_move_workspace_to_output(I3_CMD, const char *name) {
             continue;
         }
 
-        bool success = workspace_move_to_output(ws, name);
+        Output *current_output = get_output_for_con(ws);
+        if (current_output == NULL) {
+            ELOG("Cannot get current output. This is a bug in i3.\n");
+            ysuccess(false);
+            return;
+        }
+
+        Output *target_output = get_output_from_string(current_output, name);
+        if (!target_output) {
+            ELOG("Could not get output from string \"%s\"\n", name);
+            ysuccess(false);
+            return;
+        }
+
+        bool success = workspace_move_to_output(ws, target_output);
         if (!success) {
             ELOG("Failed to move workspace to output.\n");
             ysuccess(false);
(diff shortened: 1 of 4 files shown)


ograff:
The test I added passes with this fix, and fails if this fix is not included, evidenced by https://travis-ci.org/ograff/i3/jobs/361975157

Airblader:
Awesome, thanks! It looks good to me. @orestisf1993 What do you think?

If we had to use the function in more places it would make sense to once again refactor finding the output into a function (which then calls the function taking the parsed output), but we can worry about that when we come to that scenario.

src/commands.c
```diff
@@ -1990,7 +2004,12 @@ void cmd_rename_workspace(I3_CMD, const char *old_name, const char *new_name) {
             continue;
         }
 
-        workspace_move_to_output(workspace, assignment->output);
+        Output *target_output = get_output_by_name(assignment->output, true);
+        if (!target_output) {
+            ELOG("Could not get output named \"%s\"\n", assignment->output);
```
orestisf1993: This is not an error IMO. Even if it was printed with `workspace_move_to_output`. LOG should be fine.

The other `get_output_from_string` failure above (in `cmd_move_workspace_to_output`) is fine as an error though, since the output was given by the user as a command.

testcases/t/522-rename-assigned-workspace.t
```diff
@@ -82,4 +83,13 @@ cmd 'rename workspace to baz';
 is(get_output_for_workspace('baz'), 'fake-1',
     'Renaming the workspace to a number and name should move it to the assigned output');
 
+#######################################################################################################
+# Renaming a workspace so that it is assigned a directional output does not move the workspace or crash
+#######################################################################################################
```
orestisf1993: Add a newline here for consistency :)

orestisf1993:
I think we can close #3208 with this PR and move the rest of the problems mentioned in a new issue.

@ograff do you feel like opening a new issue describing the other problems you found?

ograff:
Follow up issue located at #3228.

orestisf1993:
Thank you!

PR #3223, opened 2018-04-04
Minor changes

orestisf1993:


Airblader:
I understand the idea behind using the compiler information, but it requires this to actually be checked somehow. Which I'd only want to do if we can do this automatically (in Travis). Can we get the compiler to error out on this warning in Travis?

orestisf1993:
Travis compiles with Werror. I actually made a mistake before and the build
failed. I can push a fake commit to double check that it fails.

orestisf1993:
Here: https://travis-ci.org/i3/i3/builds/362505860

orestisf1993:
Unfortunately gcc is pretty old there, are there any plans to upgrade?

Airblader:
Thanks!

I think as long as it fails with clang it's good enough for now. If we can update gcc without any issues, I don't mind doing it.

orestisf1993:
Well, updating gcc shouldn't introduce any issues. I have no idea how travis works though. Is this something that only Michael can do?

Airblader:
No, it's driven by the `.travis.yml` file. Likely the issue is just that the Travis environment has no newer version by default.

PR #3224, opened 2018-04-04
Improve scratchpad focus handling on focus request

### LINKED ISSUE #3218: Scratchpad windows can't give themself focus
I'm using 'WhatsApp Web' in a scratchpad window. WhatsApp Web can let Firefox send push notifications when someone messages you, which you can 'activatie' via dunst or alike.

The issue is that i3 ignores these activations, as can be seen here: https://github.com/i3/i3/blob/master/src/handlers.c#L783

It would be nice if this behaviour was configurable just like when 'normal' windows ask for focus.

HarmtH:
Closes #3218 

This also sets the urgency hint when a scratchpad window request focus and focus behavior is set to 'smart' or 'urgent'. That will become more useful when #628 gets attention.

### DIFF
--- a/include/configuration.h
+++ b/include/configuration.h
@@ -89,6 +89,23 @@ struct Mode {
 };
 
 /**
+ * Defines the focus behavior when a _NET_ACTIVE_WINDOW client message is
+ * received for a window.
+ * Used by the config directive focus_on_window_activation.
+ *
+ */
+typedef enum {
+    /* Focus if the target workspace is visible, set urgency hint otherwise. */
+    FOWA_SMART,
+    /* Always set the urgency hint. */
+    FOWA_URGENT,
+    /* Always focus the window. */
+    FOWA_FOCUS,
+    /* Ignore the request (no focus, no urgency hint). */
+    FOWA_NONE
+} focus_on_window_activation_t;
+
+/**
  * Holds part of the configuration (the part which is not already in dedicated
  * structures in include/data.h).
  *
@@ -185,17 +202,9 @@ struct Config {
      * flag can be delayed using an urgency timer. */
     float workspace_urgency_timer;
 
-    /** Behavior when a window sends a NET_ACTIVE_WINDOW message. */
-    enum {
-        /* Focus if the target workspace is visible, set urgency hint otherwise. */
-        FOWA_SMART,
-        /* Always set the urgency hint. */
-        FOWA_URGENT,
-        /* Always focus the window. */
-        FOWA_FOCUS,
-        /* Ignore the request (no focus, no urgency hint). */
-        FOWA_NONE
-    } focus_on_window_activation;
+    /** Defines the focus behavior when a _NET_ACTIVE_WINDOW client message is
+     * received for a window. */
+    focus_on_window_activation_t focus_on_window_activation;
 
     /** Specifies whether or not marks should be displayed in the window
      * decoration. Marks starting with a "_" will be ignored either way. */
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -764,38 +764,50 @@ static void handle_client_message(xcb_client_message_event_t *event) {
             return;
         }
 
-        /* data32[0] indicates the source of the request (application or pager) */
-        if (event->data.data32[0] == 2) {
-            /* Always focus the con if it is from a pager, because this is most
-             * likely from some user action */
-            DLOG("This request came from a pager. Focusing con = %p\n", con);
+        focus_on_window_activation_t behavior = config.focus_on_window_activation;
 
-            if (con_is_internal(ws)) {
-                scratchpad_show(con);
+        /* Handle smart focus */
+        if (behavior == FOWA_SMART) {
+            if (!con_is_internal(ws) && workspace_is_visible(ws)) {
+                behavior = FOWA_FOCUS;
             } else {
-                workspace_show(ws);
-                /* Re-set focus, even if unchanged from i3’s perspective. */
-                focused_id = XCB_NONE;
-                con_activate(con);
+                behavior = FOWA_URGENT;
             }
-        } else {
-            /* Request is from an application. */
-            if (con_is_internal(ws)) {
-                DLOG("Ignoring request to make con = %p active because it's on an internal workspace.\n", con);
-                return;
-            }
-
-            if (config.focus_on_window_activation == FOWA_FOCUS || (config.focus_on_window_activation == FOWA_SMART && workspace_is_visible(ws))) {
-                DLOG("Focusing con = %p\n", con);
-                workspace_show(ws);
-                con_activate(con);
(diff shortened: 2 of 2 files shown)


src/handlers.c
```diff
@@ -764,38 +764,47 @@ static void handle_client_message(xcb_client_message_event_t *event) {
...
-            /* Always focus the con if it is from a pager, because this is most
-             * likely from some user action */
-            DLOG("This request came from a pager. Focusing con = %p\n", con);
+        enum focus_behavior_t bhv = config.focus_on_window_activation;
 
-            if (con_is_internal(ws)) {
-                scratchpad_show(con);
+        /* Handle smart focus */
+        if (bhv == FOWA_SMART) {
+            if (!con_is_internal(ws) && workspace_is_visible(ws)) {
```
HarmtH: It seemed ```workspace_is_visible()``` always returned true for scratchpad windows, so I placed an extra check.

src/handlers.c
```diff
@@ -764,38 +764,47 @@ static void handle_client_message(xcb_client_message_event_t *event) {
             return;
         }
 
-        /* data32[0] indicates the source of the request (application or pager) */
```
orestisf1993: I feel that this comment should be kept, no?

HarmtH: There's a comment above it, and a log message below it that data32[0]==2 indicates a pager. Don't you think that's clear enough?

orestisf1993: I feel that it doesn't explain how we reach the assumption that this is a pager:

"data[0] is the source indication, 2 means pager which indicates a user action, https://standards.freedesktop.org/wm-spec/latest/ar01s03.html#idm139870830075168" I feel that this is more informative.

HarmtH: I've added the comment back and cleared it up a little bit.

include/configuration.h
```diff
@@ -186,7 +186,7 @@ struct Config {
     float workspace_urgency_timer;
 
     /** Behavior when a window sends a NET_ACTIVE_WINDOW message. */
-    enum {
+    enum focus_behavior_t {
```
Airblader: We don't usually add enum names, and certainly not a different name than the type. Please use the type name, possible renaming it to `focus_on_window_activation_t` for naming consistency.

src/handlers.c
```diff
@@ -764,38 +764,49 @@ static void handle_client_message(xcb_client_message_event_t *event) {
...
-        if (event->data.data32[0] == 2) {
-            /* Always focus the con if it is from a pager, because this is most
-             * likely from some user action */
-            DLOG("This request came from a pager. Focusing con = %p\n", con);
+        enum focus_behavior_t bhv = config.focus_on_window_activation;
```
Airblader: No need for `enum` with the change I mentioned above.

src/handlers.c
```diff
@@ -764,38 +764,49 @@ static void handle_client_message(xcb_client_message_event_t *event) {
...
+            case FOWA_URGENT:
                 DLOG("Marking con = %p urgent\n", con);
                 con_set_urgency(con, true);
-            } else
+                break;
+            case FOWA_NONE:
                 DLOG("Ignoring request for con = %p.\n", con);
+                break;
+            default:
+                DLOG("Unknown focus behavior requested.\n");
```
Airblader: Please use `assert(false)` here and add our usual `This is a bug in i3` to the log message.

include/configuration.h
HarmtH: What do you mean exactly? Add a typedef?

Airblader: Essentially, yes.

src/handlers.c
orestisf1993: Sorry for the back and forth, but just remove the `default:` case all together, see #3223. Add a case:
```
FOWA_SMART:
    assert(false);
```
since we change bhv to FOWA_FOCUS or FOWA_URGENT

src/handlers.c
```diff
@@ -764,38 +764,49 @@ static void handle_client_message(xcb_client_message_event_t *event) {
...
+            DLOG("Request indicates it comes from pager or direct user action.\n");
+            bhv = FOWA_FOCUS;
+            /* Re-set focus, even if unchanged from i3’s perspective. */
+            focused_id = XCB_NONE;
+        }
+
+        /* Set focus */
+        switch (bhv) {
+            case FOWA_SMART:
+                assert(false);
```
Airblader: Sorry, we should've been more clear here. We should not `assert(false)` without a log, so adding

    ELOG("Unhandled case for focus_on_window_activation smart. This is a bug in i3.\n");

would be good. Thanks!

src/handlers.c
```diff
@@ -764,38 +764,49 @@ static void handle_client_message(xcb_client_message_event_t *event) {
             return;
         }
 
-        /* data32[0] indicates the source of the request (application or pager) */
-        if (event->data.data32[0] == 2) {
-            /* Always focus the con if it is from a pager, because this is most
-             * likely from some user action */
-            DLOG("This request came from a pager. Focusing con = %p\n", con);
+        focus_on_window_activation_t bhv = config.focus_on_window_activation;
```
Airblader: `bhv` is now a really weird name for this. Let's just name it something more meaningful, like `behavior` (or similar). Words are better than acronyms.

include/configuration.h
```diff
@@ -88,6 +88,22 @@ struct Mode {
     modes;
 };
 
+/**
+ * Behavior when a window sends a NET_ACTIVE_WINDOW message.
```
Airblader: The atom is called `_NET_ACTIVE_WINDOW` (leading underscore), and they're called client messages (not just messages). They also don't have to be sent *by* the window and I don't like the second part that suggests some kind of syntax. I'd rather use something like

> Defines the focus behavior when a _NET_ACTIVE_WINDOW client message is received for a window.
> Used by the config directive focus_on_window_activation.

Airblader:
You added a comment to the enum, so I have a comment for that. Sorry! :-) Please also do write a comment here when you made the change so we get notified. Thanks again!

HarmtH:
Changed it :-)

include/configuration.h
HarmtH: I know, was just trying to be consistent with the other comments/style. Anyway, I've updated it (y)

Airblader:
Sorry, one more thing; we already have some testcases for this (`testcases/t/240-focus-on-window-activation.t`). It'd be great if we could cover this with a test as well. It should be fairly simple to write the testcase and you only have to extend the function defined in the testcase to allow specifying the source of the request.

Airblader:
@HarmtH Are you going to continue working on this? I think it'd be really cool to get this through the door, especially considering the work that already went into it. :-)

HarmtH:
Yes, will try to finish it soon.

Airblader:
I'm closing this for inactivity now. If you update the PR please just let us know or reopen it as a new one. Thank you!

Issue #3227, opened 2018-04-05
"scratchpad show" does not return correct success information


Airblader:
i3 4.15

Currently, `scratchpad show` always returns `success: "true"`. We should adapt this so it actually returns success only if a window has been shown. If no scratchpad window exists (case of no criteria) or none with the given criteria exists, we should return `success: "false"`.

This is motivated by #3226 which essentially duplicates #1809 where we adapted `focus` to return proper information.

PR #3230, opened 2018-04-06
Make "scratchpad show" return correct info

hwangcc23:
Fix the issue #3227(https://github.com/i3/i3/issues/3227).

### DIFF
--- a/include/scratchpad.h
+++ b/include/scratchpad.h
@@ -29,7 +29,7 @@ void scratchpad_move(Con *con);
  * can press the same key to quickly look something up).
  *
  */
-void scratchpad_show(Con *con);
+bool scratchpad_show(Con *con);
 
 /**
  * When starting i3 initially (and after each change to the connected outputs),
--- a/src/commands.c
+++ b/src/commands.c
@@ -1825,19 +1825,20 @@ void cmd_move_scratchpad(I3_CMD) {
 void cmd_scratchpad_show(I3_CMD) {
     DLOG("should show scratchpad window\n");
     owindow *current;
+    bool result = false;
 
     if (match_is_empty(current_match)) {
-        scratchpad_show(NULL);
+        result = scratchpad_show(NULL);
     } else {
         TAILQ_FOREACH(current, &owindows, owindows) {
             DLOG("matching: %p / %s\n", current->con, current->con->name);
-            scratchpad_show(current->con);
+            result |= scratchpad_show(current->con);
         }
     }
 
     cmd_output->needs_tree_render = true;
-    // XXX: default reply for now, make this a better reply
-    ysuccess(true);
+
+    ysuccess(result);
 }
 
 /*
--- a/src/scratchpad.c
+++ b/src/scratchpad.c
@@ -84,7 +84,7 @@ void scratchpad_move(Con *con) {
  * can press the same key to quickly look something up).
  *
  */
-void scratchpad_show(Con *con) {
+bool scratchpad_show(Con *con) {
     DLOG("should show scratchpad window %p\n", con);
     Con *__i3_scratch = workspace_get("__i3_scratch", NULL);
     Con *floating;
@@ -97,7 +97,7 @@ void scratchpad_show(Con *con) {
         floating->scratchpad_state != SCRATCHPAD_NONE) {
         DLOG("Focused window is a scratchpad window, hiding it.\n");
         scratchpad_move(focused);
-        return;
+        return true;
     }
 
     /* If the current con or any of its parents are in fullscreen mode, we
@@ -124,7 +124,7 @@ void scratchpad_show(Con *con) {
                  * window inside this scratch container in order to
                  * keep the focus the same within this container */
             con_activate(con_descend_tiling_focused(walk_con));
-            return;
+            return true;
         }
     }
 
@@ -141,7 +141,7 @@ void scratchpad_show(Con *con) {
             DLOG("Found a visible scratchpad window on another workspace,\n");
             DLOG("moving it to this workspace: con = %p\n", walk_con);
             con_move_to_workspace(walk_con, focused_ws, true, false, false);
-            return;
+            return true;
         }
     }
 
@@ -149,7 +149,7 @@ void scratchpad_show(Con *con) {
      * is actually in the scratchpad */
     if (con && con->parent->scratchpad_state == SCRATCHPAD_NONE) {
         DLOG("Window is not in the scratchpad, doing nothing.\n");
-        return;
+        return false;
     }
 
     /* If this was 'scratchpad show' with criteria, we check if it matches a
@@ -165,7 +165,7 @@ void scratchpad_show(Con *con) {
         if (current == active) {
             DLOG("Window is a scratchpad window, hiding it.\n");
             scratchpad_move(con);
-            return;
+            return true;
         }
     }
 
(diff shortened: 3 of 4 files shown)


Airblader:
I'm sorry, I don't understand the fix. We want to return whether we have successfully shown a window or not. That means we need to use the information coming from `scratchpad_show`.

hwangcc23:
Sorry. Misunderstood the problem.
Up.

include/scratchpad.h
```diff
@@ -28,8 +28,10 @@ void scratchpad_move(Con *con);
  * scratchpad window, this serves as a shortcut to hide it again (so the user
  * can press the same key to quickly look something up).
  *
+ * Return true if a window is successfully showning.
```
Airblader: There's a typo, but let's make change it to

> Returns true either if con was in the scratchpad and is now shown or if con was moved back into the scratchpad due to toggling.

src/commands.c
```diff
@@ -1825,19 +1825,20 @@ void cmd_move_scratchpad(I3_CMD) {
...
+    bool result = false;
 
     if (match_is_empty(current_match)) {
-        scratchpad_show(NULL);
+        result = scratchpad_show(NULL);
     } else {
         TAILQ_FOREACH(current, &owindows, owindows) {
             DLOG("matching: %p / %s\n", current->con, current->con->name);
-            scratchpad_show(current->con);
+            result = scratchpad_show(current->con);
```
Airblader: We should report a failure if there was a failure for any matched window (that's how we try to do it everywhere), so this should be `result &= …`.

src/scratchpad.c
```diff
@@ -83,8 +83,10 @@ void scratchpad_move(Con *con) {
  * scratchpad window, this serves as a shortcut to hide it again (so the user
  * can press the same key to quickly look something up).
  *
+ * Return true if a window is successfully showning.
```
Airblader: Same as above.

src/scratchpad.c
```diff
@@ -97,7 +99,7 @@ void scratchpad_show(Con *con) {
         floating->scratchpad_state != SCRATCHPAD_NONE) {
         DLOG("Focused window is a scratchpad window, hiding it.\n");
         scratchpad_move(focused);
-        return;
+        return false;
```
Airblader: This should return true. I think the true/false semantics should be that it's true whenever we successfully did something with `con`.

src/scratchpad.c
```diff
@@ -165,7 +167,7 @@ void scratchpad_show(Con *con) {
         if (current == active) {
             DLOG("Window is a scratchpad window, hiding it.\n");
             scratchpad_move(con);
-            return;
+            return false;
```
Airblader: Should be `true`.

Airblader (changes requested):
Thanks for the update! There's a couple of changes and I also think we can & should add tests for all these different cases :-)

orestisf1993:
@hwangcc23
```perl
fresh_workspace;
open_window;
cmd 'move scratchpad';
my $result = cmd 'scratchpad show';
is($result->[0]->{success}, 1, 'call to scratchpad succeeded');

kill_all_windows;
$result = cmd 'scratchpad show';
is($result->[0]->{success}, 0, 'call to scratchpad failed');
```

src/commands.c
```diff
@@ -1825,19 +1825,20 @@ void cmd_move_scratchpad(I3_CMD) {
...
+    bool result = false;
 
     if (match_is_empty(current_match)) {
-        scratchpad_show(NULL);
+        result = scratchpad_show(NULL);
     } else {
         TAILQ_FOREACH(current, &owindows, owindows) {
             DLOG("matching: %p / %s\n", current->con, current->con->name);
-            scratchpad_show(current->con);
+            result |= scratchpad_show(current->con);
```
hwangcc23: We should report a success if there is any matched window shown or hidden (due to toggled).
This is for passing the test with the criteria(The test case 2: Verify that 'scratchpad show' with the criteria returns correct info).

hwangcc23:
Up

testcases/t/298-scratchpad-show.t
```diff
@@ -0,0 +1,71 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# �~@� https://build.i3wm.org/docs/testsuite.html
```
orestisf1993: Why not extend 185-scratchpad.t? Your `â~@¢` character seems to be wrong here too.

orestisf1993: Why not extend 185-scratchpad.t? Your `â~@¢` character seems to be wrong here too.

testcases/t/298-scratchpad-show.t
```diff
@@ -0,0 +1,71 @@
...
+
+kill_all_windows;
+$result = cmd 'scratchpad show';
+is($result->[0]->{success}, 0, 'call to scratchpad failed');
+
+################################################################################
+# 2: Verify that 'scratchpad show' with the criteria returns correct info.
+################################################################################
+my $ws = fresh_workspace;
+cmd 'workspace $ws';
```
orestisf1993: What you do here: open a new workspace with a random name, then immediately switch to a workspace named `$ws` (you used single quotes). You don't need a variable or the `cmd` so I think you can remove both.

testcases/t/298-scratchpad-show.t
```diff
@@ -0,0 +1,71 @@
...
+$result = cmd '[title="nomatch"] scratchpad show';
+is($result->[0]->{success}, 0, 'call to scratchpad with non-matching criteria failed');
+
+################################################################################
+# 3: Open a scratchpad window on a workspace, switch to another workspace and
+# call 'scratchpad show' again. Verify that it returns correct info.
+################################################################################
+kill_all_windows;
+
+cmd "workspace foo-1";
```
orestisf1993: nitpick: just use `fresh_workspace`

testcases/t/298-scratchpad-show.t
```diff
@@ -0,0 +1,71 @@
...
+
+open_window;
+cmd 'move scratchpad';
+
+cmd "workspace foo-2";
```
orestisf1993: same

orestisfl (changes requested):

hwangcc23:
@orestisf1993 Thanks for the review.

Up.

testcases/t/185-scratchpad.t
```diff
@@ -471,4 +471,47 @@ is(scalar @$nodes, 0, 'no window on current ws anymore');
 
 is($scratch_nodes[0]->{scratchpad_state}, 'changed', 'scratchpad_state changed');
 
+################################################################################
+# 15: Verify that 'scratchpad show' returns correct info.
+################################################################################
+kill_all_windows;
```
orestisf1993: nitpicky but can you add a newline after the comment blocks, like the rest of the file does?

src/scratchpad.c
```diff
@@ -83,8 +83,11 @@ void scratchpad_move(Con *con) {
  * scratchpad window, this serves as a shortcut to hide it again (so the user
  * can press the same key to quickly look something up).
  *
+ * Returns true either if con was in the scratchpad and is now shown or if con
```
orestisf1993: This is slightly inaccurate because true is also returned  if there is an unfocused scratchpad on the current workspace. We don't really need to document the return value since it's a bit obvious.

orestisfl (approved):

testcases/t/185-scratchpad.t
hwangcc23: OK. Will send the PR again.

hwangcc23:
Requested changes are all done.

orestisf1993:
Thanks. I'd merge it but I'll let @Airblader have a final look.

Airblader (approved):

Airblader:
Thanks!

PR #3231, opened 2018-04-07
Fix crash when moving container to marked workspace

orestisf1993:


orestisf1993:
I've been getting this issue lately where crashing tests hang `complete-run.pl` instead of throwing an error.

Issue #3003, opened 2017-09-30
tree_flatten incorrectly closes vertical container inside stacked


orestisf1993:
Output of `i3 --moreversion 2>&- || i3 --version`:

```
Binary i3 version:  4.14-128-gc08ef361 (2017-09-29, branch "next") © 2009 Michael Stapelberg and contributors
```

URL to a logfile as per https://i3wm.org/docs/debugging.html:

https://logs.i3wm.org/logs/5702096292675584.bz2

**What I did:**

This bug is similar to https://github.com/i3/i3/issues/1053:

1. fresh workspace
2. open window
3. layout stacked
4. open window
5. split h
6. open window
7. split v
8. open window
9. focus left
10. move left

**What I saw:**

The right side now has 3 stacked containers
![2017-09-30-043403_1680x1009](https://user-images.githubusercontent.com/5778622/31041111-c67d48a6-a598-11e7-9b3b-4843f0664bde.png)

**What I expected instead:**
`H[window S[window V[window window]]]`

I think part of the problem is the use of `con_orientation` instead of `->layout`:
https://github.com/i3/i3/blob/7a7481e5784fc4db9b5724c9d2755040f5a47fed/src/tree.c#L718-L719

efenwick97:
I took a look at this issue and I think it stems from tree_flatten in src/tree.c always flattening a pair of containers. In the example posted above, in order for the desired behavior to occur, only a single container should be flattened. I'm about to open a PR that addresses this. My solution is to flatten only the current container if the layout of the child and the parent are different. If they have the same layout then both the current container and its child will be flattened.

Issue #3001, opened 2017-09-30
split + layout stacked / tabbed creates redundant containers


orestisfl:
Output of `i3 --moreversion 2>&- || i3 --version`:

```
Binary i3 version:  4.14-128-gc08ef361 (2017-09-29, branch "next") © 2009 Michael Stapelberg and contributors
```

URL to a logfile as per https://i3wm.org/docs/debugging.html:

https://logs.i3wm.org/logs/5637036128075776.bz2

**What I did:**

1. fresh workspace
2. open window
3. layout stacked
4. open window
5. split v
6. layout stacked
7. repeat 5-6

**What I saw:**

![2017-09-30-042426_1680x1009](https://user-images.githubusercontent.com/5778622/31041003-555dc8ae-a597-11e7-9da4-69fb71b69aa3.png)

Repeating this too many times makes i3 unbearably slow and memory usage skyrockets.

**What I expected instead:**
Redundant stacked / tabbed containers should not be opened.

Airblader:
IMHO this is the correct behavior. Why should we close this (valid) container that has been intentionally created? You _can_ move a window into each of those stacked containers (just do your steps, open a second window and `move up` a few times).

(Although, oddly, `move down` a single time immediately moves it all the way to the first window)

@stapelberg CC

orestisf1993:
Hmm. I understand the behavior a bit better right now but the same thing (intentionally) doesn't happen for `layout splith && splitv`.

I think it's less confusing if we apply the same to stacked / tabbed containers instead of allowing this:
![out](https://user-images.githubusercontent.com/5778622/31044281-0584bc0a-a5d5-11e7-86f4-20072b0e7a76.png)

roobre:
I'm suffering a similar issue on `Binary i3 version:  4.14.1`. When a new workspace is created in certain mode, and then I issue a layout change (for example, `layout toggle split` in tabbed container, or `layout tabbed` on a split container), a redundant container bar at top is created.

For example, if I create a new container with `workspace_layout split` (default), and then I issue `layout tabbed`, I get:

![deepin-screenshot j28535](https://user-images.githubusercontent.com/969721/32701405-013435a0-c7d5-11e7-84b9-48ae8c327849.png)

This behaviour is diferent to what i3 did before, where this redundant top bar was not shown.

Full version output:

```Binary i3 version:  4.14.1 (2017-09-24) © 2009 Michael Stapelberg and contributors
[...]

PR #3235, opened 2018-04-10
Fix for issue 3003

efenwick97:
Fix for [Issue 3003](https://github.com/i3/i3/issues/3003).

This PR modifies the behavior of the tree_flatten function in src/tree.c to allow for the flattening of a single container. If a child has a different layout than the parent of the current container only the current container will be flattened and the child will be appended to the parent. If the child and parent of the current container are the same then the children of child should be added to parent and both the current container and the child container should be flattened.

### DIFF
--- a/testcases/t/145-flattening.t
+++ b/testcases/t/145-flattening.t
@@ -72,4 +72,55 @@ $ws = get_ws($tmp);
 @nodes = @{$ws->{nodes}};
 is(@nodes, 2, 'all three windows on workspace level');
 
+################################################################################
+# Ensure that containers that contain only a single child after moving are 
+# flattened. If the child of the container does not have the same layout as the 
+# parent, append the child to the parent.
+################################################################################
+
+$tmp = fresh_workspace;
+
+my $stack1 = open_window;
+cmd 'layout stacking';
+my $stack2 = open_window;
+cmd 'split h';
+my $top = open_window;
+cmd 'split v';
+my $bottom = open_window;
+cmd 'focus left';
+cmd 'move left';
+
+$ws = get_ws($tmp);
+@nodes = @{$ws->{nodes}};
+is(@nodes, 2, 'two windows on workspace');
+is($nodes[1]->{layout}, 'stacked', 'layout is stacked');
+is(@{$nodes[1]->{nodes}}, 2, 'two nodes in stacked container');
+is($nodes[1]->{nodes}->[1]->{layout}, 'splitv', 'vertical layout');
+is(@{$nodes[1]->{nodes}->[1]->{nodes}}, 2, 'two nodes in vertical container');
+
+################################################################################
+# Ensure that containers that contain only a single child after moving are 
+# flattened. If the child of the container is the same layout as the parent
+# append the childs children to the parent.
+################################################################################
+
+$tmp = fresh_workspace;
+
+$stack1 = open_window;
+cmd 'split v';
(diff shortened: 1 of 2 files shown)


testcases/t/145-flattening.t
```diff
@@ -72,4 +72,45 @@ $ws = get_ws($tmp);
...
+is(@nodes, 2, 'two windows on workspace');
+is($nodes[1]->{layout}, 'stacked', 'layout is stacked');
+is(@{$nodes[1]->{nodes}}, 2, 'two nodes in stacked container');
+is($nodes[1]->{nodes}->[1]->{layout}, 'splitv', 'vertical layout');
+is(@{$nodes[1]->{nodes}->[1]->{nodes}}, 2, 'two nodes in vertical container');
+
+
+$tmp = fresh_workspace;
+
+my $stack1 = open_window;
```
orestisf1993: Remove `my`

testcases/t/145-flattening.t
```diff
@@ -72,4 +72,45 @@ $ws = get_ws($tmp);
...
+cmd 'split v';
+my $stack2 = open_window;
+cmd 'split h';
+my $top = open_window;
```
orestisf1993: Remove `my`

testcases/t/145-flattening.t
```diff
@@ -72,4 +72,45 @@ $ws = get_ws($tmp);
...
+cmd 'split v';
+my $bottom = open_window;
```
orestisf1993: Remove `my`

testcases/t/145-flattening.t
```diff
@@ -72,4 +72,45 @@ $ws = get_ws($tmp);
...
+is(@{$nodes[1]->{nodes}}, 2, 'two nodes in stacked container');
+is($nodes[1]->{nodes}->[1]->{layout}, 'splitv', 'vertical layout');
+is(@{$nodes[1]->{nodes}->[1]->{nodes}}, 2, 'two nodes in vertical container');
+
+
+$tmp = fresh_workspace;
+
+my $stack1 = open_window;
+cmd 'split v';
+my $stack2 = open_window;
```
orestisf1993: Remove `my`

Airblader:
@orestisf1993 Do you want to review (and merge) it? :-)

orestisf1993:
I think we should discuss what we want to do with redundant containers, see #3001, the case this PR handles (see comment in tree_flatten) and https://www.reddit.com/r/i3wm/comments/8rdwqb/what_is_this/.

Airblader:
> and https://www.reddit.com/r/i3wm/comments/8rdwqb/what_is_this/.

I don't think there's anything to change in that case. It's certainly unusual that someone would intentionally go to this layout, but it's not unreasonable (stacking stacked containers can improve the focus navigation).

orestisf1993:
Stacking stacked containers is fine, the problem appears when you use `layout stacked` (once) in an empty workspace:

a

Another problematic situation is this:
1. open 3 windows
2. `layout tabbed`
3. `move down`
a

Also, like in #3001, I still think that containers with only one non-leaf child are redundant.

Instead of continuously patching tree_flatten, making it more complex, I think we should first specify formally which containers should be considered redundant and then reimplement tree_flatten to fix its problems.

Airblader:
For now closing this until we have progressed in #3503 where this PR is tracked as well.

Issue #3220, opened 2018-04-03
i3bar: DLOGs before verbose is configured can never be printed


orestisf1993:
i3 4.15

Ironically, this includes the DLOG that announces the verbose configuration:
https://github.com/i3/i3/blob/bc760c20364482ea5d18a91fa781c7c5c7ca62cf/i3bar/src/config.c#L307-L309

I propose adding a `--verbose` flag

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

orestisf1993:
> Did you follow https://i3wm.org/docs/debugging.html?

That's why I am here!

Airblader:
> I propose adding a --verbose flag

Not sure I understand(?) We do already have `-V` (not `-v` since that's for the version).

orestisf1993:
We don't have a `-V` for i3bar:
https://github.com/i3/i3/blob/bc760c20364482ea5d18a91fa781c7c5c7ca62cf/i3bar/src/main.c#L111-L128

Airblader:
Oops. :-) I think adding such a flag is fine.

orestisf1993:
#3236

PR #3236, opened 2018-04-10
Add an i3bar flag: --verbose

hwangcc23:
Fix the issue #3220. (https://github.com/i3/i3/issues/3220)

### DIFF
--- a/docs/debugging
+++ b/docs/debugging
@@ -160,7 +160,8 @@ flood kicks.
 
 == Debugging i3bar
 
-To debug i3bar problems, add +verbose yes+ to all +bar {}+ blocks in your i3
+To debug i3bar problems, use the +--verbose+ commandline parameter,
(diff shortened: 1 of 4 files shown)


Airblader (changes requested):

hwangcc23:
Up

orestisf1993:
We should also update the docs: https://github.com/i3/i3/blob/b68816423599037788dc719db287d131fde16896/docs/debugging#L161

We could also modify https://github.com/i3/i3/blob/b68816423599037788dc719db287d131fde16896/src/main.c#L952 to use -V whenever verbose is configured since i3bar's output is included in stderr when you start i3. It won't show up in the logs though.

hwangcc23:
Up

Airblader (approved):

Airblader:
Thanks!

PR #3240, opened 2018-04-13
Resize tile px

### LINKED ISSUE #3239: [Enhancement] Resize tiled windows by px instead of ppt
I've found myself frustrated by the inability to exactly resize windows in tiles, especially when I'm resizing a terminal that I want to get to a specific text width.  With the mouse, I can do pixel-perfect resizing, but otherwise I'm stuck with ppt, and the smallest percentage point is 1, since floats aren't supported, so there's effectively no way to resize with the keyboard at that level of precision.

I have a PR almost ready with this feature; I'm ironing out the kinks now and working on test cases. I've been running it on my development workstation for a while now.

The old docs say:

> The optional pixel argument specifies by how many pixels a *floating container* should be grown or shrunk (the default is 10 pixels). The ppt argument means percentage points and specifies by how many percentage points a *tiling container* should be grown or shrunk (the default is 10 percentage points).

And I would update them to say:

> The optional pixel argument specifies by how many pixels a container should be grown or shrunk (the default is 10 pixels).  The optional ppt argument means "percentage points", and if specified it indicates that a *tiling container* should be grown or shrunk by that many points, instead of by the px value.

If a ppt value is specified, it will be preferred to the px value for tiling containers only.  So this change is somewhat backwards compatible, in that configurations that specify both px and ppt would continue to behave the same way.

An alternative solution would be to allow fractional ppt value (I had a PR for this ready as well, but it was uglier and harder to control).

andrewla:
Allows resizing of tiled containers by px instead of just by ppt.  Syntax is just appended to current resize command, and will be backwards compatible if px is not specified.

Fixes #3239

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2326,10 +2326,10 @@ resize set <width> [px | ppt] <height> [px | ppt]
 Direction can either be one of +up+, +down+, +left+ or +right+. Or you can be
 less specific and use +width+ or +height+, in which case i3 will take/give
 space from all the other containers. The optional pixel argument specifies by
-how many pixels a *floating container* should be grown or shrunk (the default
-is 10 pixels). The ppt argument means percentage points and specifies by how
-many percentage points a *tiling container* should be grown or shrunk (the
-default is 10 percentage points).
+how many pixels a container should be grown or shrunk (the default is 10
+pixels).  The optional ppt argument means "percentage points", and if
+specified it indicates that a *tiling container* should be grown or shrunk by
+that many points, instead of by the +px+ value.
 
 Notes about +resize set+: a value of 0 for <width> or <height> means "do
 not resize in this direction", and resizing a tiling container by +px+ is not
--- a/include/util.h
+++ b/include/util.h
@@ -177,3 +177,9 @@ bool parse_long(const char *str, long *out, int base);
  *
  */
 ssize_t slurp(const char *path, char **buf);
+
+/**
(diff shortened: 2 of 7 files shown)


orestisf1993:
Can you address the failing tests?

andrewla:
Took a bit of debugging to figure out why it wasn't resizing correctly in this case, and some learning about how the actual layout code works, rather than just assuming magic.

andrewla:
NB: I have fixed existing tests, but have not added anything to test new functionality.  Now that I have all the testing framework working on my machine (kudos to perlbrew!) instead of using Travis I should be able to operate faster.

That said, the backwards compatibility of the change is well-demonstrated here; as long as both ppt and px are specified the behavior is unchanged.

Airblader:
@andrewla Can you address the failing tests please?

andrewla:
Sorry, this has been on my back burner.  I fixed the existing test, but during testing of new functionality I uncovered a bug that I had some trouble hunting down -- namely that resizing in a shrinking direction does so by one less pixel than it should, due to rippling rounding errors caused by the way that window sizes are calculated.

michaelnew:
Sorry to be that guy, but... any progress on this? It seems awfully close and it would be such a shame to not get it merged. This is a feature I've been really missing for a long time.

orestisf1993:
I am continuing this in #3372 with support for other commands as well.

PR #3243, opened 2018-04-15
man: Fix title markers

avindra:
The title marker lines have to be aligned with the previous lines.
The error was caught by asciidoctor, which tends to be picker than
asciidoc.

I'm submitting on behalf of Takashi (@tiwai).


Airblader:
Thanks!

PR #3246, opened 2018-04-16
i3bar: make modifier behave like floating_modifier

### LINKED ISSUE #3234: Feature request: Allow modifier combinations for the bar modifier
The `floating_modifier` allows any combination of keys. In contrast, the `modifier` of the `bar` section only allows a very limited set of modifiers and combinations of these are entirey prohibited. This behaviour is contra-intuitive and seems inconsistent to me. It would be more useful, if I could use key combinations (for example `Mod4+Shift`) as the bar modifier.

Background:  I'm using the bar to display a lot of information. However, not all the info fits my relatively small screen, therefore I would like to have multiple bars, one appearing when pressing Mod4, the second when pressing Mod4+Shift, etc...

orestisf1993:
Fixes #3234.

### DIFF
--- a/i3bar/src/config.c
+++ b/i3bar/src/config.c
@@ -119,6 +119,7 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
+    /* Kept for backwards compatibility. */
     if (!strcmp(cur_key, "modifier")) {
         DLOG("modifier = %.*s\n", len, val);
         if (len == strlen("none") && !strncmp((const char *)val, "none", strlen("none"))) {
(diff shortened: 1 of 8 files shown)


i3bar/src/config.c
```diff
@@ -119,42 +119,6 @@ static int config_string_cb(void *params_, const unsigned char *val, size_t _len
         return 1;
     }
 
-    if (!strcmp(cur_key, "modifier")) {
```
Airblader: This breaks the compatibility. Would it be a lot of trouble to keep supporting the old values here and parse them correctly? We did the same for `wheel_up_cmd`, for example.

orestisf1993: Sure. Should I move parse_long in libi3?

Issue #3162, opened 2018-03-07
[feature request] Assign scratchpad to monitor


k1sul1:
Reposting from https://github.com/Airblader/i3/issues/198#issuecomment-371174838

---

I've been using i3(-gaps) for a little over 6 months now, and the one thing that's bugging me is that the scratchpad follows my cursor.  I've put a terminal emulator on it, and I show & hide it with a hotkey. It doesn't follow the cursor immediately, but if the scratchpad is open on my primary monitor, and I move my mouse to the secondary and then try to hide it with the hotkey, it will instead pop to my secondary monitor.

`bindsym $mod+minus scratchpad show`

That is the hotkey combination I'm using.

If my monitors were the same resolution, the scratchpad following wouldn't be an issue, but the terminal is unreadable on my laptop monitor when I'm using it on the primary monitor. I've tried searching the documentation but I haven't found anything on the subject. 

Something like this is what I'd want to do. But pretty much anything that will achieve what I want will do. 

```
workspace 1 output $external_monitor
workspace 2 output $external_monitor
workspace 3 output $external_monitor

scratchpad output $external_monitor

workspace 11 output $laptop_monitor
workspace 12 output $laptop_monitor
workspace 13 output $laptop_monitor
```
I'm not sure if the original i3 dev is willing to do such a feature, so I'm posting it here.

Output of `i3 --moreversion 2>&- || i3 --version`:

Binary i3 version:  4.14.1 (2017-09-24) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.14.1 (2017-09-24) (pid 959) abort…)
Loaded i3 config: /home/k1sul1/.config/i3/config (Last modified: Wed 07 Mar 2018 12:49:30 PM EET, 144 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3

Airblader:
I think adding a flag like `scratchpad show --focused-workspace` (we can talk about the name of course) to conditionally skip this would be fine.

Airblader:
Sorry, I should elaborate. I don't like the proposal here (and don't think it makes much sense), so my suggestion is a different one. Would that work for you as well? From my understanding it should.

k1sul1:
Yeah, I don't really care how it's done, but I'd just like to have the scratchpad stay put.

Airblader:
Alright, let's add a flag then. PRs welcome, should be pretty trivial to implement.

Airblader:
[...]

PR #3247, opened 2018-04-16
Issue 3162: Add --focused-workspace flag to scratchpad show

efenwick97:
Fix for [3162](https://github.com/i3/i3/issues/3162).

scratchpad --focused-workspace show will not move a viable scratchpad window to the current workspace.


include/commands.h
```diff
@@ -288,7 +288,7 @@ void cmd_move_scratchpad(I3_CMD);
  * Implementation of 'scratchpad show'.
  *
  */
-void cmd_scratchpad_show(I3_CMD);
+void cmd_scratchpad_show(I3_CMD, const char *focused_workspace);
```
orestisf1993: should be a bool

testcases/t/501-scratchpad.t
```diff
@@ -108,4 +108,31 @@ $second = fresh_workspace(output => 0);
...
+# scratchpads on the currently focused workspace.
+################################################################################
+
+my $ws1 = fresh_workspace;
+my $window = open_window;
+my $scratchpad_window = open_window;
+cmd "move scratchpad";
+
+my $ws2 = fresh_workspace;
+cmd "workspace $ws2";
```
orestisf1993: unnecessary

testcases/t/501-scratchpad.t
```diff
@@ -108,4 +108,31 @@ $second = fresh_workspace(output => 0);
...
+cmd "scratchpad show";
+
+# verify that calling scratchpad show on ws1 will not affect the scratchpad on
+# ws2 if focused-workspace flag is set.
+cmd "workspace $ws1";
+cmd qq|"scratchpad --focused-workspace show"|;
+is(scalar @{get_ws($ws2)->{floating_nodes}}, 1, 'one floating node on this ws');
+
+# verify that calling scratchpad show with an open scratcpad on another
```
orestisf1993: Not sure if this test is needed?

testcases/t/501-scratchpad.t
```diff
@@ -108,4 +108,31 @@ $second = fresh_workspace(output => 0);
...
+
+my $ws2 = fresh_workspace;
+cmd "workspace $ws2";
+cmd "scratchpad show";
+
+# verify that calling scratchpad show on ws1 will not affect the scratchpad on
+# ws2 if focused-workspace flag is set.
+cmd "workspace $ws1";
+cmd qq|"scratchpad --focused-workspace show"|;
+is(scalar @{get_ws($ws2)->{floating_nodes}}, 1, 'one floating node on this ws');
```
orestisf1993: None of your tests fail on the current `next` branch. Can you also check that the scratchpad window opens in $ws1? It currently doesn't do that because you show the one scratchpad window in $ws2 and no new window is opened in $ws1 at all.

testcases/t/501-scratchpad.t
```diff
@@ -108,4 +108,31 @@ $second = fresh_workspace(output => 0);
 
 verify_scratchpad_switch($first, $second);
 
+################################################################################
+# If the --focused-workspace flag is set then scratchpad show should only affect
+# scratchpads on the currently focused workspace.
+################################################################################
+
+my $ws1 = fresh_workspace;
```
orestisf1993: You should open each workspace on a different output.

include/commands.h
efenwick97: Just checking but do you mean that the flag that we eventually send should be a bool? Looking at the other flags they are all const char pointers in commands.h and then we use the result of flag != NULL as the value of the flag.

testcases/t/501-scratchpad.t
efenwick97: We don't need them. I wanted to test that the old functionality wasn't changed but I suppose that the other tests should cover that making this one redundant. Thanks for the heads up.

testcases/t/501-scratchpad.t
efenwick97: I looked into it and basically the command was failing on next as it couldn't parse the flag. Unintentionally this was making the rest of the test succeed. I now assert that the "scratchpad --focused-workspace show" command succeeds.

testcases/t/501-scratchpad.t
```diff
@@ -108,4 +108,27 @@ $second = fresh_workspace(output => 0);
 
 verify_scratchpad_switch($first, $second);
 
+################################################################################
+# If the --focused-workspace flag is set then scratchpad show should only affect
+# scratchpads on the currently focused workspace.
+################################################################################
+my $ws2 = fresh_workspace(output => 1);
+
+# Kill all but one of the scratchpad windows and focus it
```
orestisf1993: You can run `kill_all_windows`

src/scratchpad.c
```diff
@@ -139,6 +139,10 @@ void scratchpad_show(Con *con) {
             (floating = con_inside_floating(walk_con)) &&
             floating->scratchpad_state != SCRATCHPAD_NONE) {
             DLOG("Found a visible scratchpad window on another workspace,\n");
+            if (focused_workspace) {
+                DLOG("Not moving workspace, --focused-workspace flag set\n");
+                return;
```
orestisf1993: I am not sure if we want to exit here? What if there is one scratchpad window in $ws1 and one hidden. Doing `scratchpad --focused-workspace show` on $ws2 should move the hidden one in $ws2, right?

Airblader:
@efenwick97 What's the status for this PR? Are you going to continue working on it?

PR #3250, opened 2018-04-18
docs/ipc: update tree node with window_properties

rtgnx:
I have added description of window_properties field in tree node so that it can be used in
go-i3 interface.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -328,6 +328,8 @@ window (integer)::
 	This field is set to null for split containers or otherwise empty
 	containers. This ID corresponds to what xwininfo(1) and other
 	X11-related tools display (usually in hex).
+window_properties (map)::
+	X11 window properties title, instance, class, window_role and transient_for.
 urgent (bool)::
 	Whether this container (window, split container, floating container or
 	workspace) has the urgency hint set, directly or indirectly. All parent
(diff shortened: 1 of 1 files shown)


rtgnx:
Instance and transient_for fields are added now.

docs/ipc
```diff
@@ -328,6 +328,9 @@ window (integer)::
 	This field is set to null for split containers or otherwise empty
 	containers. This ID corresponds to what xwininfo(1) and other
 	X11-related tools display (usually in hex).
+window_properties (map)::
+	X11 window propertirs title, instance, class, window_role and transient_for
```
Airblader: s/propertirs/properties/

docs/ipc
```diff
@@ -328,6 +328,9 @@ window (integer)::
...
+	which is container id of transition window or +null+.
```
Airblader: There's a bunch of articles missing for this to be an actual sentence. But I don't think we should explain this here at all – they are X11 properties and can be looked up in the specifications, so let's remove this half sentence.

rtgnx: Do you want me to leave `window_properties (map)::` or remove it all ?
I've only submitted it because it's required to get this https://github.com/i3/go-i3/pull/1
merged.

Airblader: Listing the properties is fine, I just mean removing the last half sentence. So it can be

> X11 window properties title, instance, class, window_role and transient_for.

Airblader:
Thanks!

PR #3254, opened 2018-04-20
Multiple assignments of workspaces to outputs (#555)

### LINKED ISSUE #555: multiple "workspace n output <foo>"
[Originally reported by c+i3@…]
(It is possible to have multiple workspace entries in the config, with different outputs. Only the last one is used.

-> Throw an error if there is already an output defined for a workspace
e.g.:

workspace 1 output HDMI1
workspace 1 output VGA1

workspace 1 will appear on VGA1 or on current if VGA1 is inactive. The line with HDMI1 is ignored

orestisf1993:
- [x] userguide
- [ ] I followed the discussion in #555 and concluded to use the `workspace X output A B C` syntax but it might be inconsistent with https://i3wm.org/docs/userguide.html#_output_s.
- [X] Fix infinite reccursion error
- [X] Fix mouse over crash
- [X] move_workspace_to_output crash

CC @ograff

Fixes #555.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -888,7 +888,7 @@ the second screen and so on).
 
 *Syntax*:
 -------------------------------------
-workspace <workspace> output <output>
+workspace <workspace> output <output1> [output2]…
 -------------------------------------
 
 The 'output' is the name of the RandR output you attach your screen to. On a
@@ -907,12 +907,15 @@ monitor name is “Dell UP2414Q”.
 entire monitor, i3 will still use the entire area of the containing monitor
 rather than that of just the output's.)
 
+You can specify multiple outputs. The first available will be used.
+
 If you use named workspaces, they must be quoted:
 
 *Examples*:
 ---------------------------
 workspace 1 output LVDS1
-workspace 5 output VGA1
+workspace 2 output primary
+workspace 5 output VGA1 LVDS1
 workspace "2: vim" output VGA1
 ---------------------------
 
--- a/include/workspace.h
+++ b/include/workspace.h
@@ -39,6 +39,13 @@ Con *get_existing_workspace_by_name(const char *name);
 Con *get_existing_workspace_by_num(int num);
 
 /**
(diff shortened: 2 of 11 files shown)


foonathan:
What is the status of this PR? I'd love to have this feature.

orestisf1993:
Mainly needs to be reviewed. I didn't use the feature much on my machine so if you want to run this PR locally you might be able to spot possible issues since this change is a bit tricky.

orestisf1993:
@Airblader I see that you put the `waiting` label. The only thing missing AFAIK is updating the userguide and writing some missing docstrings. I can do either if you want them before reviewing.

Airblader:
The changes look fine to me at first glance, so I think we can go ahead.

orestisf1993:
up

docs/userguide
```diff
@@ -888,7 +888,7 @@ the second screen and so on).
 
 *Syntax*:
 -------------------------------------
-workspace <workspace> output <output>
+workspace <workspace> output <outputs…>
```
Airblader: We write this as `[<output>+]` in other places in the userguide. I think we should use the same pattern here.

orestisf1993: `[<XXX>+]` is only used for things that need a '+' character on the end right now, eg `[<Modifiers>+]` could be `Mod1+` or `Mod1+Shift+`. I think we should think of a new notation to not confuse users as outputs are not separated by pluses but by spaces.

orestisf1993: But, tbf, no-one will really be confused since we'll have an example right after.

Airblader: You're right, but nonetheless we have this case already in https://i3wm.org/docs/userguide.html#_manipulating_layout. Following that it should be `<output>…`. :-)

orestisf1993: Maybe `<output1> [output2]…` since you have to specify at least one output.

Airblader: GTM

PR #3258, opened 2018-04-20
i3-nagbar: add option for button that runs commands without a terminal

### LINKED ISSUE #2199: Cannot exit i3 with default config and no terminal emulator
i3 version 4.11 (2015-09-30, branch "4.11") © 2009 Michael Stapelberg and contributors
Logfile: http://logs.i3wm.org/logs/5639221461123072.bz2

With the default config and no terminal emulator, attempting to exit i3 with the exit keybind raises i3-nagbar, and clicking "Yes, exit i3" raises an error "i3-sensible-terminal could not find a terminal emulator. Please install one." Apparently i3-nagbar requires a terminal emulator (https://github.com/i3/i3/blob/next/i3-nagbar/main.c#L178). Exiting works, if the exit keybind in the config file is modified to simply do 'exec "i3-msg exit"' .

orestisf1993:
I think this is an acceptable solution: two kind of buttons. One executes buttons on a terminal, one directly. `-x` / `--button-no-terminal` is a placeholder name, we can probably find something better.

Fixes #2199.

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -147,7 +147,7 @@ bindsym Mod1+Shift+c reload
 # restart i3 inplace (preserves your layout/session, can be used to upgrade i3)
 bindsym Mod1+Shift+r restart
 # exit i3 (logs you out of your X session)
-bindsym Mod1+Shift+e exec "i3-nagbar -t warning -m 'You pressed the exit shortcut. Do you really want to exit i3? This will end your X session.' -b 'Yes, exit i3' 'i3-msg exit'"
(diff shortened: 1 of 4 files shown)


Airblader:
LGTM, do you have a proposal for the name/flag?

orestisf1993:
Nothing yet :/

Airblader:
Any ideas now? :-)

stapelberg:
I think button-no-terminal is a fine long name, and `B` is a good short name (like `b` for button, but different).

orestisf1993:
up

PR #3277, opened 2018-05-02
Resize set improvements

### LINKED ISSUE #3275: enhancement: `resize set` should accept a “width” or “height” keyword
Currently, the syntax is `resize set <width> [px | ppt] <height> [px | ppt]`, which results in commands like `resize set 0 ppt 70 ppt`.

I think we should also allow the following syntax:
```
resize set width <width> [px | ppt]
resize set height <height> [px | ppt]
resize set width <width> [px | ppt] height <height> [px | ppt]
```

### LINKED ISSUE #3276: resize set does not correctly interpret 0 for floating windows

Running i3 version: 4.15-137-g981e901b (2018-05-01, branch "next") (pid 39735)

Can add one later on in case this turns out to not be reproducible.

**What I did:**

* open urxvt
* set floating
* `i3 resize set 800 px 400 px`
* `i3 resize set 0 px 400 px`

**What I saw:**

The window was first resized to 800x400 (expected), then to a large width exceeding my screen width (unexpected).

**What I expected instead:**

The second resize command should not result in any changes.

orestisfl:
Fixes #3275
Fixes #3276

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2320,7 +2320,9 @@ If you want to resize containers/windows using your keyboard, you can use the
 *Syntax*:
 -------------------------------------------------------
 resize grow|shrink <direction> [<px> px [or <ppt> ppt]]
-resize set <width> [px | ppt] <height> [px | ppt]
+resize set [width] <width> [px | ppt]
+resize set height <height> [px | ppt]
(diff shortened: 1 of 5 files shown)


Airblader (approved):

stapelberg:
Thanks for the fix!

PR #3282, opened 2018-05-08
Replace strncpy call with memcpy when result is not NUL-terminated

orestisf1993:


orestisf1993:
Is this a new thing?

2018-05-08-170956_840x199_escrotum

Airblader:
Yeah, it is. :-)

PR #3295, opened 2018-05-20
Updated ISSUE_TEMPLATE.md

Airblader:
Same as in i3lock, I've updated the issue template as inspired by angular/angular.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,18 +1,62 @@
-Output of `i3 --moreversion 2>&- || i3 --version`:
+<!--
+PLEASE HELP US PROCESS GITHUB ISSUES FASTER BY PROVIDING THE FOLLOWING INFORMATION.
+-->
 
-_REPLACE: i3 version output_
+## I'm submitting a…
+<!-- Please check one of the following options with "x" -->
+<pre>
+[ ] Bug
+[ ] Feature Request
+[ ] Documentation Request
+[ ] Other (Please describe in detail)
+</pre>
 
-URL to a logfile as per https://i3wm.org/docs/debugging.html:
+## Current Behavior
+<!-- Describe the current behavior, e.g., »When pressing Alt+j (focus left), the window above the current window is focused.« -->
 
-_REPLACE: URL to logfile_
+## Expected Behavior
+<!-- Describe the desired behavior you expect after mitigation of the issue, e.g., »The window left next to the current window should be focused.« -->
 
-**What I did:**
+## Reproduction Instructions
+<!--
+For bug reports, please provide detailed instructions on how the bug can be reproduced.
+For feature requests you can remove this section.
 
-_REPLACE: e.g. "I’m pressing Alt+j (focus left)"_
+E.g., »Open three windows in a V[A H[B C]] layout on a new workspace«
+-->
 
-**What I saw:**
+## Environment
+<!--
+Please include your exact i3 version.
+Note that we only support the latest major release and the current development version. If you are using an older version of i3, please first update to the current release version and reproduce the issue there.
+-->
+Output of `i3 --moreversion 2>&-`:
+<pre>
+i3 version: 
+</pre>
 
-_REPLACE: e.g. "i3 changed focus to the window ABOVE the current window"_
+<!--
+For bug reports, please include your (complete) i3 config with which the issue occurs. You can either paste the file directly or provide a link to a service such as pastebin.
 
-**What I expected instead:**
-_REPLACE: e.g. "Focus should be on the window to the left"_
+If you would like to help debugging the issue, please try to reduce the config such that it is as close to the default config as possible while still reproducing the issue. This can help us bisect the root cause.
+-->
+<pre>
+</pre>
+
+<!--
+Providing a logfile can help us trace the root cause of an issue much quicker. You can learn how to generate the logfile here:
+https://i3wm.org/docs/debugging.html
+
+Providing the logfile is optional.
+-->
+<pre>
+Logfile URL:
+</pre>
+
+<!--
+Please also anser the questions below to help us process your issue faster. If you have any other information to share, please add it here as well.
+-->
+<pre>
+- Linux Distribution & Version:
+- Are you using a compositor (e.g., xcompmgr or compton):
+</pre>


orestisf1993:
LGTM

I assume we need a separate repo to see how it looks?

Airblader:
To see it in action, yes. But I think we can just go ahead and merge it to look at it (and fix then, if necessary). It doesn't affect anything else anyway.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,18 +1,57 @@
...
+<!--
+For bug reports, please include a logfile as described in
+https://i3wm.org/docs/debugging.html
+-->
+<pre><code>
+Logfile URL:
+</code></pre>
+
+<!--
+For bug reports, please provide an i3 config file with which the issue can be reproduced. Please try to get the config as closely to the default config as possible in order to make it easier to bisect the issue's root cause.
```
stapelberg: The config file is already included in the logfile. Can’t we just take it from there?

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,18 +1,57 @@
...
+Please include only the entire config either directly or through a link to a service such as pastebin.
+-->
+<pre><code>
+i3 config
+</code></pre>
+
+<!--
+Please also provide further information such as your distribution, presence of a compositing manager etc. if necessary. Providing this information allows us to process your issue faster.
```
stapelberg: This is pretty vague. If people think they have relevant extra details, they’ll provide them. If we want to know specific things, why not use a checklist? E.g., “Are you using a compositing manager such as xcompmgr?”

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,18 +1,57 @@
...
-**What I expected instead:**
-_REPLACE: e.g. "Focus should be on the window to the left"_
+<!--
+PLEASE HELP US PROCESS GITHUB ISSUES FASTER BY PROVIDING THE FOLLOWING INFORMATION.
+ISSUES MISSING IMPORTANT INFORMATION MAY BE CLOSED WITHOUT INVESTIGATION.
+-->
+
+## I'm submitting a…
+<!-- Please check one of the following options with "x" -->
+<pre><code>
```
stapelberg: Isn’t one of &lt;pre&gt; or &lt;code&gt; sufficient? :)

.github/ISSUE_TEMPLATE.md
Airblader: We could, but it's usually (at least for me) more comfortable to have a quick look at the config directly instead of looking at it through the logfile. Also, I think the config is useful (and required) in more situations than the log, and the log is something people don't provide regularly. So I'd prefer asking for the config separately.

.github/ISSUE_TEMPLATE.md
Airblader: I took it like this from the angular issue template, but yeah, seems like `pre` suffices (`code` seems to be inline). I'll update it.

.github/ISSUE_TEMPLATE.md
stapelberg: Could we make the question conditional? I.e., “If you can’t provide a log right now, please provide your configuration file.”

I’d like to keep the template as short as possible. Each piece of information we add will stop more and more people from filing bugs.

.github/ISSUE_TEMPLATE.md
Airblader: I guess the idea here is to make the user actively think about whether there is something. I agree that chances are they won't add something they wouldn't have added otherwise, though. I'll make distro and compositor into a checklist for now.

.github/ISSUE_TEMPLATE.md
Airblader: The effort required to provide the log file is significantly higher than providing the config, so if the goal is to make reporting issues easier rather than improving their initial quality, I'd opt for making the log file conditional (I also think we need it in fewer cases than the config), not the config. Would that be OK?

Airblader: @stapelberg As proposed, I've updated the template to ask for the config first and make the logfile optional.

Airblader:
I've also softened the phrasing for reducing the config to the default one and made that just an optional step to help out. I think just providing the config is the most important part.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,18 +1,61 @@
...
-
-**What I saw:**
-
-_REPLACE: e.g. "i3 changed focus to the window ABOVE the current window"_
-
-**What I expected instead:**
-_REPLACE: e.g. "Focus should be on the window to the left"_
+<!--
+PLEASE HELP US PROCESS GITHUB ISSUES FASTER BY PROVIDING THE FOLLOWING INFORMATION.
+ISSUES MISSING IMPORTANT INFORMATION MAY BE CLOSED WITHOUT INVESTIGATION.
```
stapelberg: I think “closing without investigation” is too unfriendly. Let’s just remove this sentence.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,18 +1,61 @@
...
-
-_REPLACE: i3 version output_
-
-URL to a logfile as per https://i3wm.org/docs/debugging.html:
-
-_REPLACE: URL to logfile_
-
-**What I did:**
-
-_REPLACE: e.g. "I’m pressing Alt+j (focus left)"_
```
stapelberg: The examples for “what i did”, “what i saw” and “what i expected” are no longer in the template. Should we add them? I think they make it a little easier for users to phrase their report.

.github/ISSUE_TEMPLATE.md
Airblader: Remove or rephrase? We *do* (already) automatically close issues for outdated versions (and IMHO we should do the same for missing version information), so being upfront about what happens might prevent the user from ignoring the issue template.

.github/ISSUE_TEMPLATE.md
Airblader: I've added examples back.

.github/ISSUE_TEMPLATE.md
Airblader: In any case, I've removed it for now.

stapelberg (approved):

PR #3310, opened 2018-06-15
Update i3-msg.man: added get_config and send_tick

KJoke70:
Added get_config and send_tick which are mentioned [here](https://i3wm.org/docs/ipc.html).


Airblader:
Thanks

PR #3313, opened 2018-06-22
Border artifacts when moving window

czak:
Using `hide_edge_borders both` I noticed some visual artifacts on the borders.

## How to reproduce

Given the following config:

```sh
# doesn't need to be 10, but makes the issue more visible
default_border pixel 10
hide_edge_borders both
```

1. Open 2 tiled windows
2. Swap the windows (or: move the right one to the left)

### Result

Visual artifacts on the border between windows. The artifacts go away when focus changes and borders are redrawn. After being redrawn once, the artifacts don't reappear.

before

## My fix

I updated the condition for `has_rect_changed`. Moving the window only changes its `x`, so previously this meant `has_rect_changed == false`. With the updated condition, a change in `rect.x` or `rect.y` also yields `has_rect_changed == true`.

### Result

after

## Notes

* This is my first ever look into i3's code
* I feel the solution may result in some unnecessary rendering
* If this is the right location for the fix, should `memcmp` be used instead?
* With `hide_edge_borders none`, the issue doesn't occur
    * this leads me to believe my fix is not optimal
    * `has_rect_changed` is also false with `hide_edge_borders none`, however it doesn't suffer from this issue

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -796,11 +796,8 @@ void x_push_node(Con *con) {
          * background and only afterwards change the window size. This reduces
          * flickering. */
 
-        /* As the pixmap only depends on the size and not on the position, it
-         * is enough to check if width/height have changed. Also, we don’t
-         * create a pixmap at all when the window is actually not visible
-         * (height == 0) or when it is not needed. */
-        bool has_rect_changed = (state->rect.width != rect.width || state->rect.height != rect.height);
+        bool has_rect_changed = (state->rect.x != rect.x || state->rect.y != rect.y ||
+                                 state->rect.width != rect.width || state->rect.height != rect.height);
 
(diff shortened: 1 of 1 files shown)


Airblader:
Hi, thanks so much for submitting a PR with detailed information along the way, I appreciate it! On first glance I do think the fix makes sense (clearly the assumption we had there so far was wrong), but I'll take a better look later on!

Airblader:
Scratch the part about later on. This change makes a lot of sense to me. I don't think it's worth to add complexity to decide in all detail considering feature flags etc. whether the rect changed. So this should be just fine. Thanks!

czak:
Cool, thanks @Airblader :)

PR #3322, opened 2018-07-09
If no trayoutput matches, take first

### LINKED ISSUE #3317: Tray symbols disappear after switching monitor outputs
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
I have `tray_output primary` in a `bar` block. When switching from my external monitor to my internal monitor while leaving the primary output unchanged (`xrandr --verbose --output DP1 --off --output eDP1 --auto`), the tray symbols disappear. Even when switching back (`xrandr --verbose --output eDP1 --off --output DP1 --auto --primary`), the symbols don't show up until I press `Mod1+Shift+r` (restart).

## Expected Behavior
When switching back to the external monitor (set as primary by xrandr), the tray symbols should be shown.

Also, not showing any symbols if the primary output is off is a bit confusing but you could argue that's exactly what I configured. (EDIT: I think that the behavior described in this paragraph is fine. I shouldn't have distracted from the actual problem which was described in the previous paragraph.)

## Reproduction Instructions
Described above.

## Environment
<pre>
Running i3 version: 4.15 (2018-03-10) (pid 383)to abort…)
Loaded i3 config: /home/manu/.config/i3/config (Last modified: 2018-06-29T21:00:40 CEST, 408 seconds ago)

</pre>

For testing, I've used an almost default config (copied from `/etc/i3/config`), with `tray_output primary` inserted into the `bar` block.

<pre>
- Linux Distribution & Version: Arch Linux (rolling release)
</pre>

I sometimes encounter other issues with tray symbols and switching monitors (in particular, the Pidgin symbol disappears sometimes and even restarting i3 doesn't help; only restarting Pidgin helps). However, the issue described here is the only one that I could reproduce reliably.

bebehei:
When changing the screen, the tray was searching for a matching output.
But if no screen matched, it neglected the setting and didn't show
anything.

As the default, the i3bar shows a tray on the first matching output.
Its intention is to display a tray. So it should also show a tray, if
the monitors don't match.

Fixes #3317

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -794,7 +794,7 @@ static void handle_client_message(xcb_client_message_event_t *event) {
 
             /* If no tray_output has been specified, we fall back to the first
              * available output. */
-            if (output == NULL && TAILQ_EMPTY(&(config.tray_outputs))) {
+            if (output == NULL) {
                 SLIST_FOREACH(walk, outputs, slist) {
                     if (!walk->active)
                         continue;


Airblader:
The behavior is intended as-is. Only if the user doesn't configure anything we use the first output. If the user configures something, we only show the bar if the output can be matched. Otherwise it'd be impossible to hide the bar on specific outputs. Am I missing something?

bebehei:
> Otherwise it'd be impossible to hide the bar on specific outputs.

@Airblader On the one hand side, I agree with you. But on the other side, this is "common userfault" and I don't see, why a program shouldn't fix it (a.k.a. automatically do it right). While creating this PR, I had it in my mind, that you couldn't be able anymore to disable the tray output, but I neglected it.

I had some solutions in my mind. Most of them cover solely the usage of `tray_output primary` and actually handle `tray_output <monitorname> ...` (and disabling other outputs) as a cornercase. So I dedcided between those two paths:

- Make the `tray_output` a soft setting (status quo of this PR): No matter what happens (you change GPU, swap the screens, clone your dotfiles to somewhere else, ...), you'll get a traybar. It's just there and it works. No hassle involved. You'll get a traybar. If you want it to be refined, e.g the left or right screen, you can change it via `tray_output`. The only con: You can't explicitly say anymore: "I don't want a traybar!"
- Let the `tray_output` a hard setting. (Status quo of `master`). The `tray_output` governs. If the `tray_output` is set to `VGA1` and your machine hasn't even a VGA adapter anymore, nothing will display. Also, if you're setting `tray_output primary` and your primpary display is your external one and while you take away your laptop, the primary display won't get updated to point onto your notebook's LCD, no tray output will be displayed.

So, which solution should we choose? And: What needs should we fulfill?

What would you prefer as a user? You can only have a single tray output. Would you display anyways it or rather not?! And we're talking about the tray output, not the bar itself (which could be displayed multiple times). (This is not a rethoric question).

Airblader:
Without a doubt I prefer the current behavior. It avoids the risk of breaking things for users and to me it makes sense. If people configure primary as the output but don't have a primary output, why *should* the bar appear anywhere? It's not what they configured. They always have the option to specify multiple outputs.

So all in all I think we can close this PR. However, thanks for opening it in any case!

PR #3340, opened 2018-07-29
fix height offset calculation in pango text drawing

### LINKED ISSUE #3114: Displaying Unicode Thermometer (U+1F321) in i3bar causes text to be shifted up

Running i3 version: 4.14.1 (2017-09-24) (pid 788) abort…)
Loaded i3 config: /home/joel/.config/i3/config (Last modified: Mon 08 Jan 2018 10:11:19 CET, 25 seconds ago)

https://pastebin.com/Hi7QAcjA
i3bar log:
https://pastebin.com/rPCDbTDZ

When displaying Unicode Thermometer, the text is shifted up by about half a line in i3bar.
example: 
i3 config:
```
bar {
   status_command /path/to/test.sh
}
```
test.sh:
```
echo '{"version":1} ['
while true
do
	echo '[{"full_text":"asdf"},{"full_text":"🌡 temperature"},{"full_text":"asdf"}],'
	sleep 1s
done
```
also works with non-json input,
i.e. just `echo "🌡 temperature"`

PandorasFox:
* difference is zero: dividing by two does nothing (that's fine)
* difference is positive: halve the difference, subtract it from the height to center it
* difference is negative: `abs` it, then do the same as above.

Works fine on my fresh laptop; dunno about older pango versions. I'm not exactly sure about what changed with things that are larger than given height.

Unless I'm mistaken, this shouldn't affect anything where `height <= savedFont->height`. This just changes the previous case of subtracting the (positive) height difference when something was too tall (and thus shifting everything a decent amount above the bar canvas).

This handles it *well enough* for the case "text too tall". I'm not really sure if there's a "good" solution for this, since it's usually just one glyph that's messing up the height.

Fixes #3114

### DIFF
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -111,7 +111,7 @@ static void draw_text_pango(const char *text, size_t text_len,
     pango_layout_get_pixel_size(layout, NULL, &height);
     /* Center the piece of text vertically if its height is smaller than the
      * cached font height, and just let "high" symbols fall out otherwise. */
-    int yoffset = (height < savedFont->height ? 0.5 : 1) * (height - savedFont->height);
+    int yoffset = abs(height - savedFont->height) / 2;
     cairo_move_to(cr, x, y - yoffset);
     pango_cairo_show_layout(cr, layout);
 


PandorasFox:
This is what my bar looks like _with_ the patch:

image

This is what it looks like _without_ the patch:

image

I don't particularly like this since the text in the (formerly broken) boxes isn't aligned the same as the text in 'normal' boxes. I don't think there's a real way to do that, though.

PandorasFox:
```C
    int yoffset = (height < savedFont->height ? 0.5 : 0.4) * (height - savedFont->height);

```

Produces the following:

image


I think that this is a bit closer to what's expected. I'm unsure if `0.4` is really going to work the same on every device. On my display the text appears to be perfectly aligned.

Airblader:
Thanks for the PR. I think as is we can merge it. I don't like the second proposed version because it can't really be justified as it's mostly guess work. I think for that we'd have to investigate closely with pango how this would have to be handled correctly.

PandorasFox:
Yeah, if I could repro on multiple machines it'd be easier to figure out why this happens and figure out a proper fix, but I can't get `height > savedFont->height` to trigger on any of my other machines, weirdly enough.

I also can't even get this to trigger on the machine this was happening on before, either - maybe it's some weird caching issue with newly-installed fonts or something.

PR #3342, opened 2018-07-30
Simplify tree_close_internal

orestisf1993:
I'll be running this patch on my machine to catch any possible regressions.


Airblader:
I assume you ran it for a while now? Is it good to go from your side?

orestisf1993:
Yes

PR #3370, opened 2018-08-22
Set WM_STATE to normal when focusing a window

Streetwalrus:
Some applications request iconic state when they lose focus. One notable example is games running in Wine that use the Windows exclusive full-screen API.
While this is perfectly sound behavior on a floating window manager, we do not support the concept of iconized windows, so not only do we not handle this case, but there is also no sane way to escape that state. This leaves the aforementioned games in a paused state, with the contents reduced to a tiny corner of the window, and with no audio playing. The workaround, up until now, was to enable Wine's virtual desktop, but that is pretty inconvenient to use.
To remedy this, reset WM_STATE to a normal state when the window is focused. This allows the application to do whatever it wants to do while it's not focused (pause the game, let the GPU idle, etc), and avoids breaking it with a simple workspace switch.

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -1168,6 +1168,13 @@ void x_push_changes(Con *con) {
 
                 change_ewmh_focus((con_has_managed_window(focused) ? focused->window->id : XCB_WINDOW_NONE), last_focused);
 
+                /* Some applications request iconic state when they lose focus, revert them to normal when they regain focus. */
+                if (focused->window != NULL) {
+                    long data[] = {XCB_ICCCM_WM_STATE_NORMAL, XCB_NONE};
+                    xcb_change_property(conn, XCB_PROP_MODE_REPLACE, focused->window->id,
+                                        A_WM_STATE, A_WM_STATE, 32, 2, data);
+                }
+
                 if (to_focus != XCB_NONE && to_focus != last_focused && focused->window != NULL && is_con_attached(focused))
                     ipc_send_window_event("focus", focused);
             }


psychon:
According to a quick search through the code, i3 never sets `WM_STATE` to iconic. To go from normal to iconic, a window has to send a ClientMessage to the root window / to the WM, which will then trigger the state transition. Thus, this does  not happen with i3 either.

So, how does `WM_STATE` actually end up being set to iconic?

Streetwalrus:
If I understand [the documentation](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.4) correctly, applications are free to set this on their own.

psychon:
§ 4.1.3.1 begins with "The window manager will place a WM_STATE property". From a quick search, I cannot find anything any more explicit, but: Nope, clients really should not ever change the `WM_STATE` property. For example, look at how to transition from withdrawn to iconic: The window is mapped with `WM_HINTS.initial_state` being iconic. If apps were allowed to change this property directly, this could be done a lot easier by, well, changing the property directly.

No property may be modified by both the WM and the application, because in this case there can be races where they overwrite each other's properties.

Oh, I found an explicit reference: § 4.1.3 is "Window Manager Properties" and § 4.1.2 is "Client Properties". § 4.1.3 begins with:

> The properties that were described in the previous section are those that the client is responsible for maintaining on its top-level windows. This section describes the properties that the window manager places on client's top-level windows and on the root.

Streetwalrus:
Digging into Wine's code, it is [setting the initial state hint](https://github.com/wine-mirror/wine/blob/83e481fee20bc5c6598e65f8295be1b3f11fb70b/dlls/winex11.drv/window.c#L803), or uses [XIconifyWindow](https://github.com/wine-mirror/wine/blob/83e481fee20bc5c6598e65f8295be1b3f11fb70b/dlls/winex11.drv/window.c#L2448). The latter sends a WM_CHANGE_STATE message (which i3 does receive, but not handle, though it [used to do bogus things with it](https://github.com/i3/i3/commit/d551618cf03695b765b4f6d7f30df8e01847b79f)) and also sets that same hint.
I can't seem to find a more proper way for the window manager to undo what Wine does. Maybe it's their bug, but I don't think so.

psychon:
https://github.com/wine-mirror/wine/blob/58c49279f5d5ac11c0af25053f47845203dffdec/dlls/winex11.drv/event.c#L1255
Looks like it would be enough to handle `WM_CHANGE_STATE` client messages via the following code to make it appear to wine that the window was immediately un-minimized:
```
                    long data[] = {XCB_ICCCM_WM_STATE_NORMAL, XCB_NONE};
                    xcb_change_property(conn, XCB_PROP_MODE_REPLACE, window->id,
                                        A_WM_STATE, A_WM_STATE, 32, 2, data);
```
However, I'm not sure if this will actually work as intended. It might just mean that the wine app will grab inputs immediately again.

I consider "let's pretend to you that you are minimized until you get the focus again" to be a weird hack, but I do not get to make any decisions here, so feel free to ignore me.

Airblader:
My understanding of the spec is the same as @psychon's, that is the client shouldn't be setting this hint. To avoid misunderstandings, is anyone reading this differently?

For that reason my immediate feeling here is to reject the PR and fix the client instead.

Streetwalrus:
That's also something I've considered doing. Wine doesn't seem to grab input again afterwards, though there is a visible flicker through the iconized state, which is unavoidable.
Wine seems to be following the spec (it's using XIconizeWindow to iconize itself, and reverts to normal state when the property is reset by the window manager), so we might be looking at an xlib bug?

psychon:
@Airblader Thanks to the links to wine's code that @Streetwalrus provided, it looks like Wine assumes that requesting its window to be iconified means that it is guaranteed that this actually happens. As soon as Wine sent the request, it starts ignoring things (e.g. ignores `ConfigureNotify` events). I don't think anything actually modifies `WM_STATE` directly here, it's just Wine waiting for the WM to tell it to proceed.
Since i3 just ignores the request, wine keeps believing its window is minimized.

Streetwalrus:
Yeah you're right, WM_STATE never actually gets set to iconic. I'll report this to Wine devs.

Airblader:
Yeah, that makes sense. I think the question here is whether the window manager is allowed to reject the request. The spec never explicitly mentions this and there's no protocol to indicate support or lack thereof. However, the general paradigm in X is typically that a client mustn't assume that the window manager grants anything the client requests of it.

Even assuming we're forced to implement support, it raises the question of what that looks like. We simply do not have and do not want minimized windows. Unfortunately, the icon state specifies that the client can assume their icon or at least name is displayed. This seems to force window management behavior which would be really odd.

I think the proposed fix isn't really correct, it just restores interaction, but in the meantime there's still the problem that the client assumes a state that isn't actually there. 

I would suggest to raise this to Wine and get their thoughts on why they can't see if the window manager grants the request, as is common in X anyway. What do you think?

Streetwalrus:
Yeah that's the conclusion I've reached, which is why I closed the PR. Wine should request iconization with XIconizeWindow, and then instead of assuming that the window has been iconized, handle the actual change on WM_STATE to go into iconized mode. That way, the window manager gets notified of wine's request and can decide to fulfill it or not.

Streetwalrus:
Wine ticket: https://bugs.winehq.org/show_bug.cgi?id=45690

PR #3372, opened 2018-08-23
Resize tile px

### LINKED ISSUE #3239: [Enhancement] Resize tiled windows by px instead of ppt

orestisf1993:
- Support for `resize grow|shrink left|right|up|down <px> px` (#3240)
- Support for `resize grow|shrink width|height <px> px`
- Support for `resize set <px> px`
- More accurate `resize set <ppt> ppt`

Fixes #3239
Closes #3240

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2330,12 +2330,12 @@ resize set [width] <width> [px | ppt] [height] <height> [px | ppt]
 -------------------------------------------------------
 
 Direction can either be one of +up+, +down+, +left+ or +right+. Or you can be
-less specific and use +width+ or +height+, in which case i3 will take/give
-space from all the other containers. The optional pixel argument specifies by
-how many pixels a *floating container* should be grown or shrunk (the default
-is 10 pixels). The ppt argument means percentage points and specifies by how
-many percentage points a *tiling container* should be grown or shrunk (the
-default is 10 percentage points).
+less specific and use +width+ or +height+, in which case i3 will take/give space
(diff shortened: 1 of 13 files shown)


andrewla:
Thanks for picking this up, @orestisf1993!

PR #3400, opened 2018-09-13
Respect max size from WM_NORMAL_HINTS

Synray:
This change will only affect windows that would already qualify as "dialog"
windows.

Some programs make optimizations based on the assumption that the window size
won't change, or don't properly handle resizes -- for example, quake freezes if
the window gets resized too often. Many OpenGL/Vulkan applications make similar
assumptions, since resizing means synchronizing, then recreating the swap chain
and the framebuffer.

This is my first contribution to i3 -- although I've used it since installing
Arch -- so if there is a better way or this is just a bad idea, please let me
know and I'll make the changes/drop it.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -477,6 +477,10 @@ struct Window {
     int min_width;
     int min_height;
 
+    /* Maximum size specified for the window. */
+    int max_width;
+    int max_height;
+
     /* aspect ratio from WM_NORMAL_HINTS (MPlayer uses this for example) */
     double aspect_ratio;
 };
--- a/src/floating.c
+++ b/src/floating.c
@@ -96,6 +96,18 @@ void floating_check_size(Con *floating_con) {
             floating_con->rect.height += border_rect.height;
         }
 
+        if (focused_con->window->max_width) {
+            floating_con->rect.width -= border_rect.width;
+            floating_con->rect.width = min(floating_con->rect.width, focused_con->window->max_width);
+            floating_con->rect.width += border_rect.width;
+        }
+
+        if (focused_con->window->max_height) {
+            floating_con->rect.height -= border_rect.height;
+            floating_con->rect.height = min(floating_con->rect.height, focused_con->window->max_height);
+            floating_con->rect.height += border_rect.height;
+        }
+
         if (focused_con->window->height_increment &&
             floating_con->rect.height >= focused_con->window->base_height + border_rect.height) {
             floating_con->rect.height -= focused_con->window->base_height + border_rect.height;
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -981,9 +981,18 @@ static bool handle_normal_hints(void *data, xcb_connection_t *conn, uint8_t stat
         con->window->min_height = size_hints.min_height;
     }
 
+    if ((size_hints.flags & XCB_ICCCM_SIZE_HINT_P_MAX_SIZE)) {
+        DLOG("Maximum size: %d (width) x %d (height)\n", size_hints.max_width, size_hints.max_height);
+
+        con->window->max_width = size_hints.max_width;
+        con->window->max_height = size_hints.max_height;
+    }
+
     if (con_is_floating(con)) {
         win_width = MAX(win_width, con->window->min_width);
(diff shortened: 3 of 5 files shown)


Synray:
I haven't written much in perl, but I'll look at the test cases, and see what I can do.

Synray:
I pushed my initial work on some tests. I'm having trouble figuring out why tests 5-6 fail. 
Tests 5 and 6 `resize set` a floating window to exactly its requested `max_width` and `max_height`.
I think somewhere in the code, the floating window's width gets subtracted by `border_width`.

Or am I just using `cmd 'resize'` wrong?

Airblader:
@orestisf1993 Can you help @Synray out here? I'm in pre-vacation stress right now. Feel free to merge it as well once done!

orestisf1993:
I've been investigating this for a bit. Don't know if I'll have time to solve this in the next couple of days though.

Have a nice vacation ;)

Synray:
@Airblader Have fun :)

@oretisf1993 It's also likely that I'm misunderstanding perl, since I've just started reading through "Modern Perl" yesterday. I'll keep working on it but if you figure it out please let me know, thanks!

Airblader:
I'm also OK with merging this without tests given the small change, no reason to reject it over it. We can also open an issue to add tests for this later on.

orestisf1993:
Or we could just test using `border none` like `189-floating-constraints.t` does (@Synray btw move your tests there).

Then, open an issue about correcting calculations.

testcases/t/301-floating-max-size.t
```diff
@@ -0,0 +1,123 @@
...
+my $ws = fresh_workspace;
+my $window;
+
+my $min_width = 150;
+my $min_height = 100;
+
+my $max_width = 250;
+my $max_height = 200;
+
+sub change_normal_hints {
```
orestisf1993: This function is not needed.

testcases/t/301-floating-max-size.t
```diff
@@ -0,0 +1,123 @@
+#!perl
```
orestisf1993: Move the tests to `189-floating-constraints.t`

testcases/t/301-floating-max-size.t
```diff
@@ -0,0 +1,123 @@
...
+    is($window->rect->{width}, $min_width, 'width = min_width');
+    is($window->rect->{height}, $min_height, 'height = min_height');
+}
+
+sub check_maxsize {
+    is($window->rect->{width}, $max_width, 'width = max_width');
+    is($window->rect->{height}, $max_height, 'height = max_height');
+}
+
+sub do_test {
```
orestisf1993: Since we'll only test with `border none`, remove this function as well.

testcases/t/301-floating-max-size.t
```diff
@@ -0,0 +1,123 @@
...
+# Test borderless
+cmd 'border none';
+do_test;
+
+# Test with 1-px-border
+cmd 'border 1pixel';
+do_test;
+
+
+#is(get_ws($ws)->{floating_nodes}[0]->{nodes}[0]->{window}, $window->id, 'Fixed size window opened floating');
```
orestisf1993: Remove this commented-out line and the extra newlines above

orestisf1993:
The reason for the failure is actually fairly simple, `resize set` modifies the con's rect but your code puts restrictions on the window (as it should). I had a mental hiccup the other day because I didn't understand why the test didn't fail for the min restrictions but that's simple as well. Examples with `border pixel 5`, max size 500x500, min size 200x200:
- `resize set 1000 1000` means window size 995x995: window is reduced to the maximum of 500x500, con rect becomes 510x510
- `resize set 500 500` means window size 495x495: no correction
- `resize set 200 200` means window size 195x195: window is increased to minimum of 200x200, con rect become 210x210

Synray:
That makes much more sense. Thanks for clearing that up!
I moved the tests to `189-floating-constraints.t`, anything I should add?

orestisf1993:
Ok, made the changes myself. Thanks for your work!

Issue #2587, opened 2016-11-25
add command to (focus / move) sibling containers


alexherbo2:
It would be nice to have a command to (focus / move) (next / previous) sibling container, like with the mouse, by scrolling on title bar.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
Sorry, I don't understand. Can you please specify your usecase in more detail? In particular, we already have `focus next|prev`.

alexherbo2:
I would like to bind `Super + Tab` to focus next sibling container.

I mainly use tabbed layout, so it’s binded to `focus right`.

I was thinking it would be a nice command to add, as `focus <orientation>` depends on layout presentation.

Airblader:
`focus left|right|up|down` pretty much already works like this by using "prev" and "next", just that the direction is attached to that.

I think adding `focus prev|next` with the semantics of (dynamically) using the orientation of the parent container sounds good to me. @stapelberg Thoughts?

stapelberg:
I think we need to be careful to make it clear what the scope of that command is, i.e. the container. People sometimes ask for an alt+tab feature where i3 cycles through all windows, but we really do not want to get into the business of defining an order of windows. Hence, I suggest calling the command `focus prev_in_container|next_in_container`. Thoughts?

Airblader:
I don't necessarily think `prev|next` is an issue because a command will do what it's documented to do and it's not like it will work for a while and suddenly work differently. But using your suggested name(s) is fine with me, too.

cstyan:
I'd like to take a shot at implementing this but I don't fully understand the feature. I'm assuming focus next sibling could be any child of the current containers parent container? How deep down the tree could this go?

nmschulte:
@cstyan, I believe @stapelberg clarified that this should only inspect one level away: the parent of the focused window.

But, it seems like the scroll wheel behavior does some sort of "auto-folding" (multiple depths in a single action), where one might need to also throw in a `focus parent` in addition to `prev`/`next` (directed or inferred) to get the same behavior via keys (which means to me: multiple strokes/actions).

cstyan:
[...]

Issue #1573, opened 2015-03-24
Relative focus in `for_window` not working


acrisci:
Commands that focus relative to the current window (such as `focus [direction]`) no longer work with `for_window`. This is a regression in version 4.8.

The problem is that when the window is managed,the `for_window` command runs before focus is set on the window.

To reproduce, put this in your config:

```
for_window [class="URxvt"] focus left
```

and open a terminal window. Focus remains unchanged, while it should move to the left.

See also #1319 and #1435

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

Airblader:
The problem is easy, but the solution might be tricky. The workflow is

1. Run assignments.
2. Fetch the workspace again since assignments may have moved the window.
3. Render the workspace if it's not visible and disable setting the focus.
4. Send the `new` event
5. Set the focus if feasible

We have to get 1. behind 5. somehow, but 5. depends on 4. and 3. and 3. depends on 2. which depends on 1.

saosr:
I just reproduced the problem with the next version:

"i3 version 4.10.1 (2015-03-29, branch "4.10.1") © 2009-2014 Michael Stapelberg and contributors"

PR #3407, opened 2018-09-15
tree_next refactor & enhancements

### LINKED ISSUE #2180: [feature request] Restrict focus left/right to the active output screen in multi-output setup
Hey, thanks for the great work on this excellent WM!

I think it would be great to have an option to limit _focus left/right_ only to the currently active output (the one that has a focused window). I think this optional behaviour could work the same as if there was only one output where _focus left/right_ just wraps to last/first container on the output.

orestisfl:
Still needs some work but I'd like to confirm that you agree with the direction first.

- Makes `tree_next` not recursive
- Implements `focus wrapping workspace` (maybe `focus wrapping output`? maybe both?) - Fixes #2180
- Implements `focus next|prev sibling` like (1.) in https://github.com/i3/i3/issues/2587#issuecomment-378505551 (Issue also requests move command, not implemented here)
- Implements `focus next|prev` like (2.) in https://github.com/i3/i3/issues/2587#issuecomment-378505551
- Directional focus command now supports command criteria. ~~This is needed for #1573.~~

TODO:
- [x] Document `focus wrapping workspace`
- [x] Test `focus wrapping workspace`
- [x] More tests for `focus wrapping` in general. Especially `force`.
- [x] Document `focus next|prev`
- [x] Test `focus next|prev`
- [x] Document `focus next|prev sibling`
- [x] Test `focus next|prev sibling`
- [x] Test `<criteria> focus <direction>`
- [x] Improve commit messages
- [x] ~~Fix for #1573 was kinda a random thought, need to investigate if the solution is a good idea. If yes, add tests.~~
- [x] Verify that the testsuite passes at each commit, not just the final one

### DIFF
--- a/include/commands.h
+++ b/include/commands.h
@@ -183,6 +183,12 @@ void cmd_exec(I3_CMD, const char *nosn, const char *command);
 void cmd_focus_direction(I3_CMD, const char *direction);
 
 /**
+ * Implementation of 'focus next|prev sibling'
+ *
+ */
+void cmd_focus_sibling(I3_CMD, const char *direction);
+
+/**
  * Implementation of 'focus tiling|floating|mode_toggle'.
  *
  */
--- a/src/commands.c
+++ b/src/commands.c
@@ -1213,6 +1213,21 @@ void cmd_exec(I3_CMD, const char *nosn, const char *command) {
     ysuccess(true);
 }
 
+#define CMD_FOCUS_WARN_CHILDREN                                                        \
+    do {                                                                               \
+        int count = 0;                                                                 \
+        owindow *current;                                                              \
+        TAILQ_FOREACH(current, &owindows, owindows) {                                  \
+            count++;                                                                   \
+        }                                                                              \
+                                                                                       \
+        if (count > 1) {                                                               \
+            LOG("WARNING: Your criteria for the focus command matches %d containers, " \
+                "while only exactly one container can be focused at a time.\n",        \
+                count);                                                                \
+        }                                                                              \
+    } while (0)
+
 /*
  * Implementation of 'focus left|right|up|down'.
  *
@@ -1315,12 +1330,15 @@ void cmd_focus(I3_CMD) {
         ELOG("Example: [class=\"urxvt\" title=\"irssi\"] focus\n");
 
         yerror("You have to specify which window/container should be focused");
-
+        return;
+    } else if (TAILQ_EMPTY(&owindows)) {
+        yerror("No window matches given criteria");
         return;
     }
 
+    CMD_FOCUS_WARN_CHILDREN;
+
     Con *__i3_scratch = workspace_get("__i3_scratch", NULL);
-    int count = 0;
     owindow *current;
     TAILQ_FOREACH(current, &owindows, owindows) {
         Con *ws = con_get_workspace(current->con);
@@ -1332,7 +1350,6 @@ void cmd_focus(I3_CMD) {
         /* In case this is a scratchpad window, call scratchpad_show(). */
         if (ws == __i3_scratch) {
             scratchpad_show(current->con);
-            count++;
             /* While for the normal focus case we can change focus multiple
              * times and only a single window ends up focused, we could show
              * multiple scratchpad windows. So, rather break here. */
@@ -1341,16 +1358,10 @@ void cmd_focus(I3_CMD) {
 
         LOG("focusing %p / %s\n", current->con, current->con->name);
         con_activate_unblock(current->con);
-        count++;
     }
 
-    if (count > 1)
-        LOG("WARNING: Your criteria for the focus command matches %d containers, "
-            "while only exactly one container can be focused at a time.\n",
-            count);
-
     cmd_output->needs_tree_render = true;
-    ysuccess(count > 0);
+    ysuccess(true);
 }
 
 /*
--- a/src/commands.c
+++ b/src/commands.c
@@ -1229,23 +1229,62 @@ void cmd_exec(I3_CMD, const char *nosn, const char *command) {
     } while (0)
 
 /*
- * Implementation of 'focus left|right|up|down'.
+ * Implementation of 'focus left|right|up|down|next|prev'.
  *
  */
-void cmd_focus_direction(I3_CMD, const char *direction) {
-    switch (parse_direction(direction)) {
-        case D_LEFT:
-            tree_next('p', HORIZ);
-            break;
-        case D_RIGHT:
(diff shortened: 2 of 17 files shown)


include/commands.h
```diff
@@ -181,6 +181,7 @@ void cmd_exec(I3_CMD, const char *nosn, const char *command);
  *
  */
 void cmd_focus_direction(I3_CMD, const char *direction);
+void cmd_focus_sibling(I3_CMD, const char *direction);
```
orestisf1993: No docstring

src/commands.c
```diff
@@ -1218,24 +1218,78 @@ void cmd_exec(I3_CMD, const char *nosn, const char *command) {
     ysuccess(true);
 }
 
+#define CMD_FOCUS_WARN_CHILDREN                                                        \
```
orestisf1993: Better name?

susloparovdenis:
No movement? Really want this feature

tbodt:
@stapelberg

Airblader:
I think the person to tag is @orestisf1993 to see if he wants to continue with this. :-)

orestisf1993:
I want to continue working on this later but:

> I'd like to confirm that you agree with the direction first

Airblader:
Yes, I think it's totally fine and amazing work. Of course we'll have to throughly test, but obviously you know that just as well. :-)

stapelberg:
Not entirely sure if you’re asking me, but in case you are: I’m not attached to the current implementation, so no objection from my end on refactoring it. Ideally, this would be merged shortly after a new release to maximize testing from people running the next branch.

orestisfl:
Ready

orestisfl:
> I had a dilemma about the behavior here:
> 1. Prohibit focus leaving the workspace in any case unless if get_tree_next's initial argument is a workspace. This is what this commit does (also i3-cycle).
> 2. Leave the workspace if no warp is possible (eg workspace with single container or `focus right` with `V[a b c*]`).

btw this might be inconsistent with the logic of the `force` option. With `force`, focus can leave the parent if the directional movement is different from the orientation (ie can't wrap). With `workspace` focus can't ever leave the workspace even if there are no targets to wrap to.

But, I believe this is what most people want in #2180 and the inconsistency is ok because workspaces are much bigger structures and restricting focus to be only inside them makes more sense.

stapelberg:
I have not worked with this bit of code in a long time, so I don’t feel up to doing a review. If @Airblader can’t review it, I’m willing to step in, but be prepared for slow response times.

Airblader:
I mainly assigned you as well because I figured this could well use two reviews. I'm also not very familiar with this code. :-)

stapelberg:
Testcase + community-testing sounds like the best strategy here. Let’s go ahead with it :)

ia0:
I'm trying to find out how to have focusing and movement that respect container tree and outputs, but I can't. I found about `focus_wrapping no` and this PR which is why I'm asking here. I would like to know about the following questions in that order (see below for the definition of "is it possible to have"):
1. Is it possible to have `focus output left/right` not wrap outputs? (see case 5 below)
2. Is it possible to have `focus left/right` not exit outputs? (see case 3 below)
3. Is it possible to have a border for stack and tab implicit containers? For vertical and horizontal containers we already have the "indicator" color as a hint. For stack and tab, if "default_border" is pixel, the border is not added the container is implicit. It only is when a second child is added.
4. Is it possible to have the "indicator" color for containers? Currently, when focusing a container with `focus parent`, all borders of all windows under the container have the "child_border" color.
5. Is it possible to have `move parent` and `move child left/right/up/down`? This is necessary for next question. (see cases 11 and 13 below)
6. Is it possible to have `focus left/right/up/down` not exit containers? In other words, focusing and movement should respect the tree structure. (see cases 6 and 12 below)
7. Is it possible to have `focus left/right/up/down` focus the container instead of its last active window? (see case 9 below)

```
Multiple outputs:
 1: H[a* b] c     {focus left}     H[a* b] c :no wrap (works with focus_wrapping no)
 2: H[a* b] c     {focus right}    H[a b*] c :focus inside container (works)
 3: H[a b*] c     {focus right}    H[a b*] c :no exit output (doesn't work)
 4: H[a b*] c {focus output right} H[a b] c* :focus between outputs (works)
 5: H[a b] c* {focus output right} H[a b] c* :no wrap output (doesn't work)
Single output:
 6: H[H[a b*] c]   {focus right}     H[H[a b*] c] :no exit container (doesn't work)
 7: H[H[a b*] c]   {focus parent}    H[H[a b]* c] :focus parent
 8: H[H[a b]* c]   {focus right}     H[H[a b] c*] :focus inside container (works)
 9: H[H[a b] c*]    {focus left}     H[H[a b]* c] :no automatic focus child (doesn't work)
10: H[H[a b]* c]   {focus child}     H[H[a b*] c] :focus last active child (works)
11: H[H[a b*] c]   {move parent}     H[H[a] b* c] :move parent (doesn't work)
12: H[H[a] b* c]    {move left}      H[b* H[a] c] :swap with left (doesn't work)
13: H[b* H[a] c] {move child right}  H[H[b* a] c] :insert in container (doesn't work)
```

The definition of "is it possible to have" is:
- "is it possible to configure i3 such that", if not
- "would it make sense to implement", if not
- "is it possible to use the IPC interface to", if not
- "would it make sense to add the primitive commands that would be needed to".

I care about questions 1 and 2. Questions 3 and 4 would be helpful to have. Questions 5 to 7 are for a more primitive, local, and strict mental model (i.e. easier to predict, less dependent of the global state, and without recursion).

Thanks a lot!

Al-Caveman:
@orestisfl What do you think about changing the syntax from `focus DIR sibling` into `focus sibling DIR`?

I think it makes command parsing simpler when introducing variants `focus VARIANT DIR` in the future (e.g. #3859).

I think if you do this change from now (before it gets into stable), we will have a neater path in the future with less worries about hacking things around to maintain backwards compatibility.  

But I am not sure.  Any thoughts?

Airblader:
We can also consider using `--xyz` flags which can be position-independent.

Al-Caveman:
Is it possible to:

* Make the next/prev wrap around when we reach the end of the tile (as opposed to getting saturated at the last container as it is done now)?  I think this is more efficient as sometimes the shortest path is to wrap around?

* Let next/prev continue shuffling containers to child stacked/tabbed containers, which would give the feeling that the hierarchical tabs/stacks are just tabs on the same level, just looking hierarchical?  Currently, all the child containers get the focus (as if mod+a is pressed).

PR #3410, opened 2018-09-15
docs/i3bar-protocol: Mention skipping blocks with empty full_text 

### LINKED ISSUE #3405: i3bar does not respect min_width with an empty full_text
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

When a user wants to adds "spacers" in the bar, it is intuitive to write a block with the expected **min_width** and an empty **full_text**. But i3bar seems to ignore the block if its *full_text* key is an empty string. The workaround is to add for example a single space, which does not make sense.

## Expected Behavior

**full_text** is the only mandatory key for a block (without it a *SPEC VIOLATION* is thrown) and it's totally valid (and currently supported) to provide an empty *full_text*, so i3bar must respect the *min_width* property if it was passed along within the block.

## Reproduction Instructions

Please add the following *bar* block to your i3 config:

````
bar {
  status_command "
    echo '{\"version\":1,\"click_events\":true}[[]'
    while sleep 1
    do echo ',[{\"full_text\":\"My padded text\",\"separator\":false},{\"min_width\":200,\"full_text\":\"\"}]'
    done
  "
}
````

## Environment

<pre>
Running i3 version: 4.15 (2018-03-10) (pid 1029)o abort…)
Loaded i3 config: /home/vivien/.config/i3/config (Last modified: Fri Sep 14 15:52:18 2018, 1154 seconds ago)

</pre>

<pre>
- Linux Distribution & Version: 4.18.5-arch1-1-ARCH
</pre>

This may to be related to #2768 which emphasizes on the *short_text* property.

orestisf1993:
Closes #3405.

### DIFF
--- a/docs/i3bar-protocol
+++ b/docs/i3bar-protocol
@@ -118,7 +118,8 @@ click_events::
 
 full_text::
 	The +full_text+ will be displayed by i3bar on the status line. This is the
-	only required key.
+	only required key. If +full_text+ is an empty string, the block will be
+	skipped.
 short_text::
 	Where appropriate, the +short_text+ (string) entry should also be
 	provided. It will be used in case the status line needs to be shortened


Airblader:
I'm not convinced that we should fix this in the code. Blocks with empty text being skipped has been the behavior eversince I've been with i3, and I do know that people rely on this to hide blocks (although maybe not in combination with min_width).

Opposed to the issue, I am not convinced that printing a space is all that bad to avoid this, so I wonder whether updating the documentation to mention skipping empty blocks would be less intrusive.

(This isn't a decision, just a thought)

orestisf1993:
> and I do know that people rely on this to hide blocks

Do people rely on empty text or do they completely exclude blocks from their output if they want to hide them?

orestisf1993:
Makes sense. I've updated the documentation. Do you want to also mention something under `min_width`?

vivien:
OK that's fine with me then. Even though it is not that intuitive for a "spacer" block, it is now documented that one must provide at least one space character as *full_text* to add such block in the status line.

PR #3415, opened 2018-09-20
Check g_utf8_make_valid availability

orestisf1993:


Airblader:
We usually try to stay away from conditional compilatiion. Is there a reason why we can't just update the required dependency version?

I think I'd like @stapelberg to weigh in on whether the conditional would be OK here.

stapelberg:
glib 2.52 is not yet in Debian stable, so it’s not *that* widely available yet.

Could we copy the g_utf8_make_valid implementation and use it if glib is too old instead of disabling the feature? It’s not that big a function: https://sources.debian.org/src/glib2.0/2.56.1-2/glib/gutf8.c/?hl=1773#L1773

orestisf1993:
I think copying LGPL code (instead of linking an external library) means that we should relicense as LGPL:

>  The precise terms and conditions for copying, distribution and modification follow. Pay close attention to the difference between a "work based on the library" and a "work that uses the library". The former contains code derived from the library, whereas the latter must be combined with the library in order to run.
> …
> These requirements apply to the modified work as a whole. If identifiable sections of that work are not derived from the Library, and can be reasonably considered independent and separate works in themselves, then this License, and its terms, do not apply to those sections when you distribute them as separate works. But when you distribute the same sections as part of a whole which is a work based on the Library, the distribution of the whole must be on the terms of this License, whose permissions for other licensees extend to the entire whole, and thus to each and every part regardless of who wrote it.

stapelberg:
IIUC, instead of copying the code in question into our source, we should re-distribute the code as a separate file, ideally as-is. 

See also https://github.com/requests/requests/issues/3389

orestisf1993:
Ok, we certainly fulfill "If identifiable sections of that work are not derived from the Library, and can be reasonably considered independent and separate works in themselves" but I don't know if we "distribute them as separate works" by putting it in a separate file that gets compiled in our binary and is distributed with the rest of the project.

Anyway, I've pushed the function in a separate file, tell me if it looks ok to you.

stapelberg:
Yeah, this looks good, thanks! If it was wrong after all, we can always revert.

PR #3421, opened 2018-09-25
Reject requests for WM_STATE_ICONIC

Streetwalrus:
For compatiblity reasons, Wine will request iconic state and cannot ensure that the WM has agreed on it; immediately revert to normal to avoid being stuck in a paused state.

This is a follow-up on #3370. As suggested by @psychon and discussed in the [Wine bug report](https://bugs.winehq.org/show_bug.cgi?id=45690), I've moved this to the notification handler to immediately tell clients that we do not want them to iconize. I've also made it clear that this workaround is specific to Wine.


psychon:
@Airblader `WM_CHANGE_STATE` comes from ICCCM and not from EWMH. `_NET_SUPPORTED` comes from EWMH.

From "Introduction" / "What is this spec?" in EWMH:

> [This spec] builds on the Inter-Client Communication Conventions Manual [ICCCM], 

So, since this is not an EWMH-feature, you cannot expect it to be listed in `_NET_SUPPORTED` (and I bet nothing does so). Also, since this is an ICCCM-feature, i3 does not get to not support this and still claim to be a window manager (ICCCM is the greatest common denominator, afterwards there is just X11 left and nothing will work).
Sorry.

Airblader:
That's a good point. I guess we should also not add it to _NET_SUPPORTED then, though. I don't like workarounds for clients (it's not i3's fault that wine has to deal with Windows), but I guess in this case there isn't much harm in it either.

@Streetwalrus Do you think we could add a testcase for this, though?

Streetwalrus:
I'm not sure if this is testable. Besides the notification, there's no way to figure out what state the application is in as far as I know.

Airblader:
Let's give it a go.

Streetwalrus:
Thanks!

PR #3430, opened 2018-09-29
[RFC] Correctly calculate max_aspect

Synray:
Hello again,

This looks like a copy/paste error to me, and so it's a simple fix,
but I might be wrong about that.

It passes all tests, but just in case, should I write one?
Thanks


orestisf1993:
Thanks, tests and more stuff in #3432

PR #3435, opened 2018-10-03
i3-msg/subscribe

vivien:
This PR implements the support for the SUBSCRIBE message type in i3-msg. This allows trivial scripts to use i3-msg to listen to i3 events such as workspace changes, focused windows, etc. The `-t SUBSCRIBE` option allows to subscribe to the specified payload. A new `-m` option allows to monitor indefinitely the subscribed events instead of returning right away (similarly to inotifywait).

Here are a few simple examples made possible with this new addition.

Wait until the next workspace is focused and return its number:
````
i3-msg -t subscribe '[ "workspace", "focus" ]' | jshon -e current -e num
````

Add a simple bar in the i3 configuration file printing the focused window name:
````
bar {
  position top
  status_command i3-msg -t subscribe -m '[ "window", "focus" ]' | \
      jq --unbuffered '"Window title: \(.container.name)"'
}
````

### DIFF
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -246,9 +246,6 @@ int main(int argc, char *argv[]) {
         err(EXIT_FAILURE, "IPC: write()");
     free(payload);
 
-    if (quiet)
-        return 0;
-
     uint32_t reply_length;
     uint32_t reply_type;
     uint8_t *reply;
@@ -275,8 +272,9 @@ int main(int argc, char *argv[]) {
                 errx(EXIT_FAILURE, "IPC: Could not parse JSON reply.");
         }
 
-        /* NB: We still fall-through and print the reply, because even if one
-         * command failed, that doesn’t mean that all commands failed. */
+        if (!quiet) {
+            printf("%.*s\n", reply_length, reply);
+        }
     } else if (reply_type == I3_IPC_REPLY_TYPE_CONFIG) {
         yajl_handle handle = yajl_alloc(&config_callbacks, NULL, NULL);
         yajl_status state = yajl_parse(handle, (const unsigned char *)reply, reply_length);
@@ -289,12 +287,12 @@ int main(int argc, char *argv[]) {
             case yajl_status_error:
                 errx(EXIT_FAILURE, "IPC: Could not parse JSON reply.");
         }
-
-        goto exit;
+    } else {
+        if (!quiet) {
+            printf("%.*s\n", reply_length, reply);
+        }
     }
-    printf("%.*s\n", reply_length, reply);
 
-exit:
     free(reply);
 
     close(sockfd);
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -166,16 +166,18 @@ int main(int argc, char *argv[]) {
     uint32_t message_type = I3_IPC_MESSAGE_TYPE_RUN_COMMAND;
     char *payload = NULL;
     bool quiet = false;
+    bool monitor = false;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: i3-msg/main.c [hunk 1]
```diff
@@ -275,8 +278,8 @@ int main(int argc, char *argv[]) {
                 errx(EXIT_FAILURE, "IPC: Could not parse JSON reply.");
         }
 
-        /* NB: We still fall-through and print the reply, because even if one
-         * command failed, that doesn’t mean that all commands failed. */
+        if (!quiet)
```
orestisf1993: Use braces in if blocks

i3-msg/main.c
```diff
@@ -204,19 +206,23 @@ int main(int argc, char *argv[]) {
...
             } else {
                 printf("Unknown message type\n");
-                printf("Known types: run_command, get_workspaces, get_outputs, get_tree, get_marks, get_bar_config, get_binding_modes, get_version, get_config, send_tick\n");
+                printf("Known types: run_command, get_workspaces, get_outputs, get_tree, get_marks, get_bar_config, get_binding_modes, get_version, get_config, send_tick, subscribe\n");
                 exit(EXIT_FAILURE);
             }
         } else if (o == 'q') {
             quiet = true;
+        } else if (o == 'm') {
+            monitor = true;
```
orestisf1993: We can also validate that `-m` is used with `-t subscribe` exclusively

i3-msg/main.c
```diff
@@ -166,16 +166,18 @@ int main(int argc, char *argv[]) {
...
     char *payload = NULL;
     bool quiet = false;
+    bool monitor = false;
 
     static struct option long_options[] = {
         {"socket", required_argument, 0, 's'},
         {"type", required_argument, 0, 't'},
         {"version", no_argument, 0, 'v'},
         {"quiet", no_argument, 0, 'q'},
+        {"monitor", no_argument, 0, 'm'},
```
orestisf1993: We should also update the man page

man/i3-msg.man
```diff
@@ -91,6 +101,9 @@ i3-msg border normal
 
 # Dump the layout tree
 i3-msg -t get_tree
+
+# Monitor focused windows
+i3-msg -t subscribe -m '[ "window", "focus" ]'
```
orestisf1993: Just realized that this is not correct, you can't subscribe to specific event types. What you are doing here is subscribing to the `window` event (which exists) and the `focus` event (which doesn't, but i3 silently accepts it).

We could support subscribing to specific types in i3, for example `[ "window::focus" ]`, but this is outside the scope of this PR and should be discussed in an issue first.

Also correct the commit message please.

vivien: Good catch, I've rewrote the commit message and the documentation. Thanks!

Airblader:
Since I'm on vacation I can't look at it in detail, but I don't like the terminology too much. We call it subscribe in the IPC, not monitor, so I'd somewhat like to avoid creating a new term for it, especially one not used by other tools (spy, watch). 

I realize that "s" is taken, but couldn't we use longopts and do `--subscribe`?

vivien:
“—subscribe” would be confusing since it would make a user think about a
shortcut for “-t subscribe”. -m / monitor means listen forever (basically
-t subscribe in a while loop) and is a common option in tools like
inotifywait or dbus. That being said, this is your call!

Airblader:
Yes, that's a good point. Let's keep it this way then.

PR #3441, opened 2018-10-07
add specific GitHub issue templates

stapelberg:
I learnt about this from the GitHub blog:
https://blog.github.com/2018-05-02-issue-template-improvements/

Unfortunately, it seems like one cannot set a title in the template, otherwise we could do away with the first section of the template entirely.


orestisf1993:
oh ok, I misunderstood the screenshot

PR #3442, opened 2018-10-08
commands: (rename) check workspace number

vivien:
Assuming two workspaces 1 and 2, renaming the workspace 2 to "1: foo"
will reassign the number 1 already in use to the new workspace and break
the "workspace next_on_output" command.

This patch fixes this behavior by checking if a different workspace with
the same number does not already exist when renaming a given workspace.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1954,11 +1954,17 @@ void cmd_rename_workspace(I3_CMD, const char *old_name, const char *new_name) {
     }
 
     Con *check_dest = get_existing_workspace_by_name(new_name);
+    if (!check_dest) {
+        long new_num = ws_name_to_number(new_name);
+        if (new_num != -1) {
+            check_dest = get_existing_workspace_by_num(new_num);
+        }
+    }
 
     /* If check_dest == workspace, the user might be changing the case of the
      * workspace, or it might just be a no-op. */
     if (check_dest != NULL && check_dest != workspace) {
-        yerror("New workspace \"%s\" already exists", new_name);
+        yerror("Workspace %ld named \"%s\" already exists", check_dest->num, check_dest->name);
         return;
     }
 


src/commands.c
```diff
@@ -1954,11 +1954,17 @@ void cmd_rename_workspace(I3_CMD, const char *old_name, const char *new_name) {
...
+        if (new_num != -1) {
+            check_dest = get_existing_workspace_by_num(new_num);
+        }
+    }
 
     /* If check_dest == workspace, the user might be changing the case of the
      * workspace, or it might just be a no-op. */
     if (check_dest != NULL && check_dest != workspace) {
-        yerror("New workspace \"%s\" already exists", new_name);
+        yerror("Workspace %ld named \"%s\" already exists", check_dest->num, check_dest->name);
```
orestisf1993: `Con->num` is `int`

orestisf1993:
> Assuming two workspaces 1 and 2, renaming the workspace 2 to "1: foo"
will reassign the number 1 already in use to the new workspace

Other parts of the codebase allow multiple workspaces with the same number

> and break the "workspace next_on_output" command.

how exactly?

Airblader:
Multiple workspaces with the same number are explicitly allowed. It's up to the user to avoid this if they don't want it to happen.

vivien:
Thanks for your feedback guys. At first sight the output of `i3-msg get_workspaces` seemed weird to me because I pictured *num* as a UID. Anyway, the annoying part remains the iteration over numbered workspaces.

> > Assuming two workspaces 1 and 2, renaming the workspace 2 to "1: foo" will reassign the number 1 already in use to the new workspace
>
> Other parts of the codebase allow multiple workspaces with the same number
>
> > and break the "workspace next_on_output" command.
>
> how exactly?

The command skips multiple workspaces with the same number. Try this:

```` sh
i3-msg workspace 2
i3-sensible-terminal # put something in there
i3-msg workspace 3
i3-sensible-terminal # put something in there
i3-msg workspace 4
i3-sensible-terminal # put something in there
i3-msg rename workspace 4 to '"3: foo"'
while sleep 1; do i3-msg workspace next_on_output; done
````

You'll see that workspace "3" is skipped. This is clearly unexpected from *next_on_output* which by definition must iterate over every workspaces of the current output.

vivien:
Here are other frustrating examples caused by this behavior.

1. In the example above, `i3-msg move container to workspace number 3` will move a window to the workspace '"3: foo"'. Even though the '"3: foo"' workspace is indeed correctly numbered 3, I would expect the workspace "3" to be chosen, since it has no fancy name.

2. Also since there can be multiple workspaces with the same number, one can expect `i3-msg workspace [number] X` to *iterate* over workspaces numbered X instead of limiting itself to the first matching one.

One way to have a consistent behavior and work with multiple workspaces with the same number would be to have `workspace X` stick with the first workspace explicitly named "X" (or numbered X) while `workspace number Y` would target the **next** workspace matching `(Con *)->num == Y`. This would have the effect to:

a. iterate over *every* workspaces numbered 3 with `workspace number 3`;
b. move a window to the *next* workspace numbered 3 with a binding `move container to workspace number 3` (useful when already focusing a workspace numbered 3).

orestisf1993:
I assume we are going to proceed here by fixing the issues independently instead of forbidding existing workspace numbers?

Airblader:
Yeah, I'd say so. This is intended functionality that for all I know is actively used by some people.

orestisf1993:
@vivien do you want to handle the issues you mentioned? Or should we close this and proceed?

vivien:
I can work on the iteration issue. Fixing next_on_output is the very least.
I’ll dig into that hoping it doesn’t take too long.

Airblader:
Closing due to inactivity.

PR #3451, opened 2018-10-11
Close & reopen tray selwin when needed

### LINKED ISSUE #1329: i3bar doesn't detect new primary monitor on output configuration change
[Originally reported by bug+i3wm@…]
(i3bar should automatically redetect the primary monitor when switching outputs.

Currently, I need to restart i3 every time I enable/disable the external monitor (after setting the external/internal monitor as xrandr --primary) to move the bar output to the correct display. Without restart, it stays on the old location, even if this output is disabled and no more available.

### LINKED ISSUE #2010: Positional information for the tray should be cached
Currently, we decide for each docking request which client should hold the system tray. There's also duplicated "sanity check" logic for calling `init_tray()`.

We _should_ be deciding which client gets the tray on startup and cache this information to be used in docking requests. However, we will need to recalculate this in the event that the list of outputs or the bar configuration changed.

### LINKED ISSUE #3317: Tray symbols disappear after switching monitor outputs

orestisf1993:
Fixes #1329
Fixes #3317
Fixes #2010

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -92,6 +92,9 @@ static mode binding;
 /* Indicates whether a new binding mode was recently activated */
 bool activated_mode = false;
 
+/* The output in which the tray should be displayed. */
+static i3_output *output_for_tray;
+
 /* The parsed colors */
 struct xcb_colors_t {
     color_t bar_fg;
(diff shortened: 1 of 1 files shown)


orestisf1993:
I also think it's beneficial to allow falling back to the first available output like in https://github.com/bebehei/i3/commit/7e45133a86c01f838b6cdf8dc47f873ebfc0b0f3

i3bar/src/xcb.c
```diff
@@ -1889,6 +1847,18 @@ void reconfig_windows(bool redraw_bars) {
             }
         }
     }
+
+    /* Finally, check if we want to initialize the tray or destroy the selection
+     * window. The result of get_tray_output() is cached. */
+    if ((output_for_tray = get_tray_output())) {
```
Airblader: I would slightly change this so that it's clear this isn't a case of an accidental assignment over comparison and makes it just a bit easier to read.

```suggestion
    output_for_tray = get_tray_output();
    if (output_for_tray) {
```

orestisf1993:
up

bebehei:
@orestisf1993 wow, works like a charm! :+1: Thank you!

PR #3452, opened 2018-10-12
Add title_align config directive

### LINKED ISSUE #1750: Make title alignment configurable

orestisf1993:
Continuation of #3262
Fixes #1750

@Aestek

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -585,6 +585,16 @@ workspace_layout default|stacking|tabbed
 workspace_layout tabbed
 ---------------------
 
+=== Window title alignment
+
+This option determines the window title's text alignment.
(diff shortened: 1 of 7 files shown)


Airblader:
LGTM, would you mind adding some screenshots?

orestisf1993:
center: center
left: left
right: right

Airblader:
Thanks, I appreciate it!

PR #3453, opened 2018-10-13
Fixes for undefined behaviour on signed shift

alanbarr:
Change literal 1 to unsigned to allow safe bitshift of 31.

Caught by cppcheck.


alanbarr:
Would you prefer I close this PR and let https://github.com/i3/i3/pull/3433 fix these?
I fixed the extra shift on this branch anyways.

PR #3465, opened 2018-10-20
Add "modifiers" to events sent by i3bar

### LINKED ISSUE #3461: i3bar: Provide modifier states in events
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

i3bar events include the button and co-ordinates.

## Desired Behavior

i3bar events also include the modifier state (list of active modifiers when event happened). This would allow different behaviour for Ctrl+Left Click vs. Left Click, for example.

## Environment

<pre>
i3 version: 
Running i3 version: 4.15 (2018-03-10) (pid 958)
Loaded i3 config: /home/soumya/.config/i3/config (Last modified: Sat 29 Sep 2018 12:34:38 PDT, 1502413 seconds ago)

</pre>

<pre>
- Linux Distribution & Version: ArchLinux
</pre>

soumya92:
The modifiers field is an array of modifier names, e.g. ["ctrl", "Mod3"].

Fixes #3461

### DIFF
--- a/docs/i3bar-protocol
+++ b/docs/i3bar-protocol
@@ -243,6 +243,9 @@ relative_x, relative_y::
     of the block
 width, height::
     Width and height (in px) of the block
+modifiers::
+    An array of the modifiers active when the click occurred. The order in which
+    modifiers are listed is not guaranteed.
 
 *Example*:
 ------------------------------------------
@@ -250,6 +253,7 @@ width, height::
  "name": "ethernet",
  "instance": "eth0",
  "button": 1,
+ "modifiers": ["Shift", "Mod1"],
  "x": 1320,
  "y": 1400,
  "relative_x": 12,
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -8,6 +8,7 @@
  *
  */
 #include "common.h"
+#include "yajl_utils.h"
 
 #include <stdlib.h>
 #include <unistd.h>
@@ -26,6 +27,8 @@
 #include <yajl/yajl_version.h>
 #include <yajl/yajl_gen.h>
 #include <paths.h>
+
+#include <xcb/xcb_keysyms.h>
 
 /* Global variables for child_*() */
 i3bar_child child;
@@ -588,15 +591,11 @@ static void child_click_events_initialize(void) {
     }
 }
 
-static void child_click_events_key(const char *key) {
-    yajl_gen_string(gen, (const unsigned char *)key, strlen(key));
-}
-
 /*
  * Generates a click event, if enabled.
  *
  */
-void send_block_clicked(int button, const char *name, const char *instance, int x, int y, int x_rel, int y_rel, int width, int height) {
+void send_block_clicked(int button, const char *name, const char *instance, int x, int y, int x_rel, int y_rel, int width, int height, int mods) {
     if (!child.click_events) {
         return;
     }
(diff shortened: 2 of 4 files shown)


soumya92:
Before I do that, I'd like to point out a few reasons for keeping this as is:

- The most common operation on the modifiers will be to check for the presence of one, something which is much easier in many languages when given the bitmask rather than a list of strings. Using an array for modifiers adds complexity both in i3bar and in the status_command program. e.g., in go:
```go
if mods&Mod4Mask != 0 { 
  ...
}
// vs.
for _, mod := range mods {
  if mod == "Mod4" {
    ...
  }
}
```

- The modifiers will be handled by the status_command programs (i3status et al), so keeping consistency with the config is likely not as important, since the programs will all have different ways to surface this information.

- Making modifiers an array makes the event message variable length and adds a new field type in the output. While JSON parsers will have no trouble with this, users _may_ have shell scripts with simplistic parsing, e.g. assuming that `,` always ends a field, which this could break.

Having said that, if you think array is the way to go, I'll gladly make the change and update this PR.
Thanks for reviewing!

Airblader:
Sorry for the delay, I was just on my way back from the US and need to take care of a few things first :-)

> something which is much easier in many languages when given the bitmask rather than a list of strings

I can't speak for Go, but most higher-level languages have a direct way of checking this (`in` in Python, `includes` in Javascript, …). In my experience, bitmasks are mostly used in low-level languages and thus many people aren't too familiar with them, but most IPC scripts are written with high-level languages.

Also, using the bitmask would require scripts to either import the bitmask definitions (like your Mod4Mask) from some X utility (might not be available + widely not understood by users) or hardcode it in their script while the modifier name will be known to every user (that's how they configure i3, anyway).

> The modifiers will be handled by the status_command programs (i3status et al), so keeping consistency with the config is likely not as important

I agree that consistency isn't a necessarily good argument, but making it familar & easy for users to work with it is… :-)

> Making modifiers an array makes the event message variable length

Is that an issue?

> While JSON parsers will have no trouble with this, users may have shell scripts with simplistic parsing, e.g. assuming that , always ends a field, which this could break.

Such scripts are broken to begin with, IMHO. If you want to write shell scripts, use a JSON utility like jq. We make no promises on how the JSON is formatted, either (e.g., spaces or not etc.), we're not even really in control of that.

However, I'm also happy to consult @stapelberg on his opinion. If we *did* decide to expose the bitmask, I'd at least rename it to `modifiers_mask`, but IMO I wouldn't expost the bitmask here.

stapelberg:
Thanks for your PR!

Passing `event->state` as the PR currently suggests seems like exposing too much of an X11 implementation detail. Note that `event->state` is influenced by NumLock state, for example.

> I can't speak for Go, but most higher-level languages have a direct way of checking this (`in` in Python, `includes` in Javascript, …).

Go does intentionally not have a way to check for existence in an array/slice (the argument is that writing down the loop explicitly will make it more obvious that this piece of code might be performance-critical). This should not be an argument for the contents of the IPC message, though.

> Also, using the bitmask would require scripts to either import the bitmask definitions (like your Mod4Mask) from some X utility (might not be available + widely not understood by users)

Strong +1. We should strive for the IPC to be usable with only the IPC libraries/definitions/includes/documentation, not also knowledge about how X11 works on the wire level.

> Such scripts are broken to begin with, IMHO. If you want to write shell scripts, use a JSON utility like jq. We make no promises on how the JSON is formatted, either (e.g., spaces or not etc.), we're not even really in control of that.

+1. We have advocated for many years for using correct parsing, ideally in a higher-level language than shell :).

In conclusion, I agree with @Airblader.

soumya92:
Thanks for the explanations! I've made the changes to the PR, "modifiers" is now an array of strings, where the values are the names of the modifiers as used when parsing the config.

i3bar/src/child.c
```diff
@@ -606,34 +606,52 @@ void send_block_clicked(int button, const char *name, const char *instance, int
...
 
-    child_click_events_key("button");
+    ystr("button");
     yajl_gen_integer(gen, button);
 
-    child_click_events_key("x");
+    ystr("modifiers");
+    yajl_gen_array_open(gen);
+    if (mods & ShiftMask)
+        ystr("shift");
```
Airblader: This should actually be uppercase ("Shift"), same for ctrl below.

i3bar/src/child.c
```diff
@@ -27,6 +27,8 @@
 #include <yajl/yajl_gen.h>
 #include <paths.h>
 
+#include <X11/Xlib.h>
```
Airblader: We don't wanna use Xlib for this… :-) Please use the xcb equivalents (`XCB_KEY_BUT_MASK_CONTROL`, …). The closer we get this to be like `src/config_directives.c#event_state_from_str`, the better.

i3bar/src/child.c
```diff
@@ -588,15 +590,13 @@ static void child_click_events_initialize(void) {
     }
 }
 
-static void child_click_events_key(const char *key) {
-    yajl_gen_string(gen, (const unsigned char *)key, strlen(key));
-}
+#define ystr(str) yajl_gen_string(gen, (unsigned char *)str, strlen(str))
```
Airblader: Can't this still be `const unsigned char *`?

docs/i3bar-protocol
```diff
@@ -243,13 +243,16 @@ relative_x, relative_y::
...
+modifiers::
+    An array of the modifiers active when the click occurred
 
 *Example*:
 ------------------------------------------
 {
  "name": "ethernet",
  "instance": "eth0",
  "button": 1,
+ "modifiers": ["shift", "Mod1"],
```
Airblader: Also needs to be updated, see comment below.

docs/i3bar-protocol
```diff
@@ -243,13 +243,16 @@ relative_x, relative_y::
     of the block
 width, height::
     Width and height (in px) of the block
+modifiers::
+    An array of the modifiers active when the click occurred
```
Airblader: Let's maybe add a sentence like

> The order in which modifiers are listed is not guaranteed.

i3bar/src/child.c
soumya92: Done.

i3bar/src/child.c
soumya92: Oh, I guess I picked a bad example at https://github.com/i3/i3/blob/6339427f017e1265a022e6537e16a8b4a921e52f/i3bar/src/config.c#L130. Changed to use xcb and uppercase modifiers.

i3bar/src/child.c
soumya92: Removed, imported yajl_utils.h instead :)

docs/i3bar-protocol
soumya92: Done.

docs/i3bar-protocol
soumya92: Added.

Airblader:
Thanks!

PR #3469, opened 2018-10-21
Update userguide "Focus Parent": add the default

yablonsky:
While reading the UserGuide, I had to refer to the TL;DR image on top to find the default keybind for "Focus Parent" since it's not mentioned in the text.
This pr fixes it.


Airblader:
Thanks!

PR #3482, opened 2018-10-26
Do not assume STDIN_FILENO is available for input from child

### LINKED ISSUE #3477: i3bar misuses file descriptors resulting in libev abort
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

If i3bar is started without any file attached to file descriptor zero, and the epoll mechanism is being used by libev, i3bar can overwrite the epoll descriptor.  This happens because i3bar unconditionally uses dup2 to put a pipe on the stdin (zero) file descriptor.  When libev tries to poll using the pipe descriptor, it gets an EINVAL error back from epoll_wait and aborts.

## Expected Behavior

i3bar shouldn't abort just because it is started with no file descriptor zero.  i3bar shouldn't mess with stdin (or should ensure that a stdin file descriptor exists before initializing libev).

## Reproduction Instructions

I have attached the following log (log.txt) which shows the behaviour.

[log.txt](https://github.com/i3/i3/files/2516760/log.txt)

I figure it can be fixed either by not mucking around with FD 0, or by ensuring FD 0 is a valid file descriptor before starting up libev.

I have patches for both of these (no-muck.patch.txt and preopen.patch.txt) attached.
[no-muck.patch.txt](https://github.com/i3/i3/files/2516790/no-muck.patch.txt)
[preopen.patch.txt](https://github.com/i3/i3/files/2516791/preopen.patch.txt)

## Environment

<pre>
Binary i3 version:  4.15-116-gc54f5008 (2018-03-10, branch "master") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.15-116-gc54f5008 (2018-03-10, branch "master") (pid 31838)
Loaded i3 config: /home/hmacdona/.config/i3/config (Last modified: 2018-10-25T13:28:50 EDT, 13431 seconds ago)

The i3 binary you are running: /usr/local/bin/i3

</pre>

<pre>
# i3 config file (v4)
bar {
    status_command i3status
}
</pre>

<pre>
Logfile URL:
</pre>

<pre>
- Linux Distribution & Version: Centos7
</pre>

hamishimac:
Fixes #3477 

Testcases all pass, although I did have to modify t/189-floating-constraints.t to change the "my sub" declarations to "sub" in order to get this test-case to pass (even without my changes).  Presumably my perl is too old (This is perl 5, version 16, subversion 3 (v5.16.3)).

Perhaps the variable name (stdin_fd) is ill-advised.  "input_from_child_fd" might be a clearer name.

### DIFF
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -35,6 +35,7 @@ i3bar_child child;
 
 /* stdin- and SIGCHLD-watchers */
 ev_io *stdin_io;
+int stdin_fd;
 ev_child *child_sig;
 
 /* JSON parser for stdin */
@@ -450,7 +451,7 @@ static void stdin_io_first_line_cb(struct ev_loop *loop, ev_io *watcher, int rev
     }
     free(buffer);
     ev_io_stop(main_loop, stdin_io);
-    ev_io_init(stdin_io, &stdin_io_cb, STDIN_FILENO, EV_READ);
+    ev_io_init(stdin_io, &stdin_io_cb, stdin_fd, EV_READ);
     ev_io_start(main_loop, stdin_io);
 }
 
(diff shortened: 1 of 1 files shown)


Airblader:
@stapelberg Do you wanna get this in before or after the release?

stapelberg:
After, just to be prudent.

Airblader:
Thank you!

Issue #3412, opened 2018-09-17
Enhancehement: Render left and right borders of titles in stacked mode


aksel:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
I'm using pixel borders.

Stacking vs tabbed mode renders the title borders of the focused container in different manners, and inconsistently with regards to child borders.

[These two images](https://imgur.com/a/LU5b5e7) show the differences. Tabbed render a border all around the title, whereas stacked stop at the child border.

[In this image](https://i.imgur.com/AiawCYd.png), I have increased the border to exaggerate the effect. You can see, in addition to the left and right borders of the stacked title missing, the lower horizontal border stops at the child border, with 1 pixel missing at the edges.

```
Current, tabbed: 

╔══════╗──────┬──────┐
║ tab1 ║ tab2 │ tab3 │
╠══════╝─────────────╖
║                    ║
╚════════════════════╝

Current, stacked:
┌────────────────────┐   ┌────────────────────┐
│ stack1             │   │ stack1             │
├────────────────────┤   ├════════════════════┤
│ stack2             │   │ stack 2 (focused)  │
├════════════════════┤   ├════════════════════┤
│ stack 3 (focused)  │   │ stack3             │
├════════════════════┤   ╓────────────────────╖
║                    ║   ║                    ║
║                    ║   ║                    ║
╚════════════════════╝   ╚════════════════════╝
```

## Expected Behavior
```
Desired, stacked:
┌────────────────────┐   ┌────────────────────┐
│ stack1             │   │ stack1             │
├────────────────────┤   ╔════════════════════╗
│ stack2             │   ║ stack 2 (focused)  ║
╔════════════════════╗   ╚════════════════════╝
║ stack 3 (focused)  ║   │ stack3             │
╠════════════════════╣   ╓────────────────────╖
║                    ║   ║                    ║
║                    ║   ║                    ║
╚════════════════════╝   ╚════════════════════╝
```

Borders should be rendered all around around the focused title.

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: 

Binary i3 version:  4.15.0.1 (03-13-2018) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.15.0.1 (03-13-2018) (pid 601)bort…)
[...]

PR #3485, opened 2018-10-27
Draw outer header borders for all layouts

xzfc:
#3412
[old](http://dump.bitcheese.net/files/kytamik/old.png)
[new](http://dump.bitcheese.net/files/jymizuj/new.png)


Airblader:
Thanks!

Issue #3500, opened 2018-11-04
Introduce a command to delete a container but not its children


ilikeheaps:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Desired Behavior
You could bind command that would remove a container but move all its children to its parent (with as little changes to perceived layout as possible: e.g. removing inner H from H[W1 H[W2 W3]] shouldn't change the placement of windows -- I'm not sure how that example interacts with tree_flatten though).

## Motivation
Moving windows to sides to delete a stack container seems roundabout. Also with current implementation of tree_flatten you sometimes have chains of containers and you'd have to move each child out of the chain individually to get rid of the chain.

## Other
I've got a draft of solution that kinda works (WIP, messes up the order and sizes of containers) which I'll make a PR for if this issue is accepted.

Airblader:
Can't this be implemented trivially as an IPC script? Get children of the focused container and move them (wherever you want)?

ilikeheaps:
I actually don't know, first thing I tried was C code because I was investigating `tree_flatten` at the same time. I suppose I could do it through IPC indeed. One point for including it in the base code is that I use these functions in `tree_flatten` rework (but that's only if I get it to acceptable state) - from there it's just a matter of making a little wrapper. I guess I'll try IPC route anyway in case my flatten rework isn't accepted.

Airblader:
Can we try to implement this using the IPC for now? If that turns out to be impossible, we could revisit. At the moment I'm not convinced that this is worth adding a new command yet.

ilikeheaps:
Alright, if this is too niche functionality to be included in base i3, this issue can be closed. It just felt to me that this is an operation basic enough that one would expect it out of the box.

CarloWood:
> Alright, if this is too niche functionality to be included in base i3, this issue can be closed. It just felt to me that this is an operation basic enough that one would expect it out of the box.

I switched from plasma to i3 today - and the first really annoying bug that I ran into, the very same day, is that implicit containers can pollute your tree without even being visible. I quickly end up with -say- three windows next to each other, looking like a single H-container with three windows, but actually being something different (which you can't see until you try things).

[...]

PR #3501, opened 2018-11-04
(WIP) Rework `tree_flatten`

ilikeheaps:
Complete rework of `tree_flatten`. Also includes code for https://github.com/i3/i3/issues/3500 because they share common functions - I'll remove appropriate parts if we don't want it.

## Principles
A container is redundant if at the same time:
1. It's not focused
2. None of its children is focused (optional?)
3. Its layout is split horizontal/vertical
4. Either:
    - its layout is the same as its parent
    - it has only one child

I'm currently having problems when the parent is a workspace because sometimes there is a container being implicitly created when you change the layout of a workspace and I would try to flatten it ending up in an endless loop. So I'm not flattening workspaces' children until I figure out how exactly that works.

### Reasoning for the rules
1. If a container is focused the user might want to perform some action on it.
2. Similar to (1.) but I'm not sure what's being focused when we create a new container so it might be not needed or needed only in certain cases
3. Tabbed/stack containers always are possibly desired by the user to label and structurize things
4. Assuming (3.) it just seems right

## State of the implementation
Seems to transform the tree correctly but messes up the order of containers and their sizes. I temporarily removed regular calls to tree_flatten on certain events (because they were in functions that I currently use in `tree_flatten` and ended up in endless loops) and made a binding to call it on-demand for testing.

TODO:
- scale container scales (uh, their percentage size) so that they don't actually change in absolute size
- keep the order of containers (is this the focus list?)
- decide when to call `tree_flatten`

## Alternative approach
I'm considering whether it wouldn't be possible to keep the tree flattened at all times by running some smaller function checking invariants at every change instead of walking the whole tree every few operations. I think that approach could be complicated so I'm sticking with `tree_flatten` for now.

### DIFF
--- a/include/commands.h
+++ b/include/commands.h
@@ -16,6 +16,8 @@
 /** The beginning of the prototype for every cmd_ function. */
 #define I3_CMD Match *current_match, struct CommandResultIR *cmd_output
 
+void cmd_flatten_tree(I3_CMD);
+
 /**
  * Initializes the specified 'Match' data structure and the initial state of
  * commands.c for matching target windows of a command.
--- a/include/commands.h
+++ b/include/commands.h
@@ -242,6 +242,8 @@ void cmd_exit(I3_CMD);
  */
 void cmd_reload(I3_CMD);
 
+void cmd_remove_parent(I3_CMD);
+
 /**
  * Implementation of 'restart'.
  *
--- a/include/tree.h
+++ b/include/tree.h
@@ -90,7 +90,7 @@ bool tree_restore(const char *path, xcb_get_geometry_reply_t *geometry);
  * Removes the target node while shifting its children up to its parent.
  * Assumes: target->window == NULL
  */
-void tree_remove_node(Con* target);
+void tree_remove_node(Con *target);
 
 /**
  * tree_flatten() removes pairs of redundant split containers, e.g.:
--- a/include/tree.h
+++ b/include/tree.h
@@ -87,6 +87,12 @@ bool tree_close_internal(Con *con, kill_window_t kill_window, bool dont_kill_par
 bool tree_restore(const char *path, xcb_get_geometry_reply_t *geometry);
 
 /**
+ * Removes the target node while shifting its children up to its parent.
+ * Assumes: target->window == NULL
+ */
+void tree_remove_node(Con* target);
+
+/**
  * tree_flatten() removes pairs of redundant split containers, e.g.:
  *       [workspace, horizontal]
  *   [v-split]           [child3]
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -13,6 +13,7 @@ state INITIAL:
   # We have an end token here for all the commands which just call some
   # function without using an explicit 'end' token.
   end ->
+  'flatten' -> call cmd_flatten_tree()
   '[' -> call cmd_criteria_init(); CRITERIA
   'remove_parent' -> call cmd_remove_parent()
   'move' -> MOVE
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -14,6 +14,7 @@ state INITIAL:
   # function without using an explicit 'end' token.
   end ->
   '[' -> call cmd_criteria_init(); CRITERIA
+  'remove_parent' -> call cmd_remove_parent()
   'move' -> MOVE
   'exec' -> EXEC
   'exit' -> call cmd_exit()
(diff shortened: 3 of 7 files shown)


orestisf1993:
#3235 and related discussions / issues

Airblader:
Haven't looked at the changes yet, but assumption #1 already sounds wrong to me — operations can also be done on non-focused containers.

Could you explain what exactly this rework fixes (other adding the feature from your feature request)?

ilikeheaps:
I want to fix some issues mentioned in https://github.com/i3/i3/pull/3235 - currently it's very easy to make a tree with redundant containers that won't be flattened (like `H[w1 H[w2 w3]] === H[w1 w2 w3]`). I'm not sure which issues exactly I want to solve - I'm hoping to make a mechanism with fairly simple rules that make sense and see where that ends up. I think the ones I proposed are fairly simple but I'm not sure if they make sense.

On (1.): if one wanted to create a container away from the focused point of the tree, he would do that in order to change the perceived layout; that is if he made a layout that would be flattened it would have to be a temporary structure. I see few solutions: not running `tree_flatten' during the sequence of operations or providing commands that avoid flattenable temporary states. These two could (or would) break compatibility.

Another possible solution I see is introducing some kind of "spawn invulnerability" to containers so we wouldn't collapse a freshly created container that the user wanted to actually use for something. It could be configurable so that a container can't be flattened during first X calls of `tree_flatten`. Heck, it could have an option of infinite for compatibility (and keeping the old behaviour during invulnerability but that would make code twice as big).

Essentially most solutions seem to be breaking compatibility to some extent because this `tree_flatten` is fundamentally different from the old one. The question is: do people actually use things that would be broken by this change? If you have some specific example I think it would be a good basis for further consideration of the rules. If the compatibility is absolutely needed I think the invulnerability thing would be good (without old behaviour - there isn't someone relying on the containers collapsing in a script, is there?).

ilikeheaps:
I guess I could make it so that by default only h/v splits with one child are flattened (with some restrictions so that it doesn't immediately close a container that the user wants to use) which I think would fix unintuitive behaviour like https://github.com/i3/i3/issues/3003. In that case making the behaviour that I desire optional would be trivial and wouldn't make the code any more complex. I think that making behaviour expected in https://github.com/i3/i3/issues/3001 optional would be quite simple as well. Anyways, I'll keep thinking about it. I wanted to consider two containers the same if one can always mimic the other by using appropriate parameters but resizing behaviour is important indeed (depending on how one uses i3).

Airblader:
I think that if we want to rework how tree flattening works, we should first come up with the requirements, the issues of the current implementation actually justifying any change (such as #3003), discuss what the solution should be and, most importantly, define a comprehensive set of test cases through which we can run any new implementation for it (we do already have some tests, but I'm not sure how thorough they are).

I'd like to exercise a high level of caution here because this is really at the core of what defines i3, and container manipulation is probably one of the most important use cases of existing scripts and certainly of day to day user interaction. And I don't think tree flattening is fundamentally broken at the moment, so there's no need to rush things — let's take our time thinking about this one. 

As a pretty simple first requirement: we shouldn't break *any* scenario for a user. That is, we do not remove containers that can in fact serve a purpose, but only those that are truly unnecessary in the sense that any layout derived from the unflattened structure can also be derived from the flattened one. 
The reasoning for this is quite simple: I'll take a user having to perform an extra step to close a (for them in that moment unnecessary) container over a user who cannot achieve the layout they need. :-)

Two more points, though I haven't thought them through entirely:

1. Focused containers should never be treated differently. 
2. I'd like to avoid having some kind of "pristine" flag on containers. Layout changes should behave the same regardless of whether commands are executed in sequence or whether there's another command in between that affects something entirely different. 

I think as such I'll close this PR for now and we move any discussion to the issue tracker (perhaps we open a fresh issue where we also collect a list of links to related issues like #3003). 

Note, please, that #3500 is completely orthogonal/independent of such a change as that discusses adding a new feature. We shouldn't mix these things.

ilikeheaps:
> Note, please, that #3500 is completely orthogonal/independent of such a change as that discusses adding a new feature. We shouldn't mix these things.

Sure, it only got mixed up in here because of how I organized my branches.

CarloWood:
Making a special command for a (new) tree flatten would solve all mentioned concerns of Airblader: it would not break any script or change anything for existing users. It would be a NEW command that people can bind and use if they want. I see no reason not to add such a new command.  I do agree that it shouldn't be like the function "tree_flatten" and work on the whole tree though. It should work on the focused window/container (and it's children) only.

PR #3504, opened 2018-11-05
Change config order in manpage

Foxboron:
This brings the headline for the configuration files inline with the
recent move to XDG directories.


Airblader:
Thanks!

PR #3508, opened 2018-11-06
load_layout: Correctly mark non-leaf containers

orestisf1993:


orestisf1993:
Ok, this isn't that simple because of https://github.com/i3/i3/issues/2511

orestisf1993:
Ok, made it not re-regress #2511.

PR #3511, opened 2018-11-07
For resizing, convert pixel diff to percentage, based on parent.

aksel:
Previously, it first calculated one of the containers' next percentage, and then subtracted the previous percentage to find the actual change.

Now it directly calculates the change, and subtracts and adds the change to the two affected containers.

Also, prematurely return when pixel diff is 0, as no action is necessary.

This is related to [this issue on i3-gaps](Airblader/i3#247).

### DIFF
--- a/src/resize.c
+++ b/src/resize.c
@@ -102,29 +102,15 @@ bool resize_find_tiling_participants(Con **current, Con **other, direction_t dir
 }
 
 /*
- * Calculate the given container's new percent given a change in pixels.
- *
- */
-double px_resize_to_percent(Con *con, int px_diff) {
-    Con *parent = con->parent;
-    const orientation_t o = con_orientation(parent);
-    const int total = (o == HORIZ ? parent->rect.width : parent->rect.height);
-    /* deco_rect.height is subtracted from each child in render_con_split */
-    const int target = px_diff + (o == HORIZ ? con->rect.width : con->rect.height + con->deco_rect.height);
-    return ((double)target / (double)total);
-}
-
-/*
  * Calculate the minimum percent needed for the given container to be at least 1
  * pixel.
  *
  */
 double percent_for_1px(Con *con) {
-    Con *parent = con->parent;
-    const orientation_t o = con_orientation(parent);
-    const int total = (o == HORIZ ? parent->rect.width : parent->rect.height);
-    const int target = (o == HORIZ ? 1 : 1 + con->deco_rect.height);
-    return ((double)target / (double)total);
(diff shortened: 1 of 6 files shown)


src/resize.c
```diff
@@ -102,16 +102,15 @@ bool resize_find_tiling_participants(Con **current, Con **other, direction_t dir
...
 /*
- * Calculate the given container's new percent given a change in pixels.
+ * Calculate and return the given container's percentage change given a change in pixels.
  *
  */
 double px_resize_to_percent(Con *con, int px_diff) {
     Con *parent = con->parent;
     const orientation_t o = con_orientation(parent);
     const int total = (o == HORIZ ? parent->rect.width : parent->rect.height);
-    /* deco_rect.height is subtracted from each child in render_con_split */
```
orestisf1993: Let's move the deleted comment in `percent_for_1px` where it still applies. I don't think we can avoid the `deco_rect.height` code there.

aksel:
Should (con_)size_in_orientation, as a general util function, include deco_rect.height?

orestisf1993:
For our purpose, no, since percentage (with vertical orientation) is (con's rect height + deco height) / (parent's con rect without deco height).

Maybe we need a better name?

Airblader:
I'd say if it excludes the decoration, it's more of a window than a container function, right? Either ways I don't think we need to overcomplicate this for now. I'm fine just inlining that oneliner, it's sufficiently self-explanatory.

orestisf1993:
It's not for windows, we actually use it for non-leaf containers to get the
total size in which their children must be rendered.

Airblader:
Hm, alright. `con_undecorated_size_in_orientation` or `con_rect_size_in_orientation`? Still not convinced that requires a separate function at this point in time, though :-)

orestisf1993:
Eh, we will have 3 places to call it (https://github.com/i3/i3/blob/44e8fddc28f36ea0fd15104752e004713118ba5b/src/render.c#L195) and
```c
const int total = con_rect_size_in_orientation(con->parent);
```
is a bit cleaner than
```c
const orientation_t o = con_orientation(con->parent);
const int total = (o == HORIZ ? con->parent->rect.width : con->parent->rect.height);
```

Up to you.

Airblader:
If it's more than one place we should use a function. Sorry I'm answering from mobile without looking at the PR all that much. Just coming back from London. ;)

aksel:
I have added the function to con.c, con_rect_size_in_orientation and inlined the percentage conversion.

Airblader:
And patch is downstreamed to i3-gaps. :+1: Thanks, guys!

Issue #2742, opened 2017-04-16
Input shapes with holes don't correctly allow input through


gandalf3:
Output of `i3 --moreversion 2>&- || i3 --version`:

```
Binary i3 version:  4.13-non-git © 2009 Michael Stapelberg and contributors
Running i3 version: 4.13-non-git (pid 699)rl-c to abort…)
Loaded i3 config: /home/gandalf3/.config/i3/config (Last modified: Wed 12 Apr 2017 11:17:14 PM PDT, 220408 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

URL to a logfile as per http://i3wm.org/docs/debugging.html:

http://logs.i3wm.org/logs/5744863563743232.bz2

**What I did:**

Started [peek](https://github.com/phw/peek) and clicked on [its click-through-able region](https://github.com/phw/peek/blob/4a60f9a08669b8691c0c7ede5d704474a583b3f5/src/ui/application-window.vala#L376) during recording.

**What I saw:**

The click was received by peek and not the program underneath. 
For what its worth, I observed the same behavior with [licecap](http://www.cockos.com/licecap/) running in wine.

**What I expected instead:**

Mouse events to be received by the program under peek.

----

[According the the peek developer](https://github.com/phw/peek/issues/52#issuecomment-272098774), this is likely an upstream issue.

Airblader:
Note that this is not a *bug* since i3 doesn't claim to support the Shape extension. In general there's also no reason to: tiling window managers don't have overlapping windows, that's the entire point. Floating windows in i3 are meant for dialogs and popups, not long-living applications.

That said, I see why in this specific case a floating window makes sense even for such an application. I think implementing Shape in i3 would be quite a lot of work, and would of course introduce a new dependency. I'm not really sure yet whether we'd want this.

pciavald:
+1 to support Shape in i3, even just for this peek killer-app.

stapelberg:
peek is indeed pretty slick and useful.

Two observations:

1. peek requires a compositor, so it won’t work on i3 out of the box anyway.

2. I think peek’s use of floating windows makes sense, even when applying i3’s strict philosophy — a floating window is the only way to overlay a tiling window which you want to record.

So, pull requests which implement just enough of the shape extension so that clicks are routed correctly are welcome :).

phw:
> peek requires a compositor, so it won’t work on i3 out of the box anyway.

[...]

Issue #1966, opened 2015-09-27
Draw decorations on own frame for V/H


Airblader:
@stapelberg I wonder if there is a good reason why we draw window decorations on the parent container? I understand that it's a design decision that is deeply integrated into the i3 code, but I'd still like to figure out whether it's something we could consider changing given someone does the work.

For stacked and tabbed containers it makes a lot of sense, but I'm not sure this applies to vertically or horizontally split containers as well.
- Is there an actual, technical reason for this? Can we anticipate changes other than the rendering that would become necessary?
- Is it something that could be considered to be changed?

i3 4.10.4

stapelberg:
If you have a really good reason to change it, we might be able to pull it off.

The reason we went for this design is that it unifies the code paths for stacked/tabbed containers and regular containers. I don’t think there’s any technical reason for it.

Airblader:
Thanks for the reply. I will close this issue for now as I have no immediate plans to do this move (and assume no one else does, either). The biggest reason I have in mind for this would likely be fixed by the move to cairo, too (well, once certain bugfixes have propagated enough), so I'll revisit once that happened. Given that we're talking a long time of inactivity on this issue, I'll close it.

Airblader:
Reopening because the move to cairo does not fix the usecase properly, so this would be an option.

Airblader:
@stapelberg I suspect that if we did this change, the values / meaning of things like `deco_rect` could change (we'd have to see). Is your answer, that you're generally not against this, still valid in this case? Might be something for a major version release (?)

stapelberg:
We could retain deco_rect’s semantics when serializing for the IPC interface, right?

Airblader:
I guess we could. My only pain point with it is that as a developer I don't have access to the proper information anymore, unless we add a new key for the new semantics.

stapelberg:
Yes, we should add a new field with the new semantics.

Airblader:
OK. Suggestions for a name welcome :) 

The following list is simply for myself. Please disregard.

- [ ] Come up with a good way to draw decorations for stacked/tabbed. Parent seems logical, but poses a problem due to the recursion approach when rendering decorations. Either one needs to be adapted.
- [ ] Adapt `is_pixmap_needed`, as window-less containers will no longer need a surface.
- [ ] Don't even create windows for window-less containers since we don't need to draw on them anymore.
- [ ] Adapt rendering of all types. Decorations must be respected on the actual leaf container now, not on the parent.
[...]

PR #3514, opened 2018-11-09
Add input and bounding shapes support (#2742)

xzfc:
Basic idea: if the window has a shape, set the parent container shape as the union of the window shape and the shape of the frame borders.

Related issue: #2742.

Thanks @psychon for the [initial work](https://github.com/i3/i3/issues/2742#issuecomment-433619265).

---

Screenshot of shaped `xeyes`, `oclock`, and `peek` floating windows:
screenshot

---

1. New dependency: `xcb-shape`.
2. I check for Shape 1.1 support in `src/main.c`. Whereas it is possible to support Shape 1.0 (it does not have input shapes, only bounding shapes), but I think it does not worth that.
3. Shapes in tiled windows are weird: if shaped window is in tabbed/stacked container and `focus_follows_mouse` is enabled, it may be hard not to raise window underneath when moving mouse cursor.
4. Rendering of shaped tiled windows: the similar issue with windows titlebars as with transparent windows or in i3-gaps with titlebars.

Concerning 1, should I update `debian/control`?
Concerning 3 and 4, I'm not sure what i3 should do with shaped non-floating windows. Ignore shaping of tiled windows or leave it as is?


Airblader:
Thanks for the PR! This is really cool. I'll have a look at the PR a bit later, but regarding your questions:

1. Yeah, please update it. 
2. I think that's fine. Shape 1.1 predates i3 by years, so if you have i3 on your system, chances are your server supports no Shape or 1.1. And you only lose shape support otherwise, so yeah.
3. I think I'd accept that for now. I think chances are there's more usecases where it follows user intent than cases where it is a problem. 
4. That's more of a problem, I'll get back on this later as well as there's a few things I wanna check out first.

src/handlers.c
```diff
@@ -1517,6 +1518,21 @@ void handle_event(int type, xcb_generic_event_t *event) {
         return;
     }
 
+    if (shape_supported && type == shape_base + XCB_SHAPE_NOTIFY) {
+        DLOG("shape event.\n");
```
Airblader: This log message is too unspecific. Please also log the affected window and the type of the event here.

src/handlers.c
```diff
@@ -1517,6 +1518,21 @@ void handle_event(int type, xcb_generic_event_t *event) {
...
+        xcb_shape_notify_event_t *shape = (xcb_shape_notify_event_t *)event;
+        Con *con = con_by_window_id(shape->affected_window);
+        if (con == NULL)
```
Airblader: I think we should log if this happens (again including the affected window). Also, please use `{} ` for all blocks.

src/main.c
```diff
@@ -683,6 +687,22 @@ int main(int argc, char *argv[]) {
...
+        shape_base = extreply->first_event;
+        xcb_shape_query_version_cookie_t cookie = xcb_shape_query_version(conn);
+        xcb_shape_query_version_reply_t *version =
+            xcb_shape_query_version_reply(conn, cookie, NULL);
+        shape_supported = version && version->minor_version >= 1;
+        free(version);
+    } else {
+        shape_supported = false;
+    }
+    if (!shape_supported)
```
Airblader: `{} `

src/manage.c
```diff
@@ -574,6 +574,11 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
      * cleanup) */
     xcb_change_save_set(conn, XCB_SET_MODE_INSERT, window);
 
+    /* Receive ShapeNotify events whenever the client altered it's window
```
Airblader: nit: "its", not "it's"

src/manage.c
```diff
@@ -574,6 +574,11 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
...
+     * shape. */
+    if (shape_supported)
+        xcb_shape_select_input(conn, window, 1);
```
Airblader: Please use `true` instead of `1` for booleans.

src/manage.c
```diff
@@ -574,6 +574,11 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
      * cleanup) */
     xcb_change_save_set(conn, XCB_SET_MODE_INSERT, window);
 
+    /* Receive ShapeNotify events whenever the client altered it's window
+     * shape. */
+    if (shape_supported)
```
Airblader: `{}`

src/tree.c
```diff
@@ -248,6 +248,10 @@ bool tree_close_internal(Con *con, kill_window_t kill_window, bool dont_kill_par
              * mapped. See https://bugs.i3wm.org/1617 */
             xcb_change_save_set(conn, XCB_SET_MODE_DELETE, con->window->id);
 
+            /* Stop receiving ShapeNotify events. */
+            if (shape_supported)
+                xcb_shape_select_input(conn, con->window->id, 0);
```
Airblader: `false` and `{}`

src/x.c
```diff
@@ -51,6 +51,9 @@ typedef struct con_state {
     bool need_reparent;
     xcb_window_t old_frame;
 
+    bool shaped;
```
Airblader: Could we add a one line comment to each field which quickly describes what they represent?

src/x.c
```diff
@@ -730,6 +769,32 @@ static void set_hidden_state(Con *con) {
...
+    assert(con->window);
+
+    xcb_shape_combine(conn, XCB_SHAPE_SO_SET, shape_kind, shape_kind, con->frame.id,
+                      con->window_rect.x + con->border_width,
+                      con->window_rect.y + con->border_width,
+                      con->window->id);
+    xcb_rectangle_t rectangles[4];
+    size_t rectangles_count = x_get_border_rectangles(con, rectangles);
+    if (rectangles_count)
+        xcb_shape_rectangles(conn, XCB_SHAPE_SO_UNION, shape_kind, XCB_CLIP_ORDERING_UNSORTED, con->frame.id, 0, 0, rectangles_count, rectangles);
```
Airblader: `{}`

src/x.c
```diff
@@ -793,6 +860,16 @@ void x_push_node(Con *con) {
         con->ignore_unmap++;
         DLOG("ignore_unmap for reparenting of con %p (win 0x%08x) is now %d\n",
              con, con->window->id, con->ignore_unmap);
+
+        need_reshape = true;
```
Airblader: Why do we need to reshape here?

src/x.c
```diff
@@ -898,6 +975,14 @@ void x_push_node(Con *con) {
         fake_notify = true;
     }
 
+    if (shape_supported && need_reshape && con->window != NULL) {
+        /* We need to reshape the window frame only if it already have shape. */
+        if (state->shaped)
```
Airblader: `{}` for both statements

src/x.c
```diff
@@ -898,6 +975,14 @@ void x_push_node(Con *con) {
         fake_notify = true;
     }
 
+    if (shape_supported && need_reshape && con->window != NULL) {
+        /* We need to reshape the window frame only if it already have shape. */
```
Airblader: nit: "[...] only if it already has a shape."

src/x.c
```diff
@@ -1325,3 +1410,26 @@ void x_mask_event_mask(uint32_t mask) {
...
     }
 }
+
+/*
+ * Enables or disables nonrectangular shape of the container frame.
+ */
+void x_set_shape(Con *con, xcb_shape_sk_t kind, bool enable) {
+    struct con_state *state;
+    if ((state = state_for_frame(con->frame.id)) == NULL) {
+        ELOG("window state for con not found\n");
```
Airblader: Please include the ID of the con here

src/x.c
```diff
@@ -1325,3 +1410,26 @@ void x_mask_event_mask(uint32_t mask) {
...
+        return;
+    }
+
+    assert(kind == XCB_SHAPE_SK_INPUT || kind == XCB_SHAPE_SK_BOUNDING);
```
Airblader: I think an assert is overly aggressive here. I'd just switch this construct so a switch case:

```
switch (kind) {
    case XCB_SHAPE_SK_INPUT:
        state->input_shaped = enable;
        break;
    case XCB_SHAPE_SK_BOUNDING:
        state->shaped = enable;
        break;
    default:
        // Use a log instead, also log con ID 
        ELOG("received unknown shape event type for con = ... This is a bug in i3\n");
        break;
}

if (enable) {
    x_shape_frame(con, kind);
} else {
    x_unshape_frame(con, kind);
} 
```

src/x.c
xzfc: We are applying the shape on the parent window. When the parent is changed, we need to apply the shape on the new parent.

Airblader:
FYI, for question 4 #1966 would be the solution, but that isn't exactly a trivial change. I can't see a good way to prevent this issue with simple means, so I think for now we'd have to disable Shape for tiled windows.

src/x.c
Airblader: Yeah, but that doesn't mean in new code we can't do better. Having IDs in log messages makes reading the log files a whole lot easier. :-)

xzfc: Maybe we should add ID logging into `state_for_frame` instead? It is already have `ELOG("No state found\n")` inside.

Airblader: Yeah, that would also be good. If we log here as well, we should still add it here, too, though.

xzfc:
Updated regarding to the comments, added missing `xcb_flush()` in `x_set_shape()`, and made shaping only for floating windows.

src/x.c
```diff
@@ -898,6 +985,28 @@ void x_push_node(Con *con) {
...
+            /* Remove the shape when container is no longer floating. */
+            if (state->shaped) {
+                x_unshape_frame(con, XCB_SHAPE_SK_BOUNDING);
+            }
+            if (state->input_shaped) {
+                x_unshape_frame(con, XCB_SHAPE_SK_INPUT);
+            }
+        }
+    }
+    state->was_floating = con->parent->type == CT_FLOATING_CON;
```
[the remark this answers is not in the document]
orestisf1993: This also segfaults my setup upon starting with xephyr. Got to go, no time to debug but here's a backtrace:

```
#0 0x5570b5f512c1 in x_push_node ../../i3/src/x.c:1008
#1 0x5570b5f547d2 in x_push_changes ../../i3/src/x.c:1229
#2 0x5570b5f2730e in tree_render ../../i3/src/tree.c:463
#3 0x5570b5effc94 in randr_query_outputs ../../i3/src/randr.c:955
#4 0x5570b5f01b1d in randr_init ../../i3/src/randr.c:1091
#5 0x5570b5ed462f in main ../../i3/src/main.c:755
```

src/handlers.c
```diff
@@ -1517,6 +1518,27 @@ void handle_event(int type, xcb_generic_event_t *event) {
...
+             shape->affected_window, shape->shape_kind, shape->shaped);
+
+        Con *con = con_by_window_id(shape->affected_window);
+        if (con == NULL) {
+            LOG("Not a managed window 0x%08x, ignoring shape_notify_event\n",
+                shape->affected_window);
+            return;
+        }
+
+        if (shape->shape_kind == XCB_SHAPE_SK_BOUNDING ||
```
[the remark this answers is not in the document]
xzfc: We need to track that tiled container has shape and apply the shape later when it becomes floating.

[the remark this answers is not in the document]
Airblader: Yeah, those changes wouldn't affect tracking that, would they?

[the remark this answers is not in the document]
xzfc: `state->shape` is set inside `x_set_shape` before checking if window is floating.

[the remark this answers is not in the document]
Airblader: Ah, right, I see now. My bad! Then in x_set_shape we should also use con_is_floating, though.

xzfc:
New update.

* Changes regarding to the new comments.
* Moved `shaped`, `input_shaped` booleans from `struct con_state` (`src/x.c`) to `struct Window` (`include/data.h`) as it reflects the state of application window, not the state of the container.
* Initial check for the shape in `manage_window` (to handle case when the window sets the shape once and newer updates it).
* Fixed segfault.

src/x.c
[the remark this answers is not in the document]
xzfc: Fixed segfault.

Airblader:
@xzfc Great, thanks! Could you also rebase to resolve the conflict?

orestisf1993:
I can take a look but not immediately. Let's say an upper bound of one week?

xzfc:
Travis still fails with `E: Unable to locate package clang-format-4.0`. I think it's related to #3497.
https://travis-ci.org/i3/i3/jobs/456670878#L557

Airblader:
Hm, that's certainly odd because I just restarted a Travis build on next and it goes through fine. I can't investigate it at the moment, though.

Airblader:
@stapelberg Do you happen to know why clang-format-4.0 can't be found when building this PR? Searching for it it does seem like it's not in debian:sid, but then how does it work on the next branch?

stapelberg:
Does it help to rebase the PR onto next? If no, let me know and I’ll take a look this evening.

Airblader:
We rebased a few days ago and it failed then, too. I don't think we changed anything in the time since either.

stapelberg:
Turns out clang-format-4.0 vanished from Debian testing, too, see https://packages.debian.org/search?keywords=clang-format-4.0&searchon=names&suite=all&section=all

Essentially, we’ll need another commit such as https://github.com/i3/i3/commit/eedb25765590460d328f42364cc2396a6fcd9f0e#diff-7985139ae9b6efb45373e3e36e444224, but with clang-format-6.0, I suppose?

stapelberg:
Try rebasing to next now.

src/x.c
```diff
@@ -396,6 +401,55 @@ static void x_draw_decoration_after_title(Con *con, struct deco_render_params *p
     x_draw_title_border(con, p);
 }
 
+/* Get rectangles representing the border around the child window. Some borders
```
orestisf1993:
```suggestion
/*
 * Get rectangles representing the border around the child window. Some borders
```

src/x.c
```diff
@@ -396,6 +401,55 @@ static void x_draw_decoration_after_title(Con *con, struct deco_render_params *p
...
+ * are adjacent to the screen-edge and thus not returned. Return value is the
+ * number of rectangles. */
```
orestisf1993:
```suggestion
 * number of rectangles.
 *
 */
```

xzfc:
* Added test case.
* Rebased onto next. Travis is now green. ✅
* Applied the suggested change by @orestisf1993.

algmyr:
Just wanted to leave a thank you for everyone who worked on the feature. It seems to work perfectly, ignoring non-floating windows as was mentioned earlier. Superb work!

PR #3550, opened 2018-12-09
Fix #3535 - Check for DISPLAY when requesting version information

### LINKED ISSUE #3535: i3 --moreversion should warn when DISPLAY is unset
Similar to https://github.com/i3/i3/blob/7ade46c61f3d7ef4b1cdcffa1b6ad626da0ade42/i3-dump-log/main.c#L148

tlercher:
This change shouldn't be too major.


Airblader:
Thanks!

PR #3551, opened 2018-12-10
Preserve back_and_forth during restart

xzfc:


src/workspace.c
```diff
@@ -11,9 +11,12 @@
 #include "all.h"
 #include "yajl_utils.h"
 
-/* Stores a copy of the name of the last used workspace for the workspace
- * back-and-forth switching. */
-static char *previous_workspace_name = NULL;
+/**
```
Airblader: Here it should just be `/* … */`, following the rest of the docs.

xzfc: Updated.

Airblader:
Thank you!

PR #3559, opened 2018-12-13
Drag threshold

### LINKED ISSUE #3555: Allow drag stacked/tabbed cons by its title
## I'm submitting a…
<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

* Can't drag title to resize in the following layout: `V[w T[w w]]`.
* Can't drag to move floating tabbed con without using floating_modifier.

More detailed, it is not currently allowed to drag con by its title in some cases:
* Vertical resize tiled con by dragging its title:
  * leaf container or container with single child: allowed
  * vsplit: allowed
  * hsplit: **disallowed**
  * tabbed or stacked: **disallowed**
* Move floating con by dragging its title:
  * hsplit or vsplit: allowed
  * tabbed or stacked: **disallowed**

The reason why is that both of the following actions are bound to a mouse button, and i3 needs to distinguish them:
* activate con
* initiate moving/resizing drag

## My proposal

1. Allow dragging an active title regardless of the con type.
   When an user clicked on an active title, they're clearly want to drag the con since it's already activated.

2. Allow dragging an inactive title, but use drag threshold.
   This is the same way as GUI toolkits handle drags. For example see how Firefox or Chromium handles dragging tab.

## Implementation

I'm going to implement this in #3545 or in separate PR, but just want to make sure it'll be accepted.

i3 version 4.16

xzfc:
Fixes #3555.

Complex stuff in src/resize.c (mouse warp, helpwin) will be removed in #3545. It's here to make changes incremental.


Airblader:
Thanks! @orestisf1993 Can you take a look at this?

xzfc:
@orestisf1993 Do you have time to look at this?

orestisf1993:
Sorry for the delay. I don't have much time lately, and I'll probably continue to be very busy in the next months, but I hope I can review this soon (tomorrow or some time in the next week).

In the meantime, I feel that `use_threshold`'s usage is not very clear. Can you add some comments regarding its functionality?

xzfc:
OK, thank you for the reply.

I've committed the comment on `use_threshold`.

src/drag.c
```diff
@@ -0,0 +1,251 @@
...
+    /* The original event initiated drag. */
+    const xcb_button_press_event_t *event;
+
+    /* The dimensions of con when the loop was started. */
+    Rect old_rect;
+
+    /* The callback to invoke after every pointer movement. */
+    callback_t callback;
+
+    /* Drag distance threshold exceeded or never was. */
```
orestisf1993: maybe
```suggestion
    /* Drag distance threshold exceeded at least once. */
```
?

src/click.c
```diff
@@ -311,7 +292,8 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
     else if ((dest == CLICK_BORDER || dest == CLICK_DECORATION) &&
              is_left_or_right_click) {
         DLOG("Trying to resize (tiling)\n");
-        tiling_resize(con, event, dest);
+        tree_render();
```
orestisf1993: Why the `tree_render` here?

src/resize.c
```diff
@@ -179,14 +201,10 @@ void resize_graphical_handler(Con *first, Con *second, orientation_t orientation
         helprect.width = logical_px(2);
         helprect.height = second->rect.height;
         initial_position = second->rect.x;
-        xcb_warp_pointer(conn, XCB_NONE, event->root, 0, 0, 0, 0,
```
orestisf1993: This was intended behavior, do we want to remove it?

src/click.c
xzfc: Consider this:
Layout is `V[w1 T[w2 w3]]`, `w2` is active.

User clicks and drags `w3`.

Since we updated the tree (`con_activate()` at line 253), we need to re-render the tree before returning to the event loop (`drag_pointer` inside `tiling_resize` runs its own event-loop).

This is not an issue on the current i3 version (before the PR), because such dragging is not allowed.

src/drag.c
xzfc: Bad wording on my part. I had in mind
```c
/* Drag distance threshold exceeded. If use_threshold is not set,
 * then threshold_exceeded is always true. */
```

src/resize.c
xzfc: It is moved to line 222 in this PR.

I'm going to remove it at all in #3545 because bspwm-like resize resembles current floating resize which does not require pointer warping and helpwin.

orestisf1993: My bad, I was running with Xephyr and didn't notice

src/click.c
orestisf1993: Ok, lets add an explanation:

```suggestion
        /* Consider this layout: V[w1 T[w2 w3]], w2 is active.
         * User clicks and drags w3. Since we updated the tree (con_activate()
         * above) we need to re-render the tree before returning to the event
         * loop (drag_pointer inside tiling_resize runs its own event-loop). */
        tree_render();
```

xzfc:
Can we continue on this?

I've rebased and squashed it to the current `next`.

PR #3560, opened 2018-12-14
userguide: Document mark --replace flag

CyberShadow:
- Explicitly document --replace, which was previously only mentioned
  in the command syntax.

- Improve wording: "a window can only have one mark" is slightly
  misleading because it appears to describe the limitation as a
  property of the model, whereas this actually pertains the mark
  command.


orestisf1993:
Thanks!

PR #3563, opened 2018-12-14
userguide: Un-hide a TODO block completed in 2011

CyberShadow:
The userguide contained a commented-out section for marks, which included the line:

> TODO: make i3-input replace %s

The line was added in a26a11c6099f82fc9fd8b87f8bda128408b4f7c9 (May 2011), at which point i3-input did not have a -F switch. The switch was added only in 1737a78fcd8025e11398dbcf5acd65c6a07ae86d (September 2011), but the documentation was never updated to enable the commented-out examples.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2428,6 +2428,21 @@ mark [--add|--replace] [--toggle] <identifier>
 unmark <identifier>
 ----------------------------------------------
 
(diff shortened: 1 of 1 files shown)


Airblader:
Thanks!

PR #3574, opened 2019-01-02
Remove \n from errx and die messages

pclouds:
errx() already appends \n internally. "\n" in the error message will
result in a blank line after the message. die() is just a wrapper around
errx() so it receives the same treatment.


pclouds:
This is a similar patch to one that was merged in i3lock, https://github.com/i3/i3lock/pull/216

Issue #3587, opened 2019-01-14
[enhancement] Make binding modes case sensitive


nejni-marji:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

Binding modes are case insensitive

## Desired Behavior

It would be useful if they were not case insensitive, so I could have more bindings with short names

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
Binary i3 version:  4.16 (2018-11-04) © 2009 Michael Stapelberg and contributors
</pre>

<pre>
- Linux Distribution & Version: Arch Linux
- Are you using a compositor (e.g., xcompmgr or compton): No
</pre>

Airblader:
This is technically a bit of a breaking change, but I think it would be OK in this case. Most people use variable names for modes they write on their own anyway, and I think case-sensitivity is something to expect here. Feel free to send  a PR!

PR #3590, opened 2019-01-17
Make binding modes case sensitive

nejni-marji:
Proposal for issue #3587 
Changed `strcasecmp` to `strcmp` where it's used with binding modes


Airblader:
Thanks!

PR #3591, opened 2019-01-17
Update userguide- describe border style distinction

JonWoodlief:
When setting up my config, I seeked to adjust window borders to zero while keeping my title bars. I found nothing in the docs that noted the distinction between the border styles. These couple sentences on that distinction might help clarify that.


Airblader:
Thanks!

Issue #1639, opened 2015-04-06
[enhancement] i3-msg should return with proper return codes


Airblader:
i3-msg currently always returns with `0`. How about having it return with proper error codes so that non-success cases can be handled without parsing the JSON? We should have two error codes, one for `ysuccess(false)` and one for `yerror(…)`.

However, this *can* break compatibility if people already do connect two calls with `&&` or `||` (even though it doesn't make much sense).

Just throwing it out there. If we don't want to do this, I'll just write a wrapper around i3-msg. :)

stapelberg:
For what? If you’re doing anything moderately complex, you should not use i3-msg. It’s a toy for sending the occasional IPC command and serving as an example of how IPC works.

Airblader:
My specific usecase is having a terminal in the scratchpad. I can already have a single keybinding to hide and show that terminal, now I'd also like the terminal to be opened with that shortcut if it hasn't been opened yet. With return codes, I could do something along the lines of

````
for_window[instance="__scratchpad"] floating enable
bindsym <key> exec i3-msg scratchpad show || urxvt -name __scratchpad
```

I don't think this is anywhere near "complex" and would usually expect any command-line call to return a proper error code. I can, of course, solve this particular problem in a number of ways, so I'm fine with not being able to react to the error code. It's just the first thing I'd think of.

stapelberg:
I think having to be more explicit about what should happen on your end is a good thing here (i.e. using `jq` or a higher-level language to parse the error).

As you noted, scratchpad show doesn’t return an error, and personally I wouldn’t want to make this solution only work when that terminal is the only thing in the scratchpad.

PR #3614, opened 2019-02-09
Add proper return code for i3-msg

NilsIrl:
I thought I might do a PR even though it had already been discussed at #1639.

Also maybe making the return code 1 for this situation isn't the best thing maybe -1 would be preferable since it would collide with EXIT_FAILURE.

### DIFF
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -67,6 +67,7 @@ typedef struct reply_t {
     char *errorposition;
 } reply_t;
 
+static int exit_code = 0;
 static reply_t last_reply;
 
 static int reply_boolean_cb(void *params, int val) {
@@ -100,6 +101,7 @@ static int reply_end_map_cb(void *params) {
             fprintf(stderr, "ERROR:               %s\n", last_reply.errorposition);
         }
         fprintf(stderr, "ERROR: %s\n", last_reply.error);
+        exit_code = 2;
     }
     return 1;
 }
@@ -326,5 +328,5 @@ int main(int argc, char *argv[]) {
 
     close(sockfd);
 
-    return 0;
+    return exit_code;
 }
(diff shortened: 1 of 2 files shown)


Airblader:
I think this is worthwhile, because sometimes just knowing a command failed is all you need to know. Given that @stapelberg explicitly rejected this in the past I'd like him to reconsider here, though.

stapelberg:
I’m not entirely opposed to the change (it’s simple enough and returning a non-0 error code when things went wrong is common behavior), but the rationale provided in #1639 still stands: most cases should really not be using i3-msg to begin with.

Can you outline what your use-case is please?

NilsIrl:
It is exactly was is described in #1639:
```bash
i3-msg [instance="$1"] scratchpad show || $2
```
currently what I have to do instead is this:
```bash
if [ $(i3-msg "[instance=\"^$1\$\"] scratchpad show") != '[{"success":true}]' ]; then
    $2
fi
```
Though they could be other use cases.

i3-msg/main.c
```diff
@@ -100,6 +101,7 @@ static int reply_end_map_cb(void *params) {
             fprintf(stderr, "ERROR:               %s\n", last_reply.errorposition);
         }
         fprintf(stderr, "ERROR: %s\n", last_reply.error);
+        exit_code = 1;
```
stapelberg: Let’s make this exit code 2 and document the exit code in the man page.

man/i3-msg.man
```diff
@@ -90,6 +90,15 @@ See the -m option for continuous monitoring.
...
 interface to i3.
 
+=== Exit status:
+
+0:
+if OK,
+1:
+if invalid syntax or unable to connect to ipc-socket
+2:
+if error while executing message
```
stapelberg: for clarity, how about: i3 returned an error processing your command(s)

Airblader:
@NilsIrl Sorry for the wait; if you drop a comment after making the changes we get notified and can take another look. I've reviewed it now, though. Thanks for the PR!

jwflory:
Hey, did this PR resolve #3614 too by chance?

Airblader:
I can't take a look right now but I would suspect that it doesn't. :-)

PR #3623, opened 2019-02-17
Use I3_IPC_MESSAGE_TYPE_RUN_COMMAND and define it.

holoxs:
As We know I3_IPC_MESSAGE_TYPE_COMMAND is deprecated,
that's why we need to use the function that isnt deprecated and
define it.

### DIFF
--- a/include/i3/ipc.h
+++ b/include/i3/ipc.h
@@ -71,6 +71,7 @@ typedef struct i3_ipc_header {
  *
  */
 #define I3_IPC_REPLY_TYPE_COMMAND 0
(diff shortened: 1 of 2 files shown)


include/i3/ipc.h
```diff
@@ -71,6 +71,7 @@ typedef struct i3_ipc_header {
  *
  */
 #define I3_IPC_REPLY_TYPE_COMMAND 0
+#define I3_IPC_MESSAGE_TYPE_RUN_COMMAND 0
```
orestisf1993: Constant is already defined

orestisf1993:
Closing because user is ghost

Issue #3565, opened 2018-12-14
"i3-msg restart" yields error code 1 even on successful restart


jwflory:
## I'm submitting a…
<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

```
$ i3-msg restart
$ echo $?
1
```

Restarting i3wm in place with `i3-msg` yields an error code of `1` – however, the restart appears to complete successfully with no clear error with debugging and verbosity enabled.

## Expected Behavior

```
$ i3-msg restart
$ echo $?
0
```

Yielding the right error code allows me to run commands in a systemd unit file and script certain behavior with i3wm. Currently, whenever my systemd service restarts, it fails because the error code yields `1`. However, I visibly notice the restart takes place and the desired functionality from my systemd unit file works.

This also makes it difficult for me to automate the installation of my systemd unit file with Ansible, because when the systemd service returns a failure, the running Ansible playbook halts.

## Reproduction Instructions
This happens any time I call the command after logging into i3wm from gdm.

## Environment
Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: 4.15 (2018-03-10)
</pre>

<pre>
https://github.com/jwflory/swiss-army/blob/master/roles/apps/i3wm/files/config
</pre>

<pre>
- Logfile URL: https://logs.i3wm.org/logs/5730225442783232.bz2
- Linux Distribution & Version: Fedora 28
- Are you using a compositor (e.g., xcompmgr or compton): no
</pre>

i3bot:
Sorry, we can only support the latest major version. Please upgrade from 4.15 to 4.16, verify the bug still exists, and re-open this issue.

Airblader:
I'm fairly sure we didn't change anything about this in 4.16 but would you mind double checking anyway? Thanks!

jwflory:
@Airblader Thanks, I updated to 4.16 now and confirmed the issue is still there.

I didn't realize I was out of date. It's cool @i3bot picked up on that! I wish more projects had things like this…

jk977:
[...]

PR #3624, opened 2019-02-17
Updated docs for COMMAND reply.

Iskustvo:
Note that my findings are observed from responses over sockets, I haven't looked in source code of i3.
Therefore, I am not 100% sure about i3 closing the sockets on restart/shutdown and you should correct me if I'm wrong.
However, I am sure that reading response from same socket for restart/shutdown command is invalid.
While debugging this, there was never even "i3-ipc" magic string in the socket which was successfully used for communication previously.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -137,6 +137,12 @@ TICK (10)::
 The reply consists of a list of serialized maps for each command that was
 parsed. Each has the property +success (bool)+ and may also include a
 human-readable error message in the property +error (string)+.
+Note that when you send multiple commands, separated by semicolon,
+commands are parsed and executed one by one.
+This means that commands like +restart+ and +exit+ are special cases
+because after reading them, i3 will restart/shutdown and close sockets.
+Therefore, you should NOT try to read +success (bool)+ or +error (string)+
+status for them.
 
 *Example:*
 -------------------


Airblader:
@orestisf1993 Can you take a look at this given the changes that have been made recently?

orestisf1993:
We didn't really change anything related to this.

After fixing https://github.com/i3/i3/issues/3565 I imagine that users will be able to read the `success` return value correctly, right?

docs/ipc
```diff
@@ -137,6 +137,12 @@ TICK (10)::
 The reply consists of a list of serialized maps for each command that was
 parsed. Each has the property +success (bool)+ and may also include a
 human-readable error message in the property +error (string)+.
+Note that when you send multiple commands, separated by semicolon,
+commands are parsed and executed one by one.
+This means that commands like +restart+ and +exit+ are special cases
```
stapelberg: With commit e4ecc6e4a11a8ea14c9159502c17300fbecb8eb7, the `restart` command is no longer a special case here.

docs/ipc
```diff
@@ -137,6 +137,12 @@ TICK (10)::
...
+because after reading them, i3 will restart/shutdown and close sockets.
+Therefore, you should NOT try to read +success (bool)+ or +error (string)+
+status for them.
```
stapelberg: Maybe it makes sense to mention here that you can subscribe to the shutdown event so that you can actually know when i3 is about to exit. We do that in the Go package: https://github.com/i3/go-i3/blob/36e6ec85cc5a706a3ef686b88b2e1fc722771ab6/restart_test.go#L26

Iskustvo:
@stapelberg, first of all, thanks for the fix regarding the `restart` command.

Can you please describe the current expected behavior by answering questions from [this](https://github.com/i3/i3/issues/3565#issuecomment-465357735) comment?
I expect I will have some free time for the weekend to test this behavior and update the docs accordingly.

stapelberg:
Replied to the questions in that comment. Thanks!

Iskustvo:
The `restart` command works as expected now.

However, `exit` command has the same issue - If you send `exit` command and try to read anything (expecting the return status) from the socket, you will read garbage since socket is probably already closed.
@stapelberg, is this expected outcome, should I document it as well?

P.S. Sorry for being pain in the *** :D

stapelberg:
Yeah, for `exit`, let’s document the behavior.

Iskustvo:
I somehow destroyed old fork and had to create a new one, and with it a new pull request.
Please follow the changes on #3754.

stapelberg:
Closing this in favor of #3754

Issue #3351, opened 2018-08-05
[FR] can we get an event on window resize


laur89:
## I'm submitting a…
[ ] Bug
[X] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)

Would it be possible to get a `window::resize` event over IPC?

wooque:
Any progress on this?

Airblader:
I'm not opposed to adding this event, but progress in terms of someone having worked on this, no, otherwise you'd see it here.

PR #3642, opened 2019-03-11
Send window resize event over IPC on explicit resize (#3351)

wooque:
Added IPC events when user explicitly resizes windows either using `resize shrink|grow...` or by dragging window border


wooque:
I need to get more familiar with codebase to do that and I haven't had free time lately.
I'll close this pull request for now and open new one when I produce more complete solution.

PR #3657, opened 2019-03-21
Added note which config directives could be used at runtime

jeffhuxen:


Airblader:
Thanks!

docs/userguide
```diff
@@ -680,7 +680,8 @@ for_window [class="urxvt"] border pixel 1
 for_window [title="x200: ~/work"] floating enable
 ------------------------------------------------
 
-The valid criteria are the same as those for commands, see <<command_criteria>>.
+The valid criteria are the same as those for commands, see <<command_criteria>>. Only config
```
orestisf1993: > Only config directives with a command equivalent can be executed at runtime

This is not correct, commands don't have a config equivalent, `for_window` can execute any i3 command.

docs/userguide
```diff
@@ -1846,6 +1847,9 @@ The criteria +class+, +instance+, +role+, +title+, +workspace+ and +mark+ are
 actually regular expressions (PCRE). See +pcresyntax(3)+ or +perldoc perlre+ for
 information on how to use them.
 
+Note that config directives listed under <<configuring>> cannot be changed at runtime 
+unless they happen to have a command equivalent.
```
orestisf1993: Is there any config directive with a command equivalent? Commands that make sense in a config file are not config directives. Maybe we should explain this to avoid further confusion.

Airblader: @orestisf1993 Do you want to send a followup PR?

PR #3669, opened 2019-03-29
Allow checking for duplicate bindings with -C 

### LINKED ISSUE #3660: i3 -C does not report duplicate keybindings
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

Using a test file with content:
```
# variables
set $mod Mod4
bindsym $mod+Left focus left
bindsym $mod+c reload
bindsym $mod+c restart

bar {
    status_command i3status
}
```

i3 -C does not report the duplicate keybinding. I guess this is related to https://github.com/i3/i3/issues/338 but that issue is closed, and this bug still exists in a later version than the target on that issue.

## Expected Behavior
i3 -C should report the duplicate keybinding

## Reproduction Instructions
Run `i3 -c /path/to/test -C`

## Environment

<pre>
i3 version: 
Running i3 version: 4.16.1 (2019-01-27) (pid 8317)abort…)
Loaded i3 config: /home/reeto/.config/i3/config (Last modified: Sun 24 Mar 2019 13:23:10 GMT, 545 seconds ago)

</pre>

Technically I'm running i3-gaps, but when I briefly tested, the same issue existed in regular i3.

<pre>
</pre>

orestisf1993:
Fixes #3660

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -69,6 +69,9 @@ bool parse_configuration(const char *override_configpath, bool use_nagbar) {
 /*
  * (Re-)loads the configuration file (sets useful defaults before).
  *
+ * If you specify override_configpath, only this path is used to look for a
+ * configuration file.
+ *
  */
 void load_configuration(xcb_connection_t *conn, const char *override_configpath, bool reload) {
     if (reload) {
--- a/src/config.c
+++ b/src/config.c
@@ -73,7 +73,7 @@ bool parse_configuration(const char *override_configpath, bool use_nagbar) {
  * configuration file.
  *
  */
-void load_configuration(xcb_connection_t *conn, const char *override_configpath, bool reload) {
+void load_configuration(const char *override_configpath, bool reload) {
     if (reload) {
         /* If we are currently in a binding mode, we first revert to the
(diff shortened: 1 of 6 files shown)


src/config.c
```diff
@@ -39,39 +39,22 @@ void update_barconfig(void) {
...
-    return parse_file(path, use_nagbar);
-}
-
 /*
  * (Re-)loads the configuration file (sets useful defaults before).
  *
+ * If you specify override_configpath, only this path is used to look for a
+ * configuration file.
+ *
+ * If the global variable conn is still NULL, we assume that we are just
```
Airblader: I don't particularly like guessing context as this can easily break. Maybe we should have an enum for the three different scenarios (validate, load, reload) and just pass that?

orestisf1993: In this case is more like doing the sensible thing according to i3's state:

- We can't call the nagbar if we don't know if the X server is alive
- The default font can't be loaded without an X connection

Airblader: We can still assert the connection to be set in the non-validating cases :-)

orestisf1993:
up

PR #3701, opened 2019-05-13
i3bar: If click events disabled, use whole bar for other events

### LINKED ISSUE #3700: cannot disable click events on i3bar
Last known working version: 8233454

Broken in: a9e476e to b6485c9

I [don't specify click events](https://github.com/acrisci/i3-dstatus/blob/master/i3dstatus/service.py#L84) in my statusline generator. (I also tried setting "click_events" to false) because I don't use them.

I want to be able to scroll anywhere on the top bar at all and cycle through workspaces, even if I'm over text.

To reproduce:

Scroll over a part of i3bar with text on it.

Expected:

Workspaces should cycle.

Actual:

Nothing happens. i3bar eats the click event and prevents default.

orestisf1993:
Fixes #3700


acrisci:
:+1: fixes it thanks

PR #3726, opened 2019-06-21
feat: added support for user-defined border widths in i3bar blocks

### LINKED ISSUE #3722: Support user-defined border widths per edge in i3bar protocol
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Desired Behavior

In i3-gaps, the i3bar protocol allows `border_[top|left|bottom|right]` keys to specify, per status block, the border of each edge in pixels. The exact behavior should be migrated into i3.

## Environment
<pre>
i3 version: 4.16.1
</pre>

Airblader:
This change introduces support for four new properties on the i3bar
protocol, namely "border_top", "border_right", "border_bottom"
and "border_left".

If a block is drawn with a border, these values define the width of
the corresponding edge in pixels. They all default to 1 if not
specified to preserve compatibility.

fixes #3722

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -213,7 +213,7 @@ static uint32_t predict_statusline_length(bool use_short_text) {
 
         render->width = predict_text_width(text);
         if (block->border)
-            render->width += logical_px(2);
+            render->width += logical_px(block->border_left + block->border_right);
 
         /* Compute offset and append for text aligment in min_width. */
         if (block->min_width <= render->width) {
@@ -287,8 +287,8 @@ static void draw_statusline(i3_output *output, uint32_t clip_left, bool use_focu
 
         color_t bg_color = bar_color;
 
-        int border_width = (block->border) ? logical_px(1) : 0;
         int full_render_width = render->width + render->x_offset + render->x_append;
+        int has_border = block->border ? 1 : 0;
         if (block->border || block->background || block->urgent) {
             /* Let's determine the colors first. */
             color_t border_color = bar_color;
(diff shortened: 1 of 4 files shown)


i3bar/src/xcb.c
```diff
@@ -287,8 +287,8 @@ static void draw_statusline(i3_output *output, uint32_t clip_left, bool use_focu
 
         color_t bg_color = bar_color;
 
-        int border_width = (block->border) ? logical_px(1) : 0;
         int full_render_width = render->width + render->x_offset + render->x_append;
+        bool is_border = !!block->border;
```
stapelberg: Why the change in strategy here? I liked the previous solution (assigning 0 or 1 to an integer) better than relying on the fact that bool is implemented as 0 and 1.

Airblader: I agree, I like it more as well. This is mostly just a ported patch and I wrote this years ago… :-) I've adjusted it to an explicit ternary again and also renamed it.

Airblader:
@stapelberg done

i3bar/src/xcb.c
```diff
@@ -287,8 +287,8 @@ static void draw_statusline(i3_output *output, uint32_t clip_left, bool use_focu
 
         color_t bg_color = bar_color;
 
-        int border_width = (block->border) ? logical_px(1) : 0;
         int full_render_width = render->width + render->x_offset + render->x_append;
+        bool has_border = block->border ? 1 : 0;
```
stapelberg: Can you also change the type from bool to int, please?

Airblader: @stapelberg Done

PR #3727, opened 2019-06-21
feat: support transparency (RGBA) in i3bar

### LINKED ISSUE #3723: Support transparency in i3bar
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Desired Behavior

The `i3bar -t` flag from i3-gaps should be migrated into i3 so that it can also be used there.

## Environment
<pre>
i3 version: 4.16.1
</pre>

Airblader:
We introduce a -t flag for i3bar in order to enable a mode which
supports the use of RGBA colors.

An important constraint here is that tray icons will always have
a fully transparent background.

fixes #3723

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1744,6 +1744,26 @@ bar {
 }
 --------------------------------------
 
+=== Transparency
+
+i3bar can support transparency by passing the +--transparency+ flag in the
+configuration:
+
+*Syntax*:
+--------------------------------------
+bar {
+    i3bar_command i3bar --transparency
+}
+--------------------------------------
+
+In the i3bar color configuration and i3bar status block color attribute you can
+then use colors in the RGBA format, i.e. the last two (hexadecimal) digits
+specify the opacity. For example, +#00000000+ will be completely transparent,
+while +#000000FF+ will be a fully opaque black (the same as +#000000+).
+
+Please note that due to the way the tray specification works, enabling this
+flag will cause all tray icons to have a transparent background.
+
 [[list_of_commands]]
 == List of commands
 
--- a/i3bar/include/configuration.h
+++ b/i3bar/include/configuration.h
@@ -48,6 +48,7 @@ typedef struct config_t {
 
     position_t position;
     bool verbose;
+    bool transparency;
     struct xcb_color_strings_t colors;
     bool disable_binding_mode_indicator;
     bool disable_ws;
--- a/i3bar/src/main.c
+++ b/i3bar/src/main.c
@@ -56,13 +56,14 @@ static char *expand_path(char *path) {
 }
 
 static void print_usage(char *elf_name) {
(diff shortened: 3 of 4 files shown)


docs/userguide
```diff
@@ -1744,6 +1744,25 @@ bar {
...
+i3bar can support transparency by passing the +-t+ flag in the configuration:
+
+*Syntax*:
+--------------------------------------
+bar {
+    i3bar_command -t
+}
+--------------------------------------
+
+You can now use colors in the RGBA format in order to achieve transparency.
```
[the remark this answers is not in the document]
orestisfl: I think explicitly saying that the last 2 characters specify the transparency will make this a bit easier to read.

docs/userguide
```diff
@@ -1744,6 +1744,26 @@ bar {
...
 --------------------------------------
 
+=== Transparency
+
+i3bar can support transparency by passing the +-t+ flag in the configuration:
+
+*Syntax*:
+--------------------------------------
+bar {
+    i3bar_command -t
```
stapelberg: This should read `i3bar_command i3bar -t` (you’re missing an “i3bar” in there).

docs/userguide
```diff
@@ -1744,6 +1744,26 @@ bar {
 }
 --------------------------------------
 
+=== Transparency
+
+i3bar can support transparency by passing the +-t+ flag in the configuration:
```
stapelberg: Let’s use the `--transparency` name for the flag here and elsewhere in the userguide. This is something people configure once, not enter interactively frequently, hence we should prefer the descriptive name over the convenient-to-type abbreviation.

docs/userguide
```diff
@@ -1744,6 +1744,26 @@ bar {
...
+
+*Syntax*:
+--------------------------------------
+bar {
+    i3bar_command -t
+}
+--------------------------------------
+
+You can then use colors in the RGBA format in order to achieve transparency,
```
stapelberg: Let’s turn the order around here:

In the i3bar color configuration and i3bar status block color attributes, you can then use colors in the RGBA format, i.e. the last two…

docs/userguide
Airblader: Oops. :-)

Airblader:
@stapelberg All done.

docs/userguide
```diff
@@ -1744,6 +1744,28 @@ bar {
...
+*Syntax*:
+--------------------------------------
+bar {
+    i3bar_command i3bar --transparency
+}
+--------------------------------------
+
+In the i3bar color configuration and i3bar status block color attribute you can
+then use colors in the RGBA format, i.e. the last two (hexadecimal) digits
+specify the opacity. This works in all colors, including the i3bar
```
stapelberg: The “This works in all colors” sentence is now obsolete (it’s at the beginning of the sentence already).

Airblader: Removed

PR #3743, opened 2019-07-19
Make `restart` IPC command send a reply once restart completed (!)

### LINKED ISSUE #3565: "i3-msg restart" yields error code 1 even on successful restart

stapelberg:
This is achieved by retaining the IPC connection which is sending the restart
command across the restart.

This is the cleaner fix for https://github.com/i3/go-i3/issues/3

fixes #3565

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -12,6 +12,8 @@
 #include <stdint.h>
 #include <float.h>
 #include <stdarg.h>
+#include <unistd.h>
+#include <fcntl.h>
 
 #include "shmlog.h"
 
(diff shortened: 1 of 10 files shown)


stapelberg:
Be sure to read https://github.com/i3/go-i3/issues/3#issuecomment-513348174 before diving into the code to understand the context around this change.

stapelberg:
Travis is now passing.

src/commands.c
```diff
@@ -1590,15 +1592,26 @@ void cmd_reload(I3_CMD) {
...
+    ipc_shutdown(SHUTDOWN_REASON_RESTART, exempt_fd);
     unlink(config.ipc_socket_path);
     /* We need to call this manually since atexit handlers don’t get called
      * when exec()ing */
     purge_zerobyte_logfile();
     i3_restart(false);
-
-    // XXX: default reply for now, make this a better reply
-    ysuccess(true);
+    // unreached
```
Airblader: I think we usually `assert(false)` unreachable code. Should we do it here, too?

src/main.c
```diff
@@ -236,6 +236,24 @@ static void setup_term_handlers(void) {
...
+static int parse_restart_fd(void) {
+    const char *restart_fd = getenv("_I3_RESTART_FD");
+    if (restart_fd == NULL) {
+        return -1;
+    }
+
+    char *end = NULL;
+    errno = 0;
+    long int fd = strtol(restart_fd, &end, 10);
+    if ((errno == ERANGE && (fd == LONG_MIN || fd == LONG_MAX)) ||
```
Airblader: Can we reuse `parse_long` here? Perhaps improve it to also handle `errno`?

src/commands.c
stapelberg: Done.

src/main.c
stapelberg: Thanks, I thought we had a helper for that, but couldn’t find it!

PR #3746, opened 2019-07-25
workspace_move_to_output: fix ws creation

antoinemadec:
Fixes #3745

PR #3747, opened 2019-07-28
default config: use workspace number, not just workspace

stapelberg:
This is strictly better: if the configured name does not match the current name,
the correct workspace will still be used.

When creating a new workspace, the configured name is still used.

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -117,30 +117,29 @@ set $ws8 "8"
 set $ws9 "9"
 set $ws10 "10"
 
-
 # switch to workspace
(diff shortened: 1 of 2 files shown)


Airblader:
This change would be unnecessary with #3746, right? If course it shouldn't hurt either way. I'll not merge it since I'm not sure if you want to merge it before the release.

stapelberg:
> This change would be unnecessary with #3746, right?

That issue is only related to assignments, no? This change also covers creating a new workspace by switching to it.

PR #3748, opened 2019-07-28
default config: start xss-lock+i3lock, nm-applet, pactl (volume key)

stapelberg:
This change will make things strictly better for new users (without an existing
configuration file) and has no effect on existing users.

The tools should be fairly uncontentious, I hope, especially as they only serve
as a starting point anyway: users can quickly delete what they don’t want, or
change it into what they prefer.

But having something is strictly better than having nothing :)

We make some space in the config file by removing the old paragraph about pixel
fonts, which seems rather outdated and irrelevant to me.

### DIFF
--- a/etc/config.keycodes
+++ b/etc/config.keycodes
@@ -18,12 +18,21 @@ font pango:monospace 8
 # text rendering and scalability on retina/hidpi displays (thanks to pango).
 #font pango:DejaVu Sans Mono 8
 
-# Before i3 v4.8, we used to recommend this one as the default:
-# font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
-# The font above is very space-efficient, that is, it looks good, sharp and
-# clear in small sizes. However, its unicode glyph coverage is limited, the old
-# X core fonts rendering does not support right-to-left and this being a bitmap
-# font, it doesn’t scale on retina/hidpi displays.
+# The combination of xss-lock, nm-applet and pactl is a popular choice, so
+# they are included here as an example. Modify as you see fit.
+
+# xss-lock grabs a logind suspend inhibit lock and will use i3lock to lock the
+# screen before suspend.
+exec --no-startup-id xss-lock --transfer-sleep-lock -- i3lock --nofork
+
+# NetworkManager is the most popular way to manage wireless networks on Linux,
+# and nm-applet is a desktop environment-independent system tray GUI for it.
+exec --no-startup-id nm-applet
+
+# Use pactl to adjust volume in PulseAudio.
+bindsym XF86AudioRaiseVolume exec --no-startup-id pactl set-sink-volume @DEFAULT_SINK@ +10%
+bindsym XF86AudioLowerVolume exec --no-startup-id pactl set-sink-volume @DEFAULT_SINK@ -10%
+bindsym XF86AudioMute exec --no-startup-id pactl set-sink-mute @DEFAULT_SINK@ toggle
 
 # Use Mouse+$mod to drag floating windows to their wanted position
 floating_modifier $mod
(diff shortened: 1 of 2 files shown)


etc/config.keycodes
```diff
@@ -18,12 +18,18 @@ font pango:monospace 8
...
-# font, it doesn’t scale on retina/hidpi displays.
+# xss-lock grabs a logind suspend inhibit lock and will use i3lock to lock the
+# screen before suspend.
+exec --no-startup-id xss-lock --transfer-sleep-lock -- i3lock --nofork
+
+# NetworkManager is the most popular way to manage wireless networks on Linux,
+# and nm-applet is a desktop environment-independent system tray GUI for it.
+exec --no-startup-id nm-applet
+
+# Use pactl to adjust volume in PulseAudio.
```
Airblader: I really think we should add some explanation that this is just a "starter" and may well require modification. Otherwise we'll bring question upon us to make this work for all kinds of scenarios, which I don't think is really what we want to support.

etc/config.keycodes
```diff
@@ -18,12 +18,18 @@ font pango:monospace 8
...
 
-# Before i3 v4.8, we used to recommend this one as the default:
-# font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
-# The font above is very space-efficient, that is, it looks good, sharp and
-# clear in small sizes. However, its unicode glyph coverage is limited, the old
-# X core fonts rendering does not support right-to-left and this being a bitmap
-# font, it doesn’t scale on retina/hidpi displays.
+# xss-lock grabs a logind suspend inhibit lock and will use i3lock to lock the
+# screen before suspend.
+exec --no-startup-id xss-lock --transfer-sleep-lock -- i3lock --nofork
```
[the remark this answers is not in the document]
stapelberg: I haven’t tried it, but I imagine that the worst case is you’ll get two lock screens in succession, which should be easy to recognize as double configuration.

etc/config.keycodes
stapelberg: Isn’t that always the case for all config files? :)

Where would you like to see such a comment? Asking because you’re commenting on the pactl line specifically AFAICT.

Airblader: In the past we intentionally avoided shipping configuration that depends on the presence of specific tools. Changing this for sane defaults probably makes sense, but then we should also clarify the expectation, IMHO, that we don't intend to create some kind of DE or make this work for "everyone".

stapelberg: Can you make a suggestion for such a disclaimer?

Airblader: Maybe something like this?

> Note that these commands are just an example that should run on most machines and is not intended to be a bullet-proof solution. You may have to change them for your system or personal preferences.

stapelberg: Thanks. I have shortened your suggestion a bit, because I don’t think we need to be overly defensive about it. We have shipped key bindings which rely on other software (dmenu) for many many years. We have started other i3 software for you by default for many many years (i3status in i3bar). These changes are just more of the same :)

stapelberg:
Does this work for you? I’d like to get this change in before the release.

etc/config.keycodes
Airblader: The difference being that dmenu and other i3 tools work the same on all machines, if installed. Something like pactl or alsamixer or whatever depends on the system, presence of hardware etc. etc. etc. :-)

Airblader:
@stapelberg I'm not really seeing anything like what I've proposed in the updated branch. Are you sure you pushed it correctly? In any case, this is a minor thing. If you think we can do without the hint I won't insist on it. :-)

stapelberg:
> I'm not really seeing anything like what I've proposed in the updated branch. 

Did you try shift+reload? :)

This is what I pushed (line 20):
```
# The combination of xss-lock, nm-applet and pactl is a popular choice, so
# they are included here as an example. Modify as you see fit.
```

Airblader:
Ah, I was expecting it right above the binding, but it's a bit higher up. Fair enough!

PR #3751, opened 2019-07-29
304-ipc-workspace-init: add sync to deflake

stapelberg:
Not sure whether this actually fixes the problem, but let’s give it a shot.


stapelberg:
Nevermind, the test is not actually flaky.

PR #3754, opened 2019-07-31
Updated the documentation for COMMAND reply.

Iskustvo:
I somehow destroyed old repository and had to create new one for this pull request.
Please close the pull request #3624 if/when this one is accepted.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -111,12 +111,6 @@ The following reply types are implemented:
 
 COMMAND (0)::
 	Confirmation/Error code for the RUN_COMMAND message.
-+
-Note that when sending the `restart` command, you will get a reply once
-the restart completed. All IPC connection state (e.g. subscriptions)
-will reset, and libraries must be able to cope with it. One way of
-achieving that is to close the connection, if the library already
-supports transparent reconnects.
 WORKSPACES (1)::
 	Reply to the GET_WORKSPACES message.
 SUBSCRIBE (2)::
@@ -144,6 +138,20 @@ TICK (10)::
 The reply consists of a list of serialized maps for each command that was
 parsed. Each has the property +success (bool)+ and may also include a
 human-readable error message in the property +error (string)+.
+
+NOTE: When sending the `restart` command, you will get a singular reply once the
+restart completed. All IPC connection states (e.g. subscriptions) will reset and
+all but one socket will be closed. Libraries must be able to cope with this by
+aligning their internal states. It is also recommended that libraries close
+the last remaining socket(one which replied to `restart` command) to achieve
+the full reset.
+
+NOTE: It is easiest to always send the `restart` command alone: due to i3’s
+state reset, the reply messages of preceding commands are lost, and following
+commands will not be executed.
+
+NOTE: When processing the `exit` command, i3 will immediately exit without
+sending a reply. Expect the socket to be shut down.
 
 *Example:*
 -------------------
@@ -777,7 +785,7 @@ This event consists of a single serialized map containing a property
 +change (string)+ which indicates the type of the change ("focus", "init",
 "empty", "urgent", "reload", "rename", "restored", "move"). A
 +current (object)+ property will be present with the affected workspace
-whenever the type of event affects a workspace (otherwise, it will be +null).
+whenever the type of event affects a workspace (otherwise, it will be +null+).
 
 When the change is "focus", an +old (object)+ property will be present with the
 previous workspace.  When the first switch occurs (when i3 focuses the


docs/ipc
```diff
@@ -145,6 +139,40 @@ The reply consists of a list of serialized maps for each command that was
 parsed. Each has the property +success (bool)+ and may also include a
 human-readable error message in the property +error (string)+.
 
+When sending the +restart+ command, you will get a reply AFTER
+restart is completed. All IPC connection states (e.g. subscriptions)
+will reset and all sockets will be closed.
+Only exception is the socket that sent the +restart+ request, since that
+socket is needed to report execution status of +restart+ command.
+We recommend that you close that socket as well, to really have clean state
+after the restart.
```
stapelberg: Why are you rephrasing my paragraph? I thought it was good as-is.

Can you please revert this to:
```
NOTE: When sending the `restart` command, you will get a singular reply once the
restart completed. All IPC connection state (e.g. subscriptions) will reset, and
libraries must be able to cope with it. One way of achieving that is to close
the connection, if the library already supports transparent reconnects.
```

docs/ipc
```diff
@@ -145,6 +139,40 @@ The reply consists of a list of serialized maps for each command that was
...
+
+Libraries must be able to cope with i3's internal state reset.
+For example, that means that libraries should also clear their internal
+subscription lists. To be notified BEFORE the restart happens, library
+can subscribe to +shutdown+ event.
+
+Note that when sending multiple commands, separated by a semicolon,
```
stapelberg: Thanks for the detailed description. I think we can express it in the required level of detail in fewer words, though. How about replacing these 3 paragraphs with:

```
NOTE: It is easiest to always send the `restart` command alone: due to i3’s
state reset, the reply messages of preceding commands are lost, and following
commands will not be executed.
```

docs/ipc
```diff
@@ -145,6 +139,40 @@ The reply consists of a list of serialized maps for each command that was
...
+that creates 2 strange behaviors when +restart+ is used in this way.
+
+1) Requsted commands after +restart+ won't be executed, since +restart+
+cleared the list of pending commands.
+
+2) Even when commands before +restart+ are executed, only one execution
+status will be returned since +restart+ cleared the list of previous execution
+statuses and then it returned its own execution status.
+
+Note that +exit+ command is also special because after receiving it,
```
stapelberg: For consistency with the previous paragraphs:

```
NOTE: When processing the `exit` command, i3 will immediately exit without
sending a reply. Expect the socket to be shut down.
```

docs/ipc
```diff
@@ -145,6 +139,40 @@ The reply consists of a list of serialized maps for each command that was
...
+will reset and all sockets will be closed.
+Only exception is the socket that sent the +restart+ request, since that
+socket is needed to report execution status of +restart+ command.
+We recommend that you close that socket as well, to really have clean state
+after the restart.
+
+Libraries must be able to cope with i3's internal state reset.
+For example, that means that libraries should also clear their internal
+subscription lists. To be notified BEFORE the restart happens, library
+can subscribe to +shutdown+ event.
```
stapelberg: Is there a use-case for being notified before the restart? This trick was useful for older versions of i3, but now that `restart` has a reply, should we still recommend it? If no, drop the entire paragraph.

docs/ipc
Iskustvo: The only reason I started changing your paragraph was because it broke the format on [latest docs](https://build.i3wm.org/docs/ipc.html#_reply_format).
I was unable to fix it on that spot so I moved it below, where command reply was described.
Now that my additions would be coupled with your explanation, it made sense to join them in single thought process(but it looks like I did a poor job of it).

Regarding the changes in your explanation, there are 2:
1) I wanted to highlight the AFTER part because you [asked](https://github.com/i3/i3/pull/3624/files#r305616814) me to point out the  possibility of subscribing to react BEFORE restart happens. Since we don't see the point of it now, I am OK with reverting this part.

2) I have replaced 
    > One way of achieving that is to close the connection, if the library already supports transparent reconnects.

    with a bit more description.
    I don't like that sentence and wouldn't know what to do from the client side if you haven't [explained](https://github.com/i3/i3/issues/3565#issuecomment-514245214) it to me.
I interpreted this sentence as `One way to achieve this is to close the socket`, which is certainly not the only requirement. Therefore I think we should point out specifically that client side should clear its own internal states and optionally close the last remaining socket.

If you disagree with my proposal from **2.**, I will revert this part as you requested.
Just give me the final confirmation.

docs/ipc
Iskustvo: You are correct, I can't think of any use-case where this might be relevant, I will drop it.

docs/ipc
Iskustvo: I couldn't agree more.

docs/ipc
Iskustvo: Sure, no problem.

docs/ipc
stapelberg: Feel free to clarify point ② as you outlined

Iskustvo:
I am not sure if `NOTE:` was intended to change the style, but Firefox rendered this section like this
image

stapelberg:
> I am not sure if `NOTE:` was intended to change the style, but Firefox rendered this section like this

Yeah, that’s intentional.

Thanks!

PR #3779, opened 2019-08-29
Allow custom theme for i3-nagbar

gartens:
It is now possible to set a custom color theme for i3-nagbar.

I tried to keep it as little intrusive as possible. I am looking forward to your feedback.

### DIFF
--- a/i3-nagbar/main.c
+++ b/i3-nagbar/main.c
@@ -321,16 +321,13 @@ free_resources:
     return result;
 }
 
-static int count_commas(const char *color_string) {
-    int count = 0;
-
-    for (const char *cur = color_string; *cur; cur++) {
-        if (*cur == ',') {
-            count++;
-        }
+static color_t safe_draw_util_hex_to_color(const char *color) {
+    if (color == NULL) {
+        fprintf(stderr, "Invalid color string\n");
+        exit(1);
     }
 
-    return count;
+    return draw_util_hex_to_color(color);
 }
 
 int main(int argc, char *argv[]) {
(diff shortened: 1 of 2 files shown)


i3-nagbar/main.c
```diff
@@ -390,11 +406,20 @@ int main(int argc, char *argv[]) {
                 prompt = i3string_from_utf8(optarg);
                 break;
             case 't':
-                bar_type = (strcasecmp(optarg, "warning") == 0 ? TYPE_WARNING : TYPE_ERROR);
+                if (strcasecmp(optarg, "warning") == 0) {
+                    bar_type = TYPE_WARNING;
+                } else if (strcasecmp(optarg, "custom") == 0) {
+                    bar_type = TYPE_CUSTOM;
+                }
```
barddes: Actually its my first time reading through the source code, so I might be wrong. But shouldn't exist some condition making `bar_type = TYPE_ERROR;`?

gartens: Actually `bar_type` is by default `TYPE_ERROR` (see line 376). So if the type argument is something we don't know, it is just `TYPE_ERROR`. This is also the current behavior.

orestisfl:
I'll let @Airblader reply to this but I don't think we want this feature. Just letting you know before you put more work into this.

Airblader:
Yeah, I am hesitant to merge this as it pushes i3-nagbar into a direction I'm not sure we want to go. There's plenty of good and customizable tools which are far better than i3-nagbar, which we mostly need for config errors and as an initial configuration for the logout in the default config.

gartens:
Ah I was not aware of those tools. Maybe you can give me some pointers?

Airblader:
@gartens dmenu can be used for such things, or something like zenity etc. I don't know all the tools out there, but i3 nagbar exists mostly for our internal use so we have something to use without requiring dependencies.

Closing this PR as explained above. Thanks for understanding!

PR #3783, opened 2019-09-01
fix lcov support

erwinvaneijk:
Ubuntu 18.10 comes with lcov 1.13, which is now added.


orestisfl:
Please rebase with `next` to fix travis

PR #3784, opened 2019-09-01
Added documentation for "fullscreen_mode" in GET_TREE reply.

### LINKED ISSUE #3780: Description for the `fullscreen_mode` field received from IPC request 'GET_TREE'.
## I'm submitting a…
<pre>
[ ] Bug
[ ] Feature Request
[x] Documentation Request
</pre>

## Description
When requesting `GET_TREE` over the IPC, `i3`'s response contains fields named `fullscreen_mode`.
I've got an [reasonable PR](https://github.com/Iskustvo/i3-ipcpp/pull/2) for using these fields in my IPC library, but since it is not documented in the official documentation(at least not for `GET_TREE`), I have some concerns about it:

1. Is `fullscreen_mode` field safe to use?
2. Which nodes have this flag set to `1`? (I've seen it set on some of `i3`'s internal nodes)
3. Will this field always have value or can it be `null`/unset?
4. Can it have any value other than `0` and `1`?

## Version
<pre>
Binary i3 version:  4.17.1 (2019-08-30) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.17.1 (2019-08-30) (pid 647) abort…)
</pre>

Iskustvo:
Closes #3780

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -357,6 +357,14 @@ focus (array of integer)::
 	order. Traversing the tree by following the first entry in this array
 	will result in eventually reaching the one node with +focused+ set to
 	true.
+fullscreen_mode (integer)::
+	Whether this container is in fullscreen state or not.
+    Possible values are
+        +0+ (no fullscreen),
+        +1+ (fullscreened on output) or
+        +2+ (fullscreened globally).
+    Note that all workspaces are considered fullscreened on their respective output.
+
 nodes (array of node)::
 	The tiling (i.e. non-floating) child containers of this node.
 floating_nodes (array of node)::


docs/ipc
```diff
@@ -357,6 +357,14 @@ focus (array of integer)::
 	order. Traversing the tree by following the first entry in this array
 	will result in eventually reaching the one node with +focused+ set to
 	true.
+fullscreen_mode (integer)::
+	Whether this container is in fullscreen state or not.
+	* +0+ - No fullscreen.
+	* +1+ - Local fullscreen(Window is fullscreened on its output).
```
Airblader: Can you please add spaces before the opening parenthesis and lowercase "window"? Applies to both lines.

docs/ipc
```diff
@@ -357,6 +357,14 @@ focus (array of integer)::
...
+	* +2+ - Global fullscreen(Window is globally fullscreened).
+
+NOTE: All workspaces are flagged as fullscreened on their output.
```
Airblader: This line is not indented properly.

Iskustvo: Is there a way for me to see how the final site will look after styles are applied to it?  
The generated `ipc.html` looks weird with this line indented with one Tab, that's the reason I left it unindented in the first place.

Iskustvo:
Thank you for explaining it :smiley:

I have amended the commit, but it looks like the Travis is broken at the moment.

docs/ipc
Airblader: Hm, you're right. I'm sure there's a syntax to achieve this, but I don't know it either. A practical approach would be to just rephrase from a list to this:

> Possible values are +0+ (no fullscreen), +1+ (fullscreened on output) or +2+ (fullscreened globally). Note that all workspaces are considered fullscreened on their respective output.

Airblader:
@stapelberg Any idea what's wrong on Travis? Another PR has the same issue.

orestisfl:
A new warning with gcc

docs/ipc
Iskustvo: Thank you, I have removed the list syntax. I guess linebreaks will be automatically inserted now.

orestisfl:
Please rebase with `next` to fix travis

orestisfl:
Can be merged

Iskustvo:
@Airblader, is everything ok with the changes?
Is there any issue left for this?

Airblader:
The formatting in the file looks weird still, but @orestisfl said it's fine so I'll go with that. Thanks!

Iskustvo:
If the preview is correct, it will all be in one paragraph, without lists or anything, just like you suggested.
For the record, I don't mind several objections regarding the review, you didn't have to merge it if you wanted me to change it again.

Airblader:
All it needs is to look correct once compiled which I couldn't verify from my phone, but since orestis did that's sufficient. Thank you for the patience, though!

PR #3797, opened 2019-09-20
feat: add window_type to ipc response

acheronfail:
Hi :wave: 

I wanted to have the `window_type` property available to `i3` via IPC, so I added this small change.
I'm not very familiar with C and its tooling, so please let me know if there's anything extra I should do.

I've compiled and run this locally, and it works when I test it manually (via `i3-msg`).
~~I have not yet, however, been able to run `i3`'s tests locally - the tools used are completely unfamiliar to me.~~ Nevermind, the tests and builds pass - I figured it out. :+1: 

---

Please let me know what I can do to help get this across the line! :slightly_smiling_face:

### DIFF
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -528,6 +528,36 @@ void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     else
         y(null);
 
+    ystr("window_type");
+    if (con->window) {
+        if (con->window->window_type == A__NET_WM_WINDOW_TYPE_NORMAL) {
+            ystr("normal");
+        } else if (con->window->window_type == A__NET_WM_WINDOW_TYPE_DOCK) {
+            ystr("dock");
+        } else if (con->window->window_type == A__NET_WM_WINDOW_TYPE_DIALOG) {
+            ystr("dialog");
+        } else if (con->window->window_type == A__NET_WM_WINDOW_TYPE_UTILITY) {
+            ystr("utility");
+        } else if (con->window->window_type == A__NET_WM_WINDOW_TYPE_TOOLBAR) {
+            ystr("toolbar");
+        } else if (con->window->window_type == A__NET_WM_WINDOW_TYPE_SPLASH) {
+            ystr("splash");
+        } else if (con->window->window_type == A__NET_WM_WINDOW_TYPE_MENU) {
+            ystr("menu");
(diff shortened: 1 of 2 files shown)


Airblader:
My first question is: what do you need it for? And why can't you read the atom from the window?

 i3 doesn't handle all window types so this is a partial solution anyway.

acheronfail:
> My first question is: what do you need it for?

I want to automatically move windows with the type `notification` to the top right of my display.
Currently, if I have the config:

```
for_window [class="..." window_type="notification"] floating enable, move ..., no_focus
```

I can't actually calculate _where_ the top right of the display is.

So my current solution is writing a small binary using `i3`'s IPC which listens for "new" window events, and then calculates the correct position and proceeds to send `i3` a `[con_id="..."] move position ...` command with the calculated position.

> And why can't you read the atom from the window?

I'm not a C programmer so a lot of this is over my head :sweat_smile: how would I go about doing this?

acheronfail:
Also, if it helps to explain my situation I use i3 together with KDE (like this https://userbase.kde.org/Tutorials/Using_Other_Window_Managers_with_Plasma). 

It’s Plasma’s notifications that I’m trying to move.

Airblader:
I think the usecase is fine, and in principle I don't mind exposing this, except that it's weird that it looks like we're proxying the atom but in reality we may well swallow information because we don't support every window type.

Airblader:
Yes, I guess that's fair. It would also be good to change the code so that if it isn't one of those values, we emit something like `unknown` rather than nothing.

Airblader:
Sorry, didn't see the updates because we only get notified if you comment again afterwards. Thanks for your work!

acheronfail:
@Airblader I'm not sure what i3's release process looks like, but when do you think this change would be included in a release?

Airblader:
@acheronfail We don't have a strict release cycle, but roughly something between every 6–9 months is a realistic number, depending on how much has happened.

@stapelberg Do you already have any plans about 4.18? It's been roughly half a year since 4.17.1

acheronfail:
> We don't have a strict release cycle, but roughly something between every 6–9 months is a realistic number, depending on how much has happened.

Okay, sounds good. :) Glad to hear that it could be sometime soon then!

FYI I just created a PR that documents the changes in this PR: #3931 (just want to make sure that PR and this one get placed in the same release).

stapelberg:
No specific plans yet, but let me go through the release prep work and tentatively target next week or so

PR #3432, opened 2018-10-01
 Fix aspect ratio bugs

orestisf1993:

PR #3810, opened 2019-10-01
Reintroduce aspect ratio logic for tiling containers

### LINKED ISSUE #3795: Content of workspaces (screen background) visible on others
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
i3  4.17.1: Background of one workspace is visible, when switching to a different workspace.

## Expected Behavior
When switching from one workspace, the contents that was seen on the former workspace is not visible anymore.

## Reproduction Instructions
Example reproduction:
One workspace with alsamixer that fills the screen (one tile only).
Then switch to a different workspace (I had more than one tile, but I think this does not matter here). If the used tile is not completely used by the program on the newly encountered (switched to) workspace, the contents of the former worksapce frames the new workspace.

## Environment
<pre>
Arch Linux.
Package that works is 4.16.1-1.
Package with the problem is 4.17.1-1.
</pre>

After finding the problem I downgraded to the older version to be sure, that no other package-installations have caused the problem.
I found out that with downgrading to 4.16.1-1 the problem was gone.
So the problem is a i3-problem.

<pre>
i3 version: 
Running i3 version: 4.16.1 (2019-01-27) (pid 28339)bort…)
Loaded i3 config: /home/oliver/.i3/config (Last modified: Sa 02 Mär 2019 23:50:22 CET, 17187316 seconds ago)

</pre>

See discussion on i3-mailinglist:
https://www.freelists.org/archive/i3-discuss/09-2019

Screenshots:

a) with the problem:
http://s000.tinyupload.com/?file_id=53830511177154834717

b) after downgrading, expected behaviour:
http://s000.tinyupload.com/index.php?file_id=16308606597116519931

orestisfl:
Fixes #3795

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -181,3 +181,10 @@ ssize_t slurp(const char *path, char **buf);
  *
  */
 orientation_t orientation_from_direction(direction_t direction);
+
+/**
+ * Returns the new aspect ratio if the given boundaries are valid and the
+ * width/height ratio lies outside it. Otherwise, returns -1.0.
+ *
+ */
+double aspect_ratio(double min_ar, double max_ar, double width, double height);
--- a/src/floating.c
+++ b/src/floating.c
@@ -139,13 +139,7 @@ void floating_check_size(Con *floating_con, bool prefer_height) {
              * this case according to the ICCCM. */
             double width = floating_con->rect.width - window->base_width - border_rect.width;
             double height = floating_con->rect.height - window->base_height - border_rect.height;
-            const double ar = (double)width / (double)height;
-            double new_ar = -1;
-            if (min_ar > 0 && ar < min_ar) {
-                new_ar = min_ar;
-            } else if (max_ar > 0 && ar > max_ar) {
-                new_ar = max_ar;
-            }
+            const double new_ar = aspect_ratio(min_ar, max_ar, width, height);
             if (new_ar > 0) {
                 if (prefer_height) {
                     width = round(height * new_ar);
--- a/src/render.c
+++ b/src/render.c
@@ -64,6 +64,33 @@ void render_con(Con *con) {
         inset->width -= (2 * con->border_width);
         inset->height -= (2 * con->border_width);
 
+        /* Aspect ratio for floating windows is managed in floating_check_size.
+         * The behaviour differs for tiling containers because the new
+         * dimensions must always be smaller than the original dimensions so
+         * that the window fits in the container. */
+        const double min_ar = con->window->min_aspect_ratio;
+        const double max_ar = con->window->max_aspect_ratio;
+        if (!con_is_floating(con) && con->fullscreen_mode == CF_NONE && (min_ar > 0 || max_ar > 0)) {
+            const double width = inset->width;
+            const double height = inset->height;
+            const double ar = aspect_ratio(min_ar, max_ar, width, height);
+            if (ar > 0) {
+                /* Maintain one dimension and change the other to match aspect
+                 * ratio. Also, x and y are adjusted to center the window. */
+                const double new_height = round(width / ar);
+                if (new_height > height) {
+                    /* new_width <= width always:
+                     * if w / a > h then w > h * a for a > 0 */
+                    const double new_width = round(height * ar);
+                    inset->x += round((width - new_width) / 2);
+                    inset->width = new_width;
+                } else {
+                    inset->y += round((height - new_height) / 2);
+                    inset->height = new_height;
+                }
+            }
+        }
+
         /* NB: We used to respect resize increment size hints for tiling
          * windows up until commit 0db93d9 here. However, since all terminal
          * emulators cope with ignoring the size hints in a better way than we
--- a/src/util.c
+++ b/src/util.c
@@ -518,3 +518,19 @@ ssize_t slurp(const char *path, char **buf) {
 orientation_t orientation_from_direction(direction_t direction) {
     return (direction == D_LEFT || direction == D_RIGHT) ? HORIZ : VERT;
 }
+
+/*
+ * Returns the new aspect ratio if the given boundaries are valid and the
+ * width/height ratio lies outside it. Otherwise, returns -1.0.
+ *
+ */
+double aspect_ratio(double min_ar, double max_ar, double width, double height) {
+    const double ar = width / height;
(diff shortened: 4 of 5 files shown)


xzfc:
After this commit, mpv aspect ratio hints are honored as well (unlike 4.16).

Ignoring mpv aspect ratio hints was a useful feature: mpv could display OSD outside of the movie frame. And this was consistent with i3 behavior regarding "resize increment" hints for tiled windows.

---

I've checked it using `slop -b 20` across commits:

| commit | mpv | mplayer |
| ---    | --- | ---     |
| 29f2510fa9484a5fa235de9a1c5c937292151888^ | inored  | honored, no bug |
| 29f2510fa9484a5fa235de9a1c5c937292151888  | ignored | ignored, bug |
| 6b8b3dcc8927179d45df99f61958f881f12e1ac7^ | ignored | ignored, bug |
| 6b8b3dcc8927179d45df99f61958f881f12e1ac7  | honored | honored¹, no bug |

1) Aspect ratio hints are ignored right after mplayer startup but honored after workspace switching.

orestisfl:
Thank you very much! I'll investigate.

orestisfl:
I think that the problem with mpv is that before https://github.com/i3/i3/pull/3432/commits/29f2510fa9484a5fa235de9a1c5c937292151888 i3 did not save the aspect ratio normal hints in `manage_window` so mpv had `window->aspect_ratio == 0.0`.

I don't think there is any reasonable way to correct that. I much prefer to not introduce problems to the more popular mpv player in order to fix a bug in mplayer.


@Airblader any other ideas?

Airblader:
In my understanding the bug lies with mplayer, right? As such I wouldn't try to work around its issues in i3, particularly if it causes issues with other clients (more popular or not).

orestisfl:
In this case, I don't know which is the "correct" behavior. i3 exposes a bug in mplayer but it would be possible for i3 to misbehave according to the standard.

According to the comment by Michael:
https://github.com/i3/i3/blob/f397698d43e3dd4f4625e1fed1b1d0470d0acdf6/src/render.c#L67-L75
the spec is not clear about fullscreen windows (and certainly not about "tiling" windows) but in this specific case I feel that we should ignore the aspect ratio hints (as in 4.17) like we do for the normal hints.

daviddavo:
Could this be activated with a keyboard shortcut or some kind of action? Having a window locked to an aspect ratio is pretty useful even with tiling containers

xsawyerx:
I'm suffering from the issue this PR is fixing. I was wondering if there are plans to continue the work and merge this fix or a different one?

orestisfl:
I will gladly make the changes if we reach an agreement.

@stapelberg do you have an opinion on this?

stapelberg:
Could you summarize what the question is please? I have read this issue, but it’s not 100% clear to me.

In particular, https://github.com/i3/i3/pull/3810#issuecomment-537787605 says that with your latest commit, both mpv and mplayer do work as expected (isn’t that the desired outcome?), or am I reading this wrong?

---

PS: Hey @xsawyerx! I remember reading your name many times when I was working with Perl. Thanks for all your work, and enjoy i3! :)

xsawyerx:
> Could you summarize what the question is please? I have read this issue, but it’s not 100% clear to me.

I don't know if I'm phrasing it well, but, when using mplayer, the window area, past the video border itself, does not stay black. Instead it reflects other windows backgrounds (like the workspace I moved from, the last image on the screen in full screen, etc.).

> In particular, [#3810 (comment)](https://github.com/i3/i3/pull/3810#issuecomment-537787605) says that with your latest commit, both mpv and mplayer do work as expected (isn’t that the desired outcome?), or am I reading this wrong?

I had missed that. I'm certain I'm not using the latest version, so I'll first upgrade and then check again. Apologies if I added more noise to a resolved issue.

> PS: Hey @xsawyerx! I remember reading your name many times when I was working with Perl. Thanks for all your work, and enjoy i3! :)

I've been a very happy i3 user for a long time now. I think it is I who owes *you* a thank you. :)

xsawyerx:
> > In particular, [#3810 (comment)](https://github.com/i3/i3/pull/3810#issuecomment-537787605) says that with your latest commit, both mpv and mplayer do work as expected (isn’t that the desired outcome?), or am I reading this wrong?

I think this commit is not merged, so I'm not sure upgrading from 4.17.1 (current stable Ubuntu) to 4.18.1 (official stable) will fix the issue. Maybe I'm missing something.

stapelberg:
> I think this commit is not merged, so I'm not sure upgrading from 4.17.1 (current stable Ubuntu) to 4.18.1 (official stable) will fix the issue. Maybe I'm missing something.

Correct, the commit in question is pending in this very PR.

If you want to test, check out this PR and build i3 yourself. You can use `apt build-dep i3-wm` to get all the deps, and `mkdir -p build && cd build && meson .. && ninja` to build i3 :)

orestisfl:
> Could you summarize what the question is please? I have read this issue, but it’s not 100% clear to me.
> 
> In particular, [#3810 (comment)](https://github.com/i3/i3/pull/3810#issuecomment-537787605) says that with your latest commit, both mpv and mplayer do work as expected (isn’t that the desired outcome?), or am I reading this wrong?
> 
> PS: Hey @xsawyerx! I remember reading your name many times when I was working with Perl. Thanks for all your work, and enjoy i3! :)

"Honored" for mpv is not the optimal behavior as explained because subtitles can be placed in the dark areas that appear when the aspect ratios are not honored. I wonder if mpv should report different aspect ratios here because it clearly does not need/want to be that restricted.

stapelberg:
> "Honored" for mpv is not the optimal behavior as explained because subtitles can be placed in the dark areas that appear when the aspect ratios are not honored. I wonder if mpv should report different aspect ratios here because it clearly does not need/want to be that restricted.

Okay, so it’s not optimal because a welcome side-effect of ignoring the aspect ratio hints is no longer triggered. But, it fixes behavior otherwise. Did I get this right?

I’m okay with going ahead with the fix, and requiring people to take up mpv usability improvements with mpv upstream :). Given that mpv is an active open source project, we can fix this the right way instead of having to accept a workaround.

orestisfl:
Ok, I will contact the mpv upstream

PR #3812, opened 2019-10-07
Fix "move con to parent" trick

xzfc:


src/con.c
```diff
@@ -1312,8 +1312,8 @@ bool con_move_to_mark(Con *con, const char *mark) {
         return true;
     }
 
-    if (target->type == CT_WORKSPACE) {
-        DLOG("target container is a workspace, simply moving the container there.\n");
+    if (target->type == CT_WORKSPACE && con_is_leaf(target)) {
```
orestisfl: Is
```suggestion
    if (target->type == CT_WORKSPACE && con_get_workspace(con) == target) {
```
better here?

xzfc: I don't think so.

The purpose of this branch is just to avoid the crash, so it's checking for the minimal case in which i3 crashes. The suggested change (I think you meant `!=` instead of `==`) would still fix the trick, but leave behavior inconsistent for other cases (moving into the workspace should behave the same as moving into any other container).

orestisfl: > I think you meant `!=` instead of `==`

yes

> The purpose of this branch is just to avoid the crash

I see. Given that this is the original reason of this branch and that I agree with "moving into the workspace should behave the same as moving into any other container" I'll accept this

PR #3813, opened 2019-10-07
i3 msg: add 'sync' type

antoinemadec:


Airblader:
Thanks for the PR. This is currently not included in i3-msg because this protocol is really only used in the internal test suite of i3. Would you mind letting us know what you intend to use this for with `i3-msg`?

antoinemadec:
@Airblader I'm trying to automate opening windows and setting up workspaces at startup.
To do that I am sending a bunch of i3-msg like so:
```
i3-msg 'workspace 1:email'
i3-msg 'exec --no-startup-id firefox' 
i3-msg 'layout tabbed'
```
It seems that the 'sync' does not do what I want after ```exec``` but seems to work on other messages.
I might be completely wrong and just follow prosaic superstitious instincts :smile: 

Synchronization is kinda tricky, I have to count the number of windows with **wmctrl** for ```exec``` type of commands, and I was hoping that ```i3-msg -t sync``` would do the trick for other messages.

Is my understanding correct?

Airblader:
I think you might be looking for layout saving/restoring: https://i3wm.org/docs/layout-saving.html

antoinemadec:
@Airblader, thanks this would be a nice fit for my current needs.

However, is there a way to guarantee the order of multiple i3-msg?
Would the 'sync' do this or is my understanding wrong?

orestisfl:
Multiple i3-msg commands that are run one after another do not need synchronization, they run one after another since each of them returns the result of the command.

The problem here probably is that executing firefox and firefox opening a window are two very different things. When i3 executes a process it does not wait for a window to spawn or for the process to finish. So, when you open firefox there is no guarantee for it to run in the current workspace (even though firefox implements the startup-notification protocol).

Airblader:
> When i3 executes a process it does not wait for a window to spawn or for the process to finish.

And just to go into more detail, neither one of those is feasible. The process may well not fork and thus not finish until the program actually terminates, and whether a process will ever open a window, how many it will open or when this will happen is entirely impossible to know upfront.

If you want to do something after the process has been started, chain the command after it. If you want to do something after a window has opened, use `for_window` (layout restoring is essentially the same).

Im going to close this PR because the sync protocol is used for our tests and unless there's an actual usecase to have it in i3-msg, I'd rather avoid misunderstandings like this by not adding it there. Do note that i3-msg is just a reference implementation anyway, the protocol can always be used through the IPC.

antoinemadec:
@Airblader @orestisfl , no problem, thanks a lot for the clear reponse :+1:

PR #3816, opened 2019-10-09
Match empty window properties (e. g. no title set) #3308

### LINKED ISSUE #3308: Match empty window properties (e. g. no title set)
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

We cannot check for "title unset" or something similar to match a window.

## Expected Behavior

I can check for title="^$" or title=null to match weird windows.

## Environment

<pre>
Binary i3 version:  4.15.0.1-161-g19735b5e (2018-05-04, branch "gaps-next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.15.0.1-161-g19735b5e (2018-05-04, branch "gaps-next") (pid 3720)
Loaded i3 config: $HOME/.config/i3/config (Last modified: Mon 11 Jun 2018 10:47:18 AM CEST, 726 seconds ago)

</pre>

<pre>
- Linux Distribution & Version: Ubuntu 18.04
</pre>

I've asked about this here, describing the use case: https://www.reddit.com/r/i3wm/comments/87qlqr/finding_criteria_for_heidisql_window_with_net_wm/

Looking at the code I found that it's simply not possible atm.

sandsmark:
Fixes #3308

### DIFF
--- a/testcases/t/119-match.t
+++ b/testcases/t/119-match.t
@@ -33,9 +33,7 @@ my $win = $content->[0];
 # not match it
 ######################################################################
 # TODO: specify more match types
-# we can match on any (non-empty) class here since that window does not have
-# WM_CLASS set
-cmd q|[class=".*"] kill|;
+# Try matching with an empty pattern since there isn't a WM_CLASS set.
 cmd q|[con_id="99999"] kill|;
 
 is_num_children($tmp, 1, 'window still there');
@@ -100,6 +98,20 @@ ok($left->mapped, 'left window mapped');
 is_num_children($tmp, 1, 'window opened');
 
 cmd '[title="^\w [3]$"] kill';
+wait_for_unmap $left;
+is_num_children($tmp, 0, 'window killed');
+
+######################################################################
+# check that we can match empty properties
+######################################################################
+
+$tmp = fresh_workspace;
+
+$left = open_window(name => 'class is empty', wm_class => '');
+ok($left->mapped, 'left window mapped');
+is_num_children($tmp, 1, 'window opened');
+
+cmd '[class="^$"] kill';
 wait_for_unmap $left;
 is_num_children($tmp, 0, 'window killed');
 
--- a/src/match.c
+++ b/src/match.c
@@ -92,11 +92,9 @@ bool match_matches_window(Match *match, i3Window *window) {
 #define CHECK_WINDOW_FIELD(match_field, window_field, type)                                       \
     do {                                                                                          \
         if (match->match_field != NULL) {                                                         \
-            if (window->window_field == NULL) {                                                   \
-                return false;                                                                     \
-            }                                                                                     \
-                                                                                                  \
-            const char *window_field_str = GET_FIELD_##type(window->window_field);                \
+            const char *window_field_str = window->window_field == NULL                           \
+                                               ? ""                                               \
+                                               : GET_FIELD_##type(window->window_field);          \
             if (strcmp(match->match_field->pattern, "__focused__") == 0 &&                        \
                 focused && focused->window && focused->window->window_field &&                    \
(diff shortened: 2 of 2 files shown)


sandsmark:
Does Github pick it up now (since it doesn't seem like it picked up the one in the title)? Issue #3308

Airblader:
Can you please squash the commits and also add `fixes #3308` to the actual commit message?

testcases/t/119-match.t
```diff
@@ -33,9 +33,8 @@ my $win = $content->[0];
 # not match it
 ######################################################################
 # TODO: specify more match types
-# we can match on any (non-empty) class here since that window does not have
-# WM_CLASS set
-cmd q|[class=".*"] kill|;
+# Try matching with an empty pattern since there isn't a WM_CLASS set.
+cmd q|[class=""] kill|;
```
orestisfl: The purpose of this test here is that matching a window without a `WM_CLASS` set fails (thus the window is not killed).

With your change, I think the command fails due to a parsing error making the test misleading. I think we should delete this certain test if we accept these changes.

archenemies: Out of curiosity, what is the parse error? Also, what else is wrong with the rest of this patch? It looks good to me, unless there is a better place to check for NULL.

But I think there should be some kind of test case to show that we can match windows with empty string class, so if this one is broken ...

Please let me know if any help is needed from me with this.

orestisfl: > Out of curiosity, what is the parse error?

```
$ i3-msg '[class=""] kill'
ERROR: Your command: [class=""] kill
ERROR:                       ^^^^^^^
ERROR: Expected one of these tokens: <word>
[{"success":false,"parse_error":true,"error":"Expected one of these tokens: <word>","input":"[class=\"\"] kill","errorposition":"        ^^^^^^^"}]
```

> Also, what else is wrong with the rest of this patch?

Nothing

archenemies: >     ERROR: Your command: [class=""] kill
>     ERROR:                       ^^^^^^^
>     ERROR: Expected one of these tokens: <word>

But shouldn't it be possible for users to specify the empty string class, especially with this patch? Or maybe that is a problem to be solved in separate patch. Do you know much about the parser or what is happening above?

orestisfl: Your patch modifies a different part of the code, not the parser.

We could modify the parser code to allow this syntax or just remove this specific line since it doesn't work anymore.

orestisfl: The test checks the opposite of what this PR is trying to accomplish: it tests that a window with an empty WM_CLASS is **not** matched with command criteria

orestisfl: > I think we are in agreement here. Although I think the patch is changing the behavior of i3 so that the test no longer should succeed, right? In other words ".*" is now going to match the empty string?

yes

> Should we care yet about why class="" produces a parse error, i.e. do you want that problem to be addressed in this PR?

It might be better to support it but it doesn't have to be in this patch. If #3308 is fixed, a new issue can be opened.

> I have looked at testcases/t/119-match.t, it seems that the patch's test should be inserted a little lower, where the script tests that the window gets killed properly.

yes, that's ok

sandsmark: just deleted the old test, and added a new one

sandsmark:
Sorry, completely forgot about this.

But old test case removed, new added, squashed and added reference to bug report in commit message.

Airblader:
@sandsmark Are you still working on this?

sandsmark:
didn't see that the CI failed.

Just want to officially voice my disagreement with clang-format wrt. putting absolutely everything on one huge line. :-P

orestisfl:
Yeah, this is a quirk that you have to know. I split the line into multiple ones in a way that clang-format accepts.

PR #3824, opened 2019-10-13
Use AC_REPLACE_FUNCS

### LINKED ISSUE #2610: Move compatibility functions (e.g. our memmem) into separate files, use AC_REPLACE_FUNCS
Currently, we occasionally provide implementations of functions which aren’t present on some platforms.

The only instance I can find is https://github.com/i3/i3/blob/8d739b7fe15fd351b5d703f922db01be46699a92/src/util.c#L283-L317. If there are others, please point me to them :).

We should consider moving these out into a standalone source file and then using autoconf’s [`AC_REPLACE_FUNCS` macro](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Generic-Functions.html) to compile and link this file on systems where the function is detected as missing.

orestisfl:
Fixes #2610

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -72,7 +72,8 @@ AC_CHECK_TYPES([mode_t, off_t, pid_t, size_t, ssize_t], , [AC_MSG_FAILURE([canno
 AC_FUNC_FORK
 AC_FUNC_LSTAT_FOLLOWS_SLASHED_SYMLINK
 AC_FUNC_STRNLEN
-AC_CHECK_FUNCS([atexit dup2 ftruncate getcwd gettimeofday localtime_r memchr memset mkdir rmdir setlocale socket strcasecmp strchr strdup strerror strncasecmp strndup strrchr strspn strstr strtol strtoul], , [AC_MSG_FAILURE([cannot find the $ac_func function, which i3 requires])])
+AC_CHECK_FUNCS([atexit dup2 ftruncate getcwd gettimeofday localtime_r memchr memset mkdir rmdir setlocale socket strcasecmp strchr strdup strerror strncasecmp strrchr strspn strstr strtol strtoul], , [AC_MSG_FAILURE([cannot find the $ac_func function, which i3 requires])])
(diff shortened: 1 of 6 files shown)


Airblader:
@stapelberg Can you take a look at this one?

configure.ac
```diff
@@ -72,7 +72,8 @@ AC_CHECK_TYPES([mode_t, off_t, pid_t, size_t, ssize_t], , [AC_MSG_FAILURE([canno
 AC_FUNC_FORK
 AC_FUNC_LSTAT_FOLLOWS_SLASHED_SYMLINK
 AC_FUNC_STRNLEN
-AC_CHECK_FUNCS([atexit dup2 ftruncate getcwd gettimeofday localtime_r memchr memset mkdir rmdir setlocale socket strcasecmp strchr strdup strerror strncasecmp strndup strrchr strspn strstr strtol strtoul], , [AC_MSG_FAILURE([cannot find the $ac_func function, which i3 requires])])
+AC_CHECK_FUNCS([atexit dup2 ftruncate getcwd gettimeofday localtime_r memchr memset mkdir rmdir setlocale socket strcasecmp strchr strdup strerror strncasecmp strrchr strspn strstr strtol strtoul], , [AC_MSG_FAILURE([cannot find the $ac_func function, which i3 requires])])
+AC_REPLACE_FUNCS([mkdirp strndup])
```
stapelberg: Do we not also need AC_CONFIG_LIBOBJ_DIR to tell autoconf that our replacement funcs live in libi3/? Have you tested this on systems which don’t have these?

orestisfl: Most systems don't have `mkdirp` right? I think `AC_CONFIG_LIBOBJ_DIR` is mainly for linking.

stapelberg: Indeed, yeah. Just got around to giving this a try, and mkdirp ends up in the binary as my system doesn’t have it either.

I guess we can go forward and correct later if necessary.

PR #3829, opened 2019-10-14
_con_move_to_con: focus_next only if in different workspaces

orestisfl:


testcases/t/294-focus-order.t
```diff
@@ -280,4 +280,33 @@ cmd '[id=' . $windows[0]->id . '] kill';
 kill_and_confirm_focus($windows[2]->id, 'window 2 focused after tiling killed');
 kill_and_confirm_focus($windows[3]->id, 'window 3 focused after tiling killed');
 
+######################################################################
+# cmp_tree tests
+######################################################################
+
+cmp_tree(
```
Airblader: These tests really take away a lot of boilerplate!

PR #3834, opened 2019-10-15
Reparent windows to their current position on unmanage

### LINKED ISSUE #3819: Polybar appears on wrong monitor after being hidden
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

When hiding and showing Polybars on multi-monitor setup, the polybars show up in wrong monitors. Polybar says this is an issue in i3: https://github.com/polybar/polybar/issues/1484. Also, see that issue for more details about reproducing and expected behavior.

Some more information here: https://github.com/polybar/polybar/wiki/Known-Issues#override-redirect-in-i3

Do you agree this is an issue in i3 and could it be fixed?

## Environment

<pre>
i3 version: 
Binary i3 version:  4.17.1 (2019-08-30) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.17.1 (2019-08-30) (pid 1559)abort…)
Loaded i3 config: /run/current-system/sw/etc/xdg/i3/config (Last modified: Thu 01 Jan 1970 02:00:01 AM EET, 1570718492 seconds ago)

The i3 binary you just called: /nix/store/gayzfi0a0pmbhh2mkj3cfj0yzqkpf7ml-i3-4.17.1/bin/i3
The i3 binary you are running: /nix/store/gayzfi0a0pmbhh2mkj3cfj0yzqkpf7ml-i3-4.17.1/bin/i3
</pre>

<pre>
- Linux Distribution & Version: NixOS 19.09
</pre>

Airblader:
When unmanaging a window we need to reparent it back to the root
window. With this patch we do so at the current position of the
container such that we don't throw away the geometry position.

This fixes cases where clients withdraw a window and reparent it
later on, expecting to keep the same geometry (in particular dock
clients) but then end up on a wrong output.

fixes #3819

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -229,7 +229,7 @@ bool tree_close_internal(Con *con, kill_window_t kill_window, bool dont_kill_par
             xcb_change_window_attributes(conn, con->window->id,
                                          XCB_CW_EVENT_MASK, (uint32_t[]){XCB_NONE});
             xcb_unmap_window(conn, con->window->id);
-            cookie = xcb_reparent_window(conn, con->window->id, root, 0, 0);
+            cookie = xcb_reparent_window(conn, con->window->id, root, con->rect.x, con->rect.y);
 
             /* Ignore X11 errors for the ReparentWindow request.
              * X11 Errors are returned when the window was already destroyed */


Airblader:
Opening as draft until this is verified. Also, I have tracked the (0, 0) position back to when it was introduced and I think there was no specific rationale behind it; it probably just seemed easiest. Interestingly, openbox also uses (0, 0) specifically for dock clients but not others (I believe), but this may also just be an oversight.

Airblader:
Thanks for the confirmation!

PR #3835, opened 2019-10-15
Fix crash with moving container that is to be flattened

### LINKED ISSUE #3831: tree_flatten causes crash when moving parent
4.17-199-ga638e0408

See testcase:
```perl
use i3test;

cmp_tree(
    msg => 'TODO',
    layout_before => 'a H[V[b* c]]',
    layout_after => 'TODO',
    cb => sub {
        cmd 'focus parent, focus parent, move down';

        does_i3_live;
    });

done_testing;
```

https://logs.i3wm.org/logs/5648198538362880.bz2

orestisfl:
Fixes #3831

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -237,8 +237,8 @@ static void move_to_output_directed(Con *con, direction_t direction) {
     /* force re-painting the indicators */
     FREE(con->deco_render_params);
 
-    tree_flatten(croot);
     ipc_send_window_event("move", con);
+    tree_flatten(croot);
     ewmh_update_wm_desktop();
 }
 
@@ -384,7 +384,7 @@ end:
     /* force re-painting the indicators */
     FREE(con->deco_render_params);
 
-    tree_flatten(croot);
     ipc_send_window_event("move", con);
+    tree_flatten(croot);
     ewmh_update_wm_desktop();
 }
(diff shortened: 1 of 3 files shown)


src/move.c
```diff
@@ -237,8 +237,8 @@ static void move_to_output_directed(Con *con, direction_t direction) {
     /* force re-painting the indicators */
     FREE(con->deco_render_params);
 
-    tree_flatten(croot);
```
Airblader: Actually, what is the rationale for this? If we flatten only afterwards we change what we send on the IPC. Is that intentional?

orestisfl: It's not optimal but I don't see a way around it currently. For example, in the case where `con` is deleted what would we want to send on the IPC?

For a second I thought that it could be safe to just keep the first child before the flattening and send it if `!con_exists(con)` but this fails the new test:
```diff
diff --git a/src/move.c b/src/move.c
index 77007c6b9..313e1961a 100644
--- a/src/move.c
+++ b/src/move.c
@@ -384,7 +384,11 @@ end:
     /* force re-painting the indicators */
     FREE(con->deco_render_params);
 
-    ipc_send_window_event("move", con);
+    Con * to_send = TAILQ_FIRST(&(con->nodes_head));
     tree_flatten(croot);
+    to_send = con_exists(con) ? con : to_send;
+    assert(con_exists(to_send));
+
+    ipc_send_window_event("move", to_send);
     ewmh_update_wm_desktop();
 }
```

In the resulting layout of the included test there is no single container that would be considered to have moved: `b` and `c` where inside `H` and `V` but none of these exist any more.

PR #3859, opened 2019-11-06
Adds `surface` focusing mode

### LINKED ISSUE #3738: Better handling of stack/tab swamps
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

Currently, as far as _keyboard-based_ focus navigation is concerned, tabs/stacks/tiles are all treated like the same thing (tiles), so that when we use keys to navigate tiles, we also end up shuffling tabs/stacks.

The suggested solution by `i3` is that we press `mod+a+...+a` frequently enough, until we select the entire collection of tabs/stacks, then we move around.  This has two problems:

* `mod+a+...+a` has the asymptotic worst run time complexity of **O(n)**, where **n** is total number of levels/hierarchies in the tabs/stacks tree.  Ideally, if there are **m**-many windows distributed in a perfectly balanced tree, **n=log m**, which is still very convenient even in the ideal case for a tabs/stacks user.  The only users that do not suffer this, are those who do not use stacks/tabs.
* It's practically even worse than **O(n)**, because this number assumes that the user does not forget (while paying attention to a more important task) that he is now jumping over a bunch of stacked/tabbed windows.  More realistically, what happens in many cases is this:

   1. User tries to move from a tile to another tile on the other screen.
   2.  His attention is mainly on that other screen for that other task.
   3.  He has forgotten that, in the middle of the path of tile-jumps exists a collection of tabbed/stacked tiles (aka a **swamp**).
   4.  As he is moving to the destination tile on the other screen, he crosses that swamp and gets stuck for a while.
   5. He gets slowed down in that swamp, as he does this:
       1. Accidentally shuffles some stacks/tabs, 
       2. Then corrects them,
       3. Then presses `mod+a+...+a`, with that annoying **O(n)** to finally make a clean jump out of the swamp of stacks/tabs.

We might call the current behaviour to be _"having stack/tile swamps"_.  I am very confident that tabs/stacks are semantically useful features, which makes me wonder, which one came first:

   * Did most users abandon stacks/tiles because they don't need them?
   * Or did they abandon them because of the stacks/tabs swamps?   I conjecture this came first for most users (and I confirm it is my case).  But, unlike users that have the liberty of being able to ignore stacks/tabs, I cannot due to my workflow.  So currently I use them, and suffer stack/tab swamps.

**Note:** The mouse-based focus navigation does not suffer tab/stack swamps.  We can consider this feature request as the solution to make the keyboard-based focus navigation as good as the mouse-based focus navigation in this regard (handling stack/tab swamps).

## Desired Behavior

As I described [#3724](https://github.com/i3/i3/issues/3724#issuecomment-511272866), I suggest the following:

* A tile is the box that you see in front of you.  Tiles are browsed by, say, `mod+up/down/left/right`.
* If a number of tiles are stacked/tabbed, do not treat them as separate, but rather treat them as a _merged/grouped_ set of tiles.  This way, when you do `mod+up/down/left/right`, you do not shuffle the tabs/tacks, but rather jump over the collection of the merged/grouped tabbed/stacked tiles as a single unit.
* In order to navigate/shuffle tabs/stacks, introduce tab/stack-specific commands, such as, maybe `focus up-stack`, `focus down-stack`, `focus left-stack`, `focus right-stack`, etc, and the same for `*-tab`.
* For backwards compatibility, the default key binding of, say, `focus up-stack` could be the key binding chosen for `focus up`.  This way, users who like the old behavior will not feel of the change, while others, such as me, will benefit greatly.

This will effectively reduce the asymptotic worst run-time complexity of tab/swamp jumping down to **O(1)** from **O(n)**.

## Environment
i3 4.16.1-1

<pre>
- Linux Distribution & Version: archlinux
</pre>

al-caveman:
- Adds 'focus estranged left|right|up|down|next|prev'
    - Modifies sibling focus from: 'focus next|prev sibling' into:
    'focus sibling next|prev' for consistency with the estranged one.

Fixes #3738


Al-Caveman:
Thanks to the latest generalization, the code size increases by only 6 lines.

When I generalized `cmd_focus_direction`, I thought of another implementation that doesn't use `get_tree_next_sibling`.  But the function `get_tree_next_sibling` is still in `src/tree.c` and its header file.  The only reason is that it is used used in `src/click.c`.  

I didn't look into `src/click.c` to see why it needs `get_tree_next_sibling`.  I am thinking:

* Is there room to remove `get_tree_next_sibling` from `src/click.c`, and hence totally delete the function in `src/tree.c` and its header?  This way, this PR will result in code shrinkage while adding an extra feature: lazy focus (previously I called it estranged focus).
* Maybe my implementation of `cmd_focus_sibling` in the now-generalized `cmd_focus_direction` is missing some details?  I personally can't see why my simpler sibling implementation is worse.  But maybe I am wrong.

Airblader:
I'm a bit constrained on time and I think @orestisfl understands this code the best at the moment anyway, perhaps he can take a look and help you further on this for now.

My only comment right now is that we'll need to squash the commits into one proper commit in the end, but of course there's many other things to be done as well (userguide, tests). :-) I'm not yet sure about the "lazy" (for the record, I dislike "estranged" even more). We should try to think of a more descriptive name.

Al-Caveman:
And I just did my 1st github squash ever 😀.  I don't know for sure, but by the looks of it, my heart says that something is not right.

Airblader:
> my heart says that something is not right.

Yeah, it is definitely not correct at the moment. :-/

Al-Caveman:
Fixed commit.  Now my PR looks as if I always was a very smart guy.

I guess the trick is to always forcefully push lies to my github fork?  I.e. 1st rewrite history to look smart, then push such lie by `push -f`?

Airblader:
Yes, squashing rewrites the history of your feature branch and therefore requires rewriting the history. :-)

Al-Caveman:
now i no longer touch the implementation of `cmd_focus_sibling`, nor alter its syntax.  i only extended `cmd_focus_direction` to have an extra argument `mode` that is either `normal` or `surface.

o, and, instead of calling the new mode `lazy` or `estranged`, i called it `surface` - thoughts?  its naming comes from the imagination: it lets you move around the tiles as if you are a bird flying over the surface (visible tiles), without causing to unhide any hidden tiles.

**synopsis:** this feature will allow navigating throughout stacked/tabbed/normal windows without getting stuck in tab/stack swamps (without even needing `mod+a` to get out of the swamp).  it's neat.  i cannot live without this.  imo no one should.  for self-torture you may read #3738.

a neat side effect of this is that, once it is coupled with `next|prev sibling`, we can effectively configure `i3` such that tabs vs. stacks is only a matter of appearance (no other impact).  e.g. we can consider both stacks/tabs to be (say) themes of tabs, so users shuffle them accordingly to (say) minimize pixel wastage.

Airblader:
I'm OK with the complexity of the change, but haven't reviewed it yet. I'm not too happy with the naming yet, though. To be continued. :)

Al-Caveman:
how about `visible` or `visible-only` (or `only-visible`.. dunno which one is more english (or should it be plural?  e.g. `visibles`?  `visibles-only`))?

stapelberg:
Closing this PR as the author is banned from all i3 community spaces after interactions such as (but not limited to) https://www.freelists.org/post/i3-discuss/squigz-abusing-op-power-in-i3-i3offtopic

PR #3871, opened 2019-11-28
added marks field to get_tree reply in ipc documentation

geshido:


Airblader:
@geshido Can you rebase into current next? We had to push a fix for Travis.

geshido:
@Airblader rebased

PR #3889, opened 2020-01-01
Only swallow windows once

### LINKED ISSUE #3888: Multiple placeholders swallows same window
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

When multiple placeholders have overlapping swallow criteria, and only one window is opened that matches several placeholders, all matching placeholders seems to swallow each other.

## Expected Behavior

Only one placeholder swallowing, as was the case prior to commit 0845d7b264fdd8aa65b0d56c100ed716db06c63a.

## Reproduction Instructions

Append the following layout, and start one urxvt.
```{"layout":"splith","nodes":[{"layout":"splitv","nodes":[{"swallows":[{"class":"^URxvt$"}]}]},{"layout":"splitv","nodes":[{"swallows":[{"class":"^URxvt$"}]}]},{"layout":"splitv","nodes":[{"swallows":[{"class":"^URxvt$"}]}]}]}```

## Environment

<pre>
i3 version: Binary i3 version:  4.17-235-gd341b91b © 2009 Michael Stapelberg and contributors
Running i3 version: 4.17-235-gd341b91b (pid 1094) abort…)
Loaded i3 config: /home/lars/.config/i3/config (Last modified: sön 29 dec 2019 13:44:36, 14885 seconds ago)

</pre>

<pre>
# i3 config file (v4)
# litteraly empty
</pre>

<pre>
Logfile URL:
</pre>

<pre>
- Linux Distribution & Version: slackware 14.2
</pre>

### PR DESCRIPTION
**Pull request by izzel**
fixes #3888. Some windows "update" some of their properties like window names when their dimensions change. They don't actually update the properties, but the handlers get called in any case. Which calls remanage_window again. There's also the fact that a window can change its name after it gets mapped and get re-swallowed. So this PR only allows a window to get swallowed once.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -494,6 +494,9 @@ struct Window {
     /* Time when the window became managed. Used to determine whether a window
      * should be swallowed after initial management. */
     time_t managed_since;
+
+    /* The window has been swallowed. */
+    bool swallowed;
 };
 
 /**
--- a/src/manage.c
+++ b/src/manage.c
@@ -273,6 +273,7 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     DLOG("Initial geometry: (%d, %d, %d, %d)\n", geom->x, geom->y, geom->width, geom->height);
 
     /* See if any container swallows this new window */
+    cwindow->swallowed = false;
     Match *match = NULL;
     Con *nc = con_for_window(search_at, cwindow, &match);
     const bool match_from_restart_mode = (match && match->restart_mode);
@@ -361,6 +362,8 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
             match_free(match);
             FREE(match);
         }
+
+        cwindow->swallowed = true;
     }
 
     DLOG("new container = %p\n", nc);
@@ -696,6 +699,11 @@ out:
  *
  */
 Con *remanage_window(Con *con) {
+    /* Make sure this windows hasn't already been swallowed. */
+    if (con->window->swallowed) {
+        run_assignments(con->window);
+        return con;
+    }
     Match *match;
     Con *nc = con_for_window(croot, con->window, &match);
     if (nc == NULL || nc->window == NULL || nc->window == con->window) {
@@ -741,5 +749,6 @@ Con *remanage_window(Con *con) {
         ewmh_update_wm_desktop();
     }
 
+    nc->window->swallowed = true;
     return nc;
 }

### CODE REVIEW

**Comment by Airblader** [comment]
> There's also the fact that a window can change its name after it gets mapped and get re-swallowed.

I think this is a case where we do want it to happen, though, as this is essentially working as expected.

> Some windows "update" some of their properties like window names when their dimensions change.

Ugh, that is frustrating. Perhaps it'd be better to instead only handle an update if something actually changed?

**Comment by izzel** [comment]
I should've said "after the window gets mapped and swallowed by a placeholder, and then later re-swallowed by another placeholder". Do we want windows to get swallowed by multiple placeholders? If so, could you give me an example of where someone would want that?

> Ugh, that is frustrating. Perhaps it'd be better to instead only handle an update if something actually changed?

I agree. Should I include it in this pull request or in another?

Also it has been... awhile, and I'm still a bit new to git and github. Do I need to merge with the latest next branch in my local repo?

**Comment by wlhlm** [comment]
I'm affected by this bug (as described in #3888) and would like to see this fix get merged (tested and works for me). The maintainers marked this as "waiting", but haven't yet answered the questions asked by the submitter above.

>> There's also the fact that a window can change its name after it gets mapped and get re-swallowed.
>
> I think this is a case where we do want it to happen, though, as this is essentially working as expected.

I personally think that a window getting swallowed again after it already replaced a placeholder is surprising behavior.

Regardless of whether we want to allow multiple swallows, or not, I think we should store whether a placeholder was swallowed a window (as this patch implements) in both cases. If we allow multiple swallows, the old window location should be replaced by a placeholder again so as to not jumble up the layout and for that we would need to know if it was already swallowed.

Here is a regression test for #3888:
```diff
diff --git a/testcases/t/542-layout-restore-remanage.t b/testcases/t/542-layout-restore-remanage.t
index 26b50835..fd45beee 100644
--- a/testcases/t/542-layout-restore-remanage.t
+++ b/testcases/t/542-layout-restore-remanage.t
@@ -83,4 +83,67 @@ is($nodes[0]->{name}, 'different_title', 'test window got swallowed');
 
 close($fh);
 
+############################################################
+# Make sure window only gets swallowed once
+############################################################
+# Regression, issue #3888
+$ws = fresh_workspace;
+
+($fh, $filename) = tempfile(UNLINK => 1);
+print $fh <<'EOT';
+// vim:ts=4:sw=4:et
+{
+    // splith split container with 2 children
+    "layout": "splith",
+    "type": "con",
+    "nodes": [
+        {
+            "type": "con",
+            "swallows": [
+               {
+               "class": "^foo$"
+               }
+            ]
+        },
+        {
+            // splitv split container with 2 children
+            "layout": "splitv",
+            "type": "con",
+            "nodes": [
+                {
+                    "type": "con",
+                    "swallows": [
+                       {
+                        "class": "^foo$"
+                       }
+                    ]
+                },
+                {
+                    "type": "con",
+                    "swallows": [
+                       {
+                        "class": "^foo$"
+                       }
+                    ]
+                }
+            ]
+        }
+    ]
+}
+EOT
+$fh->flush;
+cmd "append_layout $filename";
+
+$window = open_window wm_class => 'foo';
+
+change_window_title($window, "different_title");
+
+@content = @{get_ws_content($ws)};
+
+subtest 'regression test that window gets only swallowed once', sub {
+    is($content[0]->{nodes}[0]->{window}, $window->id, 'first placeholder swallowed window');
+    isnt($content[0]->{nodes}[1]->{nodes}[0]->{window}, $window->id, 'second placeholder did not swallow window');
+    isnt($content[0]->{nodes}[1]->{nodes}[1]->{window}, $window->id, 'thid placeholder did not swallow window');
+};
+
 done_testing;
```

Airblader:
> I personally think that a window getting swallowed again after it already replaced a placeholder is surprising behavior.

You can use layout restoring to transition windows between different arrangements, and for that this would be a crucial aspect. Layout restoring is specifically not limited to the startup of the window manager.

> If we allow multiple swallows, the old window location should be replaced by a placeholder agai

This would interfere with some use cases as well (see above), and may even be perceived as a bug if sufficient time has passed in between. 

Whether a window has been swallowed before is not visible to the user, so I wouldn't expect them to memorize forever which window has or hasn't been to predict behavior.

Thanks for the regression test nonetheless :-)

wlhlm:
> You can use layout restoring to transition windows between different arrangements, and for that this would be a crucial aspect. Layout restoring is specifically not limited to the startup of the window manager.

I was unaware of this aspect, and would have to rethink my suggestion above to take this into consideration. However, you say "Layout restoring is specifically not limited to [...]", but I can't find this documented anywhere? All the [layout restore docs](https://i3wm.org/docs/layout-saving.html) say is:

> When an application opens a window that matches the specified swallow criteria, it will be placed in the corresponding placeholder window.

It only talks about *new* windows getting swallowed by placeholders, not ones that already exist when restoring a layout, and my own experimentation with `append_layout` confirms that (with the addition of the behavior implemented by #3759). From what I can find elsewhere, transitioning windows between arrangements works by "unmapping", and then mapping them again, but there seem to be some drawbacks to this method as you wrote on [Reddit](https://www.reddit.com/r/i3wm/comments/deechr/i_made_a_little_script_to_minimizerestore_windows/f2utw9y/):

> Note that un- and remapping a window externally causes i3 to recreate its container (it treats it like a new window), so this will cause window rules to run and remove all container information such as marks.

I feel like it would make more sense to explicitly support this use case for example by adding an argument to the `append_layout` command like `append_layout --consider-existing-windows <layout.json>` so that we don't loose container information as when unmapping a window. In this case i3 could just ignore the "swallowed" property (as implemented by this PR) on existing windows.

Airblader:
Well, you're completely right. I had entirely misremembered that change. So this actually looks fine to me. However, since you already wrote a test for it as well I think it would be great to include it as well. Can you maybe open this PR with the original commit by @izzel and another one to add your test?

PR #3898, opened 2020-01-08
Fixed location of the IPC socket in the docs

dvarrazzo:
Socket gets created by default in `$XDG_RUNTIME_DIR` rather than in `/tmp/`, see 8a40dc0011828f57661a6bd9e2df09c138f4a1a9


Airblader:
Thank you!

PR #3904, opened 2020-01-12
Document floating properties in get_tree reply

sophacles:
Update `doc/ipc` get_tree reply docs to include: `sticky` and `floating`.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -440,6 +440,9 @@ focus (array of integer)::
 	order. Traversing the tree by following the first entry in this array
 	will result in eventually reaching the one node with +focused+ set to
 	true.
+sticky (bool)::
+	Whether this window is "sticky". If it is also floating, this window will
+	be present on all workspaces on the same output.
 fullscreen_mode (integer)::
 	Whether this container is in fullscreen state or not.
     Possible values are


docs/ipc
```diff
@@ -363,14 +363,22 @@ focus (array of integer)::
 	order. Traversing the tree by following the first entry in this array
 	will result in eventually reaching the one node with +focused+ set to
 	true.
+sticky (bool)::
+	Whether this window is "sticky". If it is also floating, this window will
+	be present on all workspaces
```
Airblader: … on all workspaces on the same output. :-)

docs/ipc
```diff
@@ -363,14 +363,22 @@ focus (array of integer)::
...
 	Whether this container is in fullscreen state or not.
     Possible values are
         +0+ (no fullscreen),
         +1+ (fullscreened on output) or
         +2+ (fullscreened globally).
     Note that all workspaces are considered fullscreened on their respective output.
-
+floating (string)::
+	Whether the window was floating, and how the floating was enabled. Can be one of
+	    "auto_off" - floating off
```
orestisfl, line 378: Whitespace seems off here

orestisfl (changes requested):

orestisfl:
What's the status here?

sophacles:
I was cleaning up some old todos and found that this was still open. Sorry on the long delay. It seems that since I opened this the `floating` field was documented elsewhere so I just removed my changes related to that. This now only has the `sticky` field documentation.

docs/ipc
sophacles: Someone else documented `floating` in the interim of me posting and me updating, so resolving this.

Airblader (approved):
Thanks!

Airblader:
@orestisfl It seems like I can't actually merge this since your review is still showing a requested change (weird, I would've thought I can override this…)

orestisfl (approved):

orestisfl:
Sorry, just noticed this.

PR #3905, opened 2020-01-12
Respect match criteria for exec command

### LINKED ISSUE #3903: "exec" command not working with [command_criteria]
**

## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

When writing a bindsym with [command_criteria], whenever I use the "exec" command, it ignores the criteria and runs 100% of the time. Using any other command than "exec" doesn't cause this problem, which leads me to believe there's something specifically wrong with the "exec" command.

## Expected Behavior

The "exec" command runs only for windows narrowed down by the [command_criteria].

## Reproduction Instructions

In my case, these two sets of bindsyms showed the error:
```i3
bindsym $mod+a [con_id=__focused__ floating] exec "i3-msg border normal"
bindsym $mod+b [con_id=__focused__ tiling]   exec "i3-msg border none"
```
The above two commands run regardless of whether the focused window is floating or tiled.
The expected behavior is for $mod+a to only work on floating windows, and $mod+b on tiled windows.

```i3
bindsym $mod+c [con_id=__focused__ floating] border normal
bindsym $mod+d [con_id=__focused__ tiling]   border none
```
The above two commands abide by the criteria, i.e. $mod+c only works on floating windows, and $mod+d only on tiled windows.

(The use of "border normal" and "border none" is completely arbitrary, it's just to illustrate which bindsyms get executed and which don't)

## Environment

<pre>
i3 version: 4.17-239-gd21c3a09
</pre>

<pre>
https://pastebin.com/1veHqCzK
</pre>

<pre>
Logfile URL: https://logs.i3wm.org/logs/5673511062339584.bz2
</pre>

<pre>
- Linux Distribution & Version: Arch Linux, Kernel x86_64 Linux 5.4.10-arch1-1
- Are you using a compositor (e.g., xcompmgr or compton): yes, picom
</pre>

Airblader:
We currently do not evaluate match criteria for the exec command
since generally executing the same command multiple times is
unlikely to make sense.

However, it does make sense when the match is empty and this should
prevent the command from running, which currently does not happen.

For consisteny we execute the command as many times as there are
matched criteria, but print a warning if it matches more than one
container.

fixes #3903

---

This PR is not yet tested.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1207,10 +1207,26 @@ void cmd_kill(I3_CMD, const char *kill_mode_str) {
 void cmd_exec(I3_CMD, const char *nosn, const char *command) {
     bool no_startup_id = (nosn != NULL);
 
-    DLOG("should execute %s, no_startup_id = %d\n", command, no_startup_id);
-    start_application(command, no_startup_id);
+    HANDLE_EMPTY_MATCH;
 
-    ysuccess(true);
+    int count = 0;
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        count++;
+    }
+
+    if (count > 1) {
+        LOG("WARNING: Your criteria for the exec command match %d containers, "
+            "so the command will execute this many times.\n",
+            count);
+    }
+
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        DLOG("should execute %s, no_startup_id = %d\n", command, no_startup_id);
+        start_application(command, no_startup_id);
+    }
+
+    ysuccess(count > 0);
 }
 
 #define CMD_FOCUS_WARN_CHILDREN                                                        \


Airblader:
I'll worry about the build failure once verified.

stapelberg:
Looks good thus far

src/commands.c
```diff
@@ -1207,8 +1207,23 @@ void cmd_kill(I3_CMD, const char *kill_mode_str) {
...
-    start_application(command, no_startup_id);
+    HANDLE_EMPTY_MATCH;
+
+    int count = 0;
+    owindow *current;
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        count++;
+    }
+
+    if (count > 1) {
```
orestisfl:
```suggestion
    if (count != 1) {
```

Airblader: This is actually exactly what we don't want, because using command criteria to make an empty match so the command is not executed is the very use-case from the issue :-)

orestisfl: I just want to make sure that the user is warned when their command is **not** executed because of command criteria

Airblader: That shouldn't be a warning, though. You're right that we need to do something however. I'll adjust the reply in that case.

src/commands.c
```diff
@@ -1207,10 +1207,26 @@ void cmd_kill(I3_CMD, const char *kill_mode_str) {
...
+            "so the command will execute this many times.\n",
+            count);
+    }
+
+    TAILQ_FOREACH(current, &owindows, owindows) {
+        DLOG("should execute %s, no_startup_id = %d\n", command, no_startup_id);
+        start_application(command, no_startup_id);
+    }
+
+    ysuccess(count > 0);
```
orestisfl:
```suggestion
    if (count == 0) {
        yerror("Your criteria for the exec command do not match any container, so the command will not execute");
        } else {
            ysuccess(true);
        }
```

Airblader: Not executing the command is not an error, it is expected behavior.

orestisfl: You used `ysuccess(count > 0);`, the command will return with an error

Airblader: I think this comes down to how we interpret the difference between `ysuccess(false)` and `yerror()`. To me the former is a well-behaved case where nothing happens but in an expected way, while the second one is meant to draw attention to the user. I don't think we have this defined anywhere, however (and it's certainly not handled consistently at the moment).

orestisfl: Yes, but with `ysuccess(false)` the i3-msg command exits with 1. This, from a user perspective, is a failure that is not explained.

Airblader: Ah, true. Then we should probably just do `ysuccess(true)`. Can you send a PR? I don't have access to my machine until the weekend.

PR #3909, opened 2020-01-17
Do not error on exec not matching a container

orestisfl:
See #3903, #3905

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1226,7 +1226,7 @@ void cmd_exec(I3_CMD, const char *nosn, const char *command) {
         start_application(command, no_startup_id);
     }
 
-    ysuccess(count > 0);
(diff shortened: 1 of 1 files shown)


src/commands.c
```diff
@@ -1219,14 +1219,17 @@ void cmd_exec(I3_CMD, const char *nosn, const char *command) {
         LOG("WARNING: Your criteria for the exec command match %d containers, "
             "so the command will execute this many times.\n",
             count);
+    } else if (count == 0) {
+        LOG("WARNING: Your criteria for the exec command do not match a container, "
```
Airblader: I still don't agree that this is a warning. We don't log for any other command that it isn't being executed due to an empty match.

orestisfl:
up

PR #3918, opened 2020-01-25
Fix fno-common problems with gcc10

### LINKED ISSUE #3914: i3 4.17.1 fails to link with gcc 10 due to default -fno-common
### I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
i3 4.17.1 fails to link in the latest Fedora Rawhide buildroot (which just got gcc 10) due to gcc switching the default to `-fno-common` (see also: https://gcc.gnu.org/gcc-10/porting_to.html#common). This causes a plethora of linking errors due to multiple defined constants, as can be seen in the last 3/4 of the latest test build in Fedora Rawhide: https://kojipkgs.fedoraproject.org//work/tasks/6876/40806876/build.log

## Expected Behavior
i3 should link with gcc 10/with `-fno-common`.

## Reproduction Instructions
Compile i3 4.17.1 with gcc 10 or add `-fno-common` to `CFLAGS` and `LDFLAGS` for gcc <= 9.

orestisfl:
Fixes #3914

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
 script:
   - docker run -v $PWD:/usr/src/i3/ -w /usr/src/i3 ${BASENAME} ./travis/check-safe-wrappers.sh
   - docker run -v $PWD:/usr/src/i3/ -w /usr/src/i3 ${BASENAME} ./travis/check-formatting.sh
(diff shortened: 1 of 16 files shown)


stapelberg:
Thanks for fixing this!

PR #3928, opened 2020-02-04
Improved documentation for "window_properties" field in GET_TREE reply.

Iskustvo:
Pointed out the fact that whole `window_properties` field or any of its attributes might be missing from JSON reply for GET_TREE request.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -348,7 +348,8 @@ window (integer)::
 	containers. This ID corresponds to what xwininfo(1) and other
 	X11-related tools display (usually in hex).
 window_properties (map)::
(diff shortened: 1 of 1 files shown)


Airblader:
Can you please rebase instead of merging to keep the history clean?

Iskustvo:
Yes, of course.
I was just curious to see how it looks when conflicts are resolved directly on GitHub :D
I have rebased the change properly now.

P.S. Shouldn't `window_type` be part of `window_properties` instead of separate field?

Airblader:
> Shouldn't window_type be part of window_properties instead of separate field?

Yeah, probably, but that would be a breaking change and there's no real benefit, so it's not worth it to change.

PR #3942, opened 2020-02-16
configure.ac: test for iconv_open with #include <iconv.h>

stapelberg:
The previously used AC_SEARCH_LIBS uses AC_LANG_CALL, which is discouraged
because it intentionally violates the C type system: it has no way of specifying
function parameters, so it does not include the required headers to suppress
type warnings:
https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Generating-Sources.html

The new code does what AC_SEARCH_LIBS does, but uses AC_LANG_PROGRAM instead of
AC_LANG_CALL. It explicitly includes iconv.h and calls iconv_open with the
correct number and type of arguments.

This fixes compilation on systems where libiconv’s iconv.h is discovered first,
but glibc also provides iconv. Previously, this would result in configure
discovering glibc’s iconv, and make failing to compile because -liconv wasn’t
specified, but iconv.h pulled in libiconv.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,8 +84,11 @@ AC_SEARCH_LIBS([ev_run], [ev], , [AC_MSG_FAILURE([cannot find the required ev_ru
 
 AC_SEARCH_LIBS([shm_open], [rt], [], [], [-pthread])
 
-AC_SEARCH_LIBS([iconv_open], [iconv], ,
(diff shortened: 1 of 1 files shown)


Airblader:
@stapelberg Do we want to put this in before the release?

stapelberg:
> @stapelberg Do we want to put this in before the release?

Yes please

PR #3947, opened 2020-02-18
Use OVER operator for drawing text

### LINKED ISSUE #3604: i3bar does not seem to have subpixel rendering
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

The main problem is that i3bar does not seem to render fonts using subpixel rendering. I tried to make a comparison with another bar (lemonbar), which instead seems to work fine. I encounter this problem only on i3bar, all my other applications work as expected. I tried to modify the `.Xresources` file, but it does not work. If I change, for example, the dpi value in that file, I see the effects on i3bar, so I know that it will look for settings in the `.Xresources`. The content of such file is the following:
```
Xft.autohint: 0
Xft.lcdfilter: lcddefault
Xft.hintstyle: hintslight
Xft.hinting: 1
Xft.antialias: 1
Xft.rgba: rgb
```
I don't know if this is a bug or I am missing some configuration, but I cannot find a way to make it work.

## Expected Behavior

i3bar should render fonts using subpixel rendering, resulting in a sharper result of the characters.

## Reproduction Instructions

As I already mentioned, I don't know if this is a bug or I misconfigured something, but it affects only i3bar every time I start it.

## Environment

<pre>
i3 version: 
Binary i3 version:  4.16 (2018-11-04) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.16 (2018-11-04) (pid 586)to abort…)
Loaded i3 config: /home/paolo/.config/i3/config (Last modified: Mon 28 Jan 2019 12:04:02 AM CET, 38239 seconds ago)

</pre>

This is the part in the i3 config where I set the font and the bar:
<pre>
font pango:Hack Nerd Font Mono 7
bar {
	font pango:Hack Nerd Font Mono 9
	colors {
			background $darkblack
			statusline $darkwhite
			separator $cyan
			focused_workspace  $blue $darkblue $darkblack
			active_workspace   $blue $blue $darkwhite
			inactive_workspace $darkblack $darkblack $white
			#urgent_workspace   $darkblack $darkblack $white
		}
	status_command i3blocks
	position top
	mode dock
	modifier None
}
</pre>

I followed the guide to debug i3bar but the only log I found in the file is the following:
<pre>
ERROR block_reap:328: [record] bad exit code 1
</pre>

<pre>
                   -`                    paolo@arch
                  .o+`                   ----------------
                 `ooo/                   OS: Arch Linux x86_64
                `+oooo:                  Host: K56CB 1.0
               `+oooooo:                 Kernel: 4.20.3-arch1-1-ARCH
               -+oooooo+:                Uptime: 11 hours, 19 mins
             `/:-:++oooo+:               Packages: 1447 (pacman)
            `/++++/+++++++:              Shell: bash 5.0.0
           `/++++++++++++++:             Resolution: 1366x768
          `/+++ooooooooooooo/`           WM: i3
         ./ooosssso++osssssso+`          Theme: Adwaita [GTK2], oomox-gruvbox-numix [GTK3]
        .oossssso-````/ossssss+`         Icons: Adwaita [GTK2], oomox-gruvbox-numix [GTK3]
       -osssssso.      :ssssssso.        Terminal: st
      :osssssss/        osssso+++.       Terminal Font: Hack Nerd Font
     /ossssssss/        +ssssooo/-       CPU: Intel i7-3517U (4) @ 3.000GHz
   `/ossssso+/:-        -:/+osssso+-     GPU: Intel 3rd Gen Core processor Graphics Controll
  `+sso+:-`                 `.-/+oso:    GPU: NVIDIA GeForce GT 740M
 `++:.                           `-/+/   Memory: 4081MiB / 7861MiB
 .`                                 `/
- Are you using a compositor (e.g., xcompmgr or compton): yes, compton
</pre>

xzfc:
Bring back commit 16160462a30f186f5b72bb551ba2188670d4e45c which is reverted by commit eba177342f00ee6efdc3eff05665f317e1a837ca.

Advantages of OVER over SOURCE:

* Subpixel font rendering (#3604).
* No transparent garbage around color emoji characters.

---

See https://github.com/i3/i3/pull/2925#issuecomment-556411728

**Renders**:
[before.png](https://user-images.githubusercontent.com/5121426/74693396-9fe17380-51e3-11ea-9037-2393d8a0b86a.png)
[after.png](https://user-images.githubusercontent.com/5121426/74693395-9d7f1980-51e3-11ea-8619-d79028cd6b23.png)

Fixes #3604

### DIFF
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -105,7 +105,7 @@ static void draw_text_pango(const char *text, size_t text_len,
         pango_layout_set_text(layout, text, text_len);
 
     /* Do the drawing */
-    cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
+    cairo_set_operator(cr, CAIRO_OPERATOR_OVER);
     cairo_set_source_rgb(cr, pango_font_red, pango_font_green, pango_font_blue);
     pango_cairo_update_layout(cr, layout);
     pango_layout_get_pixel_size(layout, NULL, &height);


stapelberg:
Can you clarify which cairo version contains the fix, and ideally link to the fix? The only commit referenced in the issue that I can find is https://gitlab.freedesktop.org/cairo/cairo/commit/99427c3f4f6ce7ce3c95c4caa4d2b8ff7c0093d9, but that doesn’t mention OVER?

Airblader (commented):
What happened to doing this conditionally on the alpha channel? This is no longer a pure i3-gaps problem.

xzfc:
@Airblader 
> What happened to doing this conditionally on the alpha channel? This is no longer a pure i3-gaps problem.

As I can see, current i3-next doesn't support RGBA fonts (aka hole-punching), and thus there is no `pango_font_alpha` variable.

xzfc:
@stapelberg
It seems the cairo commit `99427c3f` is the fix of the original issue with SOURCE described in #2908 (text isn't drawn if color emoji is present), which is different from the issues described in the current PR.

stapelberg:
> It seems the cairo commit `99427c3f` is the fix of the original issue with SOURCE described in #2908 (text isn't drawn if color emoji is present), which is different from the issues described in the current PR.

To ensure I understand correctly: the original motivation for this change is gone because of commit 99427c3f.

Can you outline why it’s now safe to re-introduce this change, given that it had to be reverted previously? This needs to go into the commit message.

xzfc:
> To ensure I understand correctly: the original motivation for this change is gone because of commit 99427c3f.

Yes.
| PR    | OP     | comment |
| ----- | ------ | - |
| old   | `SOURCE` | Text isn't drawn in cairo `1.15.8`..`1.15.10` if color emoji is present. |
| #2908 | `OVER`   | Original motivation: fix for cairo `1.15.8`. |
| #3604 | `SOURCE` | The previous patch has been reverted in i3 since it causes a problem in i3-gaps in case if the text alpha is 0 (https://github.com/i3/i3/pull/2925#issuecomment-328358822). |
| #3947 | `OVER`   | This PR. |

> Can you outline why it’s now safe to re-introduce this change, given that it had to be reverted previously? This needs to go into the commit message.

I think it has always been safe to use `OVER` in i3, and I've provided a fix for i3-gaps (https://github.com/i3/i3/pull/2925#issuecomment-556411728).

Airblader:
I won't stand in the way of this change due to the obvious benefit, but I still don't find it satisfactory that something in cairo simply doesn't seem to work correctly (perhaps I'm missing the bigger point as to why this cannot work correctly) and i3 needs to work around that.

With the idea of bringing i3-gaps into i3 this will also affect i3 (I was not aware the alpha channel on fonts wasn't in i3 yet – I thought it was), and the fix for i3-gaps isn't really a fix, but just reduces the issue a bit more (correct me if I'm wrong, but drawing emojis with alpha < 1.0 would still not work).

orestisfl (changes requested):
> Can you outline why it’s now safe to re-introduce this change, given that it had to be reverted previously? This needs to go into the commit message.

@xzfc Can you add this into the commit message?

stapelberg:
Is this still a problem with modern versions of cairo?

I just tried to reproduce it, and when I use i3 with https://github.com/i3/i3/pull/5246 applied, this is the result.png that I get from the test script you provided in https://github.com/i3/i3/pull/2925#issuecomment-556411728 (Thanks! Very useful):

result

Aside from my dpi leaking into the test, I don’t see any artifacts?

PR #3948, opened 2020-02-18
clang-format: bring back ForeachMacros

xzfc:
`ForeachMacros` was disabled in 4211274fcd028a8e33a084e5695290ae0e9f3020 due to the breakage of `include/queue.h`. The currently used version, clang-format-6.0 doesn't break it.


src/con.c
```diff
@@ -684,9 +684,9 @@ bool con_exists(Con *con) {
  */
 Con *con_by_frame_id(xcb_window_t frame) {
     Con *con;
-    TAILQ_FOREACH(con, &all_cons, all_cons)
-    if (con->frame.id == frame)
-        return con;
+    TAILQ_FOREACH (con, &all_cons, all_cons)
+        if (con->frame.id == frame)
+            return con;
```
xzfc: The main benefit is the proper indentation in places like this.

orestisfl: This might be a bit too much to ask but if we are doing big formatting changes, can you also add curly braces in changed lines where they are missing?

orestisfl: Nvm, I did it myself

PR #3950, opened 2020-02-21
Small fixes

xzfc:


Airblader:
Thanks! Could you file the floating resize commit as a separate PR please and also describe the case a bit better?

xzfc:
@Airblader
> Could you file the floating resize commit as a separate PR please and also describe the case a bit better?

Done: #3950 (also updated the commit).

PR #3953, opened 2020-02-22
Makefile: add bear target for using clangd/ccls language servers

stapelberg:
After installing clangd and/or ccls (Emacs eglot defaults to ccls),
run e.g. make bear -j32 and restart your editor (or just its language server).


orestisfl:
I don't see why this has to be included in the Makefile

stapelberg:
It doesn’t have to be, but it seems nice. What’s the harm in having it? If it turns out to be hard to maintain, we can remove it

PR #3954, opened 2020-02-22
Make floating-tiling resize code consistent with plain tiling resize

xzfc:
I've copied the conditions from a plain tiling resize case (see `/* 8: … */` later in function `route_click()`) because I see no reason why they should be different.

Also, I've moved the call to `tree_render()` from `route_click()` to `resize_graphical_handler()`. So that `tree_render()` would be called in both cases.

---

As a consequence, the following bug is fixed:

1. Suppose you have a floating split container `H[a b]`.
2. Drag a border between `a` and `b`.

Old behavior: floating resize.
New behavior: tiling resize.

### DIFF
--- a/src/click.c
+++ b/src/click.c
@@ -238,18 +238,18 @@ static int route_click(Con *con, xcb_button_press_event_t *event, const bool mod
 
         /*  5: resize (floating) if this was a (left or right) click on the
          * left/right/bottom border, or a right click on the decoration.
-         * also try resizing (tiling) if it was a click on the top */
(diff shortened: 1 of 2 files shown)


Airblader:
> Suppose you have a floating split container

This is already something we do not support, it just happens to mostly work. :-)

> I see no reason why they should be different

That said, I don't see a reason against it either.

@orestisfl Can you maybe review this change? I think as of late you're most familiar with it and I just recovered from being sick and have too much on my plate to catch up on for the moment.

PR #3955, opened 2020-02-23
Assume xcb_cursor_context_new never fails

xzfc:
As I can see in `libxcb-cursor` code, the only way of `xcb_cursor_context_new()` returning non-zero is memory allocation failure ([src]); and `libxcb-cursor` already [falls back to `xcb_create_glyph_cursor`](https://gitlab.freedesktop.org/xorg/lib/libxcb-cursor/blob/0.1.3/cursor/load_cursor.c#L214) if it can't load the cursor.

Can we just assume that `xcursor_supported` is always true, and refactor out places [like this](https://github.com/i3/i3/blob/4.18/src/xcb.c#L47-L61)?

[src]: https://gitlab.freedesktop.org/xorg/lib/libxcb-cursor/blob/0.1.3/cursor/cursor.c#L131-132

### DIFF
--- a/i3-nagbar/main.c
+++ b/i3-nagbar/main.c
@@ -38,10 +38,6 @@ xcb_visualtype_t *visual_type = NULL;
 
 #include "i3-nagbar.h"
 
-/** This is the equivalent of XC_left_ptr. I’m not sure why xcb doesn’t have a
- * constant for that. */
-#define XCB_CURSOR_LEFT_PTR 68
-
 #define MSG_PADDING logical_px(8)
 #define BTN_PADDING logical_px(3)
 #define BTN_BORDER logical_px(3)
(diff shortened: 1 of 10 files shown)


Airblader:
SGTM. @stapelberg Do you know of a specific platform that may not have xcursor?

stapelberg:
The usual culprit for not having advanced X11 features are virtual X servers such as the various VNC implementations, e.g. tigervnc.

It would probably be worthwhile to collect a list of X extensions those support. Not sure if their support varies based on compilation-time options, though.

xzfc:
Tested under tigervnc (it seems it supports `Xcursor.theme` defined in xresources) and under Xephyr (it seems it is not; plain default X11 cursors were used). Still fine in both cases.

PR #3957, opened 2020-02-28
clang-format: use TypenameMacros

xzfc:
This commit removes line breaks and extra empty lines introduced in commit fff3f79da9a87a1f790c6328f6615422f2b69b47.


xzfc:
Oh, this option has been added in clang-format-9.0. Should we switch to a more recent version? On Debian Sid, the last available versions are 9 and 10; Ubuntu Bionic has 9.

orestisfl:
I agree we can update clang-format

orestisfl (approved):

PR #3960, opened 2020-02-29
Remove comment referencing old source code

ammgws:
Behaviour was changed in f691a55923850a4d315450925fc98733d07b69c9, so the comment is invalid.


Airblader:
Much appreciated :)

PR #3961, opened 2020-02-29
i3-nagbar: Use _PATH_BSHELL

ammgws:
It seems like this may have been forgotten in f691a55923850a4d315450925fc98733d07b69c9


Airblader:
Thanks!

PR #3963, opened 2020-03-01
i3-input: Make frame configurable with arguments

panovotn:
Add command line arguments to the i3-input tool setting
i3-input frame options. New arguments listed below.

    -C/--fg-color       foreground color (eg. #ffffff)
    -B/--bg-color       background color (eg. #ffffff)
    -W/--border-color   border color     (eg. #ffffff)
    -S/--padding        padding in (logical) pixels
    -b/--border         border width in (logical) pixels

### DIFF
--- a/i3-input/main.c
+++ b/i3-input/main.c
@@ -35,8 +35,6 @@ xcb_visualtype_t *visual_type = NULL;
 #include "i3-input.h"
 
 #define MAX_WIDTH logical_px(500)
-#define BORDER logical_px(2)
-#define PADDING logical_px(2)
 
 /* IPC format string. %s will be replaced with what the user entered, then
  * the command will be sent to i3 */
@@ -57,6 +55,13 @@ static int limit;
 xcb_window_t root;
 xcb_connection_t *conn;
 xcb_screen_t *root_screen;
+struct frame_options {
+    int border;
+    int padding;
+    color_t border_color;
+    color_t fg_color;
+    color_t bg_color;
+};
 
 /*
  * Having verboselog(), errorlog() and debuglog() is necessary when using libi3.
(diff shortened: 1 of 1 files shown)


orestisfl:
@Airblader are we interested in this?

orestisfl:
I feel that i3-input has better third-party alternatives and that we should put it in maintenance mode. Maybe even officially deprecate it. That said, this is not a horrible addition of complexity.

Airblader:
I agree that we don't intend i3-input to compete with the other tools and we see it as a bit of a reference implementation rather than a full fledged tool, but I've seen people use it quite happily over time. I think adding this is within reason in terms of complexity.

I don't think we'll actually deprecate i3-input anytime soon, though.

i3-input/main.c
```diff
@@ -386,10 +394,15 @@ int main(int argc, char *argv[]) {
         {"prefix", required_argument, 0, 'p'},
         {"format", required_argument, 0, 'F'},
         {"font", required_argument, 0, 'f'},
+        {"fg-color", required_argument, 0, 'C'},
```
orestisfl, line 397: i3lock uses lower-case 'c', maybe have some consistency since it's available? https://github.com/i3/i3lock/blob/f3b061233e28ad0cf920c9421adc2459b6920812/i3lock.c#L1025

```suggestion
        {"fg-color", required_argument, 0, 'c'},
```

orestisfl: Or even remove the short options since they don't generally make much sense.

i3-input/main.c
```diff
@@ -530,6 +558,7 @@ int main(int argc, char *argv[]) {
         free(event);
     }
 
+    FREE(frameopts);
```
orestisfl, line 561: Does it need to be freed? It was never malloced in the first place

orestisfl (changes requested):

PR #3965, opened 2020-03-02
refresh tree and update WM on workspace moving to output

medisun:
There was a problem while using multi-head setup with 2 monitors (via intel-virtual-output). Focusing a workspace going crazy after moving workspace to that virtual output. A wrong workspace was focused instead of. Commands like these were used ```wmctrl -a 'HttpRequester'```, ```jumpapp -i Chromium /usr/bin/chromium```

### DIFF
--- a/src/workspace.c
+++ b/src/workspace.c
@@ -1044,6 +1044,9 @@ void workspace_move_to_output(Con *ws, Output *output) {
         workspace_show(ws);
     }
 
+    tree_flatten(croot);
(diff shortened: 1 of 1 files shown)


Airblader:
Hi,

thanks for your contribution. I have to admit that I can't quite understand your description of what is happening. Would you mind describing step by step what you did to reproduce this issue? If this is something we can reproduce it helps us understand the fix and also perhaps a test could be written for it.

PR #3969, opened 2020-03-05
clean up math calculating width of last tab

ghost:


Airblader (changes requested):
Cleaning up the math is fine, but child is a pointer and the arrow operator has to he used. Can you please update the commit? Thanks!

6144:
should be fixed now.

Airblader:
Could you please squash the two commits into a single one? Thanks!

6144:
i don't know how to do that.

Airblader:
You use `git rebase -i HEAD~2` to tell git that you want to interactively rebase the last two commits. This brings up an editor with two lines that say "pick" and then the respective commit.

Change the first word of the second of your two commits to "squash" and then save. Git will now squash both commits together and bring up another editor to tell it what the (combined) commit message should be. By default it just puts both underneath each other. Edit this to a single, useful commit message and save.

You can use `git log` to make sure everything is correct. Then you'll have to push your branch again. Since the rebase rewrites the history, you'll need to force-push it `git push --force`.

Airblader (approved):

PR #3970, opened 2020-03-07
ipc: always include marks property in TREE reply

### LINKED ISSUE #3967: IPC TREE reply does not include marks property if there are none
Recently, the marks field was added to the documentation for the ipc. The manual states that each node must contain the listed fields, but marks are only included if they are present.

Concretely, when a node does not contain any marks, I expect it to contain the "marks" property with a value of an empty list; however, the actual behaviour is to not include the "marks" property.

If this is indeed an issue, I believe that the fix is just to remove the condition at: https://github.com/i3/i3/blob/next/src/ipc.c#L413

ianyfan:
Closes #3967

### DIFF
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -410,17 +410,13 @@ void dump_node(yajl_gen gen, struct Con *con, bool inplace_restart) {
     ystr("urgent");
     y(bool, con->urgent);
 
-    if (!TAILQ_EMPTY(&(con->marks_head))) {
-        ystr("marks");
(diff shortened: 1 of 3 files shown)


xzfc:
Can you also fix the failing test? You can execute `make check` to run tests locally.

ianyfan:
I have amended the offending test cases; let me know if there's a more preferable way to do it: it's my first time doing this.

Airblader:
Thanks!

Issue #3818, opened 2019-10-10
i3bar: protocol for workspace buttons


xzfc:
## Proposal

I am proposing to add `ws_command` option to i3bar. It should be similar to `status_command` but for controlling workspace buttons instead of statusline blocks.

This would allow the user to define arbitrary complex customizations without introducing actual complexity into the i3bar code or i3bar configuration syntax.

Examples of customizations are:
* Show empty workspaces #2617
* Hide certain workspaces #2333
* Groups of workspaces #1808
* Sort workspaces alphabetically #3548

Examples of per-button customizations:
* `"min_width":32` (related to #3711)
* `"separator":true` — personally I find it is useful to visually separate groups of workspaces.
* `"color":[...]` — could be used to distinguish fake empty workspaces.

## Diagram

![i3bar_ws](https://user-images.githubusercontent.com/5121426/66557530-92372580-eb41-11e9-8aba-111bca6e0859.png)

## Example configurations

```
bar {
	ws_command ~/bin/ws_command_default.sh
}
```

`ws_command_default.sh` - recreate the default i3bar behavior.
```sh
#!/bin/sh
i3-msg -t subscribe -m '["workspace"]' | {
	i3-msg -t get_workspaces;
	while read; do i3-msg -t get_workspaces; done;
}
```

`ws_command_hide.sh` - hide workspace named `foo` unless it is visible.
```sh
#!/bin/sh
i3-msg -t subscribe -m '["workspace"]' | {
	i3-msg -t get_workspaces;
	while read; do i3-msg -t get_workspaces; done;
} | jq -c '[ .[] | select(.name != "foo" or .visible) ]'
```

`ws_command_show_empty.sh` - show empty workspaces `foo` and `bar` on `LVDS1` even if they do not exist at the moment.
```sh
#!/bin/sh
i3-msg -t subscribe -m '["workspace"]' | {
	i3-msg -t get_workspaces;
	while read; do i3-msg -t get_workspaces; done;
} | jq -c '
	def fake_ws(name): {
		num: -1, name: name,
		visible: false,
		focused: false,
		urgent: false,
		output: "LVDS1",
		"colors":["#191919", "#111111", "#444444"], # dimmed inactive_workspace
	};
	. + [ fake_ws("foo"), fake_ws("bar") ] | unique_by(.name)
'
```

orestisfl:
I like the idea and since we have the infrastructure to support IPC and JSON parsing this will not be very complex. @Airblader will have to approve though.

[...]

Issue #2333, opened 2016-05-01
enhancement: allow i3bar to not display some workspaces


woodensquares:
I have been using this

https://gist.github.com/woodensquares/c1afc4fb56b4d9d21fa261fb4b28b092

to allow me to have some workspaces hidden from i3bar, not sure if there is any interest in having this be generally available, if there is please let me know. 

As a side note for fun I have taken a stab also at writing a test that exercises this but I am having some issues.

The test creates a couple of workspaces, with windows in each, and I was planning to send button presses and then verify which window was displayed, which would help infer if a to-be-ignored workspace was displayed or not, but even if I have xephyr up with, say, workspace 1 + 2 and 2 focused, and use xtest_button_press/release(1, 500, 500); it will switch to workspace 1, when I would think it should stay on workspace 2 (as I am not clicking anywhere near the top-left i3bar workspace 1 pager)

Is xtest_button_press supposed to work for this kind of test? the only use I have found in the test suite is to test binding to a specific button nothing about checking behavior with mouse position.

Airblader:
Related to #1808 and #2214.

Airblader:
Personally I think hiding workspaces from i3bar is OK, but I don't like the approach with "hide them if their name ends a certain way". I think what @acrisci proposed, which is having underscore(s) in the beginning, is more consistent with other things (e.g., display of marks). Another approach would be a regular expression.

To judge the usefulness of the feature it'd be good if you could also explain what you are using this for, i.e., why you want to hide workspaces?

woodensquares:
to be honest I had a hardcoded 'starts with _' for months that I have been using, but when thinking about sharing it I thought using a trailing symbol (whatever the user decided) would be better because trailing characters seem to be ignored by i3 for "num" (this way you can do scripts creating a new "hidden" workspace automatically by incrementing num of the last hidden, and appending your ignore characters)

I will write a blog post about my use of i3 in the near future, but basically for me my linux desktop is running only i3, terminals if I need something, emacs and virtualbox, and everything else I do pretty much is in virtualbox vms. I have customized i3bar status scripts that display all the VMs currently running, and have them set up so I can click on the vm name and i3 is switched to the workspace of the vm in question.

This means that having the VM workspaces displayed in i3bar is redundant and just takes space, I'd rather have all my virtualbox VMs in hidden workspaces and switch to them via my custom i3bar status handlers. I do think this use is a bit niche, but it's working well for me and I was wondering if others might be interested.

[...]

PR #3986, opened 2020-04-03
WIP: Support layers for managing many workspaces

bastidest:
## Proposal
I want to manage large numbers of workspaces (>20), which can be grouped into different layers (default, work, uni, ...) as they are completely different tasks. I rarely switch **between layers**, but I switch **inside layers** very frequently.

Every layer has a similar structure: WS1 = webbrowser, WS2 = emacs, WS3 = docs, ...

One workspace shall be reachable from every layer, as it contains email, password manager, ...

The idea is to define layers, which make switching workspaces inside them easier. 

First of all: The original behavior obviously needs to stay the same. If one chooses to define a layer, it needs to be added to the configuration. A layer is defined with the directive `layer <name> <from> <to>`.

## Example
output
```
# switch to workspace
bindsym $mod+1 workspace number 1
bindsym $mod+2 workspace number 2
# ...
# layer configuration
layer uni 11 19
layer work 21 29
bindsym $mod+F1 layer default
bindsym $mod+F2 layer uni
bindsym $mod+F3 layer work
# ...
```
This defines a layer "uni" starting at workspace 11, ending at workspace 19 (inclusive) and a layer "work" starting at 21, ending at 29. Also there are keybindings for switching to the defined layers.

### Switching
- At startup, layer "default" is enabled, wich has no influence on any behavior
- Pressing `$mod+F1` will activate layer "uni", but do nothing else
- Upon pressing `$mod+1`, the relative offset of the layer will be added to the switching command: `workspace number 1` -> `workspace number 11`
- Upon pressing `$mod+2`, the relative offset of the layer will be added to the switching command: `workspace number 2` -> `workspace number 12`
- ...
- Upon pressing `$mod+9`, the relative offset of the layer will be added to the switching command: `workspace number 9` -> `workspace number 19`
- Upon pressing `$mod+0`, the switch will happen normally, as the last mapped workspace is `19`

### Moving
"I want to move the current container to the 2nd workspace of layout 'work'" == "I want to move the current container to workspace 22"
- `$mod+F3` == `layer work`
- `$mod+shift+2` == `move container to workspace number 2`

## TODOs
- [ ] See if there is any form of acceptance here, or if I need to maintain this seperately
- [ ] Tests
- [ ] Concept Improvements

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -408,6 +408,7 @@ i3bar_i3bar_SOURCES = \
 	i3bar/include/configuration.h \
 	i3bar/include/ipc.h \
 	i3bar/include/mode.h \
+	i3bar/include/layer.h \
 	i3bar/include/outputs.h \
 	i3bar/include/parse_json_header.h \
 	i3bar/include/trayclients.h \
@@ -420,6 +421,7 @@ i3bar_i3bar_SOURCES = \
 	i3bar/src/ipc.c \
 	i3bar/src/main.c \
 	i3bar/src/mode.c \
+	i3bar/src/layer.c \
 	i3bar/src/outputs.c \
 	i3bar/src/parse_json_header.c \
 	i3bar/src/workspaces.c \
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -91,6 +91,7 @@ extern TAILQ_HEAD(statusline_head, status_block)
 #include "util.h"
 #include "workspaces.h"
 #include "mode.h"
+#include "layer.h"
 #include "trayclients.h"
 #include "xcb.h"
 #include "configuration.h"
--- a/i3bar/include/layer.h
+++ b/i3bar/include/layer.h
@@ -17,10 +17,10 @@
 
 /* Name of current binding mode and its render width */
 struct layer {
-  i3String *name;
-  int name_width;
-  long long from;
-  long long to;
+    i3String *name;
+    int name_width;
+    long long from;
+    long long to;
 };
 
 typedef struct layer layer;
--- a/i3bar/include/layer.h
+++ b/i3bar/include/layer.h
@@ -1,0 +1,32 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3bar - an xcb-based status- and ws-bar for i3
+ * © 2010 Axel Wagner and contributors (see also: LICENSE)
+ *
+ * mode.c: Handle "layer" events and show current layer in the bar
+ *
+ */
+#pragma once
+
+#include <config.h>
+
+#include <xcb/xproto.h>
+
+#include "common.h"
+
+/* Name of current binding mode and its render width */
+struct layer {
+  i3String *name;
+  int name_width;
+  long long from;
+  long long to;
+};
+
+typedef struct layer layer;
+
+/*
+ * Start parsing the received JSON string
+ *
+ */
+void parse_layer_json(char *json);
--- a/i3bar/include/xcb.h
+++ b/i3bar/include/xcb.h
@@ -48,6 +48,18 @@ struct xcb_color_strings_t {
     char *binding_mode_bg;
     char *binding_mode_fg;
     char *binding_mode_border;
+    char *layer_bg;
+    char *layer_fg;
+    char *layer_border;
+    char *focus_ws_layer_bg;
+    char *focus_ws_layer_fg;
+    char *focus_ws_layer_border;
+    char *active_ws_layer_bg;
+    char *active_ws_layer_fg;
+    char *active_ws_layer_border;
+    char *inactive_ws_layer_bg;
+    char *inactive_ws_layer_fg;
+    char *inactive_ws_layer_border;
 };
 
 typedef struct xcb_colors_t xcb_colors_t;
--- a/i3bar/include/xcb.h
+++ b/i3bar/include/xcb.h
@@ -143,3 +143,9 @@ void redraw_bars(void);
  *
  */
 void set_current_mode(struct mode *mode);
+
+/*
+ * Set the current layer
+ *
+ */
+void set_current_layer(layer *current);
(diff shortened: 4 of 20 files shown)


rayvermey:
This is really a very nice add on for i3! 
Would be nice if this can be incorporated in the main i3!
Ray

MatheusMuriel:
Wow its very nice!

NightH4nter:
This would be a pretty good addition. Ideally if it's gonna be compatible with i3-gaps.

existme:
Great addition, thank you. Key binding for many workspaces opens up i3 for lots of new possibilities

matthinc:
Good idea! I really hope that this will become an official feature!

NightH4nter:
Probably another TODO is to make it possible to move a window from a workspace on one layer to a workpace on another layer.

bastidest:
@NightH4nter That already works. To move a window to another layer, just switch to the layer and press your "move workspace" command.

Example: "I want to move the current container to workspace 23" -> $mod+FN2, $mod+shift+3

NightH4nter:
> @NightH4nter That already works. To move a window to another layer, just switch to the layer and press your "move workspace" command.
> 
> Example: "I want to move the current container to workspace 23" -> $mod+FN2, $mod+shift+3

Okay, I understand. Nice!

jpsala:
I like it!

Airblader:
Thanks for bringing this up and making a PR! We have seen questions surrounding this topic a few times in the past and I can see the value and am open to such a concept. However, I'm not really happy with the specific proposal in this PR as it uses a very static configuration, and I can already see that people would want to modify this at runtime. 

Out of curiosity: what happens if I have layers 1—10, 11—20, 21—30 and while in the first layer I open workspace 15? Simply using an offset will overlap this into another layer, won't it?

A feature like this comes with quite a bit of configuration and/or command surface, which means it is a significant complexity driver. Therefore I think we need to carefully discuss an approach for this. 

For example, we could omit any kind of upfront config and instead just have a `layer foo` command which creates the layer on the fly. All workspace management happens entirely inside the layer, no matter the workspace number etc. Then there is a single `move workspace to layer` command. Immediate questions coming to my mind:

1. Do we even really need to support moving windows between layers individually? This would be a significant driver in the command surface. Using command criteria like marks, con_id etc would allow this without modification on the command syntax.
2. How to handle a new window which relates to layer A while we are on layer B? Do we show it in layer A (= not visible to user), layer B (= breaks layer assignment for the app), automatically switch to layer A and show it there (= can be annoying)? Does it have to be configurable?
3. Do we treat window activations on other layers just like we currently do window activations on non visible workspaces?
4. How to handle specific requests to focus a window on another layer? Automatically switch to it? Or reject the request? I think this one is tricky, the correct answer probably is "it depends".
5. On EWMH/ICCCM side we *must* expose all workspaces regardless of group I think or else we'd break specs. This will cause bars to always show all workspaces unless they implement specific support for it (unfortunately I don't think there is a spec for this to implement against). i3bar would of course be such a client that needs to handle it. 
6. Given that, can something like stickiness even be kept in a spec-compliant way? I assume we don't want sticky across layers (would also be even more complex), but not doing so probably breaks its spec as the window will not be shown on its output. We have to read the spec to see if this is OK, or discuss not allowing sticky on inactive layers.
7. Do we keep workspace to output assignments as-is? This means this config is valid across all layers individually.

There's probably a whole lot more.

---

In summary I hope you can see this isn't too trivial of an issue. However, I do hear the support for it in the community. Therefore I will close this PR and ask for someone to open a feature request issue with a proposal and ideally addressing various scenarios with examples. A topic like this needs to he spec'd out in a lot of detail before we can ship it. At this stage the PR is not yet helpful I think as we first need to find a good direction. Thanks!

orestisfl:
I believe this functionality can be implemented through an IPC script, which would be more flexible and maintainable.

> This will cause bars to always show all workspaces unless they implement specific support for it

For this, we have https://github.com/i3/i3/issues/2333 / https://github.com/i3/i3/issues/3818

Airblader:
> I believe this functionality can be implemented through an IPC script

This was my initial thought also, but I'm actually not sure how convenient that would be. But definitely worth a shot and in fact this has been done before (I never tried it out myself though).

bastidest:
@Airblader 

Thank you for taking the time to read the proposal! I too was surprised by the community feedback.

> Out of curiosity: what happens if I have layers 1—10, 11—20, 21—30 and while in the first layer I open workspace 15? Simply using an offset will overlap this into another layer, won't it?

As there is only one active layer at a time, the concept of "other layers" does not exist. When layer one (1-10) is enabled and you switch to WS15, you switch to WS15. The layer only affects commands in the range.

> A feature like this comes with quite a bit of configuration and/or command surface, which means it is a significant complexity driver. Therefore I think we need to carefully discuss an approach for this.

I completely agree on that, it was meant as a immediate solution for me and as input for the community.

> Then there is a single `move workspace to layer` command.

This is the exact opposite of my idea, as the layers are completely transparent. Adding another non-concept would require too much modifications. For more elaborate concepts there is always IPC.

> Do we even really need to support moving windows between layers individually?

See above.

> How to handle a new window which relates to layer A while we are on layer B?

Again. No handling needed, as the layers are a transparent concept.

> Do we treat window activations on other layers just like we currently do window activations on non visible workspaces?

Exactly. The only thing the layers do is to give a visual indication of the current layer range and make switching inside a layer easier.

> How to handle specific requests to focus a window on another layer?

Valid point. From my perspective, I'd never touch the current active layer when doing anything.

I think you get my point. My idea was to keep the layers as non-invasive as possible in order to not change any existing behavior.

@orestisfl 
> I believe this functionality can be implemented through an IPC script, which would be more flexible and maintainable.

In fact this was my first shot at the problem, but I quickly hit obstacles. Im sure it can be done, but having to rely on a external daemon for correct window switching behavior is not acceptable to me. Also I don't know how far the i3bar can be customized to show the current active layer with their respective workspaces.

Thank you everyone for taking the time. I will definetly keep the fork updated, as I now depend on it 😅

Ricocotam:
Do you plan opening an issue so we can all progress on this awesome feature ?

bastidest:
> Do you plan opening an issue so we can all progress on this awesome feature ?

Will do in this week.

bastidest:
I created an issue for further discussing and clarifying this idea. Leave your thoughts on that topic there. #4024

rayvermey:
Thanks, I will.

Ray

Verzonden vanuit Outlook Mobile<https://aka.ms/blhgte>

________________________________

PR #3988, opened 2020-04-04
implementation for bisect command

jerojasro:
bisect horizontal | vertical | h | v

takes a tabbed/stacked container, splits its currently-used space into
two new tabbed/stacked containers, puts currently-focused window in one
of them, the other windows in the other one.

### DIFF
--- a/include/commands.h
+++ b/include/commands.h
@@ -159,6 +159,12 @@ void cmd_floating(I3_CMD, const char *floating_mode);
 void cmd_move_workspace_to_output(I3_CMD, const char *name);
 
 /**
+ * Implementation of 'bisect v|h|vertical|horizontal'.
+ *
+ */
+void cmd_bisect(I3_CMD, const char *direction);
+
+/**
  * Implementation of 'split v|h|t|vertical|horizontal|toggle'.
  *
  */
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -30,6 +30,7 @@ state INITIAL:
   'open' -> call cmd_open()
   'fullscreen' -> FULLSCREEN
   'sticky' -> STICKY
+  'bisect' -> BISECT
   'split' -> SPLIT
   'floating' -> FLOATING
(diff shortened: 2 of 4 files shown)


jerojasro:
I see the build failed, but I couldn't find any actionable information (for me) in the logs it generated about the failure.

Can someone check those logs and please let me know if I broke a test, and if so which one?

Thanks,

orestisfl:
Don't worry about this build failure. @stapelberg maybe you know what's wrong with the check-spelling script here?

Airblader:
Isn't this trivially implemented as a script or even just a sequence of commands? I'm not convinced this is worth a separate command.

jerojasro:
> Isn't this trivially implemented as a script or even just a sequence of commands? I'm not convinced this is worth a separate command.

@Airblader I hadn't considered the approach you suggest, because it didn't occur to me to move all of the i3-command-chaining to an external script and use `i3-msg`; I'll take a stab at it and validate your confidence in this being "trivially implemented". From the top of my mind, I think it might be needlessly hard to perform the sanity checks I already do in the code, but nonetheless I'll check.

Is this the (alternative) approach you have in mind for achieving the functionality I want? (external script + `i3-msg`)

Thanks for your comment.

stapelberg:
> Don't worry about this build failure. @stapelberg maybe you know what's wrong with the check-spelling script here?

Sent a fix in https://github.com/i3/i3/pull/3994

Airblader:
> Is this the (alternative) approach you have in mind for achieving the functionality I want?

Possibly, yes; that would be simplest. If some things are too difficult to do this way, you can use a proper IPC implementation and script a solution (i3-msg is mostly a reference implementation for simple usage of sending commands). Most IPC implementations offer easy ways to query various state in the i3 tree.

orestisfl:
@jerojasro Let us know if you have any news with your IPC script or if you believe that what you are trying to achieve should be integrated into i3 in some way.

I'll close this PR for the time being

PR #4006, opened 2020-04-10
Extend tiling/floating criteria with optional auto/user values

### LINKED ISSUE #3588: Make `for_window[floating|tiling]` when floating status is updated
## I'm submitting a…
<pre>
[ ] Bug
[ ] Feature Request
[ ] Documentation Request
[X] Other (Please describe in detail): 
Configuration for use case. I'm not sure if the following syntax is correct and I believe this is not a bug but I think this use case is not in the documentation either.
</pre>

## Current Behavior
`for_window  [title="myterminal"] [floating]      resize set 800 600`

## Expected Behavior
Upon executing `floating toggle`, if there's a window with `title="myterminal"` in floating mode, resize it to 800 600. Not for all windows, just the one(s) with that title.

## Reproduction Instructions
Opening window with title="myterminal" (in tiled mode) and pressing $mod+Shift+Space (`floating toggle`): window does enter in floating mode but is not resized

## Environment
```
$ i3 -v
i3 version 4.16 (2018-11-04) © 2009 Michael Stapelberg and contributors

$ i3 --moreversion 2>&- 
Binary i3 version:  4.16 (2018-11-04) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.16 (2018-11-04) (pid 883)
Loaded i3 config: /home/nexor/.config/i3/config (Last modified: Tue 15 Jan 2019 08:36:44 PM -02, 549 seconds ago)

```

<pre>
- Linux Distribution & Version: Linux 4.20.0-arch1-1-ARCH x86_64
</pre>

orestisfl:
Fixes #3588

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1900,8 +1900,18 @@ con_id::
 	to match only the currently focused window.
 floating::
 	Only matches floating windows. This criterion requires no value.
+floating_from::
+	Like +floating+ but this criterion takes two possible values: "auto"
+	and "user". With "auto", only windows that were automatically opened as
+	floating are matched. With "user", only windows that the user made
+	floating are matched.
 tiling::
 	Only matches tiling windows. This criterion requires no value.
+tiling_from::
+	Like +tiling+ but this criterion takes two possible values: "auto" and
+	"user". With "auto", only windows that were automatically opened as
+	tiling are matched. With "user", only windows that the user made tiling
+	are matched.
 
 The criteria +class+, +instance+, +role+, +title+, +workspace+ and +mark+ are
 actually regular expressions (PCRE). See +pcresyntax(3)+ or +perldoc perlre+ for
--- a/parser-specs/config.spec
+++ b/parser-specs/config.spec
@@ -181,16 +181,18 @@ state NO_FOCUS_END:
 
 # Criteria: Used by for_window and assign.
 state CRITERIA:
-  ctype = 'class'       -> CRITERION
-  ctype = 'instance'    -> CRITERION
-  ctype = 'window_role' -> CRITERION
-  ctype = 'con_id'      -> CRITERION
-  ctype = 'id'          -> CRITERION
-  ctype = 'window_type' -> CRITERION
-  ctype = 'con_mark'    -> CRITERION
-  ctype = 'title'       -> CRITERION
-  ctype = 'urgent'      -> CRITERION
-  ctype = 'workspace'   -> CRITERION
+  ctype = 'class'         -> CRITERION
+  ctype = 'instance'      -> CRITERION
+  ctype = 'window_role'   -> CRITERION
+  ctype = 'con_id'        -> CRITERION
+  ctype = 'id'            -> CRITERION
+  ctype = 'window_type'   -> CRITERION
+  ctype = 'con_mark'      -> CRITERION
+  ctype = 'title'         -> CRITERION
+  ctype = 'urgent'        -> CRITERION
+  ctype = 'workspace'     -> CRITERION
+  ctype = 'floating_from' -> CRITERION_FROM
+  ctype = 'tiling_from'   -> CRITERION_FROM
   ctype = 'tiling', 'floating'
       -> call cfg_criteria_add($ctype, NULL); CRITERIA
   ']'
@@ -198,6 +200,22 @@ state CRITERIA:
 
 state CRITERION:
   '=' -> CRITERION_STR
+
+state CRITERION_FROM:
+  '=' -> CRITERION_FROM_STR_START
+
+state CRITERION_FROM_STR_START:
+  '"' -> CRITERION_FROM_STR
+  kind = 'auto', 'user'
+    -> call cfg_criteria_add($ctype, $kind); CRITERIA
+
+state CRITERION_FROM_STR:
+  kind = 'auto', 'user'
+    -> CRITERION_FROM_STR_END
+
+state CRITERION_FROM_STR_END:
+  '"'
+    -> call cfg_criteria_add($ctype, $kind); CRITERIA
 
 state CRITERION_STR:
   cvalue = word
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -842,6 +842,7 @@ static void handle_client_message(xcb_client_message_event_t *event) {
             DLOG("The window was requested to be visible on all workspaces, making it sticky and floating.\n");
 
             floating_enable(con, false);
+            con->floating = FLOATING_AUTO_ON;
 
             con->sticky = true;
             ewmh_update_sticky(con->window->id, true);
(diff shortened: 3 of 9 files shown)


src/handlers.c
```diff
@@ -1156,6 +1157,19 @@ static bool handle_strut_partial_change(Con *con, xcb_get_property_reply_t *prop
     return true;
 }
 
+/*
+ * Handles the _I3_FLOATING_WINDOW property to properly run assignments for
+ * floating window changes.
```
xzfc, line 1162: Why listen to our own properties instead of just calling `remanage_window()` from `floating_enable()`/`floating_disable()`?

orestisfl:
@stapelberg the only other way I see to support `floating=auto` is to extend the parser, but I don't think it's really worth it and we can always do it later, maintaining backwards compatibility, if we find more use cases.

src/handlers.c
orestisfl: See discussion in #3588, suggestions welcome :)

orestisfl: Added a comment in the docstring.

Obviously, this is a bit hacky solution, maybe we can make the parser re-entrant.

orestisfl: After having a quick look, I am not sure if I want to invest the time to modify the parser. So far, I've found that besides the static variables in `command_parser.c`, `owindows` in `commands.c` also needs changes.

stapelberg:
> the only other way I see to support `floating=auto` is to extend the parser, but I don't think it's really worth it and we can always do it later, maintaining backwards compatibility, if we find more use cases.

I think I found a way: https://github.com/i3/i3/commit/a0ba91bcd618310f6881b89d17492908a126cb2c

Feel free to cherry-pick this into your PR.

orestisfl:
@stapelberg That's nice but unfortunately fails in 2 cases.

1. Default `floating` followed by other criteira

```diff
diff --git a/testcases/t/201-config-parser.t b/testcases/t/201-config-parser.t
index a8d45325..7cacff4f 100644
--- a/testcases/t/201-config-parser.t
+++ b/testcases/t/201-config-parser.t
@@ -99,11 +99,15 @@ is(parser_calls($config),
 
 $config = <<'EOT';
 for_window [class="^Chrome"] floating enable
+for_window [floating class="==Class=="] nop floating
 EOT
 
 $expected = <<'EOT';
 cfg_criteria_add(class, ^Chrome)
 cfg_for_window(floating enable)
+cfg_criteria_add(floating, NULL)
+cfg_criteria_add(class, ==Class==)
+cfg_for_window(nop floating)
 EOT
 
 is(parser_calls($config),
```

2. `[floating="auto"] …` (with double quotes)

stapelberg:
> 1. Default `floating` followed by other criteira

I think this issue can only be addressed through repetition.

I think at this point, we should probably use `floating_from` as a new criterion which takes a string value, and then parse that string as user or auto.

orestisfl:
Why not go with the current solution?

The problem with parsing is that it won't be easy to show a parsing error.

stapelberg:
> Why not go with the current solution? 

By current solution, you’re referring to floating_user and floating_auto, right? I’m not too keen on adding so many new criteria just for this feature. If we ever expand the list, to, say, include floating-from-ipc, we’d need another 2 criteria.

> The problem with parsing is that it won't be easy to show a parsing error.

But that’s the same as with the other criteria right now, correct? E.g. if you supply an invalid regexp.

orestisfl:
Supplying an invalid regex is different than mistyping one of the few static possibilities.

Besides, if the possible values are within a discrete set of possibilities, handling them after the parser is not really different to handling them in the parser.

stapelberg:
I still think we should use a floating_from instead of floating_user, but if you want to make floating_from only accept the discrete set of values, go ahead. I think it can be done with a few additional states, using the same syntax as we use for strings.

orestisfl:
In your approach, we can re-use `floating` but just make it accept a `word` instead of `'auto', 'user'`. So, no need for `floating_from`. This deals with both quotes and the first issue I raised, but not the i3-nagbar error.

Do you want to do it this way?

stapelberg:
> In your approach, we can re-use `floating` but just make it accept a `word`

But that’s not backwards-compatible, is it?

orestisfl:
I am sorry, I forgot why that doesn't play well with other criteria.

Anyway, what's the conclusion here?

stapelberg:
I think this is how it should be done: https://github.com/i3/i3/commit/abaa1f8f914cbdc897c0ac13ef6832b52dda1627

Let me know if there’s anything that’s still broken with that approach, or anything to improve (I don’t see any easy improvements now).

orestisfl:
Yes, this doesn't work with `[tiling_from=user]` (without quotes) like other criteria

stapelberg:
It’s kind of unfortunate (for consistency’s sake) that we ever accepted the variant without quotes, but it’s too late to rectify that now :)

This is now addressed in https://github.com/i3/i3/commit/08b25407f0ef32f7fdfdd25db1443dc22573d0cc

orestisfl:
up

docs/userguide
```diff
@@ -1899,9 +1899,17 @@ con_id::
 	interface. Handy for scripting. Use the special value +\_\_focused__+
 	to match only the currently focused window.
 floating::
```
stapelberg, line 1901: The documentation update needs to be moved to floating_from and tiling_from, respectively.

src/handlers.c
```diff
@@ -1156,6 +1157,25 @@ static bool handle_strut_partial_change(Con *con, xcb_get_property_reply_t *prop
...
+ * windows which are triggered by user commands (floating enable | disable). In
+ * that case, we can't call run_assignments because it will modify the parser
+ * state when it needs to parse the user-specified action, breaking the parser
+ * state for the original command.
+ *
+ */
+static bool handle_i3_floating(Con *con, xcb_get_property_reply_t *prop) {
+    DLOG("floating change for con %p\n", con);
+
+    remanage_window(con);
```
stapelberg, line 1174: Can this result in an infinite loop, if a user toggles floating status from an assignment?

parser-specs/config.spec
```diff
@@ -181,24 +181,42 @@ state NO_FOCUS_END:
...
+  ctype = 'window_type'   -> CRITERION
+  ctype = 'con_mark'      -> CRITERION
+  ctype = 'title'         -> CRITERION
+  ctype = 'urgent'        -> CRITERION
+  ctype = 'workspace'     -> CRITERION
+  ctype = 'floating_from' -> CRITERION_FROM
+  ctype = 'tiling_from'   -> CRITERION_FROM
   ctype = 'tiling', 'floating'
-      -> call cfg_criteria_add($ctype, NULL); CRITERIA
+       -> call cfg_criteria_add($ctype, NULL); CRITERIA
```
stapelberg, line 197: Messed up the whitespace in this line, sorry. Can you correct it while you’re at it please?

src/handlers.c
orestisfl: `for_window` rules run once per window

```
for_window [tiling_from="user"] mark --add tiling_user, floating enable
for_window [floating_from="user"] mark --add floating_user, floating disable
```

Here, each rule will call the other one but it will only happen once

PR #4007, opened 2020-04-10
Match WM_CLIENT_MACHINE

### LINKED ISSUE #3981: Allow `for_window` to match against `WM_CLIENT_MACHINE`
## I'm submitting a...
<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
`for_window` cannot match against `WM_CLIENT_MACHINE`

## Desired Behavior
Matching with `for_window` against `WM_CLIENT_MACHINE`.

## Use Case
When using lxc containers (or possibly other remote means) to start gui programs this match would allow to apply common rules to all windows launched from this remote instance.
This would be useful when running multiple instances of ie. firefox from different containers.

xzfc:
* Add `machine` criterion to match `WM_CLIENT_MACHINE`. Closes #3981
* Add `%machine` title format placeholder.
* Add `machine` to the IPC and layout saving/restoring.

Also, would it make sense to support special values `__local__` and `__remote__` to match corresponding clients? Use-case is to indicate windows from remote clients, like this:
`for_window [machine=__remote__] title_format "[%machine] %title"`
Alternative is to hard-code the machine hostname into the user config.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1896,6 +1896,10 @@ window_type::
 	Compare the window type (_NET_WM_WINDOW_TYPE). Possible values are
 	+normal+, +dialog+, +utility+, +toolbar+, +splash+, +menu+, +dropdown_menu+,
 	+popup_menu+, +tooltip+ and +notification+.
+machine::
+	Compares the name of the machine the client window is running on
+	(WM_CLIENT_MACHINE). Usually, it is equal to the hostname of the local
+	machine, but it may differ if remote X11 apps are used.
 id::
 	Compares the X11 window ID, which you can get via +xwininfo+ for example.
 title::
@@ -1933,9 +1937,9 @@ tiling_from::
 	tiling are matched. With "user", only windows that the user made tiling
 	are matched.
 
-The criteria +class+, +instance+, +role+, +title+, +workspace+ and +mark+ are
-actually regular expressions (PCRE). See +pcresyntax(3)+ or +perldoc perlre+ for
-information on how to use them.
+The criteria +class+, +instance+, +role+, +title+, +workspace+, +machine+ and
++mark+ are actually regular expressions (PCRE). See +pcresyntax(3)+ or +perldoc
+perlre+ for information on how to use them.
 
 [[exec]]
 === Executing applications (exec)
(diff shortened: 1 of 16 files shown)


src/window.c
```diff
@@ -464,3 +464,21 @@ void window_update_motif_hints(i3Window *win, xcb_get_property_reply_t *prop, bo
...
+ */
+void window_update_machine(i3Window *win, xcb_get_property_reply_t *prop) {
+    if (prop == NULL || xcb_get_property_value_length(prop) == 0) {
+        DLOG("WM_CLIENT_MACHINE not set.\n");
+        FREE(prop);
+        return;
+    }
+
+    FREE(win->machine);
+    win->machine = strndup((char *)xcb_get_property_value(prop), xcb_get_property_value_length(prop));
```
orestisfl:
```suggestion
    win->machine = sstrndup((char *)xcb_get_property_value(prop), xcb_get_property_value_length(prop));
```

src/handlers.c
```diff
@@ -1084,6 +1084,16 @@ static bool handle_class_change(Con *con, xcb_get_property_reply_t *prop) {
...
 
+/*
+ * Handles the WM_CLIENT_MACHINE property for assignments and criteria selection.
+ *
+ */
+static bool handle_machine_change(Con *con, xcb_get_property_reply_t *prop) {
+    window_update_machine(con->window, prop);
+    con = remanage_window(con);
+    return true;
+}
```
xzfc, line 1097: I'm not sure that i3 should listen to changes in the property `WM_CLIENT_MACHINE` since it is unlikely to change.
However, I make it similar to e.g. `WM_CLASS` which is should not be changed too (?).

orestisfl (dismissed):
Can you resolve the conflicts and update the release notes as well?

docs/userguide
```diff
@@ -2584,6 +2588,9 @@ and the following placeholders which will be replaced:
 +%instance+::
     The X11 window instance (first part of WM_CLASS). This corresponds to the
     +instance+ criterion, see <<command_criteria>>.
++%machine+::
```
Airblader, line 2591: This seems like an orthogonal feature, so at least it should be documented in the release notes separately.

orestisfl:
up

Airblader (approved):

orestisfl (approved):

ReneFroger:
Does this issue say you something @orestisfl https://github.com/i3/i3/issues/5616 ?

PR #4018, opened 2020-04-12
Sort dock clients by class and instance

### LINKED ISSUE #3491: order of bar blocks in the config file is not respected
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

With multiple bar blocks defined in the config file, the order they are drawn by i3 isn't deterministic.

With the following config, restarting i3 multiple times will display bars in non-deterministic orders:

````
bar {
  position top
  workspace_buttons yes
}

bar {
  position top
  workspace_buttons no
}
````

## Expected Behavior

With the previous example, we expect the bar with workspace buttons to be always first, at the very top.

## Reproduction Instructions

Test the previous example and trigger `i3-msg restart` multiple times while watching the workspace buttons at top of the screen.

## Environment

<pre>
(Getting version from running i3, press ctrl-c to abort…Running i3 version: 4.15 (2018-03-10) (pid 1026)
Loaded i3 config: /home/vivien/.config/i3/config (Last modified: Wed 31 Oct 2018 11:21:34 AM EDT, 315 seconds ago)

</pre>

<pre>
</pre>

<pre>
</pre>

<pre>
- Linux Distribution & Version: 4.18.16-arch1-1-ARCH
</pre>

### PR DESCRIPTION
**Pull request by orestisfl**
See commit message

Fixes #3491

### DIFF
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -1816,6 +1816,8 @@ void reconfig_windows(bool redraw_bars) {
                                                                       bar_height);
 
             /* Set the WM_CLASS and WM_NAME (we don't need UTF-8) atoms */
+            char *class;
+            int len = sasprintf(&class, "%s%ci3bar%c", config.bar_id, 0, 0);
             xcb_void_cookie_t class_cookie;
             class_cookie = xcb_change_property(xcb_connection,
                                                XCB_PROP_MODE_REPLACE,
@@ -1823,8 +1825,8 @@ void reconfig_windows(bool redraw_bars) {
                                                XCB_ATOM_WM_CLASS,
                                                XCB_ATOM_STRING,
                                                8,
-                                               (strlen("i3bar") + 1) * 2,
-                                               "i3bar\0i3bar\0");
+                                               len,
+                                               class);
 
             char *name;
             sasprintf(&name, "i3bar for output %s", walk->name);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: i3bar/src/xcb.c [hunk 1]
```diff
@@ -1816,15 +1803,17 @@ void reconfig_windows(bool redraw_bars) {
                                                                       bar_height);
 
             /* Set the WM_CLASS and WM_NAME (we don't need UTF-8) atoms */
+            char *class;
+            int len = sasprintf(&class, "%s%ci3bar%c", config.bar_id, 0, 0);
```
Airblader, line 1807: This is a change we should make in a separate commit so that it also lands separately in the list of changes.

src/con.c
```diff
@@ -132,6 +132,30 @@ static void _con_attach(Con *con, Con *parent, Con *previous, bool ignore_focus)
         goto add_to_focus_head;
     }
 
+    if (parent->type == CT_DOCKAREA) {
+        /* Insert dock client to the correct position, sorting by class and
```
Airblader, line 136: This doesn't seem quite correct, as the sorting happens lexicographically even for i3bar. This leading to an order which reflects the order of the bar blocks is an artifact of that which only holds true if they don't specify a custom `id`.

orestisfl: Yes, this is what I was implying. I've updated the comment to make it more explicit why this is happening.

orestisfl:
up

PR #4025, opened 2020-04-18
x_draw_decoration: always clear parent pixmap when rendering the first child

xzfc:
See the commit message.

## Steps to reproduce:

1. Enable transparent borders and titlebars, and use a compositor.

2. Make this layout: `H[x V[a b]]`
  * `a` is 32-bit window, has `border pixel 50`
  * `b` has `border normal 50`

3. Resize `a` (top right window) and see visual garbage in the background.

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -590,11 +590,6 @@ void x_draw_decoration(Con *con) {
         }
     }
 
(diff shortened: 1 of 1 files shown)


orestisfl (approved):
SGTM, I see no problems. Do you want me to reproduce?

PR #4032, opened 2020-04-22
release.sh fixes from last release

stapelberg:


Airblader (approved):

release.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/zsh
 # This script is used to prepare a new release of i3.
 
-export RELEASE_VERSION="4.17"
-export PREVIOUS_VERSION="4.16"
+export RELEASE_VERSION="4.19"
```
orestisfl, line 4: Should have been 4.18.1, right?

stapelberg: We don’t have a clear process for updating the release.sh versions after a bugfix release. I decided to just update them for the upcoming next major release.

It’d probably be best to pass these in via a command-line flag anyway :)

PR #4033, opened 2020-04-22
i3-dmenu-desktop: Manually search for topdir in .desktop deduplication

### LINKED ISSUE #4031: i3-dmenu-desktop: Nondeterministic .desktop entries ordering (unable to reliably override system/default version)
## I'm submitting a…
<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
I have two `.desktop` entries for `thunderbird`. One is located in `/usr/share/applications/thunderbird.desktop` (system one), the other is in `/home/user/.local/share/applications/thunderbird.desktop` (user generated one). I'm using `i3-dmenu-desktop` to display and run desktop entries.

When displaying an entry for `thunderbird` or running it, one of those two `.desktop` entries is being selected in a nondeterministic fashion. Sometimes the user generated one is being used, sometimes the system one.

## Expected Behavior
User configuration should take precedence over the global system one, so the `/home/user/.local/share/applications/thunderbird.desktop` should be used every time.

While it is remotely possible that I've been very lucky for the last few years, I'd say that this change was introduced recently in 3a672bc930237e1b10c81a4c3a8d51eb3c4ae3f5. After rolling it back, everything seems to be working fine again (of course I might be very lucky... again ;)).

Additionally both the commit message and Perl documentation (`follow_fast`) mention duplicates and both `.desktop` entries have the same file name.

## Reproduction Instructions
Copy any `.desktop` entry from `/usr/share/applications/` to `/home/user/.local/share/`. Change the value of `Name` in the `.local/share` version and invoke `i3-dmenu-desktop` several times. You will notice that the name will change randomly.

## Environment
<pre>
Binary i3 version:  4.18-85-g0bce0d86 (2020-04-15, branch "next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.18-85-g0bce0d86 (2020-04-15, branch "next") (pid 42362)
Loaded i3 config: /home/user/.config/i3/config (Last modified: Tue 21 Apr 2020 13:27:27 CEST, 14286 seconds ago)

</pre>

<pre>
- Linux Distribution & Version: Debian (testing)
- Are you using a compositor (e.g., xcompmgr or compton): Yes, compton
</pre>

_edit: Readability improved._

orestisfl:
Fixes #4031

### DIFF
--- a/i3-dmenu-desktop
+++ b/i3-dmenu-desktop
@@ -13,6 +13,7 @@ use POSIX qw(locale_h);
 use File::Find;
 use File::Basename qw(basename);
 use File::Temp qw(tempfile);
+use List::Util 'first';
 use Getopt::Long;
 use Pod::Usage;
 use v5.10;
@@ -123,6 +124,8 @@ for my $dir (split(':', $xdg_data_dirs)) {
 # Cleanup the paths, maybe some application does not cope with double slashes
 # (the field code %k is replaced with the .desktop file location).
 @searchdirs = map { s,//,/,g; $_ } @searchdirs;
+# Also remove any trailing slashes.
+@searchdirs = map { s,/*$,,g; $_ } @searchdirs;
 
 # To avoid errors by File::Find’s find(), only pass existing directories.
 @searchdirs = grep { -d $_ } @searchdirs;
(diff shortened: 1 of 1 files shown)


orestisfl:
@stapelberg There don't seem to be any historical reasons to not use `basename` here

i3-dmenu-desktop
```diff
@@ -131,10 +131,7 @@ find(
     {
         wanted => sub {
             return unless substr($_, -1 * length('.desktop')) eq '.desktop';
-            my $relative = $File::Find::name;
-
-            # + 1 for the trailing /, which is missing in ::topdir.
-            substr($relative, 0, length($File::Find::topdir) + 1) = '';
+            my $relative = basename($File::Find::name);
```
stapelberg: I don’t think this is a safe change to make.

The previous code made the path relative to the search dir (e.g. search dir ~/.local/share/applications, relative=wine/Programs/DigitalSimulatorV5.57/DigitalSimulator.desktop), which is different than the basename.

Granted, most .desktop files will reside directly in the search dir, in which case there won’t be any difference. But, it seems like subdirectories are both supported and used in practice, so we should keep supporting them.

In which way is the current code buggy? Maybe there’s a different way to fix this.

orestisfl: Why do we need the relative path?

orestisfl: I see

See the commit for the bug:
> Since 3a672bc, using follow or follow_fast, does not set $File::Find::topdir, breaking our deduplication.

orestisfl:
@stapelberg I have adjusted the code according to your comments. Notice however that i3-dmenu-desktop does not respect subdirectories mainly because of the uses of `basename` in this line:

https://github.com/i3/i3/blob/440268f4545990698ddb968b1ca506eb7080ad03/i3-dmenu-desktop#L154

It seems that if I remove this, it works as you describe.

I can push that change as well.

stapelberg (approved):

Issue #3612, opened 2019-02-03
Directory or repository of third-party scripts and such


eater:
## I'm submitting a…

<pre>
[ ] Bug
[ ] Feature Request
[x] Documentation Request
[ ] Other (Please describe in detail)
</pre>

The design philosophy of i3 intentionally sets out that all non-foundational functionality should be implemented via IPC or other tools external to the codebase. That's great! 

In order to encourage and streamline that, and make the end-user experience better, it would be incredibly helpful if there were a single place where users could find scripts that have already been created, and share new ones that they have implemented. The current system involves searching Reddit and maybe asking on IRC, and neither of those methods provides very good coverage of what actually exists.

Many other projects have simple directories of add-ons. Random example: https://github.com/todotxt/todo.txt-cli/wiki/Todo.sh-Add-on-Directory.  It could be hosted as a Github wiki page, and I volunteer to help maintain that.

Airblader:
We do have the user-contributed articles section on the website, but maintaining that is pretty painful. I'm fairly sure we agreed in the past that keeping community-maintained scripts somewhere is a great idea that we support, though I'm not sure right now where we had that discussion.

Ultimately, of course, nothing stops anyone from creating such a repository. I don't think we'd want to host a repository in the i3 organization ourselves because it'd be a lot of effort to maintain it. I don't see an issue with linking to such a repository from the website and/or reddit, though. @stapelberg Your thoughts?

stapelberg:
As long as it’s clearly marked that the directory/list is community-maintained, I’d be happy to link to it.

I do agree that it would be preferable for such a directory/list to live outside the official i3 org for the time being. If it becomes a no-brainer at some point to bless it as official, I’m happy to do that, but I want to be cautious with regards to what we sign up to maintain :).

Does that make sense?

Airblader:
Great. Let's keep this open and if someone wants to start such a repository they can let us know so we can link to it in appropriate places.

eater:
All right, I have kicked this off here: https://github.com/noodlor/i3-resources

I look forward to feedback!

Airblader:
@eater Thanks! Feel free to send a PR to document this in places that seem useful. You can also post about it on our subreddit, we can make it sticky there.

orestisfl:
We can mention https://github.com/altdesktop/i3ipc-python/tree/master/examples somewhere

ghost:
[...]

PR #4038, opened 2020-04-24
contrib: Add auto-mark script

zsugabubus:
For lightning fast navigation between windows, one can use marks. It can
be orders of magnitude faster than stepping through intermediate windows
and workspaces to reach target window.

However, marking (ad-hoc) windows manually is tedious, so this script is
here to automatize this task.

---

I just would like to share this scriptlet, maybe someone will find it useful. If you do not like it just close the PR.

### DIFF
--- a/contrib/auto-mark
+++ b/contrib/auto-mark
@@ -1,0 +1,29 @@
+#!/bin/sh -e
+scope=$1
+shift 2>&- ||:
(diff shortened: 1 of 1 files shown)


orestisfl:
The script is fine but I don't think that we should extend the `contrib/` folder this way, since most scripts there are not scripts that should be used directly by the users.

What we could do, is create a new folder `examples/` with various IPC scripts (https://github.com/i3/i3/issues/3612) with proper documentation and maintenance.
If we ever decide to do that, this PR will be linked to the issue so we can revive the automark script.

PR #4067, opened 2020-05-04
etc/config: don't focus notifications in default config

Hi-Angel:
People usually don't want notifications to take focus, because that introduces lots of problems *(cursor moves as someone recently reported, transparency goes to the wrong window, keyboard layout suddenly changes…)*.

While the correct way to fix that would probably be in a core, but I figured it gonna bring many discussions about changing default behavior, and then someone has to do the job, etc… I figured adding the `no_focus` switch for notifications in the default configuration should be a good compromise.

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -192,6 +192,8 @@ bar {
         status_command i3status
 }
 
+no_focus [window_type="notification"]
+
 #######################################################################
 # automatically start i3-config-wizard to offer the user to create a
 # keysym-based config which used their favorite modifier (alt or windows)


Hi-Angel:
Probably worth noting: this directive tested to be working with KDE notifications and ones of `notification-daemon` *(i.e. no DE)*

orestisfl:
I don't agree with the approach here. i3 already has criteria to decide if focus needs to be set or not in `src/manage.c` (`set_focus`). If we fix an issue by changing the config, not every user is going to see the change.

Isn't the issue https://github.com/i3/i3/issues/3979 you are referring fixed already?

Hi-Angel:
> Isn't the issue #3979 you are referring fixed already?

It is fixed. I am not sure why are you asking though, I only devoted to that a short sentence…

> I don't agree with the approach here. i3 already has criteria to decide if focus needs to be set or not in `src/manage.c` (`set_focus`). If we fix an issue by changing the config, not every user is going to see the change.

Okay, so you're saying it should be addressed in the core. I'll open a new issue on this in that case.

orestisfl:
I am not sure what the issue is in the first place but yes let's track the discussion in a new issue

Hi-Angel:
> I am not sure what the issue is in the first place

Hmm… The issue is that notifications gets focused by default. The #3979 you're referring to was about `no_focus` directive not working in some cases. These are a bit orthogonal: if a user didn't use `no_focus` against notifications they wouldn't stumble upon #3979.

What I am suggesting here is that notification shouldn't take focus by default.

> but yes let's track the discussion in a new issue

Okay. Writing steps-to-reproduce will take some time though. I just figured there seems to be some misunderstanding, so I'm replying here.

orestisfl:
We have to actually confirm if `no_focus` is needed after https://github.com/i3/i3/pull/4012 because it seems to have been added to treat the bug. Will the notification be focused without `no_focus`? If yes, then the window is asking for focus and i3 shouldn't discriminate against it (or alternatively there is another bug), e.g. by a non-standard rule. If no, then there is no bug to be fixed.

Issue #3670, opened 2019-03-29
Makefile try to compile i3-config-wizard before i3-config-parser.stamp on many threads building


MarvelousBlack:
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

A failure occurred in build
```
 GEN      man/i3-config-wizard.xml
  GEN      man/i3-migrate-config-to-v4.xml
  GEN      man/i3-sensible-editor.xml
  GEN      man/i3-sensible-pager.xml
  GEN      man/i3-sensible-terminal.xml
  GEN      man/i3-dump-log.xml
  GEN      man/i3-dmenu-desktop.1
  GEN      man/i3-save-tree.1
  GEN      docs/debugging.html
  GEN      docs/lib-i3test.html
  GEN      docs/lib-i3test-test.html
  GEN      docs/hacking-howto.html
  GEN      docs/userguide.html
../../i3/i3-config-wizard/main.c:109:10: fatal error: GENERATED_config_enums.h: No such file or directory
 #include "GENERATED_config_enums.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:2774: i3-config-wizard/i3_config_wizard-main.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/build/i3-gaps-next-git/src/i3/build'
make: *** [Makefile:1332: all] Error 2
 
```
When building on many threads concurrently, the generated Makefile may try to compile i3-config-wizard before i3-config-parser.stamp and result an error where GCC cannot find the to-be-generated header file. This error happens at a low possibility
## Expected Behavior

Successfully make without error.
## Reproduction Instructions

```
autoreconf --force --install
../configure  --prefix=/usr  --sysconfdir=/etc  --disable-sanitizers 
make CPPFLAGS+="-U_FORTIFY_SOURCE" -j$(nproc)
```
In my environment $(nproc) is 24.
## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: Binary i3 version:  4.16.1-152-g767a6a52 (2019-03-28, branch "gaps-next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.16.1-152-g767a6a52 (2019-03-28, branch "gaps-next") (pid 544)
[...]

PR #4069, opened 2020-05-05
Makefile.am: make sure i3-config-wizard depends on libi3.a

### LINKED ISSUE #4020: i3 has issues building when parallel builds are enabled.
(From https://aur.archlinux.org/packages/i3-git/#comment-739440)

When parallel builds are enabled, i3 build can fail with various errors. One such error:
[error.log](https://github.com/i3/i3/files/4480799/error.log)

Workaround is to add `-j1` to `make`.

I'm experiencing this with Manjaro linux running on Ryzen 3700X.

Hi-Angel:
i3-config-wizard uses libi3.a as part of its build process. When
parallel build is enabled, build of i3-config-wizard may start before
libi3.a finished building. Fix this by adding dependency on libi3.a

Fixes: #4020

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -442,7 +442,8 @@ i3_config_wizard_i3_config_wizard_SOURCES = \
 	i3-config-wizard/xcb.h
 
 i3_config_wizard_i3_config_wizard_DEPENDENCIES = \
-	$(config_parser_SOURCES)
+	$(config_parser_SOURCES) \
+	$(top_builddir)/libi3.a
 
 test_inject_randr15_CPPFLAGS = \
 	$(AM_CPPFLAGS)


orestisfl (approved):
Travis will fail because of another issue so don't be concerned

orestisfl:
Hmm, I am now having second thoughts that this is valid approach since I can see the `libi3.a` -> `i3-config-wizard/i3-config-wizard` dependency before your patch (see https://github.com/i3/i3/issues/3670#issuecomment-555875797 for a cool way to visualize).

I noticed that I can't trigger the bug on my machine even without your patch. Maybe you can try a few more times and see if it actually fixes it? Thank you.

Hi-Angel:
Sure. Yeah, it is hard with floating bugs.

Hi-Angel:
Okay, so, I ran build a hundred of times with the following script

```bash
#!/bin/sh
set -e
for _ in {0..100}; do
    make clean && make -j8
done
```

and no build failures happened to me.

orestisfl:
Ok, for me the bug is specific to my `makepkg` configuration for some reason. I can't trigger it manually but if I run `makepkg` on the `i3-git` package it triggers.

Anyway, I can now confirm that your patch fixes the problem for me as well.

Hi-Angel:
> I can see the `libi3.a` -> `i3-config-wizard/i3-config-wizard` dependency before your patch (see [#3670 (comment)](https://github.com/i3/i3/issues/3670#issuecomment-555875797) for a cool way to visualize).

The comment was posted in 2019, perhaps something could have changed since then. I just tried making a similar graph, but I'm getting empty output from `make2graph`, probably [this issue](https://github.com/lindenb/makefile2graph/issues/27). So can't check that.

Overall makes me wonder if it's worth switching from autotools to Meson. Many projects did so, it improves build system maintainability [and performance](https://mesonbuild.com/Simple-comparison.html).

stapelberg:
> Overall makes me wonder if it's worth switching from autotools to Meson. Many projects did so, it improves build system maintainability [and performance](https://mesonbuild.com/Simple-comparison.html).

Yes, this has been in the back of my mind, too. I have seen how much quicker meson is, and the language doesn’t look bad — certainly an improvement over autotools :)

I haven’t gotten around to writing a meson build file for i3 yet, though. If someone wanted to send a pull request to get the ball rolling, that would be cool.

Hi-Angel:
*(rebased)*

stapelberg (approved):

stapelberg:
Thank you!

orestisfl:
> > I can see the `libi3.a` -> `i3-config-wizard/i3-config-wizard` dependency before your patch (see [#3670 (comment)](https://github.com/i3/i3/issues/3670#issuecomment-555875797) for a cool way to visualize).
> 
> The comment was posted in 2019, perhaps something could have changed since then. I just tried making a similar graph, but I'm getting empty output from `make2graph`, probably [this issue](https://github.com/lindenb/makefile2graph/issues/27). So can't check that.
> 
> Overall makes me wonder if it's worth switching from autotools to Meson. Many projects did so, it improves build system maintainability [and performance](https://mesonbuild.com/Simple-comparison.html).

make2graph works for me, tried it today

jilljenn:
> The comment was posted in 2019, perhaps something could have changed since then. I just tried making a similar graph, but I'm getting empty output from `make2graph`, probably [this issue](https://github.com/lindenb/makefile2graph/issues/27). So can't check that.

OK I got it! You need to have your bash in English first: `export LC_ALL=C`

Issue #4087, opened 2020-05-17
get the build address sanitizer-clean


stapelberg:
While working on issue #4086, I noticed that we currently seem to have a bunch of unaddressed address sanitizer warnings, and at least some of these are legit memory leaks.

I’ll send a few changes to make asan easier to work with first, then try to clean up the issues it finds.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

orestisfl:
Can you share the logs?

stapelberg:
When building with `-O2`, a few places similar to https://github.com/i3/i3/pull/4088 are reported.

Aside from that, I’m getting reports about libxcb:

```
=================================================================
==3628304==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1024 byte(s) in 32 object(s) allocated from:
    #0 0x7f0566c79b3a in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x7f05669f0b17  (/usr/lib/libxcb.so.1+0xeb17)

SUMMARY: AddressSanitizer: 1024 byte(s) leaked in 32 allocation(s).
```

In general, the following files are flagged:
```
The following test logfiles contain AddressSanitizer or LeakSanitizer reports:
	testsuite-2020-05-17-15-52-40-4.18-282-gca6d6821/i3-log-for-113-urgent.t
	testsuite-2020-05-17-15-52-40-4.18-282-gca6d6821/i3-log-for-165-for_window.t
	testsuite-2020-05-17-15-52-40-4.18-282-gca6d6821/i3-log-for-166-assign.t
	testsuite-2020-05-17-15-52-40-4.18-282-gca6d6821/i3-log-for-167-workspace_layout.t
	testsuite-2020-05-17-15-52-40-4.18-282-gca6d6821/i3-log-for-170-force_focus_wrapping.t
	testsuite-2020-05-17-15-52-40-4.18-282-gca6d6821/i3-log-for-174-border-config.t
	testsuite-2020-05-17-15-52-40-4.18-282-gca6d6821/i3-log-for-176-workspace-baf.t
[...]

Issue #4086, opened 2020-05-16
Create a meson.build file for evaluating meson


stapelberg:
I had mentioned this before, but hadn’t actually filed an issue yet, so here goes.

The meson build system is very attractive. We should see if it works for all the things we need from a build system.

This issue tracks progress. I’m working on this right now.

Milestones:
* [x] i3* binaries built successfully
* [x] test suite ran successfully
* [x] release tarball created successfully
* [x] successful docs build
* [x] make install produces identical contents
* [x] release tarball contents match 4.18.1 release tarball (except for build system changes of course)
* [x] travis CI passes
* [x] fix release build detection in libi3/is_debug_build.c
* [x] fix and re-enable address sanitizer on travis
  * further cleanups tracked in https://github.com/i3/i3/issues/4087
* [x] install-from-tarball (as opposed to install-from-git) works with included docs/mans
* [ ] portability tested: iconv, mkdirp, strndup
* [x] release.sh works
* [x] travis CI exercises minimum supported meson version (0.45.1 for ubuntu 18.04): the ubuntu package we build post-submit cover this configuration
* [ ] complete-run.pl integrated with meson’s protocol: tap setting on test() targets

Work to split out for backwards compatibility:
* [x] xmacro changes: https://github.com/i3/i3/pull/4089
* [x] AnyEvent-I3 changes: https://github.com/i3/i3/pull/4090
* [x] `\` changes in complete-run.pl.in and lib/i3test.pm.in
* [x] flatten build dir structure: https://github.com/i3/i3/pull/4093

Final switch immediately before 4.19 release:
* [x] remove all autotools files
* [x] remove I3_VERSION file
* [x] remove autotools files from .gitignore
* [x] remove autotools build from .travis.yml
* [x] make sure all `TODO(autotools-removal)` occurrences are resolved

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

stapelberg:
https://github.com/stapelberg/i3/commit/836017a597200c587daa827cfebf63be492b8e19 is what I have so far. There are a few detail TODO items left, but overall it works and should allow you to get a good impression of meson.

I hope I can clean up the rest tomorrow and send a pull request.

stapelberg:
[...]

PR #4088, opened 2020-05-17
free(socket_path) to suppress leak sanitizer false-positive with -O2

stapelberg:
related to #4086 (triggered by our meson config)
related to #4087 (sanitizer cleanup tracking bug)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -376,6 +376,11 @@ int main(int argc, char *argv[]) {
                     char *socket_path = root_atom_contents("I3_SOCKET_PATH", NULL, 0);
                     if (socket_path) {
                         printf("%s\n", socket_path);
+                        /* With -O2 (i.e. the buildtype=debugoptimized meson
+                         * option, which we set by default), gcc 9.2.1 optimizes
+                         * away socket_path at this point, resulting in a Leak
+                         * Sanitizer report. An explicit free helps: */
+                        free(socket_path);
                         exit(EXIT_SUCCESS);
                     }
 


src/main.c
```diff
@@ -376,6 +376,11 @@ int main(int argc, char *argv[]) {
                     char *socket_path = root_atom_contents("I3_SOCKET_PATH", NULL, 0);
                     if (socket_path) {
                         printf("%s\n", socket_path);
+                        // With -O2 (i.e. the buildtype=debugoptimized meson
+                        // option, which we set by default), gcc 9.2.1 optimizes
+                        // away socket_path at this point, resulting in a Leak
+                        // Sanitizer report. An explicit free helps:
+                        free(socket_path);
```
orestisfl, line 383: I don't understand your explanation here. It seems that this is a leak because we have to free the results of `root_atom_contents`. So, instead I propose to free them in both paths.

orestisfl (changes requested):

src/main.c
stapelberg: What do you mean by both paths? This path is `socket_path != NULL`, so it doesn’t make sense to free in the other part, where `socket_path` is already `NULL`?

orestisfl: Sorry, it was late.

Anyway, this is not a false-positive, right? We are supposed to free `socket_path`.

stapelberg: No worries :)

Well, we terminate the process right after, so the OS will free that memory. Leaving cleanup to the OS is a widely used strategy.

The nuance here revolves around whether `socket_path` is reachable at process exit or not. When building with `buildtype=debug` (-g), `socket_path` is still considered reachable at `exit` time, which matches the intuitive interpretation: the variable is still in scope, so the memory should be reachable.

When building with `buildtype=debugoptimized` (-g -O2), `socket_path` is optimized out so aggressively that the leak sanitizer cannot find it and hence must consider it unreachable at `exit` time.

It seems possible to me that future versions of the leak sanitizer can deal with this better, or compiler optimizations might be relaxed when the leak sanitizer is used. For now, we just do the extra `free(socket_path)`, which is neither incorrect, nor required (under normal operation).

orestisfl: Ok I see. I think that freeing before exit is good either way so I don't think we should justify it with so much text (harder on the reader) but if you want to keep the comment, use `/*` for consistency.

stapelberg: I’d like to keep the comment, because an explicit free is at least an inconsistency. Without the comment, contributors might assume that all memory always needs to be freed before exiting the process (which is laborious, error-prone and unnecessary), or go the other way and remove the explicit `free()` for consistency, without realizing that it will break further down the road.

Updated the comment marks.

orestisfl (approved):

PR #4089, opened 2020-05-17
xmacro: declare in header files, instantiate instead of include

stapelberg:
This works better with meson, where .h files can be declared as being part of an
executable easily, but I couldn’t find a way to declare
e.g. include/atoms.xmacro as a dependency.

related to #4086


orestisfl:
Just for my understanding, wouldn't you be able to just rename these files to be header files, everything else the same?

stapelberg:
I haven’t tried that. I think the answer might be yes, but I’m not sure if it will stay that way.

In general, I thought we were following https://en.wikipedia.org/wiki/X_Macro, but we weren’t. With this change, our usage of x macros is closer to how they are documented on wikipedia, which I think means it’s a cleaner way and more likely to be working in every environment.

orestisfl (approved):
Thanks!

PR #4093, opened 2020-05-18
autotools: place binaries in top level of build dir, not in subdirs

stapelberg:
meson only supports the top level (no subdirs), so this makes the transition
easier.

For this to work with autotools, we need to *disable* the subdir-objects option,
that autotools wants us to enable for forward-compatibility.

This results in a bunch of warnings at autoreconf-time, but we don’t care, given
that we intend to switch away from autotools. Both build systems working next to
each other (as best as they can) is more important.

related to #4086

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,12 +5,12 @@ echo-version:
 
 bin_PROGRAMS = \
 	i3 \
-	i3bar/i3bar \
-	i3-config-wizard/i3-config-wizard \
-	i3-dump-log/i3-dump-log \
-	i3-input/i3-input \
-	i3-msg/i3-msg \
-	i3-nagbar/i3-nagbar
+	i3bar \
+	i3-config-wizard \
+	i3-dump-log \
+	i3-input \
+	i3-msg \
+	i3-nagbar
 
 install-exec-hook:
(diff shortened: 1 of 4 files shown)


orestisfl (approved):
Are you sure that these warnings can be ignored? Are we intending to keep both build systems?

stapelberg:
Pretty sure. The commit message of the pull request I’m about to send has more details on the migration plan. Quoting it:

> Migration plan
> 
> This commit adds the meson build files to the tree, but does not remove
> autotools yet. For the development phase, we will keep both build systems
> functional (and built on travis).
> 
> Then, just before the i3 v4.19 release, we will remove autotools from the tree
> and the release tarball will require meson to compile.
> 
> This way, we incentivize maintainers to change, while also offering them an easy
> way out (if desired) by reverting the most recent commit. In practice, switching
> a distribution package from autotools to meson should only be a few line change,
> easier than applying the provided patch :). Take a look at the debian/ changes
> in this commit for an example.
> 
> meson is broadly available everywhere that i3 is available: Both xorg-server and
> systemd gained meson build files in 2017, so we can follow suit:
> https://anholt.livejournal.com/52574.html
> https://in.waw.pl/~zbyszek/blog/systemd-meson.html

Issue #2611, opened 2016-12-19
Switch to AX_IS_RELEASE/git-version-gen?


stapelberg:
Currently, we use custom code to check whether the build is a release build or not: https://github.com/i3/i3/blob/8d739b7fe15fd351b5d703f922db01be46699a92/configure.ac#L32-L52

We should consider using [`AX_IS_RELEASE([dash-version])`](https://www.gnu.org/software/autoconf-archive/ax_is_release.html).

We should also consider whether our custom version file handling can be replaced by the more commonly used [git-version-gen](https://github.com/kergoth/autoconf/blob/master/build-aux/git-version-gen)

stapelberg:
Closing in favor of #4086 (switching to meson)

Issue #2246, opened 2016-03-08
enhancement: reproducible builds: asciidoc


stapelberg:
Once https://github.com/asciidoc/asciidoc/pull/9 is merged and included in a released version, we should set `-a footer-style=revdate` to have our documentation built reproducibly.

elextr:
@stapelberg one way to speed the commit to asciidoc is for you to test it with all backends and say it works.

rumpelsepp:
Maybe https://github.com/asciidoctor/asciidoctor provides that? That project seems to be much more active.

elextr:
@rumpelsepp an open source project is only active if people are active, the change is awaiting somebody testing it.

aerostitch:
@rumpelsepp : the idea behind this is also to be able to have the Debian packages that use asciidoc to generate their doc to be able to build in a reproducible way after that. So make everybody move to something else don't seem like the best solution ever! ;)

rumpelsepp:
I don't understand these strange comments here. Asciidoctor is just a dropin replacement that would compile the existing sources to the same results; and it provides a reproducible command line switch. I just have posted a pointer that *might* be useful. You can use it or not; and it is archived for the future if someone else wants to step up. That is also how open source works. Please stop fooling around.

Airblader:
This PR has been merged now.

PR #4094, opened 2020-05-18
add meson build files

stapelberg:
Motivation:

• faster builds (on an Intel Core i9-9900K):
  ( ../configure --disable-sanitizers && make -j8; )
  19,47s user 2,78s system 395% cpu 5,632 total

  ( meson .. -Dmans=true -Ddocs=true -Dprefix=/usr && ninja; )
  38,67s user 3,73s system 1095% cpu 3,871 total

• more approachable build system configuration in the
  python-esque meson domain specific language instead of
  the autotools m4 macro language

• built-in language server support thanks to ninja:
  the required compile_commands.json is built automatically
  and only needs to be linked from the source dir, e.g.:
  ln -s build/compile_commands.json .

Changes:

• the embedded vcs version info format changed from e.g.
  4.18-282-gabe46f69 (2020-05-16, branch "next")
  to:
  4.18-282-gabe46f69
  I think it’s better to lose a little bit of detail for
  the gained cleanliness of using meson’s vcs_tag()

• Drop unused xcb-event dependency.

• We can no longer enable sanitizers and debug options
  based on whether we are in a release or non-release build,
  because our new version logic runs at ninja build time,
  not at meson configure time.

  The new behavior is probably for the better in terms of
  what people expect, and we can make the CI use address sanitizer
  explicitly to ensure it is still exercised.

• We lose the AX_EXTEND_SRCDIR behavior, i.e. including the
  path component of the parent of the source dir in all paths.
  This was a trick we used for easier debugging, so that stack
  traces would contain e.g. ../i3-4.18.1/src/main.c, instead of
  just src/main.c.

  The other mechanism (_i3_version symbol) that we have for including
  the version number in the “backtrace full” (but not merely
  “backtrace”) output of gdb still works.

• Release tarballs now use tar.xz. Why not.

Migration plan

This commit adds the meson build files to the tree, but does not remove
autotools yet. For the development phase, we will keep both build systems
functional (and built on travis).

Then, just before the i3 v4.19 release, we will remove autotools from the tree
and the release tarball will require meson to compile.

This way, we incentivize maintainers to change, while also offering them an easy
way out (if desired) by reverting the most recent commit. In practice, switching
a distribution package from autotools to meson should only be a few line change,
easier than applying the provided patch :). Take a look at the debian/ changes
in this commit for an example.

meson is broadly available everywhere that i3 is available: Both xorg-server and
systemd gained meson build files in 2017, so we can follow suit:
https://anholt.livejournal.com/52574.html
https://in.waw.pl/~zbyszek/blog/systemd-meson.html

How do I?

For producing a coverage report, enable the b_coverage meson base option
and run ninja coverage-html:
% cd build
% meson .. -Db_coverage=true
% ninja
% ninja test
% ninja coverage-html
See also https://mesonbuild.com/howtox.html#producing-a-coverage-report

For using the address sanitizer, memory sanitizer or undefined behavior
sanitizer, use the b_sanitize meson base option:
% cd build
% meson .. -Db_sanitize=address
% ninja
See also https://mesonbuild.com/Builtin-options.html#base-options

related to #4086


.gitignore
```diff
@@ -40,8 +40,19 @@ docs/*.html
...
+/.clangd
 LAST_VERSION
 
+# We recommend building in a subdirectory called build.
+# If you chose a different directory name,
+# it is up to you to arrange for it to be ignored by git,
+# e.g. by listing your directory in .git/info/exclude.
+/build
+
+# TODO(https://github.com/i3/i3/issues/4086): remove autotools from .gitignore:
```
orestisfl: Maybe we can agree on a common TODO format that, when searched, should output all autotools-related TODOs. Eg let's agree to search for `TODO.+autotools`

.travis.yml
```diff
@@ -31,7 +31,10 @@ install:
 script:
   - docker run -v $PWD:/usr/src/i3/ -w /usr/src/i3 ${BASENAME} ./travis/check-safe-wrappers.sh
   - docker run -v $PWD:/usr/src/i3/ -w /usr/src/i3 ${BASENAME} ./travis/check-formatting.sh
+  # autoconf build has to keep working until i3 4.19 is released,
```
orestisfl: Add a TODO

orestisfl (changes requested):
Nice. Do you want me to check more extensively for problems here? I haven't used meson yet and I didn't checkout this PR. I assume we are going to merge & sort out problems later?

orestisfl:
Can you update [the release notes](https://github.com/i3/i3/blob/next/RELEASE-NOTES-next) as well?

orestisfl:
This also closes a bunch of build-related issues. Are we going to close them with this PR or with the PR that removes autotools?

.gitignore
stapelberg: Sounds good. I have changed to `TODO(autotools-removal)` here and elsewhere

.travis.yml
stapelberg: Done

stapelberg:
> Can you update [the release notes](https://github.com/i3/i3/blob/next/RELEASE-NOTES-next) as well?

Done

stapelberg:
> This also closes a bunch of build-related issues. Are we going to close them with this PR or with the PR that removes autotools?

Do you have a list? I think we can close them in this issue, as meson is the preferred build system once this is merged.

stapelberg:
> Nice. Do you want me to check more extensively for problems here? I haven't used meson yet and I didn't checkout this PR. I assume we are going to merge & sort out problems later?

I’m generally in favor of merging now and sorting out details later.

However, do feel free to give this a spin, and see if you notice any obvious issues. We can always decide to just capture them in the issue for now and fix later :)

orestisfl (changes requested):
Please update https://github.com/i3/i3/blob/ef8935b1db51a8135e869020717ff5c36feee392/docs/hacking-howto#L24

orestisfl:
> > This also closes a bunch of build-related issues. Are we going to close them with this PR or with the PR that removes autotools?
> 
> Do you have a list? I think we can close them in this issue, as meson is the preferred build system once this is merged.

Hmm, I think they were mostly fixed after all. Does your PR fix this https://github.com/i3/i3/issues/2833?

meson.build
```diff
@@ -0,0 +1,619 @@
...
+    # 'buildtype=debugoptimized',
+  ],
+  meson_version: '>=0.46.0',
+)
+
+cc = meson.get_compiler('c')
+add_project_arguments(cc.get_supported_arguments(['-Wunused-value']), language: 'c')
+
+# https://github.com/mesonbuild/meson/issues/2166#issuecomment-629696911
+meson.add_dist_script('meson/meson-dist-script')
```
orestisfl, line 23: > > Nice. Do you want me to check more extensively for problems here? I haven't used meson yet and I didn't checkout this PR. I assume we are going to merge & sort out problems later?
> 
> I’m generally in favor of merging now and sorting out details later.
> 
> However, do feel free to give this a spin, and see if you notice any obvious issues. We can always decide to just capture them in the issue for now and fix later :)

I'll try on arch later. On ubuntu 18.04 this fails with: `meson.build:6:0: ERROR: Meson version is 0.45.1 but project requires >=0.46.0.` is there any reason we require a recent version of meson?

Probably `add_dist_script`. From the documentation I read:
> add_dist_script(script_name, arg1, arg, ...) causes the script given as argument to run during dist operation after the distribution source has been generated but before it is archived. Note that this runs the script file that is in the staging directory, not the one in the source directory. If the script file can not be found in the staging directory, it is a hard error. This command can only invoked from the main project, calling it from a subproject is a hard error. **Available since 0.48.0**. Before 0.49.0, the function only accepted a single argument. Since 0.54.0 the MESON_SOURCE_ROOT and MESON_BUILD_ROOT environment variables are set when dist scripts are run. (Since 0.55.0) The output of configure_file, files, and find_program as well as strings.

With this patch the build succeeds:
```diff
diff --git a/meson.build b/meson.build
index 098515cb..924784f1 100644
--- a/meson.build
+++ b/meson.build
@@ -13,14 +13,14 @@ project(
     # TODO(https://github.com/i3/i3/issues/4087): switch to
     # 'buildtype=debugoptimized',
   ],
-  meson_version: '>=0.46.0',
+  meson_version: '>=0.45.0',
 )
 
 cc = meson.get_compiler('c')
 add_project_arguments(cc.get_supported_arguments(['-Wunused-value']), language: 'c')
 
 # https://github.com/mesonbuild/meson/issues/2166#issuecomment-629696911
-meson.add_dist_script('meson/meson-dist-script')
+# meson.add_dist_script('meson/meson-dist-script')
 
 ################################################################################
 # Version handling
```

Maybe we can check meson version + add a warning? Most builders don't care for `dist` I think.

meson.build
```diff
@@ -0,0 +1,619 @@
+# -*- mode: meson -*-
+
+# Style objective: be consistent with what mesonbuild.com documents/uses, and/or
+# the meson book: https://meson-manual.com/
+
+project(
+  'i3',
+  'c',
+  version: '4.18.1',
```
orestisfl, line 9: Shouldn't this be the next version?

meson.build
```diff
@@ -0,0 +1,619 @@
...
+  default_options: [
+    'c_std=c11',
+    'warning_level=1',  # enable all warnings (-Wall)
+    # TODO(https://github.com/i3/i3/issues/4087): switch to
```
orestisfl, line 13: Is this a 4.19 milestone TODO?

meson.build
```diff
@@ -0,0 +1,619 @@
...
+    # 'buildtype=debugoptimized',
+  ],
+  meson_version: '>=0.46.0',
```
orestisfl: Either 0.48 for `add_dist_script` or less (works at least on 0.45, see other comment)

meson.build
orestisfl: > > Nice. Do you want me to check more extensively for problems here? I haven't used meson yet and I didn't checkout this PR. I assume we are going to merge & sort out problems later?
> 
> I’m generally in favor of merging now and sorting out details later.
> 
> However, do feel free to give this a spin, and see if you notice any obvious issues. We can always decide to just capture them in the issue for now and fix later :)

I'll try on arch later. On ubuntu 18.04 this fails with: `meson.build:6:0: ERROR: Meson version is 0.45.1 but project requires >=0.46.0.` is there any reason we require a recent version of meson?

Probably `add_dist_script`. From the documentation I read:
> add_dist_script(script_name, arg1, arg, ...) causes the script given as argument to run during dist operation after the distribution source has been generated but before it is archived. Note that this runs the script file that is in the staging directory, not the one in the source directory. If the script file can not be found in the staging directory, it is a hard error. This command can only invoked from the main project, calling it from a subproject is a hard error. **Available since 0.48.0**. Before 0.49.0, the function only accepted a single argument. Since 0.54.0 the MESON_SOURCE_ROOT and MESON_BUILD_ROOT environment variables are set when dist scripts are run. (Since 0.55.0) The output of configure_file, files, and find_program as well as strings.

With this patch the build succeeds:
```diff
diff --git a/meson.build b/meson.build
index 098515cb..924784f1 100644
--- a/meson.build
+++ b/meson.build
@@ -13,14 +13,14 @@ project(
     # TODO(https://github.com/i3/i3/issues/4087): switch to
     # 'buildtype=debugoptimized',
   ],
-  meson_version: '>=0.46.0',
+  meson_version: '>=0.45.0',
 )
 
 cc = meson.get_compiler('c')
 add_project_arguments(cc.get_supported_arguments(['-Wunused-value']), language: 'c')
 
 # https://github.com/mesonbuild/meson/issues/2166#issuecomment-629696911
-meson.add_dist_script('meson/meson-dist-script')
+# meson.add_dist_script('meson/meson-dist-script')
 
 ################################################################################
 # Version handling
```

Maybe we can check meson version + add a warning? Most builders don't care for `dist` I think.

orestisfl (changes requested):

meson.build
stapelberg: We use 4.18.1 in configure.ac as well. release.sh updates it. I suggest keeping this as-is.

meson.build
stapelberg: There’s no urgency here. We can release 4.19 without this change.

meson.build
stapelberg: Done. There were 2 conditionals I needed to add (depends arg for `test` target).

stapelberg:
> Please update hacking-howto

Done.

stapelberg:
> Hmm, I think they were mostly fixed after all. Does your PR fix this #2833?

Not entirely sure. That problem seems to be about old versions of git. There is a chance that it is fixed accidentally due to meson invoking git differently. We’ll have to have the affected users report back :)

orestisfl:
Ok, I propose closing and asking for feedback.

> > Hmm, I think they were mostly fixed after all. Does your PR fix this #2833?
> 
> Not entirely sure. That problem seems to be about old versions of git. There is a chance that it is fixed accidentally due to meson invoking git differently. We’ll have to have the affected users report back :)

orestisfl:
Other build-related issues:

* https://github.com/i3/i3/issues/2611
* https://github.com/i3/i3/issues/2246

.dockerignore
```diff
@@ -0,0 +1 @@
+.git
```
orestisfl, line 1: Is this now obsolete?

https://github.com/i3/i3/blob/ef8935b1db51a8135e869020717ff5c36feee392/travis/docker-build-and-push.sh#L10

meson.build
```diff
@@ -0,0 +1,633 @@
...
+  # at which point Ubuntu 20.04 is reasonable to expect.
+  meson_version: '>=0.45.0',
+)
+
+cc = meson.get_compiler('c')
+add_project_arguments(cc.get_supported_arguments(['-Wunused-value']), language: 'c')
+
+if meson.version().version_compare('>=0.48.0')
+  # https://github.com/mesonbuild/meson/issues/2166#issuecomment-629696911
+  meson.add_dist_script('meson/meson-dist-script')
```
orestisfl, line 27: Maybe also some kind of warning?

stapelberg:
I think #2246 is related, but independent. #2611 is obsolete indeed, closed it.

meson.build
```diff
@@ -0,0 +1,633 @@
...
+  version: '4.18.1',
+  default_options: [
+    'c_std=c11',
+    'warning_level=1',  # enable all warnings (-Wall)
+    # TODO(https://github.com/i3/i3/issues/4087): switch to
+    # 'buildtype=debugoptimized',
+  ],
+  # TODO: we can target more recent versions of meson starting with i3 v4.20,
+  # at which point Ubuntu 20.04 is reasonable to expect.
+  meson_version: '>=0.45.0',
```
orestisfl, line 19: But why not keep support for older ones as well? 18.04 has standard support until 2023. Not saying that it is our duty to keep i3 working on ancient software but historically i3 has been much more conservative than this.

I've tried and this works:

```suggestion
  meson_version: '>=0.44.0',
```

Maybe we should make sure in our CI that we are also testing with the minimal supported meson version.

orestisfl (changes requested):

.dockerignore
stapelberg: Ah, yeah, good catch. Done.

meson.build
stapelberg: Done. I hope the messages actually describe the failure mode — I haven’t tested this.

meson.build
stapelberg: Okay, fair enough. I have softened the TODO, let’s see how this goes. We can revisit if it becomes a burden.

Why the suggestion for 0.44.0? Is there a popular Linux version that doesn’t have 0.45.0?

orestisfl: The suggestion was more like a FYI. Ideally we would like to support the minimal supported version. Practically, using something that we can have on our CI from apt is less of a burden.

meson.build
orestisfl: Works for me: `Message: meson <0.48.0 detected, dist tarballs will not be filtered`

meson.build
stapelberg: Let’s target 0.45.0 then. I added a TODO item to https://github.com/i3/i3/issues/4086 for this.

.travis.yml
```diff
@@ -31,7 +31,10 @@ install:
 script:
   - docker run -v $PWD:/usr/src/i3/ -w /usr/src/i3 ${BASENAME} ./travis/check-safe-wrappers.sh
   - docker run -v $PWD:/usr/src/i3/ -w /usr/src/i3 ${BASENAME} ./travis/check-formatting.sh
+  # # TODO(autotools-removal): autoconf build has to keep working until i3 4.19
+  # is released, so we keep building it in travis.
```
orestisfl, line 35: More generally speaking: do you think that our travis needs a bit of house-keeping?

I think we should use a build matrix (also see https://github.com/i3/i3status/pull/414) that tests several configuration in parallel. For example, we should check that everything works under debian, ubuntu:latest, ubuntu:18.04. In this case we would also add meson & autools. I think what happens right now is that everything runs sequentially, increasing the total time.

We can also move to github actions but honestly this is partially change for change's sake but I do find github actions easier to work with and AFAIK they offer faster parallelism.

stapelberg: Yes, I think there are some improvements we could do on our CI setup.

One thing to keep in mind: it should stay roughly as fast as it is now (or faster of course), so we need to be careful with adding too many variations.

I haven’t used GitHub actions much (only started today), but I like what I see so far. It seems faster and easier to use. I think we should switch to GitHub actions before investing any more time into our travis setup.

All of this is independent of this particular PR, though :)

orestisfl (approved):
Great!

I was a bit skeptical about this change at first (why change what works?) but, even though I know autotools better than mason, this looks much cleaner and easier to work with and I think it will save us time in the long term.

stapelberg:
Glad to hear. This matches my thinking exactly.

PR #4098, opened 2020-05-19
More CI fixes

stapelberg:
related to #4086

### DIFF
--- a/travis/docs.sh
+++ b/travis/docs.sh
@@ -3,20 +3,24 @@
 set -e
 set -x
 
-for f in $(grep '\.html$' debian/i3-wm.docs | grep -v 'docs/refcard.html' | grep -v 'docs/lib-i3test')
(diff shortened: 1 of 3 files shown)


travis/docs.sh
```diff
@@ -3,20 +3,24 @@
 set -e
 set -x
 
-for f in $(grep '\.html$' debian/i3-wm.docs | grep -v 'docs/refcard.html' | grep -v 'docs/lib-i3test')
+# TODO: install the docs via meson, inject styles with an option
```
orestisfl, line 6: Is this TODO also not important?

orestisfl (approved):

travis/docs.sh
stapelberg: The current way works, the TODO is for a cleaner solution longer term. We can fix it when it next breaks, if we want :)

PR #4100, opened 2020-05-19
meson: only install docs/man pages when they exist (release tarball)

stapelberg:
related to #4086

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -125,20 +125,22 @@ if get_option('docs')
   endforeach
 
 else
-  install_data(
-    [
-      'docs/hacking-howto.html',
-      'docs/userguide.html',
-      'docs/ipc.html',
-      'docs/multi-monitor.html',
-      'docs/wsbar.html',
-      'docs/testsuite.html',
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -125,20 +125,22 @@ if get_option('docs')
...
-      'docs/multi-monitor.html',
-      'docs/wsbar.html',
-      'docs/testsuite.html',
-      'docs/i3bar-protocol.html',
-      'docs/layout-saving.html',
-      'docs/debugging.html',
-    ],
-    install_dir: docdir,
-  )
+  if run_command('[', '-f', 'docs/hacking-howto.html', ']').returncode() == 0
```
orestisfl, line 128: While we are at it, why not check for everything with a for loop?

orestisfl (changes requested):

meson.build
stapelberg: Sounds like overkill to me :) I’d like to keep these files as simple as possible.

orestisfl (approved):

PR #4107, opened 2020-05-24
Add situational exit codes

### LINKED ISSUE #3705: Return a i3-input exit status indicating whether a command was executed
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

Currently i3-input returns 0 whenever no error occured in the program itself.

## Desired Behavior

i3-input could return 0 when the specified command was executed successfully and other values (e.g. 1 and 2) when the user canceled the process (by using the escape key) or the specified command failed respectively.

## Environment

<pre>
Running i3 version: 4.16.1 (2019-01-27) (pid 1660)abort…)
Loaded i3 config: /home/rittelle/.config/i3/config (Last modified: Sat 18 May 2019 08:06:23 AM CEST, 268412 seconds ago)

The i3 binary you just called: /nix/store/mf89f7cf3nnvzrv4ybyym476v8w02z68-i3-gaps-4.16.1/bin/i3
The i3 binary you are running: /nix/store/mf89f7cf3nnvzrv4ybyym476v8w02z68-i3-gaps-4.16.1/bin/i3
</pre>
Yes, I‘m using gaps, but skimming over the i3-input code on the next branch, the current behavior seems to be the same.

Arc676:
Closes #3705.

I used `autoconfigure` to generate the `Makefile` and compile as specified in the instructions, but the generated binary seems to be a memory leak check wrapped around the actual executable so I can't test the exit codes. It seems there's a memory leak in one of the dependencies and whenever I run the executable, no matter what happens within `i3-input` I get the following output regarding the memory leaks and exit code 1.

How can I just run the newly compiled `i3-input` without the memory leak checker?

```
==417634==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 3840 byte(s) in 6 object(s) allocated from:
    #0 0x7fe22a0dbfa0 in __interceptor_realloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:163
    #1 0x7fe229315661  (/usr/lib/libfontconfig.so.1+0x21661)

Indirect leak of 7040 byte(s) in 220 object(s) allocated from:
    #0 0x7fe22a0dbb3a in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x7fe229301780  (/usr/lib/libfontconfig.so.1+0xd780)

Indirect leak of 3635 byte(s) in 287 object(s) allocated from:
    #0 0x7fe22a0628e6 in __interceptor_strdup /build/gcc/src/gcc/libsanitizer/asan/asan_interceptors.cc:445
    #1 0x7fe229314e45 in FcValueSave (/usr/lib/libfontconfig.so.1+0x20e45)

Indirect leak of 3328 byte(s) in 104 object(s) allocated from:
    #0 0x7fe22a0dbd48 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:153
    #1 0x7fe229315c09  (/usr/lib/libfontconfig.so.1+0x21c09)

Indirect leak of 2304 byte(s) in 72 object(s) allocated from:
    #0 0x7fe22a0dbd48 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:153
    #1 0x7fe2293150d9  (/usr/lib/libfontconfig.so.1+0x210d9)

Indirect leak of 576 byte(s) in 18 object(s) allocated from:
    #0 0x7fe22a0dbd48 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:153
    #1 0x7fe229314f77  (/usr/lib/libfontconfig.so.1+0x20f77)

Indirect leak of 144 byte(s) in 3 object(s) allocated from:
    #0 0x7fe22a0dbb3a in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x7fe22930f03e in FcLangSetCreate (/usr/lib/libfontconfig.so.1+0x1b03e)

SUMMARY: AddressSanitizer: 20867 byte(s) leaked in 710 allocation(s).
```

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -26,7 +26,7 @@ working. Please reach out to us in that case!
     to their definition order in the config file
   • update i3bar config when necessary (reduces redraws on bar mode changes)
   • mention rofi in default config file
-  • i3-input: added different exit codes for when i3-input fails
+  • i3-input: add different exit codes for when i3-input fails
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -26,6 +26,7 @@ working. Please reach out to us in that case!
     to their definition order in the config file
   • update i3bar config when necessary (reduces redraws on bar mode changes)
   • mention rofi in default config file
+  • i3-input: added different exit codes for when i3-input fails
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
--- a/i3-input/main.c
+++ b/i3-input/main.c
@@ -32,6 +32,14 @@ xcb_visualtype_t *visual_type = NULL;
 #define BORDER logical_px(2)
 #define PADDING logical_px(2)
 
+/* Exit codes for i3-input:
+ * 0 if i3-input exited successfully and the command was run
+ * 1 if the user canceled input
+ * 2 if i3-input fails for any other reason */
+const int EXIT_OK = 0;
+const int EXIT_CANCEL = 1;
+const int EXIT_ERROR = 2;
+
 /* IPC format string. %s will be replaced with what the user entered, then
  * the command will be sent to i3 */
 static char *format;
@@ -186,11 +194,11 @@ static void finish_input(void) {
     /* prefix the command if a prefix was specified on commandline */
     printf("command = %s\n", full);
 
-    ipc_send_message(sockfd, strlen(full), 0, (uint8_t *)full);
+    int ret = ipc_send_message(sockfd, strlen(full), 0, (uint8_t *)full);
 
     free(full);
 
-    exit(0);
+    exit(ret == 0 ? EXIT_OK : EXIT_ERROR);
 }
 
 /*
@@ -239,7 +247,7 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
         return 1;
     }
     if (sym == XK_Escape) {
-        exit(0);
+        exit(EXIT_CANCEL);
     }
 
     /* TODO: handle all of these? */
@@ -297,7 +305,7 @@ static xcb_rectangle_t get_window_position(void) {
     xcb_intern_atom_reply_t *nswc_reply = xcb_intern_atom_reply(conn, nswc_cookie, NULL);
     if (nswc_reply == NULL) {
         ELOG("Could not intern atom _NET_SUPPORTING_WM_CHECK\n");
-        exit(-1);
+        exit(EXIT_ERROR);
     }
     A__NET_SUPPORTING_WM_CHECK = nswc_reply->atom;
     free(nswc_reply);
@@ -392,7 +400,7 @@ int main(int argc, char *argv[]) {
                 break;
             case 'v':
                 printf("i3-input " I3_VERSION);
-                return 0;
+                return EXIT_OK;
             case 'p':
                 /* This option is deprecated, but will still work in i3 v4.1, 4.2 and 4.3 */
                 fprintf(stderr, "i3-input: WARNING: the -p option is DEPRECATED in favor of the -F (format) option\n");
@@ -420,7 +428,7 @@ int main(int argc, char *argv[]) {
                 printf("\n");
                 printf("Example:\n");
                 printf("    i3-input -F 'workspace \"%%s\"' -P 'Switch to workspace: '\n");
-                return 0;
+                return EXIT_OK;
         }
     }
     if (!format) {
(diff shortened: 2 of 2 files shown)


orestisfl:
To avoid the leak checker you should use `--disable-sanitizers` during the `./configure` step

i3-input/main.c
```diff
@@ -32,6 +32,18 @@ xcb_visualtype_t *visual_type = NULL;
...
+ * 0 if i3-input exited successfully and the command was run
+ * 1 if the user canceled input
+ * 2 if i3-input failed to send the command via IPC
+ * 3 if the keyboard was not found
+ * 4 if _NET_SUPPORTING_WM_CHECK atom interning fails */
+const int EXIT_OK = 0;
+const int EXIT_CANCEL = 1;
+const int EXIT_IPC_FAIL = 2;
+const int EXIT_KEYBOARD_FAIL = 3;
+const int EXIT_INTERN_FAIL = 4;
```
orestisfl: Maybe we should avoid special purpose exit codes:
https://tldp.org/LDP/abs/html/exitcodes.html
https://stackoverflow.com/questions/1101957/are-there-any-standard-exit-status-codes-in-linux

i3-input/main.c
```diff
@@ -239,7 +251,7 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
         return 1;
     }
     if (sym == XK_Escape) {
-        exit(0);
+        exit(EXIT_CANCEL);
```
orestisfl, line 250: This is a bit of a breaking change that could affect some existing user scripts that check for i3-input failure (everything other than zero is a failure in scripts, regardless of what special name we give to it)

orestisfl (changes requested):
Thank you for your PR

Can you also update the changelog https://github.com/i3/i3/blob/next/RELEASE-NOTES-next with `i3-input: add situational exit codes`

i3-input/main.c
Arc676: Good point. I'll shift the exit codes up to 3-6 to avoid the overlap with standard exit codes.

i3-input/main.c
Arc676: I understand. In the original feature request, an alternate exit code for "user canceled" was included; this is clearly incompatible with the existing expectation that simply canceling the operation doesn't count as a failure. How do you believe this conflict should be resolved? One possibility would be for "user canceled" to only yield a nonzero exit code if a command line argument is specified to request this, but I'd be curious to know what you think.

i3-input/main.c
orestisfl: On the other hand, many utilities like git seem to be using codes >= 1. Maybe we don't need to change anything and there doesn't seem to be a very established practice.

orestisfl: But I do believe that we don't need so specific errors. I propose:
0 -> success
1 -> cancel (if we do it)
2 -> other errors

i3-input/main.c
orestisfl: @Airblader WDYT?

Arc676: Since the PR description specifies that it implements the original feature request, for now I've committed changes including the separate exit code for user canceled and rebased the PR branch with the new commits on the destination branch. As discussed previously, I've reduced the number of custom exit codes to just 2 cases. The failure in the Travis build appears to be on unit testing which can likely be attributed to the aforementioned breaking change.

The simplest way to both fulfill the feature request and prevent breaking changes seems to make this new behavior opt-in via command line argument. I'll implement this after getting confirmation that this would be in line with the design of `i3-input` so as to avoid adding irrelevant commits to this PR.

orestisfl (approved):

orestisfl:
Can you also squash the commits together? Rebasing with `next` is also a good idea because it will make travis pass more easily.

RELEASE-NOTES-next
```diff
@@ -26,6 +26,7 @@ working. Please reach out to us in that case!
     to their definition order in the config file
   • update i3bar config when necessary (reduces redraws on bar mode changes)
   • mention rofi in default config file
+  • i3-input: added different exit codes for when i3-input fails
```
orestisfl: And a very small nitpick

```suggestion
  • i3-input: add different exit codes for when i3-input fails
```

orestisfl (changes requested):

orestisfl (approved):

orestisfl:
Ok nvm, thank you!

Arc676:
No problem. So I noticed in the end there wasn't a final decision (at least not in this thread) regarding the exit code for when the user cancels. Is this a change in behavior for `i3-input` for future releases?

orestisfl:
I implicitly accepted the exit code = 1 for canceling, hoping it will not be too much backwards incompatible and probably safer as well.

Issue #4110, opened 2020-05-28
i3bar: allow "power savings" feature to be disabled


ammgws:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
i3bar has a "power savings" feature that pauses the bar via SIGSTOP when it is hidden or
obscured by a fullscreen app, and this behaviour is not configurable.

## Desired Behavior
Power savings mode should be configurable.

## Background Information
This "power savings" feature  becomes  a problem for users who
use status line generators such as i3status-rust that run threads/proccesses that need
to run at all times. A [hacky way](https://github.com/greshake/i3status-rust/pull/701) to prevent this from happening is to set
stop_signal to a signal that won't cause the status line generator program to stop,
however I do not believe that is a good solution.

This has surprised users a fair amount of time over the past few years, to show a few:
greshake/i3status-rust#503
greshake/i3status-rust#694 (comment)
https://faq.i3wm.org/question/4631/dont-sigstop-when-in-hide-mode.1.html
i3#3242 (comment)
https://www.reddit.com/r/i3wm/comments/gm6yfh/is_there_a_painless_way_to_keep_i3bar_children/

## Implementation
I have added a new config option to the bar section that allows the power savings feature to be disabled here: https://github.com/ammgws/i3/commit/c7d4fb56a5fd4c5bd19f7848b3b3fb5c3bb4c60a

orestisfl:
I am pretty sure you can do this by setting the stop signal to zero

ammgws:
Yeah IIRC the i3 child_stop/child_cont function checks if the signal is > 0, but I thought that was just a check to see if the signal is valid rather than a way to intentionally disable the feature. Plus it's undocumented, and would mean having to add support for that in every affected status line generator program. To me it seems like this obscure feature should be able to be disabled on the i3 side, as mentioned in #3242 as well.

orestisfl:
We can bless the feature in the documentation but setting something to zero sounds good enough to me

ammgws:
The caveat would be that adding support for disabling the power saving feature would then need to be implemented by each and every status line generator instead of being able to disable it on the i3 side, but I think I've only seen i3status/i3status-rust/i3block/i3pystatus mentioned with respect to this issue so it might not be that many.

Do you need to talk to the other i3 devs first? Otherwise if you believe just documenting the feature is enough then I will send through a PR for that.

[...]

PR #4115, opened 2020-05-31
i3bar: improve docs for battery saving feature

ammgws:
Document that `stop_signal` can be set to `0` to disable the battery saving feature, as discussed in #4110.


orestisfl (approved):
SGTM but usually @Airblader has more to say about text

Airblader (approved):

Airblader:
LGTM as well

PR #4122, opened 2020-06-06
Introduce GET_BINDING_STATE IPC command

stapelberg:
fixes #3892

PR #4128, opened 2020-06-13
i3-msg: add missing option for `get_binding_state`

ammgws:
Was introduced in #4122 but the appropriate logic was missing from i3-msg itself.


stapelberg (approved):

stapelberg:
Thanks!

PR #4146, opened 2020-06-29
Fixed a typo in the user's guide

krnlks:


orestisfl (approved):

orestisfl:
Thanks!

nikitabobko:
No, it's not a typo. i3 doesn't use default vim bindings for arrows. It shifts them to the right by one key.

Proof that `k` is down by default, not `j`: https://github.com/i3/i3/blob/next/etc/config#L45

I reverted this commit: https://github.com/i3/i3/pull/5390

PR #4189, opened 2020-09-01
allow ppt values in move direction and move position commands

### LINKED ISSUE #3519: Add ppt support to move position (similar to resize)
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

Right now, the move syntax only supports px:

```
move position <pos_x> [px] <pos_y> [px]
```

## Desired Behavior

`resize` learned to use `ppt` to support percentage values here: https://github.com/i3/i3/pull/3023

It would be useful (and also more consistent) to support ppt (percentage-based) measures for `move`.

Example use case: I'm trying to top-middle align a window to take up 80% of the screen width.

`resize set 80 ppt 40 ppt` works great, but would be nice if `move absolute position 10 ppt 0 ppt` also worked (move 10% from the left).

## Environment

<pre>
i3 version: 4.16 (2018-11-04) 
</pre>

<pre>
</pre>

acheronfail:
This adds support for `ppt` values to be used in the following `move` commands:

```
move <left|right|down|up> [<amount> [px|ppt]]
move position <pos_x> [px|ppt] <pos_y> [px|ppt]
```

Fixes #3519

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -245,7 +245,7 @@ you open a new terminal, it will open below the current one.
 
 So, how can you open a new terminal window to the *right* of the current one?
 The solution is to use +focus parent+, which will focus the +Parent Container+ of
-the current +Container+. In default configuration, use +$mod+a+ to navigate one 
+the current +Container+. In default configuration, use +$mod+a+ to navigate one
 +Container+ up the tree (you can repeat this multiple times until you get to the
 +Workspace Container+). In this case, you would focus the +Vertical Split Container+
 which is *inside* the horizontally oriented workspace. Thus, now new windows will be
@@ -1415,7 +1415,7 @@ button6::
     Scroll wheel right.
 button7::
     Scroll wheel left.
-    
+
 
 Please note that the old +wheel_up_cmd+ and +wheel_down_cmd+ commands are deprecated
 and will be removed in a future release. We strongly recommend using the more general
@@ -1682,8 +1682,8 @@ When +strip_workspace_name+ is set to +yes+, any workspace that has a name of
 the form "[n][:][NAME]" will display only the number.
 
 The default is to display the full name within the workspace button. Be aware
-that the colon in the workspace name is optional, so `[n][NAME]` will also 
-have the workspace name and number stripped correctly. 
+that the colon in the workspace name is optional, so `[n][NAME]` will also
+have the workspace name and number stripped correctly.
 
 *Syntax*:
 ------------------------------
@@ -2140,12 +2140,15 @@ Use the +move+ command to move a container.
 # Moves the container into the given direction.
 # The optional pixel argument specifies how far the
 # container should be moved if it is floating and
-# defaults to 10 pixels.
-move <left|right|down|up> [<px> px]
+# defaults to 10 pixels. The optional ppt argument
+# means "percentage points", and if specified it indicates
+# how many points the container should be moved if it is
+# floating rather than by a pixel value.
+move <left|right|down|up> [<amount> [px|ppt]]
 
 # Moves the container to the specified pos_x and pos_y
 # coordinates on the screen.
-move position <pos_x> [px] <pos_y> [px]
+move position <pos_x> [px|ppt] <pos_y> [px|ppt]
 
 # Moves the container to the center of the screen.
 # If 'absolute' is used, it is moved to the center of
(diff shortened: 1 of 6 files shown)


docs/userguide
```diff
@@ -2138,14 +2138,14 @@ Use the +move+ command to move a container.
 *Syntax*:
 -----------------------------------------------------
 # Moves the container into the given direction.
-# The optional pixel argument specifies how far the
+# The optional amount argument specifies how far the
```
orestisfl: Please briefly describe what `ppt` means for both cases, like in [Resizing containers/windows
](https://i3wm.org/docs/userguide.html#resizingconfig)

src/commands.c
```diff
@@ -1492,34 +1492,36 @@ void cmd_sticky(I3_CMD, const char *action) {
...
-            DLOG("floating move with %ld pixels\n", move_px);
+            DLOG("floating move with %ld %s\n", amount, mode);
             Rect newrect = current->con->parent->rect;
+            Con *output = con_get_output(current->con);
+            bool is_ppt = mode && strcmp(mode, "ppt") == 0;
 
             switch (direction) {
                 case D_LEFT:
-                    newrect.x -= move_px;
+                    newrect.x -= is_ppt ? output->rect.width * ((double)amount / 100.0) : amount;
```
orestisfl, line 1514: Here and below, why `output->rect.width`?

orestisfl (changes requested):

src/commands.c
acheronfail: The intention here is that a command like:

```
move right 20 ppt
```

Will move the floating container to the right by 20% of the screen's width.
I used `con_get_output` and extracted the output's width in this fashion since I had seen it done in a similar way here: https://github.com/i3/i3/blob/04c489043c5caf2d4c07f16bf9606de581dc9800/src/commands.c#L674

orestisfl: Ok sorry, I misread the line before

acheronfail:
I only added some more text into the documentation but the build failed this time, I think it might have just been a dodgy build? All the tests pass locally for me.

Any chance you could re-run it and see if that fixes it? (Unless I'm missing something else?)

acheronfail:
Just ran `git commit --amend --no-edit` to re-trigger the build.

src/commands.c
```diff
@@ -1492,34 +1492,36 @@ void cmd_sticky(I3_CMD, const char *action) {
...
 
     TAILQ_FOREACH (current, &owindows, owindows) {
-        DLOG("moving in direction %s, px %ld\n", direction_str, move_px);
+        DLOG("moving in direction %s, %ld %s\n", direction_str, amount, mode);
         if (con_is_floating(current->con)) {
-            DLOG("floating move with %ld pixels\n", move_px);
+            DLOG("floating move with %ld %s\n", amount, mode);
             Rect newrect = current->con->parent->rect;
+            Con *output = con_get_output(current->con);
+            bool is_ppt = mode && strcmp(mode, "ppt") == 0;
```
orestisfl: nit: move this outside the loop and make it `const`

src/commands.c
```diff
@@ -1492,34 +1492,36 @@ void cmd_sticky(I3_CMD, const char *action) {
...
+void cmd_move_direction(I3_CMD, const char *direction_str, long amount, const char *mode) {
     owindow *current;
     HANDLE_EMPTY_MATCH;
 
     Con *initially_focused = focused;
     direction_t direction = parse_direction(direction_str);
 
     TAILQ_FOREACH (current, &owindows, owindows) {
-        DLOG("moving in direction %s, px %ld\n", direction_str, move_px);
+        DLOG("moving in direction %s, %ld %s\n", direction_str, amount, mode);
```
orestisfl: nit: move outside, it doesn't seem to change in the loop

orestisfl (approved):
Sounds good, thanks.

Just some minor nitpicks

acheronfail:
@orestisfl thanks for looking into that for me. :slightly_smiling_face: I should have addressed those minor nits now.

orestisfl:
Ok, final things:

1. Add a small line in https://github.com/i3/i3/blob/next/RELEASE-NOTES-next describing your change
2. Squash all commits into one
3. Rebase into next to make travis easier to pass

acheronfail:
Thanks for the help! I think that should be everything you asked for. :slightly_smiling_face:

orestisfl:
Thanks!

PR #4195, opened 2020-09-13
i3bar: add support for nonprimary output

### LINKED ISSUE #4083: Support for nonprimary outputs
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

I have several screens. But depending on what I do, I don't always set up the same one as primary.  I'm using a bar on each screen though. I set up the first one with `output primary` but I'd like to specify a bar for all secondary screens. Sot it would be great to have an option `outpout nonprimary` or `output secondary`.

## Desired Behavior

To be able to specify a bar for all nonprimary screens.

## Environment

<pre>
i3 version: Binary i3 version:  4.18 (2020-02-17) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.18 (2020-02-17) (pid 9341)o abort…)
Loaded i3 config: /home/$USER/.config/i3/config (Last modified: jeu. 14 mai 2020 09:12:03 CEST, 1071 seconds ago)

</pre>

<pre>
- Linux Distribution & Version: Ubuntu 18.04
</pre>

mschuwalow:
Adds support for a `nonprimary` output to i3bar. This is useful when configuring your bar using `primary`, as you can have one config for your main display and another for any additional display.

Fixes #4083

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -29,6 +29,7 @@ working. Please reach out to us in that case!
   • i3-input: add different exit codes for when i3-input fails
   • allow ppt values in move direction and move position commands
   • i3bar: add coordinates relative to the current output in i3bar click events
+  • i3bar: add “nonprimary” output option
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
--- a/docs/userguide
+++ b/docs/userguide
@@ -1484,9 +1484,16 @@ options for different outputs by using multiple 'bar' blocks.
 To make a particular i3bar instance handle multiple outputs, specify the output
 directive multiple times.
 
+These output names have a special meaning:
+
+primary::
+    Selects the output that is configured as primary in the X server.
+nonprimary::
+    Selects every output that is not configured as primary in the X server.
+
 *Syntax*:
 ---------------
-output primary|<output>
+output primary|nonprimary|<output>
 ---------------
 
 *Example*:
--- a/i3bar/src/outputs.c
+++ b/i3bar/src/outputs.c
@@ -192,11 +192,12 @@ static int outputs_end_map_cb(void *params_) {
 
     /* See if we actually handle that output */
     if (config.num_outputs > 0) {
+        const bool is_primary = params->outputs_walk->primary;
         bool handle_output = false;
         for (int c = 0; c < config.num_outputs; c++) {
-            if (strcasecmp(params->outputs_walk->name, config.outputs[c]) == 0 ||
-                (strcasecmp(config.outputs[c], "primary") == 0 &&
-                 params->outputs_walk->primary)) {
+            if ((strcasecmp(params->outputs_walk->name, config.outputs[c]) == 0) ||
+                (strcasecmp(config.outputs[c], "primary") == 0 && is_primary) ||
+                (strcasecmp(config.outputs[c], "nonprimary") == 0 && !is_primary)) {
                 handle_output = true;
                 break;
             }


wlhlm:
Thanks for taking a stab at this. I've been interested in being able to configure i3bar on non-primary outputs for a while.

Associated issues: #4083, https://github.com/i3/i3/issues/2311#issuecomment-213966121

~~Though, looking at your code, it looks quite limited to me: IIUIC, this will pick the first non-primary output and stop. This makes sense if you have two or less outputs, otherwise it will just pick the first non-primary output from the list and that depends on how the list is assembled (not sure if deterministic). I think it makes more sense if the `nonprimary` option displays the bar on all non-primary outputs not just the first one. If you have more then two outputs and want to display the bar only  on a single non-primary output, I thinks it's better to hardcode the output.~~

mschuwalow:
As far as I can tell this works with any number of outputs. I'm currently using this with 3 monitors and everything works as expected https://github.com/mschuwalow/nixos-config/blob/master/profiles/desktop/i3.nix#L455.

My understanding of that part of the code is that one map will be sent for each output, so it's enough for us to check whether we are supposed to handle it. In the case of `nonprimary` that means short-circuiting as soon as we see an output that is not marked as primary.

wlhlm:
> As far as I can tell this works with any number of outputs. I'm currently using this with 3 monitors and everything works as expected

OK, I misread the code then. Good to hear! 👍

docs/userguide
```diff
@@ -1486,7 +1486,7 @@ directive multiple times.
 
 *Syntax*:
 ---------------
-output primary|<output>
+output primary|nonprimary|<output>
```
orestisfl, line 1496: Please briefly explain what nonprimary means. I.e., every output except the primary

i3bar/src/outputs.c
```diff
@@ -195,8 +195,8 @@ static int outputs_end_map_cb(void *params_) {
         bool handle_output = false;
         for (int c = 0; c < config.num_outputs; c++) {
             if (strcasecmp(params->outputs_walk->name, config.outputs[c]) == 0 ||
-                (strcasecmp(config.outputs[c], "primary") == 0 &&
-                 params->outputs_walk->primary)) {
+                (strcasecmp(config.outputs[c], "primary") == 0 && params->outputs_walk->primary) ||
+                (strcasecmp(config.outputs[c], "nonprimary") == 0 && !params->outputs_walk->primary)) {
```
orestisfl: I think this will be  a bit more readable with a const:
```c
const bool is_primary = params->outputs_walk->primary
```

```suggestion
                (strcasecmp(config.outputs[c], "primary") == 0 && is_primary ||
                (strcasecmp(config.outputs[c], "nonprimary") == 0 && !is_primary)) {
```

orestisfl (changes requested):
Thanks! Minor changes.

Also, question: Have you verified that the bar location is updated when you change the primary output? It should in theory work.

mschuwalow:
> Thanks! Minor changes.
> 
> Also, question: Have you verified that the bar location is updated when you change the primary output? It should in theory work.

Yes, I can confirm that it works. :+1:

orestisfl (approved):
Great!

Last chore:
1. Add a line in https://github.com/i3/i3/blob/next/RELEASE-NOTES-next that describes the change
2. Squash all commits into one

mschuwalow:
> Great!
> 
> Last chore:
> 
> 1. Add a line in https://github.com/i3/i3/blob/next/RELEASE-NOTES-next that describes the change
> 2. Squash all commits into one

Done. Thanks for guiding me through the process :smile:

RELEASE-NOTES-next
```diff
@@ -28,6 +28,7 @@ working. Please reach out to us in that case!
   • mention rofi in default config file
   • i3-input: add different exit codes for when i3-input fails
   • allow ppt values in move direction and move position commands
+  • i3bar: add “nonprimary” output option
```
orestisfl, line 32: Changed the release notes a bit, this is more consistent with how we write them.

orestisfl (commented):
Thanks!

PR #4197, opened 2020-09-15
Update README.md

zero77:
This quickly shows all package versions across different distros at a glance.


orestisfl (approved):
Sounds good :)

PR #4208, opened 2020-09-23
Allow splitting into any kind of layout

### LINKED ISSUE #3001: split + layout stacked / tabbed creates redundant containers

### LINKED ISSUE #4207: Support for `split tabbed|stacked` command in addition to `split h|v|t`
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

There are `split h` and `split v` commands which create nested split frame of desired direction. However there's no way to create tabbed (and, for consistency, stacking) subframes this way.

## Desired Behavior

An ability to issue `split tabbed` command so a tabbed nested frame is created right away. Currently this can be emulated by issuing e.g. `split v` (which creates a subframe) followed by `layout tabbed` (which changes its type to tabbed), but that's not entirely convenient.

## Environment

<pre>
i3 version: 
Binary i3 version:  4.18.2 (2020-07-26) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.18.2 (2020-07-26) (pid 24354)bort…)
Loaded i3 config: /home/marakasov/.config/i3/config (Last modified: Wed Sep 23 03:28:31 2020, 72769 seconds ago)
</pre>

<pre>
- ~~Linux~~ Distribution & Version: FreeBSD 12.1
</pre>

AMDmi3:
This should solve #4207, I've successfully tested this as a patch to 4.18.2.

This change introduces `tabbed`, `stacking`, `stacked` modes to `split` command in a backward compatible way (`t` still works as `toggle`).

Additionaly, this changes behavior which replaces the layout of current container (instead of creating a nested container) if the container only has one child: while before it only worked for splits, now it works for all layouts which is IMO more consistent.

Fixes #4207 
Closes #3001

### DIFF
--- a/include/commands.h
+++ b/include/commands.h
@@ -159,7 +159,7 @@ void cmd_floating(I3_CMD, const char *floating_mode);
 void cmd_move_workspace_to_output(I3_CMD, const char *name);
 
 /**
- * Implementation of 'split v|h|t|vertical|horizontal|toggle'.
+ * Implementation of 'split v|h|t|vertical|horizontal|toggle|tabbed|stacked|stacking'.
  *
  */
 void cmd_split(I3_CMD, const char *direction);
--- a/include/tree.h
+++ b/include/tree.h
@@ -33,11 +33,11 @@ void tree_init(xcb_get_geometry_reply_t *geometry);
 Con *tree_open_con(Con *con, i3Window *window);
 
 /**
- * Splits (horizontally or vertically) the given container by creating a new
+ * Splits (using the given layout) the given container by creating a new
  * container which contains the old one and the future ones.
  *
  */
-void tree_split(Con *con, orientation_t orientation);
+void tree_split(Con *con, layout_t layout);
 
 /**
  * Moves focus one level up. Returns true if focus changed.
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -202,9 +202,9 @@ state STICKY:
   action = 'enable', 'disable', 'toggle'
       -> call cmd_sticky($action)
 
-# split v|h|t|vertical|horizontal|toggle
+# split v|h|t|vertical|horizontal|toggle|tabbed|stacked|stacking
 state SPLIT:
-  direction = 'horizontal', 'vertical', 'toggle', 'v', 'h', 't'
+  direction = 'horizontal', 'vertical', 'toggle', 'tabbed', 'stacked', 'stacking', 'v', 'h', 't'
       -> call cmd_split($direction)
 
 # floating enable|disable|toggle
--- a/src/commands.c
+++ b/src/commands.c
@@ -1127,14 +1127,14 @@ void cmd_move_workspace_to_output(I3_CMD, const char *name) {
 }
 
 /*
- * Implementation of 'split v|h|t|vertical|horizontal|toggle'.
+ * Implementation of 'split v|h|t|vertical|horizontal|toggle|tabbed|stacked|stacking'.
  *
  */
 void cmd_split(I3_CMD, const char *direction) {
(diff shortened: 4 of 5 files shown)


AMDmi3:
There seem to be a broken test related to 

> Additionaly, this changes behavior which replaces the layout of current container (instead of creating a nested container) if the container only has one child: while before it only worked for splits, now it works for all layouts which is IMO more consistent.

Can revert the bahavior or fix the test, depending on which behavior is desired.

orestisfl:
> There seem to be a broken test related to
> 
> > Additionaly, this changes behavior which replaces the layout of current container (instead of creating a nested container) if the container only has one child: while before it only worked for splits, now it works for all layouts which is IMO more consistent.
> 
> Can revert the bahavior or fix the test, depending on which behavior is desired.

So, the behavior of `layout stacked` is affected? This is certainly not desirable.

AMDmi3:
> So, the behavior of `layout stacked` is affected?

It is not, it only affects splits.

orestisfl:
Sounds good, this will also fix https://github.com/i3/i3/issues/3001.

Can you fix the test and include new ones that test the new behavior please?

AMDmi3:
Sorry, short on time right away, will do this next week

wlhlm:
Thanks for taking a stab at implementing this! :heart: This fixes one of the annoyances I've had with i3's windowing.

<details><summary>What follows is a patch that is better handled in a separate PR</summary>
Here is a patch that draws indicators for the new splits (as it already does for horizontal and vertical). For stacked, it draws the indicator over the top border, mirroring the vertical split indicator. I don't this think will be confused with the vertical split as that will always split at the bottom (as far as I know), opposite of where the stacked split indicator would be drawn. The tabbed indicator also draws over the top border, but only for half the width, sort of mimicking a tab.

```diff
diff --git a/src/x.c b/src/x.c
index def79ef2..5348764c 100644
--- a/src/x.c
+++ b/src/x.c
@@ -588,6 +588,12 @@ void x_draw_decoration(Con *con) {
             } else if (p->parent_layout == L_SPLITV) {
                 draw_util_rectangle(&(con->frame_buffer), p->color->indicator,
                                     br.x, r->height + (br.height + br.y), r->width + br.width, -(br.height + br.y));
+            } else if (p->parent_layout == L_STACKED) {
+                draw_util_rectangle(&(con->frame_buffer), p->color->indicator,
+                                    br.x, br.y, r->width + br.width, br.height);
+            } else if (p->parent_layout == L_TABBED) {
+                draw_util_rectangle(&(con->frame_buffer), p->color->indicator,
+                                    br.x, br.y, (r->width + br.width) / 2, br.height);
             }
         }
     }

```

Here is a preview with increased border size. Tabbed split is on the left, and stacked on the right:

Screenshot-2020-10-26-110024
</details>

Airblader:
@wlhlm Your patch seems to fit issue #1067 more than this PR.

wlhlm:
Huh, I totally forgot to check if this has been discussed before. Though I'm not surprised that it has.

lukasfink1:
Has there been any progress been made on this yet?

PR #4214, opened 2020-10-06
Add coordinates relative to the current output in i3bar click events

burik666:
Currently i3bar click events provide x and y coordinates relative to all monitors.
I've added coordinates relative to the current output.

```
+-----------+-----------+
|           |   i3bar   |
|           +-----------+
|   HDMI-0  |    DP-0   |
| 1920x1080 | 2560x1080 |
+-----------+-----------+
```
<details>
  <summary>xrandr output</summary>

```
Screen 0: minimum 8 x 8, current 4480 x 1080, maximum 32767 x 32767
DVI-D-0 disconnected (normal left inverted right x axis y axis)
HDMI-0 connected 1920x1080+0+0 (normal left inverted right x axis y axis) 476mm x 268mm
   1920x1080     60.00*+  59.94    50.00    60.05    60.00    50.04  
   1680x1050     59.95  
   1440x900      59.89  
   1360x765      60.00  
   1280x1024     75.02    60.02  
   1280x800      59.81  
   1280x720      60.00    59.94    50.00  
   1152x864      60.00  
   1024x768      75.03    70.07    60.00  
   800x600       75.00    72.19    60.32    56.25  
   720x576       50.00  
   720x480       59.94  
   640x480       75.00    72.81    59.94    59.93  
DP-0 connected primary 2560x1080+1920+0 (normal left inverted right x axis y axis) 798mm x 334mm
   2560x1080     59.98*+  74.99    50.00  
   1920x1080     74.91    60.00    59.94    50.00  
   1680x1050     59.95  
   1600x900      60.00  
   1280x1024     75.02    60.02  
   1280x800      59.81  
   1280x720      60.00    59.94    50.00  
   1152x864      59.96  
   1024x768      75.03    60.00  
   800x600       75.00    60.32  
   720x576       50.00  
   720x480       59.94  
   640x480       75.00    59.94    59.93  
DP-1 disconnected (normal left inverted right x axis y axis)
```
</details>

When you click in the top right corner of the DP-0, i3bar will provide something like this:
```json
{
  "x": 4480,
  "y": 10,
  "output_x": 2560,
  "output_y": 10,
}
```
This is useful for creating a `rofi` menu or something else.
`rofi -show run -location 1 -xoffset ${I3_OUTPUT_X} -yoffset ${I3_OUTPUT_Y}`

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -28,6 +28,7 @@ working. Please reach out to us in that case!
   • mention rofi in default config file
   • i3-input: add different exit codes for when i3-input fails
   • allow ppt values in move direction and move position commands
+  • i3bar: add coordinates relative to the current output in i3bar click events
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
--- a/docs/i3bar-protocol
+++ b/docs/i3bar-protocol
@@ -260,6 +260,8 @@ button::
 relative_x, relative_y::
     Coordinates where the click occurred, with respect to the top left corner
     of the block
+output_x, output_y::
+    Coordinates relative to the current output where the click occurred
 width, height::
     Width and height (in px) of the block
 modifiers::
@@ -273,10 +275,12 @@ modifiers::
  "instance": "eth0",
  "button": 1,
  "modifiers": ["Shift", "Mod1"],
- "x": 1320,
+ "x": 1925,
  "y": 1400,
  "relative_x": 12,
  "relative_y": 8,
+ "output_x": 5,
+ "output_y": 1400,
  "width": 50,
  "height": 22
 }
--- a/i3bar/include/child.h
+++ b/i3bar/include/child.h
@@ -85,4 +85,4 @@ bool child_want_click_events(void);
  * Generates a click event, if enabled.
  *
  */
-void send_block_clicked(int button, const char *name, const char *instance, int x, int y, int x_rel, int y_rel, int width, int height, int mods);
+void send_block_clicked(int button, const char *name, const char *instance, int x, int y, int x_rel, int y_rel, int out_x, int out_y, int width, int height, int mods);
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -634,7 +634,7 @@ static void child_click_events_initialize(void) {
  * Generates a click event, if enabled.
  *
  */
-void send_block_clicked(int button, const char *name, const char *instance, int x, int y, int x_rel, int y_rel, int width, int height, int mods) {
+void send_block_clicked(int button, const char *name, const char *instance, int x, int y, int x_rel, int y_rel, int out_x, int out_y, int width, int height, int mods) {
     if (!child.click_events) {
         return;
     }
@@ -685,6 +685,12 @@ void send_block_clicked(int button, const char *name, const char *instance, int 
 
     ystr("relative_y");
     yajl_gen_integer(gen, y_rel);
+
+    ystr("output_x");
+    yajl_gen_integer(gen, out_x);
+
+    ystr("output_y");
+    yajl_gen_integer(gen, out_y);
 
     ystr("width");
     yajl_gen_integer(gen, width);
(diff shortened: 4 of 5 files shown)


orestisfl:
So, isn't this something that can be calculated by the script that calls rofi? Plus, can't rofi be called with a global position?

burik666:
@orestisfl, 

These coordinates are difficult to calculate by a script.
<details><summary>output_xy.sh</summary>

```bash
#!/usr/bin/env bash
# from https://superuser.com/a/1238384

OFFSET_RE="\+([-0-9]+)\+([-0-9]+)"

# Get the window position
eval "$(xdotool getmouselocation --shell)"

# Loop through each screen and compare the offset with the window
# coordinates.
monitor_index=0
while read name width height xoff yoff
do
    if [ "${X}" -ge "$xoff" \
      -a "${Y}" -ge "$yoff" \
      -a "${X}" -lt "$(($xoff+$width))" \
      -a "${Y}" -lt "$(($yoff+$height))" ]
    then
        monitor=$name
        OUTPUT_X=$((X - xoff))
        OUTPUT_Y=$((Y - yoff))
        break
    fi
    ((monitor_index++))
done < <(xrandr | grep -w connected |
    sed -r "s/^([^ ]*).*\b([-0-9]+)x([-0-9]+)$OFFSET_RE.*$/\1 \2 \3 \4 \5/" |
    sort -nk4,5)

# If we found a monitor, echo it out, otherwise print an error.
if [ ! -z "$monitor" ]
then
    echo $OUTPUT_X $OUTPUT_Y
    exit 0
else
    echo "Couldn't find any monitor for the current window." >&2
    exit 1
fi
```

</details>


rofi only supports positioning relative to the current screen. https://github.com/davatorium/rofi/wiki/WindowPosition

orestisfl:
@Airblader if you agree that this is something that is hard to work around & worthy to implement, I will merge it.

I would however prefer `relative_{x,y}` than `output_{x,y}`

burik666:
@orestisfl,
`relative_{x,y}` already used.

https://i3wm.org/docs/i3bar-protocol.html#_click_events
```
relative_x, relative_y
    Coordinates where the click occurred, with respect to the top left corner of the block
```

burik666:
@Airblader, what do you think about this?

RELEASE-NOTES-next
```diff
@@ -28,6 +28,7 @@ working. Please reach out to us in that case!
   • mention rofi in default config file
   • i3-input: add different exit codes for when i3-input fails
   • allow ppt values in move direction and move position commands
+  • i3bar: add coordinates relative to the crurrent output in i3bar click events
```
orestisfl:
```suggestion
  • i3bar: add coordinates relative to the current output in i3bar click events
```

orestisfl (changes requested):
I'd say we can go ahead

I found a typo.

Have you tested the exact correctness of the coordinates? I'd like to run it a bit myself as well, just making sure that you get exactly the correct number.

orestisfl:
Other things: Do not introduce merge commits in PRs, use `rebase` instead. One simple way to fix this right now is `git reset --hard origin/next && git cherry-pick 3a9ac8c6fa50ebf6fde67209d6fd9d7bebd6adfb`.

Include the extra information in the PR description inside the git commit message.

orestisfl:
Oh and the typo is in the commit message as well :smiley:

burik666:
@orestisfl,

Yes, I tested with different xrandr configurations.
This works great.

Airblader:
> @Airblader, what do you think about this?

I haven't had a chance to look at the PR from the code-level, but completely trust @orestisfl in that regard. As for the feature, I think it makes sense, the naming is fine (given that relative_ is already used), so let's get it shipped.

That said, @orestisfl , I think we may want to be a bit selective with which features we merge into stable now so we don't merge things right before the next release without having some testing period. I think this one is probably fine as (apart from potential crashes) it shouldn't affect any existing things, but you can make that call.

orestisfl (approved):
Great!

Last thing is to just copy-paste the PR description code in the commit message. The reason is to not have commit context (e.g. when using `git blame`) depend on GitHub.

burik666:
@orestisfl, thanks a lot for your help with this PR.

orestisfl:
Thank you for your contribution!

PR #4226, opened 2020-10-22
i3bar: fix render empty blocks with min_width

burik666:
This change allows render blocks of any width.

```json
{
  "full_text": "",
  "min_width": 100,
  "background": "#00ff00"
}

{
  "full_text": "",
  "min_width": 1,
  "background": "#0000ff"
}
```

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -50,3 +50,4 @@ working. Please reach out to us in that case!
   • fix Xorg memory leak with i3bar
   • fix named workspace assignments on output changes
   • fix named workspace assignment precedence on workspace renames
+  • i3bar: fix render empty blocks with min_width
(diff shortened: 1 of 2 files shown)


6A61736F6E206E61646572:
I think I have come across this before but don't have access to a machine at the moment to check. Do you have any before/after screenshots?

burik666:
@ohk2kt3t4, 
Before that, similar blocks not rendered.

If you specify `full_text:" "` then the block width will always be greater than the width of one character, even if `"min_width": 1`.

Airblader:
Thanks for the PR. However, this isn't something that needs fixed because the current behavior is intentional and documented:

> If full_text is an empty string, the block will be skipped.
> — https://i3wm.org/docs/i3bar-protocol.html

burik666:
@Airblader, May change this behavior to:
>  If full_text is an empty string and min_width is not set, the block will be skipped.

?

Airblader:
I haven't yet seen an argument for why that should be done. Please raise a feature request where you explain what you want to achieve so we can discuss what change that would require and whether we'd accept such a change. The proposed fix here, however, would be breaking.

PR #4230, opened 2020-10-26
Implement context-sensitive tab-clicking behavior

### LINKED ISSUE #4225: Discrepancy in focusing split tab by mouse versus by keyboard
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

Not sure if this is a bug or a feature, so I'm filing it as a feature.

Consider a layout like `T[H[app1 app2] app3]` with `app3` having focus, and `app1` having inactive focus within the `H`.

`focus left` switches to the first tab and focuses `app1`.

If instead the mouse is used to change the tab, the entire `H` container and both its children become focused on click.

## Desired Behavior

It would be nice / save a click if `app1` were focused in this case as well, since selecting the parent container is likely not to be what the user is trying to do.

## Environment

```
Binary i3 version:  4.18.2-2-g5e15527e (2020-07-26, branch "gaps") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.18.2-2-g5e15527e (2020-07-26, branch "gaps") (pid 14064)
Loaded i3 config: /home/tbrindus/.config/i3/config (Last modified: Sun 04 Oct 2020 08:26:18 PM EDT, 1385171 seconds ago)
```

Non-gaps i3 behaves the same here.

Xyene:
This commit makes it so that when clicking on a tab for the first time,
its inactive-focused child is focused. This matches i3 behavior when
scrolling on tabs.

A second click on the tab titlebar focuses the tab container itself.

Subsequent clicks cycle between these two states.

Closes #4225.

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -13,6 +13,7 @@ strongly encouraged to upgrade.
 
   • i3-nagbar: position on focused monitor by default
   • i3-nagbar: add option to position on primary monitor
+  • alternate focusing tab/stack children-parent containers by clicking on their titlebars
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
--- a/src/click.c
+++ b/src/click.c
@@ -218,8 +218,20 @@ static void route_click(Con *con, xcb_button_press_event_t *event, const bool mo
         goto done;
     }
 
-    /* 2: focus this con. */
-    con_activate(con);
+    /* 2: focus this con or one of its children. */
+    Con *con_to_focus = con;
+    if (in_stacked && dest == CLICK_DECORATION) {
+        /* If the container is a tab/stacked container and the click happened
+         * on a tab, switch to the tab. If the tab contents were already
+         * focused, focus the tab container itself. If the tab container was
+         * already focused, cycle back to focusing the tab contents. */
+        if (was_focused || !con_has_parent(focused, con)) {
+            while (!TAILQ_EMPTY(&(con_to_focus->focus_head))) {
+                con_to_focus = TAILQ_FIRST(&(con_to_focus->focus_head));
+            }
+        }
+    }
+    con_activate(con_to_focus);
 
     /* 3: For floating containers, we also want to raise them on click.
      * We will skip handling events on floating cons in fullscreen mode */


src/click.c
```diff
@@ -218,8 +218,18 @@ static void route_click(Con *con, xcb_button_press_event_t *event, const bool mo
...
-    con_activate(con);
+    /* 2: focus this con or one of its children. */
+    Con *con_to_focus = con;
+    if (in_stacked && dest == CLICK_DECORATION) {
+        /* If the container is a tab/stacked container and the click happened
+         * on a tab, switch to the tab. If the tab contents were already
+         * focused, focus the tab container itself. If the tab container was
+         * already focused, cycle back to focusing the tab contents. */
+        if (was_focused || !con_has_parent(focused, con)) {
+            con_to_focus = con_descend_focused(con_descend_direction(con, D_LEFT));
```
orestisfl: Unfortunately, that's not correct but it shouldn't be too hard. You will need to write a custom `con_descend_focused` that does not stop at `focused`.

For a case where your approach fails:
`T[ H[ a* b ] ]` after two clicks becomes `T[ H[ a b* ] ]`

orestisfl (changes requested):

src/click.c
Xyene: Ah, I figured something was wonky about that -- I've switched to a custom version now.

Xyene:
@orestisfl gentle ping -- is this something that would be wanted upstream, and if so, is there anything else I should do on my end?

orestisfl:
Sorry for the wait.

I am in favor of the change.

Can you also update this file please: https://github.com/i3/i3/blob/next/RELEASE-NOTES-next

Airblader:
@orestisfl Same here, I'd rather wait with this until after the release. The change is small, but focus handling has been a cause of many bugs in the past, so giving this a bit of time on next before going into main would be good.

Xyene:
Sounds good! I assume I should wait until after 4.19 is released to update https://github.com/i3/i3/blob/next/RELEASE-NOTES-next?

Xyene:
Now that 4.19 has been released, I've gone ahead and updated https://github.com/i3/i3/blob/next/RELEASE-NOTES-next.

RELEASE-NOTES-next
```diff
@@ -11,7 +11,7 @@ strongly encouraged to upgrade.
  │ Changes in i3 v4.20        │
  └────────────────────────────┘
 
-  • placeholder
+  • allow focusing split tab containers by double-clicking their titlebars
```
orestisfl: Does this sound better?

```suggestion
  • alternate focusing tab/stack children-parent containers by clicking on their titlebars
```

orestisfl (approved):
Can you please rebase so that we can merge this?

Also potential suggestion

RELEASE-NOTES-next
Xyene: Yes, this sounds better to me too -- thanks.

Xyene:
I've rebased and updated RELEASE-NOTES-next.

orestisfl (approved):
Thank you!

PR #4231, opened 2020-10-27
i3bar: Add default bar_id

orestisfl:
Instead of erroring, request the list of bar configs from i3 and use the
first one.

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -14,6 +14,7 @@ strongly encouraged to upgrade.
   • i3-nagbar: position on focused monitor by default
   • i3-nagbar: add option to position on primary monitor
   • alternate focusing tab/stack children-parent containers by clicking on their titlebars
+  • i3bar: use first bar config by default
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
--- a/i3bar/src/main.c
+++ b/i3bar/src/main.c
@@ -56,9 +56,9 @@ static char *expand_path(char *path) {
 }
 
 static void print_usage(char *elf_name) {
-    printf("Usage: %s -b bar_id [-s sock_path] [-t] [-h] [-v] [-V]\n", elf_name);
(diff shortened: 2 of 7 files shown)


i3bar/src/main.c
```diff
@@ -133,13 +133,6 @@ int main(int argc, char **argv) {
         }
     }
 
-    if (!config.bar_id) {
-        /* TODO: maybe we want -f which will automatically ask i3 for the first
-         * configured bar (and error out if there are too many)? */
```
Airblader, line 138: I tend to think that erroring if multiple bar blocks are specified but no bar_id was given is better because we don't really want the order bar blocks in the config to matter, right?

orestisfl: I prefer the simpler approach (which is also more useful) since users are probably not meant to launch i3bar from the command line. The order of the bar blocks matter in the sense that they define the bar-ids.

Airblader: I disagree with it being "more useful", but it is simpler and not something the average user will be doing. The log message should make things clear. So fine by me.

> The order of the bar blocks matter in the sense that they define the bar-ids.

Only if you don't define the ID yourself, and even then that ID has no implication. ;-)

RELEASE-NOTES-next
```diff
@@ -14,6 +14,7 @@ strongly encouraged to upgrade.
   • i3-nagbar: position on focused monitor by default
   • i3-nagbar: add option to position on primary monitor
   • alternate focusing tab/stack children-parent containers by clicking on their titlebars
+  • i3bar: when starting from commandline, use first bar config by default
```
Airblader: The phrasing is not entirely correct, because we don't actually detect where it is being launched from. It's just the use-case you have in mind. ;-)

Airblader (approved):

RELEASE-NOTES-next
orestisfl: Up. Also made changes to the manpage and `--help`

Airblader (approved):

Issue #4157, opened 2020-07-22
semicolon in i3-msg commands add trailing space to new marks


budRich:
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

from a terminal:  
```
$ i3-msg "mark itsamark ;"
[{"success":true}]
$ i3-msg -t get_marks
["itsamark "]
$ i3-msg "mark itsamark;"
[{"success":true}]
$ i3-msg -t get_marks
["itsamark"]
$ i3-msg "mark itsamark ; [title=__focused__] focus"
[{"success":true},{"success":true}]
$ i3-msg -t get_marks
["itsamark "]
```

## Expected Behavior
```
$ i3-msg "mark itsamark ;"
[{"success":true}]
$ i3-msg -t get_marks
["itsamark"]
$ i3-msg "mark itsamark ; [title=__focused__] focus"
[{"success":true},{"success":true}]
$ i3-msg -t get_marks
["itsamark"]
```

## Reproduction Instructions
enter the following in a terminal  (assuming no other marks are set)
```
$ i3-msg "mark itsamark ;"
[{"success":true}]
$ i3-msg -t get_marks
["itsamark "]
$ i3-msg "mark itsamark;"
[{"success":true}]
$ i3-msg -t get_marks
["itsamark"]
$ i3-msg "mark itsamark ; [title=__focused__] focus"
[{"success":true},{"success":true}]
$ i3-msg -t get_marks
["itsamark "]
```

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
Binary i3 version:  4.18.1 (2020-04-22) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.18.1 (2020-04-22) (pid 74719)bort…)
Loaded i3 config: /home/bud/.config/i3/config (Last modified: Tue 21 Jul 2020 03:05:42 PM CEST, 85579 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
</pre>

<pre>
- Linux Distribution & Version: Arch
- Are you using a compositor (e.g., xcompmgr or compton): No
</pre>

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

stapelberg:
[...]

PR #4237, opened 2020-10-28
USERGUIDE: i3-msg and whitespace

budRich:
#4157

---

maybe i3-msg should have its own subsection in the userguide?


stapelberg:
Thanks for sending this PR.

I’ll think some more about where/how to best integrate this.

i3-msg in general should not get too much attention in the userguide, as it’s only an example program for testing the IPC interface. Maybe the manpage or docs/ipc are more appropriate, but we should add a link for sure.

Given timing, let’s push this until after the 4.19 release.

PR #4244, opened 2020-11-03
i3-nagbar: add possibility to open bar on focused monitor

### LINKED ISSUE #3692: i3-nagbar option (default?) to appear on _current_ monitor
## I'm submitting a…
<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
Use the first monitor or the one defined as _primary_ with XRANDR

## Desired Behavior
Dynamically get the current monitor and display it there.
The most similar issue I've found is #1405, but it is not the same.

Unrelated extra: I would like a _"replace"_ flag to replace previous "bars", so if I call it several times it will not creates lot of "bars"

## Environment

<pre>
i3 version:   4.14.1
</pre>

<pre>
- Linux Distribution & Version:  ElementaryOS Juno
- compositor? I don't think so
</pre>

ekarpp:
Closes #3692

i3-nagbar will open by default on the monitor with input focus and using the flag -p on the primary monitor.

### DIFF
--- a/i3-nagbar/main.c
+++ b/i3-nagbar/main.c
@@ -266,13 +266,9 @@ static int handle_expose(xcb_connection_t *conn, xcb_expose_event_t *event) {
 }
 
 /**
- * Return the position and size the i3-nagbar window should use.
- * This will be the primary output or a fallback if it cannot be determined.
+ * Tries to position the rectangle on the primary output.
  */
-static xcb_rectangle_t get_window_position(void) {
-    /* Default values if we cannot determine the primary output or its CRTC info. */
-    xcb_rectangle_t result = (xcb_rectangle_t){50, 50, 500, font.height + 2 * MSG_PADDING + BAR_BORDER};
-
+static void set_window_position_primary(xcb_rectangle_t *result) {
     xcb_randr_get_screen_resources_current_cookie_t rcookie = xcb_randr_get_screen_resources_current(conn, root);
     xcb_randr_get_output_primary_cookie_t pcookie = xcb_randr_get_output_primary(conn, root);
 
@@ -314,14 +310,61 @@ static xcb_rectangle_t get_window_position(void) {
         goto free_resources;
     }
 
-    result.x = crtc->x;
-    result.y = crtc->y;
+    result->x = crtc->x;
+    result->y = crtc->y;
     goto free_resources;
 
 free_resources:
     free(res);
     free(primary);
-    return result;
+}
+
+/**
+ * Tries to position the rectangle on the output with input focus.
+ * If unsuccessful, try to position on primary output.
+ */
+static void set_window_position_focus(xcb_rectangle_t *result) {
+    bool success = false;
+    xcb_get_input_focus_reply_t *input_focus = NULL;
+    xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    /* To avoid the input window disappearing while determining its position */
+    xcb_grab_server(conn);
+
+    input_focus = xcb_get_input_focus_reply(conn, xcb_get_input_focus(conn), NULL);
+    if (input_focus == NULL || input_focus->focus == XCB_NONE) {
+        LOG("Failed to receive the current input focus or no window has the input focus right now.\n");
+        goto free_resources;
+    }
+
+    geometry = xcb_get_geometry_reply(conn, xcb_get_geometry(conn, input_focus->focus), NULL);
+    if (geometry == NULL) {
+        LOG("Failed to received window geometry.\n");
+        goto free_resources;
+    }
+
+    coordinates = xcb_translate_coordinates_reply(
+        conn, xcb_translate_coordinates(conn, input_focus->focus, root, geometry->x, geometry->y), NULL);
+    if (coordinates == NULL) {
+        LOG("Failed to translate coordinates.\n");
+        goto free_resources;
+    }
+
+    LOG("Found current focus at x = %i / y = %i.\n", coordinates->dst_x, coordinates->dst_y);
+    result->x = coordinates->dst_x;
+    result->y = coordinates->dst_y;
+    success = true;
+
+free_resources:
+    xcb_ungrab_server(conn);
+    free(input_focus);
+    free(coordinates);
+    free(geometry);
+    if (!success) {
+        LOG("Could not position on focused output, trying to position on primary output.\n");
+        set_window_position_primary(result);
+    }
 }
 
 int main(int argc, char *argv[]) {
@@ -360,6 +403,7 @@ int main(int argc, char *argv[]) {
 
     argv0 = argv[0];
 
+    bool position_on_primary = false;
     char *pattern = sstrdup("pango:monospace 8");
     int o, option_index = 0;
     enum { TYPE_ERROR = 0,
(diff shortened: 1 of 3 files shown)


i3-nagbar/main.c
```diff
@@ -269,10 +269,43 @@ static int handle_expose(xcb_connection_t *conn, xcb_expose_event_t *event) {
...
-static xcb_rectangle_t get_window_position(void) {
+static xcb_rectangle_t get_window_position(bool position_on_focused) {
     /* Default values if we cannot determine the primary output or its CRTC info. */
     xcb_rectangle_t result = (xcb_rectangle_t){50, 50, 500, font.height + 2 * MSG_PADDING + BAR_BORDER};
 
+    xcb_get_input_focus_reply_t *input_focus = NULL;
+    xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    if (position_on_focused) {
```
orestisfl: I feel that this will be more readable if we split this in two different functions

orestisfl (changes requested):

i3-nagbar/main.c
```diff
@@ -314,14 +310,55 @@ static xcb_rectangle_t get_window_position(void) {
...
+        conn, xcb_translate_coordinates(conn, input_focus->focus, root, geometry->x, geometry->y), NULL);
+    if (coordinates == NULL) {
+        LOG("Failed to translate coordinates.\n");
+        goto free_resources;
+    }
+
+    LOG("Found current focus at x = %i / y = %i.\n", coordinates->dst_x, coordinates->dst_y);
+    result->x = coordinates->dst_x;
+    result->y = coordinates->dst_y;
+    goto free_resources;
```
orestisfl:
```suggestion
```

i3-nagbar/main.c
```diff
@@ -314,14 +310,55 @@ static xcb_rectangle_t get_window_position(void) {
...
+
+free_resources:
+    xcb_ungrab_server(conn);
+    free(input_focus);
+    free(coordinates);
+    free(geometry);
```
orestisfl, line 363: I wonder if it is preferential to fall-back to primary if we fail here. WDYT?

orestisfl (changes requested):

i3-nagbar/main.c
ekarpp: I think this is a nice idea. I suppose when the primary output is preferred, the output with input focus could also be used as a fallback?

orestisfl: Not really important but maybe no since `-p` is an opt-in option

i3-nagbar/main.c
```diff
@@ -464,7 +514,14 @@ int main(int argc, char *argv[]) {
...
 
-    xcb_rectangle_t win_pos = get_window_position();
+    /* Default values if we cannot determine the preferred window position. */
+    xcb_rectangle_t win_pos = (xcb_rectangle_t){50, 50, 500, font.height + 2 * MSG_PADDING + BAR_BORDER};
+    if (position_on_primary) {
+        set_window_position_primary(&win_pos);
+    } else {
+        set_window_position_focus(&win_pos);
+    }
+
```
orestisfl, line 522:
```suggestion
```

orestisfl: Travis failed because of this

i3-nagbar/main.c
```diff
@@ -314,14 +310,63 @@ static xcb_rectangle_t get_window_position(void) {
...
     free(primary);
-    return result;
+}
+
+/**
+ * Tries to position the rectangle on the output with input focus.
+ * If unsuccessful, try to position on primary output.
+ */
+static void set_window_position_focus(xcb_rectangle_t *result) {
+    bool error = false;
```
orestisfl:
```suggestion
    bool success = false;
```
Instead of having multiple `error` assignments, just have a `success` one and ...

i3-nagbar/main.c
```diff
@@ -314,14 +310,63 @@ static xcb_rectangle_t get_window_position(void) {
...
+        conn, xcb_translate_coordinates(conn, input_focus->focus, root, geometry->x, geometry->y), NULL);
+    if (coordinates == NULL) {
+        LOG("Failed to translate coordinates.\n");
+        error = true;
+        goto free_resources;
+    }
+
+    LOG("Found current focus at x = %i / y = %i.\n", coordinates->dst_x, coordinates->dst_y);
+    result->x = coordinates->dst_x;
+    result->y = coordinates->dst_y;
```
orestisfl, line 356:
```suggestion
    result->y = coordinates->dst_y;
    success = true;
```

i3-nagbar/main.c
```diff
@@ -314,14 +310,63 @@ static xcb_rectangle_t get_window_position(void) {
...
+    xcb_get_geometry_reply_t *geometry = NULL;
+    xcb_translate_coordinates_reply_t *coordinates = NULL;
+
+    /* To avoid the input window disappearing while determining its position */
+    xcb_grab_server(conn);
+
+    input_focus = xcb_get_input_focus_reply(conn, xcb_get_input_focus(conn), NULL);
+    if (input_focus == NULL || input_focus->focus == XCB_NONE) {
+        LOG("Failed to receive the current input focus or no window has the input focus right now.\n");
+        error = true;
```
orestisfl:
```suggestion
```

i3-nagbar/main.c
```diff
@@ -314,14 +310,63 @@ static xcb_rectangle_t get_window_position(void) {
...
+        goto free_resources;
+    }
+
+    geometry = xcb_get_geometry_reply(conn, xcb_get_geometry(conn, input_focus->focus), NULL);
+    if (geometry == NULL) {
+        LOG("Failed to received window geometry.\n");
+        error = true;
```
orestisfl:
```suggestion
```

i3-nagbar/main.c
```diff
@@ -314,14 +310,63 @@ static xcb_rectangle_t get_window_position(void) {
...
+        goto free_resources;
+    }
+
+    coordinates = xcb_translate_coordinates_reply(
+        conn, xcb_translate_coordinates(conn, input_focus->focus, root, geometry->x, geometry->y), NULL);
+    if (coordinates == NULL) {
+        LOG("Failed to translate coordinates.\n");
+        error = true;
```
orestisfl:
```suggestion
```

orestisfl (changes requested):

ekarpp:
I guess this Travis failure is unrelated to the changes?

orestisfl:
It's a formatting error, check my comments

ekarpp:
Sorry, I should have clarified, I meant the newer Travis failure. I already pushed a change based on your comments. The new failure was caused by (I think?) a parse error in one of the tests.

orestisfl (approved):
LGTM. @Airblader any input? Is there any shorter sequence of X calls that can get us the information needed here?

Airblader:
> Is there any shorter sequence of X calls that can get us the information needed here?

I don't think so, this seems fine, and frankly is less work than the primary output which requires RandR. 

I actually think we should make this the default, not a flag, though. I also think we could just remove the primary placement, though I'd keep it behind the flag for now just in case someone prefers it. We could open an issue for removing it in the release after, and maybe when using the flag print a warning to stderr to ask users to provide feedback if they want to keep using that?

Airblader:
I need to read more carefully, this is already the new default, so ignore that part. As for deprecating that primary flag we can also handle this outside of the PR. Sorry :-)

orestisfl:
@ekarpp can you rebase with next and also copy the extra information in the commit message:

```
Closes #3692

i3-nagbar will open by default on the monitor with input focus and using the flag -p on the primary monitor.
```

ekarpp:
Rebased, thanks for helping me with this PR.

orestisfl:
Thanks!

PR #4252, opened 2020-11-11
create_workspace_on_output: Prevent duplicate workspace nums

### LINKED ISSUE #4238: Duplicate workspace numbers
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

It is possible to create two i3 workspaces with the same number, using an `i3-go` script and a secondary display.

## Expected Behavior

It should not be possible to create two workspaces with the same number, no matter how you use `i3-ipc`.

## Reproduction Instructions

I'm sorry I don't have an easier way to reproduce this. I use a script called [`i3-autoname-ws`](https://github.com/ajeetdsouza/i3-autoname-ws) that uses `i3-go` to update the names of workspaces with the icons of windows that are open in them.

If I connect another monitor and set it as a primary display while this script is running, I see two workspaces with the same number. I can verify this by querying the workspace tree using `i3-go` - there are indeed duplicate workspace numbers in the tree.

## Environment

```
i3 version: 

Binary i3 version:  4.18.2 (2020-07-26) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.18.2 (2020-07-26) (pid 1460)abort…)
Loaded i3 config: /home/ajeetdsouza/.config/i3/config (Last modified: Fri 23 Oct 2020 01:16:14 AM IST, 771297 seconds ago)

```

## Config file
- I don't think the config file is needed to reproduce the bug, but it can be found [here](https://github.com/ajeetdsouza/dotfiles/blob/master/i3/.config/i3/config).
- To run i3-autoname-ws:
  - Install Go
  - Clone and cd into [`i3-autoname-ws`](https://github.com/ajeetdsouza/i3-autoname-ws/)
  - `go run .`

```
```

orestisfl:
Fixes #4238

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -17,5 +17,4 @@ strongly encouraged to upgrade.
  │ Bugfixes                   │
  └────────────────────────────┘
 
-  • placeholder
-
+  • when initializing new outputs, avoid duplicating workspace numbers
--- a/src/workspace.c
+++ b/src/workspace.c
@@ -134,7 +134,7 @@ Con *workspace_get(const char *num) {
 
     /* We set workspace->num to the number if this workspace’s name begins with
      * a positive number. Otherwise it’s a named ws and num will be 1. */
-    const long parsed_num = ws_name_to_number(num);
+    const int parsed_num = ws_name_to_number(num);
 
     Con *output = get_assigned_output(num, parsed_num);
     /* if an assignment is not found, we create this workspace on the current output */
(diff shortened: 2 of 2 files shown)


Airblader (approved):
LGTM but I'd already view this as not important enough to push it in before the release.

orestisfl:
I'd say its a pretty trivial change to cause a bug but I don't mind

Airblader:
The devil's in the detail, and focus handling has traditionally caused many, many bugs already. This behavior existed for a long time, so I don't think we need to rush changing it.

orestisfl:
This is not about focus handling, its about duplicate workspaces. I can even separate the janitorial changes so that this change is only a single line fix.

Airblader:
I had the wrong issue in mind at first, but I'd still just not merge anything other than important bugfixes for now. It's not only about risking bugs, it's also about giving users a chance to experience the change and voice their opinion before releasing it.

ajeetdsouza:
@Airblader now that the release has been made, can this PR be merged?

Airblader:
Yes, @orestisfl just needs to rebase it and then we can merge it.

PR #4259, opened 2020-11-16
i3-sensible-terminal: prioritize terminals with good accessibility

### LINKED ISSUE #4256: enhancement - Adjust i3-sensible-terminal to Prioritize Accessible Terminals for Blind Users
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

Currently i3 prioritizes non-accessible terminals that do not function properly with orca and speech-dispatcher for blind users.

## Desired Behavior

This can be resolved by adjusting the prioritization of terminals in the i3-sensible-terminal file.
I have tested this with our blindarch distro and have confirmed it does work in our testing branch.

The attached file is updated to reflect the changes that would make this work better for blind users.
We simply change the order of some terminals that have orca support to be first in line.
Mate terminal is currently considered to be the most accessible.

## Environment

<pre>
i3 version: 
4.18.3

</pre>

<pre>
Arch Linux - latest version (building custom ISO)
no
</pre>

[i3-sensible-terminal.zip](https://github.com/i3/i3/files/5541911/i3-sensible-terminal.zip)

stapelberg:
We never did (and still do not) guarantee any order in which terminal emulators
are tried. Quoting from the i3-sensible-terminal man page:

> Please don’t complain about the order: If the user has any preference, they
> will have $TERMINAL set or modified their i3 configuration file.

This commit moves mate-terminal to the beginning of the list, which is
considered the most accessible terminal emulator among blind users.

fixes https://github.com/i3/i3/issues/4256

### DIFF
--- a/i3-sensible-terminal
+++ b/i3-sensible-terminal
@@ -8,7 +8,13 @@
 # We welcome patches that add distribution-specific mechanisms to find the
 # preferred terminal emulator. On Debian, there is the x-terminal-emulator
 # symlink for example.
-for terminal in "$TERMINAL" x-terminal-emulator urxvt rxvt termit terminator Eterm aterm uxterm xterm gnome-terminal roxterm xfce4-terminal termite lxterminal mate-terminal terminology st qterminal lilyterm tilix terminix konsole kitty guake tilda alacritty hyper; do
+#
+# Invariants:
+# 1. $TERMINAL must come first
+# 2. Distribution-specific mechanisms come next, e.g. x-terminal-emulator
+# 3. The terminal emulator with best accessibility comes first.
+# 4. No order is guaranteed/desired for the remaining terminal emulators.
+for terminal in "$TERMINAL" x-terminal-emulator mate-terminal gnome-terminal terminator xfce4-terminal urxvt rxvt termit Eterm aterm uxterm xterm roxterm termite lxterminal terminology st qterminal lilyterm tilix terminix konsole kitty guake tilda alacritty hyper; do
     if command -v "$terminal" > /dev/null 2>&1; then
         exec "$terminal" "$@"
     fi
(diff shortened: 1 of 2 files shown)


Airblader:
> We never did (and still do not) guarantee any order in which terminal emulators are tried.

I +1 this change for the a11y benefit, but I remember multiple PRs adding new emulators to the list where we asked to add them to the end of the list as to avoid this issue.

Airblader (approved):

stapelberg:
> I +1 this change for the a11y benefit, but I remember multiple PRs adding new emulators to the list where we asked to add them to the end of the list as to avoid this issue.

Yes, and adding to the end of the list is what I would want to keep doing. The two statements aren’t in conflict with each other, are they?

Airblader:
I understand that you're getting at "not guaranteeing it doesn't mean we can't try to keep it stable if we add something", but I think realistically we raised the appearance of a stable order, and a note inside i3-sensible-terminal is probably not something most users will ever (have) read.

I fully expect this change to cause "problems" for a number of people. And note inside a script or not, that is not usually something we like doing. That said, the benefit outweighs the cost; it's not often we get to do accessibility improvements in i3 with such little effort.

stapelberg:
> realistically we raised the appearance of a stable order

How do you mean? Are you saying we raised the appearance because of how i3-sensible-terminal behaves? Or is there some other place where we should have specified that there is no order?

> and a note inside i3-sensible-terminal

It’s a note inside the manpage, not inside the script :)

> I fully expect this change to cause "problems" for a number of people.

Yeah, people who haven’t configured their terminal might get a different installed one by default. I’m okay with that, especially as i3 shouldn’t be prescriptive regarding which terminal is best (or even just okay) :)

Airblader:
> How do you mean? Are you saying we raised the appearance because of how i3-sensible-terminal behaves?

Yes, we've kept it stable for so long that I think people rely on it. I know that when I first started using i3 I had multiple emulators installed and never bothered or even thought about replacing i3-sensible-terminal.

> I’m okay with that, especially as i3 shouldn’t be prescriptive regarding which terminal is best (or even just okay) :)

Then why do we prioritize accessible ones with this change…? ;-)

Anyway, I think this isn't worth spending a lot of time talking about given that we both agree on changing this. :-)

PR #4260, opened 2020-11-16
userguide: clarify home row / vi navigational key relation

stapelberg:
Before this commit, the userguide mentioned “compatibility with most keyboard
layouts”, but that seems to not have been the intention of vi author(s).

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -53,12 +53,23 @@ existing window (rotated displays).
 
 image:two_terminals.png[Two terminals]
 
-To move the focus between the two terminals, you can use the direction keys
-which you might know from the editor +vi+. However, in i3, your homerow is used
-for these keys (in +vi+, the keys are shifted to the left by one for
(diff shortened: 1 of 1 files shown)


docs/userguide
```diff
@@ -53,12 +53,23 @@ existing window (rotated displays).
...
+
+|===
+| `$mod+j` | left
+| `$mod+k` | down
+| `$mod+l` | up
+| `$mod+;` | right
+|===
+
+Note that this differs by one key from the popular text editor `vi`, which was
+https://twitter.com/hillelogram/status/1326600125569961991[developed on an
```
Airblader, line 69: Looks a bit odd to me that we link to Twitter from the userguide, though it's just anecdotal, I guess.

Airblader (approved):

docs/userguide
stapelberg: Yep, we can gladly exchange this link with a more stable / authoritative one when we come across a better one. Maybe the twitter link is okay, though. We’ll see :)

PR #4269, opened 2020-11-23
Use docdir for all docs when building (Fixes #4268)

### LINKED ISSUE #4268: Inconsistent use of docdir in meson.build
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior / Expected Behavior

When building with docs enabled, some docs are installed unconditionally to datadir/doc/i3 instead of to docdir/i3.

## Reproduction Instructions

Roughly speaking my Gentoo ebuild for i3-gaps configures i3-gaps (though this is reproducable with vanilla i3, too) like so:
`meson ../i3-4.19 -Ddocdir="/usr/share/doc/i3-gaps-4.19" -Ddocs=true`

Then after installation Portage produces this message:

```
 * The ebuild is installing to one or more unexpected paths:
 *
 *   /usr/share/doc/i3
 *
 * Please fix the ebuild to use correct FHS/Gentoo policy paths.
```

This caused by lines 646 and 659 which both read:
```
install_dir: join_paths(get_option('datadir'), 'doc', 'i3'),
```
... which can be corrected simply by using docdir as the install_dir rather than that join_paths call, as is done elsewhere in meson.build.

## Environment

<pre>
i3 version: 4.19
</pre>

(Rest of bug report template seems to be irrelevant for a build time issue)

ajakk:
Trivial build fix to use docdir for doc installation instead of datadir/doc/i3 for lib-i3test.html and lib-i3test-test.html.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -643,7 +643,7 @@ if get_option('docs')
       '@OUTPUT@',
     ],
     install: true,
-    install_dir: join_paths(get_option('datadir'), 'doc', 'i3'),
+    install_dir: docdir,
   )
 
   custom_target(
@@ -656,7 +656,7 @@ if get_option('docs')
       '@OUTPUT@',
     ],
     install: true,
-    install_dir: join_paths(get_option('datadir'), 'doc', 'i3'),
+    install_dir: docdir,
   )
 endif
 


stapelberg (approved):

stapelberg:
Sorry for the slow reply, and thanks for sending the fix!

PR #4280, opened 2020-12-08
Clear pixmap to black before starting to draw

### LINKED ISSUE #3481: x_draw_decoration should fill pixmap before starting to draw
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
In `x_draw_decoration`, we simply draw our decoration to the surface.

## Expected Behavior
To avoid issues with clients like xeyes using Shape, we should fill the surface first (e.g., with black) to avoid visual garbage. See https://github.com/i3/i3/issues/2742#issuecomment-433330426 for details.

Note that this issue becomes void if #3479 is implemented first.

## Environment

i3 4.15

tbgiles:
This bug appears to still be kicking around, so here's a stab at it.

Added a call to clear the framebuffer before writing to it. In addition to running the test suite, also tested with `xeyes`, and background was black as expected.

I've updated the release notes to mention this bugfix.

fixes #3481

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -22,3 +22,4 @@ strongly encouraged to upgrade.
   • when initializing new outputs, avoid duplicating workspace numbers
   • fix workspaces not moving to assigned output after output becomes available
   • fix duplicate bindcode after i3-config-wizard
+  • clear pixmap before drawing to prevent visual grabage in clients using 'Shape'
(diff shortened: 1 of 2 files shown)


Airblader (approved):
I can't test it myself at the moment, but the change looks good to me. Looks like it'll need a rebase, though.

orestisfl (approved):

Issue #3721, opened 2019-06-17
Support custom height for i3bar


Airblader:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Desired Behavior

The i3bar config should support a `height` configuration that specifies the height of i3bar. This should work exactly as in i3-gaps (in fact, ideally we'd just migrate the feature from there).

*Edit:* As discussed, let's instead make this a `vertical_padding` kind of option rather than an exact height.

## Environment
Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: 4.16.1
</pre>

stapelberg:
I’m not saying I’m against this, but can you explain a little more what the use case is? Currently, the height is derived from the font size. If the height is increased, does that influence the font size? If not, what is displayed in the extra space?

Airblader:
I think originally I added the height because of the custom border widths on status blocks (see issue https://github.com/i3/i3/issues/3722) to avoid overlapping the borders and the text. I've also seen people use it simply to make their bars taller or thinner according to preference. All content is simply vertically centered within the bar. It has no effect on font size (that would defeat the purpose, we might as well change the font size then).

A possible alternative approach to `height` would be a user-defined vertical padding of the bar, that might more accurately reflect its idea and somewhat reduce the dependency between font size and height.

stapelberg:
> A possible alternative approach to `height` would be a user-defined vertical padding of the bar, that might more accurately reflect its idea and somewhat reduce the dependency between font size and height.

I think I’d prefer this approach over the height approach.

stapelberg:
Attempting backwards compatibility as you outlined seems fine to me.

Airblader:
@stapelberg Another thought: ideally we could get package maintainers to merge i3-gaps packages into the i3 package, but in doing so if i3 has no `height`, it's a breaking change no matter what compat approach I take in i3-gaps for the time being. Hm…

skontar:
Hi. This is my most favourite feature from gaps. It allows me to have reasonably big tray icons and workspace buttons with a smaller text and bit of padding everywhere.

I use something like:
```
    font pango: FontAwesome, Hack 11.5
    height 32
```
The result then is something like this:
![Screenshot_2019-06-18_10-58-14](https://user-images.githubusercontent.com/10827040/59668001-00816200-91b8-11e9-92ea-79eb02fc3292.png)

[...]

PR #4287, opened 2020-12-20
Add title_height and bar { height } config parameters

ShimmerGlass:
This adds config options to set the height of the title window decoration and/or i3bar height.
With this users can add a bit more padding around the bar/title text, which can look better, depending on taste.

The defaults are of course kept as-is, so the default appearance won't change unless these parameters are specified.

Example of how I use `title_height`:
image
_(note: the border on top comes from another change on my fork, ignore it)_

### DIFF
--- a/i3bar/include/configuration.h
+++ b/i3bar/include/configuration.h
@@ -55,6 +55,7 @@ typedef struct config_t {
     char *fontname;
     i3String *separator_symbol;
     TAILQ_HEAD(tray_outputs_head, tray_output_t) tray_outputs;
+    int height;
     int tray_padding;
     int num_outputs;
     char **outputs;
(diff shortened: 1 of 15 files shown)


Airblader:
For the bar height there is #3721 which discusses (and accepts) a different approach for setting the bar height, not the one implemented here (or in i3-gaps). That feature is marked accepted due to the (long-term) plan of integrating i3-gaps, but the same does not apply to something like `title_height`, so let's please remove that from this PR and open an issue for it instead to discuss it first.

Do you want to change this PR to the implementation of `bar {height}` discussed in that issue?

ShimmerGlass:
@Airblader, I wasn't aware of this issue. I asked a few questions there, and I am closing this for https://github.com/i3/i3/pull/4288, which I am ok to change to a padding setting once the discussion on the issue is resolved.
I will open an issue for the title height setting specifically, along with a draft PR to support it.

ShimmerGlass:
* Draft: https://github.com/i3/i3/pull/4289
* Issue: https://github.com/i3/i3/issues/4290

PR #4289, opened 2020-12-20
Support custom padding/height for window titles

ShimmerGlass:


orestisfl:
@Airblader According to https://github.com/i3/i3/issues/4290#issuecomment-752944861 do we keep or close this PR?

Airblader:
Let's close the PR for now but keep the issue open. If we decide to add this we can always reopen the PR.

PR #4300, opened 2020-12-30
Remove unused header file and references to it

abeutot:
Hello,

Found what looks like remainings from the past while looking through the i3 codebase.


docs/hacking-howto
```diff
@@ -400,8 +400,8 @@ the correct state.
 Then, it looks through all bindings and gets the one which matches the received
 event.
 
-The bound command is parsed by the cmdparse lexer/parser, see +parse_cmd+ in
-+src/cmdparse.y+.
+The bound command is parsed by the cmdparse lexer/parser, see +parse_command+ in
++src/commands_parser.c+.
```
orestisfl:
```suggestion
The bound command is parsed by the i3 parser, see +parse_command+ in
+src/commands_parser.c+.
```

orestisfl (approved):
Thanks! Minor nitpick

docs/hacking-howto
abeutot: I updated with the requested changes.

PR #4307, opened 2021-01-03
Revert "Clear pixmap to black before starting to draw"

orestisfl:
Reverts i3/i3#4280

Adds black bars in my configuration:

before
2021-01-03-011031_1919x456_scrot

after
2021-01-03-011116_1919x384_scrot

Re-opened #3481

CC @tbgiles


tbgiles:
Looks like the borders are being drawn before the background, I can take another look

Airblader (approved):

Issue #3976, opened 2020-03-22
[question] i3bar status line intended behaviour


ammgws:
## I'm submitting a…

<pre>
[ ] Bug
[ ] Feature Request
[ ] Documentation Request
[x] Other (Please describe in detail)
</pre>

## Current Behavior
Multiple lines are shown at once.

## Expected Behavior
Only the last received line should be shown.

## Reproduction Instructions
Set bar to `status_command tail -f /tmp/1` and run `printf "aaa\nbbb\n" >/tmp/1` before starting i3/i3bar.

## Environment
Output of `i3 --moreversion 2>&-`:
<pre>
Binary i3 version:  4.18 (2020-02-17) © 2009 Michael Stapelberg and contributors

Running version: < 4.2-200
</pre>

## Background

Came across this when attempting to solve https://github.com/swaywm/sway/issues/4496.
The user(s) state that since it is a line protocol, it should only output one line at a time, and I was intending to go that way but found that it would change the behaviour from i3's. So I am checking here whether this behaviour is intentional or not.

Comparison between sway and i3:

`printf "XXX" >/tmp/1` and start i3/sway with bar config of `status_command tail -f /tmp/1`:

| XXX  | sway | i3 |
| ------------- | ------------- | ------------- |
| aaa | aaa | aaa |
| aaa\n | aaa | aaa |
| aaa\nbbb | aaa⏎bbb (one line, newline not rendered) | aaa\nbbb (multiline output) |
| aaa\nbbb\n | aaa⏎bbb (one line, newline not rendered) | aaa\nbbb (multiline output) |

`printf "XXX" >>/tmp/1` after i3/swaybar is already running:
| XXX  | sway | i3 |
| ------------- | ------------- | ------------- |
| aaa | aaa | aaa |
| aaa\n | aaa | aaa\n (multiline output) |
| aaa\nbbb | bbb | aaa\nbbb (multiline output) |
| aaa\nbbb\n | bbb | aaa\nbbb (multiline output) |

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

Airblader:
Thanks for the detailed information!

> So I am checking here whether this behaviour is intentional or not.

[...]

PR #4311, opened 2021-01-04
i3bar: Add protocol for workspace buttons

### LINKED ISSUE #1808: Feature request: add "groups of workspaces"
Hi,

**The problem**: This is a request resulting from user experience. When using my machine I quite often do 2 or 3 different things at the same time (let's call these things tasks), switching between tasks as they stall / change priority. Suppose I have the following tasks, with their corresponding desktops containing the relevant windows:

Task A: desktop 1,5,6 - some development project
Task B: desktop 7,8 - some windows to communicate with people
Task C: desktop 2, 3, 10, 11 - system administration

i3 allows to create lots of desktops, which is convenient as some tasks require 3 or 4 desktops by themselves, and frequent switching between them. However it's tiresome to have to remember desktop numbers for each task. E.g. when I'm on task B and want to switch to A, I'll have to remember that the relevant windows are in desktops 1, 5, 6, and I'll always lose some time finding the right ones.

**Suggestion**: it would be helpful to be able to identify groups of desktops, and "hide" irrelevant desktops when necessary. I'm not quite sure how this could be implemented -- perhaps it's something we could discuss -- but it would be helpful e.g. if I'm on task B, to have only desktops 7 and 8 visible on the status bar, and to be able to "hit some keys" to switch to a different task, say C, and have desktops 2, 3, 10 and 11 replace task B's desktops in the status bar.

It corresponds to some sort of grouping and it sounds like it could make things easier to focus on, especially when there is a lot of "context switching" going on.

I'd be happy to hear your thoughts on this, and I'd also be happy to give this a shot if you think it could be useful.

Cheers
vgrevendo

### LINKED ISSUE #2333: enhancement: allow i3bar to not display some workspaces

### LINKED ISSUE #2617: feature request : show empty workspaces
Is there a way to show all workspaces, even empty ?
bspwm allows it

### LINKED ISSUE #3548: sort i3bar workspaces alphabetically
## I'm submitting a…
<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
I use workspaces dynamically, based on project I'm working on. Say, I'm working for project "foo", I will have workspaces: foo-1, foo-2, foo-3, etc devoted for this project.  I tend to have about 30 workspaces open at any time.

Creating a new workspace always puts it the new one at the end of the list in i3bar.  I want the "foo" workspaces to be near one another, and sequential.

## Desired Behavior
I want the workspaces to be ordered alphabetically in i3bar.

Setting a workspace number explicitly, does not work well in my use case.  I used to set the numbers to force sorting using helper script, but it's rather slow.

A small hack to i3bar (see below) works much better.

## Environment
I'm running i3 on debian/testing (4.16), but this problem exists in master and next branches also.
Please include your exact i3 version.
</pre>

## Proposed Fix
I have hacked on i3bar to implement this feature [in my fork](https://github.com/bartman/i3/tree/bart/i3bar-hacks)...
https://github.com/bartman/i3/commit/cc1d8fdf99be8cf50c05c35502e0b3bd5429a4c6

After adding the following to `~/.i3/config`, I get the behaviour I want.

<pre>
bar {
    sort_workspace_by_name yes
}
</pre>

Note that the default behavior is "no" and i3bar behaves as before.

### LINKED ISSUE #3818: i3bar: protocol for workspace buttons

orestisfl:
## Notes

- Adds `focus workspace` command to handle focusing of workspaces. Alternatively, could be named `workspace show`.
- Examples from #3818 work. jq's `--unbuffered` **must** be used for this to work properly. I don't think we can fix this from our side in a portable manner so users will have to make sure their code outputs to stdout without buffering.

## TODOs

- [x] Document this, including usage examples
- [x] Include `contrib/` script(s)
- Workspace JSON parsing:
  - [x] Support "primary" as output name
  - [x] When no "output" is given, default to the primary or first output
  - [x] Don't exit on parsing errors
  - [x] (python script) Fix JSON-parsing related crash when monitor changes
- [x] (python script) Workspace buttons just disappear randomly.
- [x] Release notes with all changes & fixes
- [x] ~~(can be done in separate PR) `getline` can be used instead of `get_buffer` everywhere in `child.c` to fix  https://github.com/i3/i3/issues/3976~~
- [x] Separate focus command changes in new commit w/ separate release notes

## Python script for debugging

```python
#!/usr/bin/env python
import json
import time
from subprocess import check_output

import i3ipc


def cb(_=None, e=None):
    d = json.loads(check_output(["i3-msg", "-t", "get_workspaces"]).decode())

    if not any(ws.get("name") == "foo" for ws in d):
        # foo = {"name": "foo", "output": "whatever"}
        foo = {"name": "foo"}
        d.append(foo)

    print(json.dumps(d))


i3 = i3ipc.Connection()
i3.on(i3ipc.Event.WORKSPACE, cb)
i3.on(i3ipc.Event.OUTPUT, cb)
cb()
i3.main()
```
## Issues

Closes #3818 (parent issue)
Fixes https://github.com/i3/i3/issues/1808
Fixes https://github.com/i3/i3/issues/2333
Fixes https://github.com/i3/i3/issues/2617
Fixes https://github.com/i3/i3/issues/3548


orestisfl:
@Airblader without an in-depth review, any general comments on this before I invest more time?

Airblader:
@orestisfl No, you're good to go ahead from my point of view.

i3bar/src/child.c
```diff
@@ -497,22 +530,37 @@ static void stdin_io_first_line_cb(struct ev_loop *loop, ev_io *watcher, int rev
...
-        set_statusline_error("status_command not found or is missing a library dependency (exit %d)", exit_status);
-    else
-        set_statusline_error("status_command process exited unexpectedly (exit %d)", exit_status);
+    if (exit_status == 126) {
+        set_statusline_error("%s is not executable (exit %d)", command_type, exit_status);
+    } else if (exit_status == 127) {
+        set_statusline_error("%s not found or is missing a library dependency (exit %d)", command_type, exit_status);
+    } else {
+        set_statusline_error("%s process exited unexpectedly (exit %d)", command_type, exit_status);
+    }
```
orestisfl: These lines will also set a statusline error for workspace button misconfiguration which is "unfair" to the statusline. We should at least make sure that it doesn't replace all other status blocks.

orestisfl:
I am done changing i3bar's code for this, so first wave of reviewing can start now.

andersfylling:
@orestisfl maybe marking the PR as ready, instead of draft, would make people aware it's ready for review :+1:

ymolists:
@orestisfl  Thank you so much for your great efforts. I hope you find the energy and time to finish this to the completion with everything that is happening these days.

If not i hope there is someone out thee that is able to move this forward ? :-)

orestisfl:
I started working on this again recently and I'll probably be able to continue this December.

ammgws:
Since the only thing left seems to be documentation, has anyone been daily driving this without issues?

orestisfl:
> Since the only thing left seems to be documentation, has anyone been daily driving this without issues?

It's not only documentation, there were some problems with this code.

I'll have to find some more time to work on this.

orestisfl:
Besides some non-code TODOs I think this works as expected now

i3bar/src/ipc.c
```diff
@@ -24,24 +24,33 @@ ev_io *i3_connection;
 
 const char *sock_path;
 
-typedef void (*handler_t)(char *);
+typedef void (*handler_t)(const unsigned char *, size_t);
+
+/*
+ * Returns true when i3bar is configured to require workspace information from
```
stapelberg: > Returns true when i3bar is configured to require workspace information from i3
 
How about:

Returns true when i3bar is configured to read workspace information from i3 via JSON over the i3 IPC interface, as opposed to reading workspace information from the workspace_command via JSON over stdout.

i3bar/src/xcb.c
```diff
@@ -500,6 +500,55 @@ static int predict_button_width(int name_width) {
...
+         * allowed to change workspace names as long as it provides a valid ID. */
+        sasprintf(&buffer, "[con_id=%lld] focus workspace", ws->id);
+        goto done;
+    }
+
+    if (ws->canonical_name == NULL) {
+        return;
+    }
+
+    /* To properly handle workspace names with double quotes in them, we need
```
stapelberg: Can you refactor the quoting into a separate function? That would make both functions easier to read, I think.

docs/i3bar-protocol
```diff
@@ -98,6 +98,8 @@ You can find an example of a shell script which can be used as your
 +status_command+ in the bar configuration at
 https://github.com/i3/i3/blob/next/contrib/trivial-bar-script.sh
 
+TODO: maybe same for workspace_command
```
stapelberg: Are you going to update this? Can you also update docs/userguide and docs/ipc accordingly?

I think it would also be good to state an example clearly somewhere. AFAICT, the motivating use case is to either show workspaces that don’t exist (but are configured and possibly named), or to filter workspaces in some way for further grouping or “secret” workspaces.

stapelberg (commented):
Haven’t looked through everything in detail yet, but I think some more documentation would be good to understand the high-level picture first

docs/i3bar-protocol
orestisfl: Yes, will do

i3bar/src/xcb.c
orestisfl: Simplified it a bit by removing the quote-counting bit

orestisfl:
I think the high-level picture is perfectly outlined in https://github.com/i3/i3/issues/3818 but I can write down more stuff if required at this stage.

orestisfl:
Something I realize (a little bit too late) is that while the status line protocol is ["an infinite array (so it should be parsed by a streaming JSON parser)"](https://i3wm.org/docs/i3bar-protocol.html#_the_protocol) the existing workspace protocol between i3 and i3bar is effectively [jsonlines](https://jsonlines.org). The workspace protocol is based on the ipc interface implementation which has the extra benefit of being able to subscribe to workspace events from i3, make any changes to the received data and then then pipe the modified data to i3bar.

This approach has the downside of having a different protocol between `workspace_command` and `status_command` however the benefits are that 1. The `workspace_command` protocol is easier to implement for clients since it is based on subscribing to workspace events from i3 2. We don't need to branch-off parsing the input based on who supplies the data (client or i3).

I think we should continue with the current approach, just wanted to write this down.

i3bar/src/child.c
```diff
@@ -28,13 +29,29 @@
...
-#define DLOG_CHILD DLOG("%s: pid=%ld stopped=%d stop_signal=%d cont_signal=%d click_events=%d click_events_init=%d\n", \
-                        __func__, (long)child.pid, child.stopped, child.stop_signal, child.cont_signal, child.click_events, child.click_events_init)
-
-/* stdin- and SIGCHLD-watchers */
-ev_io *stdin_io;
-int stdin_fd;
-ev_child *child_sig;
+i3bar_child ws_child = {0};
+
+#define DLOG_CHILD(c)                                                                                                              \
```
stapelberg, line 34: Can you convert this code from a macro into a regular function please? That makes it easier to work with (you can properly set breakpoints in a debugger, for example). Capture __func__ as early as possible and just call a regular function from the macro, turning it into a one-liner.

i3bar/src/child.c
```diff
@@ -598,34 +753,85 @@ void start_child(char *command) {
...
+    ws_child.stop_signal = SIGSTOP;
+    ws_child.cont_signal = SIGCONT;
+
+    int pipe_in[2]; /* pipe we read from */
+    if (pipe(pipe_in) == -1) {
+        err(EXIT_FAILURE, "pipe(pipe_in)");
+    }
+
+    ws_child.pid = fork();
+    switch (ws_child.pid) {
```
stapelberg: Can you refactor the fork-related code out into a common function used by `start_child` and `start_ws_child` please?

i3bar/src/child.c
```diff
@@ -706,62 +912,111 @@ void send_block_clicked(int button, const char *name, const char *instance, int
...
         killpg(child.pid, SIGTERM);
     }
+    if (ws_child.pid > 0) {
+        if (ws_child.cont_signal > 0 && ws_child.stopped)
+            killpg(ws_child.pid, ws_child.cont_signal);
+        killpg(ws_child.pid, SIGTERM);
+    }
+}
+
+static void _kill(i3bar_child *c) {
```
stapelberg: Could you rename this to `kill_and_wait`, and then refactor the common logic between `kill_child_at_exit` and this function please?

(Can you remove the underscore in front of the function name? I vaguely remember that not being necessarily safe in C. I think `static void` vs `void` already is enough to distinguish internal vs exported functions.)

i3bar/src/child.c
```diff
@@ -706,62 +912,111 @@ void send_block_clicked(int button, const char *name, const char *instance, int
...
+    DLOG_CHILD(*c);
+
+    if (c->pid > 0) {
```
stapelberg: Revert conditional to save indentation:

```c
if (c->pid <= 0) {
  return;
}
// …
```

i3bar/src/child.c
```diff
@@ -706,62 +912,111 @@ void send_block_clicked(int button, const char *name, const char *instance, int
...
+ */
+void kill_ws_child(void) {
+    _kill(&ws_child);
 }
 
 /*
  * Sends a SIGSTOP to the child process (if existent)
  *
  */
 void stop_child(void) {
```
stapelberg: Here and below: This should now be `stop_children`, right? The common logic could be factored into a `static void stop_child(void)`

i3bar/src/workspaces.c
```diff
@@ -216,43 +232,41 @@ static yajl_callbacks workspaces_callbacks = {
...
 
     /* FIXME: Proper error handling for JSON parsing */
     switch (state) {
         case yajl_status_ok:
             break;
         case yajl_status_client_canceled:
-        case yajl_status_error:
-            ELOG("Could not parse workspaces reply!\n");
-            exit(EXIT_FAILURE);
+        case yajl_status_error:;
```
stapelberg: remove extra ;

src/commands.c
```diff
@@ -1481,20 +1481,26 @@ void cmd_focus(I3_CMD) {
...
              * times and only a single window ends up focused, we could show
              * multiple scratchpad windows. So, rather break here. */
             break;
         }
 
-        LOG("focusing %p / %s\n", current->con, current->con->name);
-        con_activate_unblock(current->con);
+        if (focus_workspace) {
+            LOG("focusing workspace %p / %s - %p / %s\n", current->con, current->con->name, ws, ws->name);
+            workspace_show(ws);
```
stapelberg, line 1501: `con_activate_unblock` also calls `workspace_show`. Can you explain (probably best in a comment in the source) why the conditional code path is needed for `focus workspace` here?

stapelberg (commented):
I went through this in detail now and left a few comments.

Overall the change looks good, and I think I might even have a use-case for it in https://github.com/stapelberg/wsmgr-for-i3 (making its workspaces persistent).

Nevertheless, documentation is still missing: we need a description of the feature in docs/i3bar-protocol, with maybe one or two examples of what can be done with this protocol (AFAICT, filtering workspaces or injecting workspaces, or are there any other broad kinds of use-cases?), and a JSON example of what to send.

Let me know if you want to continue working on this, or if you’d appreciate some help to carry it over the finish line. Thanks!

orestisfl:
> Let me know if you want to continue working on this, or if you’d appreciate some help to carry it over the finish line. Thanks!

Yes, I'll keep working on this. Thank you for your review!

i3bar/src/child.c
orestisfl: I did some changes but the code is not exactly identical and it's not easy to refactor without affecting readability

i3bar/src/child.c
orestisfl: Actually, this will remove the correct line numbers which is pretty much the point of these macros, I think we are better off if I revert this change

src/commands.c
orestisfl: Because we just want to focus the workspace, not necessarily the matched container

i3bar/src/child.c
stapelberg: Have you considered making the code identical by always setting up a stdin for the ws child, too (which then goes unused)?

Either way, everything starting from the `fcntl(status_child.stdin_fd, F_SETFL, O_NONBLOCK);` call looks pretty identical to me and could probably be factored out into a function that takes an i3bar_child.

stapelberg (commented):
This looks good overall code-wise. The only thing missing is the documentation update, right?

orestisfl:
Documentation up :heavy_check_mark:

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
+i3bar workspace buttons protocol
+================================
+
+This document explains the protocol in which i3bar expects input for
+configuring workspace buttons. This is a available since i3 version 4.23.
```
stapelberg: “This feature”

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+Programs defined by the +workspace_command+ configuration option for i3bar can
+modify the workspace buttons displayed by i3bar. The command should constantly
+print in its standard output a stream of messages following the protocol
+defined in this page.
+
+If you are looking for the status line protocol instead, see https://i3wm.org/docs/i3bar-protocol.html.
+
+== The protocol
+
+Each message should be a newline-delimeted JSON array. The array is in the same
```
stapelberg: typo: delimited

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+format as the +GET_WORKSPACES+ ipc event, see
+https://i3wm.org/docs/ipc.html#_workspaces_reply.
+
+As an example, this is the output of the +i3-msg -t get_workspaces+ command:
+------------------------------
+[{"id":94131549984064,"num":1,"name":"1","visible":false,"focused":false,"rect":{"x":0,"y":0,"width":2560,"height":1421},"output":"HDMI-A-0","urgent":false},{"id":94131550477584,"num":2,"name":"2","visible":true,"focused":true,"rect":{"x":0,"y":0,"width":2560,"height":1421},"output":"HDMI-A-0","urgent":false},{"id":94131550452704,"num":3,"name":"3:some workspace","visible":false,"focused":false,"rect":{"x":0,"y":0,"width":2560,"height":1421},"output":"HDMI-A-0","urgent":false}]
+------------------------------
+
+Each element in the array represents a workspace. i3bar creates on workspace
```
stapelberg: typo: on → one

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+fetching the current workspace information with +GET_WORKSPACES+, modifying the
+JSON array in the response according to your needs and then printing it to the
+standard output. However, you are free to build a new message from the ground
+up.
+
+=== Workspace objects in detail
+
+The documentation of +GET_WORKSPACES+ should be sufficient to understand the
+meaning of each property but here we provide extra notes for each property and
+it's meaning with respect to i3bar.
```
stapelberg: typo: its meaning (no apostrophe)

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+
+All properties but +name+ are optional.
+
+id (integer)::
+    If it is included it will be used to switch to that workspace when you
```
stapelberg, line 77: Can you change these lines from indentation with 4 spaces to 1 tab for consistency with the lines below please?

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+    click the corresponding button. If it's not provided, the +name+ will be
+    used.
```
stapelberg: Can you add something along the lines of:

You can use the id field to present workspaces under a modified name.

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+	Defaults to the primary output if not included.
+
+== Examples
+
+These example scripts require the https://stedolan.github.io/jq/[jq] utility to
+be installed but otherwise just use the standard +i3-msg+ utility included with
+i3. However, you can write your own scripts in your preferred language and
+you can optionally use one of the
+https://i3wm.org/docs/ipc.html#_see_also_existing_libraries[pre-existing i3
+libraries]
```
stapelberg, line 108: I think we should make this stronger:

However, you can write your own scripts in your preferred language, with the help of one of the pre-existing i3 libraries

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+------------------------------
+#!/bin/sh
+i3-msg -t subscribe -m '["workspace", "output"]' | {
+	i3-msg -t get_workspaces;
+	while read; do i3-msg -t get_workspaces; done;
+} | jq --unbuffered -c '[ .[] | select(.name != "foo" or .focused) ]'
+------------------------------
+
+Important! Make sure you use the +--unbuffered+ flag with +jq+, otherwise you
+might not get the changes in real-time but whenever they are flashed, which
```
stapelberg: typo: flushed

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,150 @@
...
+
+== The protocol
+
+Each message should be a newline-delimeted JSON array. The array is in the same
+format as the +GET_WORKSPACES+ ipc event, see
+https://i3wm.org/docs/ipc.html#_workspaces_reply.
+
+As an example, this is the output of the +i3-msg -t get_workspaces+ command:
+------------------------------
+[{"id":94131549984064,"num":1,"name":"1","visible":false,"focused":false,"rect":{"x":0,"y":0,"width":2560,"height":1421},"output":"HDMI-A-0","urgent":false},{"id":94131550477584,"num":2,"name":"2","visible":true,"focused":true,"rect":{"x":0,"y":0,"width":2560,"height":1421},"output":"HDMI-A-0","urgent":false},{"id":94131550452704,"num":3,"name":"3:some workspace","visible":false,"focused":false,"rect":{"x":0,"y":0,"width":2560,"height":1421},"output":"HDMI-A-0","urgent":false}]
```
stapelberg: Can you run this through `jq` to pretty-print it? You can remove the "rect" field and go from 3 workspaces to just two to keep it short.

i3bar/src/workspaces.c
```diff
@@ -156,44 +154,62 @@ static int workspaces_string_cb(void *params_, const unsigned char *val, size_t
...
         new_workspace->num = -1;
-        new_workspace->name = NULL;
-        new_workspace->visible = 0;
-        new_workspace->focused = 0;
-        new_workspace->urgent = 0;
-        memset(&new_workspace->rect, 0, sizeof(rect));
-        new_workspace->output = NULL;
+        /* Better default for custom user commands. i3 always includes these
+         * properties so they are inefectual in that case. */
+        new_workspace->visible = true;
```
stapelberg: Can you explain in the comment why this is the better default? 

AFAICT, i3bar uses a different font color for visible vs. not visible (white vs gray), so if you’re injecting workspaces that don’t exist yet, it would actually be better to have visible==false.

stapelberg (approved):
Thanks for adding the docs. I tried all the examples and they all worked :)

Can you also let me know how you’d like this to be merged in terms of git history? Squashed into a single commit or rebased to keep commit history? If rebased, can you clean up the commits in the way you would like them to appear later? (e.g. the intermediate refactorings during our PR review doesn’t need to be part of the repository history later)

i3bar/src/workspaces.c
orestisfl: Actually this makes no sense, I'll revert it. My misunderstanding for what "visible" means.

orestisfl:
Thanks for the review.

I've split the change into two commits.

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,184 @@
...
+    used. You can use the +id+ field to present workspaces under a modified
+    name.
+num (integer)::
+    The only use of a workspace's number is if the +strip_workspace_numbers+
+    setting is enabled.
+name (string)::
+    The only required property. If an +id+ is provided you can freely change
+    the +name+ as you wish, effectively renaming the buttons of i3bar.
+visible (boolean)::
+    Defaults to +true+ if not included.
```
stapelberg: This line is now obsolete: visible defaults to false if not included.

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,184 @@
+i3bar workspace buttons protocol
+================================
+
+This document explains the protocol in which i3bar expects input for
+configuring workspace buttons. This feature is a available since i3 version
```
stapelberg: remove extra “a” in front of available

docs/i3bar-workspace-protocol
```diff
@@ -0,0 +1,184 @@
...
+4.23.
+
+Programs defined by the +workspace_command+ configuration option for i3bar can
```
stapelberg: Instead of “Programs”, I think the singular “The program” would be clearer here. It’s typically just one, and for folks who configure multiple bar {} directives, they already know the workspace_command is per bar directive.

stapelberg (commented):
Just a few small doc updates and then I can merge this :)

docs/i3bar-workspace-protocol
orestisfl: Thanks, good catch

stapelberg (approved):

PR #4321, opened 2021-01-15
i3-dump-log -f: switch from pthreads to UNIX sockets

### LINKED ISSUE #4117: switch away from pthreads for i3-dump-log -f
I just managed to lose my i3 session (there were many factors to this), and the root cause was that i3 got stuck in `pthread_cond_broadcast` after I killed i3-dump-log processes that might still have been paused.

Regardless of what I did, I should never manage to make the mechanism, and hence i3, deadlock.

I suggest we switch away from pthreads to something more proven, like perhaps UNIX Sockets for transfering the log contents, which we already use in the IPC interface.

I’ll tentatively assign this to me in the expectation that I can work on this soon.

stapelberg:
Fixes #4117

### DIFF
--- a/i3-dump-log/main.c
+++ b/i3-dump-log/main.c
@@ -25,22 +25,16 @@
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/un.h>
 
-#if !defined(__OpenBSD__)
-static uint32_t offset_next_write;
-#endif
 static uint32_t wrap_count;
 
 static i3_shmlog_header *header;
 static char *logbuffer,
     *walk;
 static int ipcfd = -1;
-
-static volatile bool interrupted = false;
-
-static void sighandler(int signal) {
-    interrupted = true;
-}
 
 static void disable_shmlog(void) {
     const char *disablecmd = "debuglog off; shmlog off";
@@ -188,22 +182,25 @@ int main(int argc, char *argv[]) {
 
     /* NB: While we must never write, we need O_RDWR for the pthread condvar. */
     int logbuffer_shm = shm_open(shmname, O_RDWR, 0);
-    if (logbuffer_shm == -1)
+    if (logbuffer_shm == -1) {
         err(EXIT_FAILURE, "Could not shm_open SHM segment for the i3 log (%s)", shmname);
+    }
 
-    if (fstat(logbuffer_shm, &statbuf) != 0)
+    if (fstat(logbuffer_shm, &statbuf) != 0) {
         err(EXIT_FAILURE, "stat(%s)", shmname);
+    }
 
-    /* NB: While we must never write, we need PROT_WRITE for the pthread condvar. */
-    logbuffer = mmap(NULL, statbuf.st_size, PROT_READ | PROT_WRITE, MAP_SHARED, logbuffer_shm, 0);
-    if (logbuffer == MAP_FAILED)
+    logbuffer = mmap(NULL, statbuf.st_size, PROT_READ, MAP_SHARED, logbuffer_shm, 0);
+    if (logbuffer == MAP_FAILED) {
         err(EXIT_FAILURE, "Could not mmap SHM segment for the i3 log");
+    }
 
     header = (i3_shmlog_header *)logbuffer;
 
-    if (verbose)
+    if (verbose) {
         printf("next_write = %d, last_wrap = %d, logbuffer_size = %d, shmname = %s\n",
                header->offset_next_write, header->offset_last_wrap, header->size, shmname);
+    }
     free(shmname);
     walk = logbuffer + header->offset_next_write;
 
@@ -233,25 +230,38 @@ int main(int argc, char *argv[]) {
         return 0;
     }
 
-    /* Handle SIGINT gracefully to invoke atexit handlers, if any. */
-    struct sigaction action;
-    action.sa_handler = sighandler;
-    sigemptyset(&action.sa_mask);
-    action.sa_flags = 0;
-    sigaction(SIGINT, &action, NULL);
+    char *log_stream_socket_path = root_atom_contents("I3_LOG_STREAM_SOCKET_PATH", NULL, 0);
+    if (log_stream_socket_path == NULL) {
+        errx(EXIT_FAILURE, "could not determine i3 log stream socket path: possible i3-dump-log and i3 version mismatch");
+    }
 
-    /* Since pthread_cond_wait() expects a mutex, we need to provide one.
-     * To not lock i3 (that’s bad, mhkay?) we just define one outside of
-     * the shared memory. */
-    pthread_mutex_t dummy_mutex = PTHREAD_MUTEX_INITIALIZER;
-    pthread_mutex_lock(&dummy_mutex);
-    while (!interrupted) {
-        pthread_cond_wait(&(header->condvar), &dummy_mutex);
-        /* If this was not a spurious wakeup, print the new lines. */
-        if (header->offset_next_write != offset_next_write) {
-            offset_next_write = header->offset_next_write;
-            print_till_end();
+    int sockfd = socket(AF_LOCAL, SOCK_STREAM, 0);
+    if (sockfd == -1) {
+        err(EXIT_FAILURE, "Could not create socket");
+    }
+
+    (void)fcntl(sockfd, F_SETFD, FD_CLOEXEC);
+
+    struct sockaddr_un addr;
+    memset(&addr, 0, sizeof(struct sockaddr_un));
+    addr.sun_family = AF_LOCAL;
+    strncpy(addr.sun_path, log_stream_socket_path, sizeof(addr.sun_path) - 1);
+    if (connect(sockfd, (const struct sockaddr *)&addr, sizeof(struct sockaddr_un)) < 0) {
+        err(EXIT_FAILURE, "Could not connect to i3 on socket %s", log_stream_socket_path);
+    }
+
+    /* Same size as the buffer used in log.c vlog(): */
+    char buf[4096];
+    for (;;) {
+        const int n = read(sockfd, buf, sizeof(buf));
+        if (n == -1) {
+            err(EXIT_FAILURE, "read(log-stream-socket):");
         }
+        if (n == 0) {
+            exit(0); /* i3 closed the socket */
+        }
+        buf[n] = '\0';
+        swrite(STDOUT_FILENO, buf, n);
     }
 
 #endif
--- a/libi3/create_socket.c
+++ b/libi3/create_socket.c
@@ -1,0 +1,70 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3 - an improved dynamic tiling window manager
+ * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
+ *
+ */
+#include "libi3.h"
+
+#include <unistd.h>
+#include <libgen.h>
+#include <err.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+
+/*
+ * Creates the UNIX domain socket at the given path, sets it to non-blocking
+ * mode, bind()s and listen()s on it.
+ *
+ * The full path to the socket is stored in the char* that out_socketpath points
+ * to.
+ *
+ */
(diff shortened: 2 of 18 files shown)


src/ipc.c
```diff
@@ -1529,49 +1513,7 @@ ipc_client *ipc_new_client_on_fd(EV_P_ int fd) {
  *
  */
 int ipc_create_socket(const char *filename) {
```
orestisfl: Maybe having this function is now redundant?

Airblader (approved):
This isn't really my area of expertise, so all I can say is "looks OK to me" :-)

orestisfl:
I'll try to review this during the weekend

i3-dump-log/main.c
```diff
@@ -233,25 +226,38 @@ int main(int argc, char *argv[]) {
...
+    if (log_stream_socket_path == NULL) {
+        errx(EXIT_FAILURE, "could not determine i3 log stream socket path: possible i3-dump-log and i3 version mismatch");
+    }
+
+    int sockfd = socket(AF_LOCAL, SOCK_STREAM, 0);
+    if (sockfd == -1) {
+        err(EXIT_FAILURE, "Could not create socket");
+    }
+
+    (void)fcntl(sockfd, F_SETFD, FD_CLOEXEC);
```
orestisfl, line 243: nitpick for consistency:
```suggestion
    fcntl(sockfd, F_SETFD, FD_CLOEXEC);
```

i3-dump-log/main.c
```diff
@@ -233,25 +226,38 @@ int main(int argc, char *argv[]) {
...
+
+    struct sockaddr_un addr;
+    memset(&addr, 0, sizeof(struct sockaddr_un));
+    addr.sun_family = AF_LOCAL;
+    strncpy(addr.sun_path, log_stream_socket_path, sizeof(addr.sun_path) - 1);
+    if (connect(sockfd, (const struct sockaddr *)&addr, sizeof(struct sockaddr_un)) < 0)
```
orestisfl: Nit: Add curly braces

i3-dump-log/main.c
```diff
@@ -233,25 +226,38 @@ int main(int argc, char *argv[]) {
...
+        err(EXIT_FAILURE, "Could not connect to i3 on socket %s", log_stream_socket_path);
+
+    // Same size as the buffer used in log.c vlog():
```
orestisfl:
```suggestion
    /* Same size as the buffer used in log.c vlog(): */
```

i3-dump-log/main.c
```diff
@@ -233,25 +226,38 @@ int main(int argc, char *argv[]) {
...
+    char buf[4096];
+    for (;;) {
+        const int n = read(sockfd, buf, sizeof(buf));
+        if (n == -1) {
+            err(EXIT_FAILURE, "read(log-stream-socket):");
+        }
+        if (n == 0) {
+            // TODO: reconnect when i3 restarts, handle transient failures
```
orestisfl: Open TODO? I would prefer an issue rather than a comment.

src/log.c
```diff
@@ -370,3 +372,53 @@ void purge_zerobyte_logfile(void) {
...
+    client->fd = fd;
+    TAILQ_INSERT_TAIL(&log_clients, client, clients);
+
+    DLOG("log: new client connected on fd %d\n", fd);
+}
+
+void log_broadcast_to_clients(const char *message, size_t len) {
+    log_client *current;
+    TAILQ_FOREACH (current, &log_clients, clients) {
+        // TODO: In case slow connections turn out to be a problem here
```
orestisfl: Maybe an `XXX` instead of a `TODO` here? Also C-style comments.

i3-dump-log/main.c
```diff
@@ -233,25 +226,38 @@ int main(int argc, char *argv[]) {
...
+            exit(0);  // i3 closed the socket
```
orestisfl:
```suggestion
            exit(0);  /* i3 closed the socket */
```

libi3/create_socket.c
```diff
@@ -0,0 +1,70 @@
...
+    /* Unlink the unix domain socket before */
+    unlink(resolved);
+
+    if ((sockfd = socket(AF_LOCAL, SOCK_STREAM, 0)) < 0) {
+        perror("socket()");
+        free(resolved);
+        return -1;
+    }
+
+    (void)fcntl(sockfd, F_SETFD, FD_CLOEXEC);
```
orestisfl, line 47:
```suggestion
    fcntl(sockfd, F_SETFD, FD_CLOEXEC);
```

libi3/create_socket.c
```diff
@@ -0,0 +1,70 @@
...
+ */
+int create_socket(const char *filename, char **out_socketpath) {
+    int sockfd;
+
+    char *resolved = resolve_tilde(filename);
+    DLOG("Creating UNIX socket at %s\n", resolved);
+    char *copy = sstrdup(resolved);
+    const char *dir = dirname(copy);
+    if (!path_exists(dir))
+        mkdirp(dir, DEFAULT_DIR_MODE);
```
orestisfl:
```suggestion
    if (!path_exists(dir)) {
        mkdirp(dir, DEFAULT_DIR_MODE);
    }
```

libi3/create_socket.c
```diff
@@ -0,0 +1,70 @@
...
+    free(copy);
+
+    /* Unlink the unix domain socket before */
+    unlink(resolved);
+
+    if ((sockfd = socket(AF_LOCAL, SOCK_STREAM, 0)) < 0) {
```
orestisfl:
```suggestion
    int sockfd = socket(AF_LOCAL, SOCK_STREAM, 0);
    if (sockfd < 0) {
```

libi3/nonblock.c
```diff
@@ -0,0 +1,20 @@
...
+ *
+ */
+void set_nonblock(int sockfd) {
+    int flags = fcntl(sockfd, F_GETFL, 0);
+    if (flags & O_NONBLOCK) {
+        return;
+    }
+    flags |= O_NONBLOCK;
+    if (fcntl(sockfd, F_SETFL, flags) < 0)
+        err(-1, "Could not set O_NONBLOCK");
```
orestisfl:
```suggestion
    if (fcntl(sockfd, F_SETFL, flags) < 0) {
        err(-1, "Could not set O_NONBLOCK");
    }
```

src/log.c
```diff
@@ -370,3 +372,53 @@ void purge_zerobyte_logfile(void) {
...
+    int fd;
+    if ((fd = accept(w->fd, (struct sockaddr *)&peer, &len)) < 0) {
+        if (errno != EINTR) {
+            perror("accept()");
+        }
+        return;
+    }
+
+    /* Close this file descriptor on exec() */
+    (void)fcntl(fd, F_SETFD, FD_CLOEXEC);
```
orestisfl, line 397:
```suggestion
    fcntl(fd, F_SETFD, FD_CLOEXEC);
```

orestisfl (changes requested):
Some nitpicks for now.

src/log.c
stapelberg: Why the suggestion? fcntl returns an error, so to signal to static analysis tools, we should add `(void)` to signal we know we’re not handling an error here. Is that no longer necessary?

i3-dump-log/main.c
stapelberg: Issue #4325 filed.

stapelberg:
PTAL

src/log.c
orestisfl: I don't think that the compiler complains about unused return values and the rest of the codebase does not use this idiom. However, I am fine with changing our approach with this hereafter.

orestisfl: And, technically, to be consistent with this practice we would technically have to apply it to common functions that are usually not checked like `printf` which can return `< 0` on failure.

stapelberg: > I don't think that the compiler complains

No, I think it was `clang-analyze`, but we removed that from our CI in 1a2ce3edaf4cb81acf1cda504151c8d96ff27a60 (2017), so I doubt anyone runs it…

> And, technically, to be consistent with this practice we would technically have to apply it to common functions that are usually not checked like `printf` which can return `< 0` on failure.

I don’t think it makes sense to check (or annotate) e.g. `printf`. What are you going to do when printing fails? Printing an error will likely fail, too ;)

I think the annotation should be placed where reviewers would otherwise ask to handle the error, to convey that we thought about an error check and decided against it.

src/main.c
```diff
@@ -634,6 +637,11 @@ int main(int argc, char *argv[]) {
             config.ipc_socket_path = sstrdup(config.ipc_socket_path);
     }
 
+    if (config.log_stream_socket_path == NULL) {
+        /* Chose a file name in /tmp/ based on the PID */
+        config.log_stream_socket_path = get_process_filename("log-stream-socket");
```
orestisfl: This variable does not seem to need to reside in `config.`

Applying this diff compiles:

```diff
diff --git a/include/configuration.h b/include/configuration.h
index eaf71829..11cdde0d 100644
--- a/include/configuration.h
+++ b/include/configuration.h
@@ -93,7 +93,6 @@ struct Config {
     i3Font font;
 
     char *ipc_socket_path;
-    char *log_stream_socket_path;
     char *restart_state_path;
 
     layout_t default_layout;
diff --git a/src/main.c b/src/main.c
index 49c23521..54b38f44 100644
--- a/src/main.c
+++ b/src/main.c
@@ -637,11 +637,6 @@ int main(int argc, char *argv[]) {
             config.ipc_socket_path = sstrdup(config.ipc_socket_path);
     }
 
-    if (config.log_stream_socket_path == NULL) {
-        /* Chose a file name in /tmp/ based on the PID */
-        config.log_stream_socket_path = get_process_filename("log-stream-socket");
-    }
-
     if (config.force_xinerama) {
         force_xinerama = true;
     }
@@ -862,7 +857,10 @@ int main(int argc, char *argv[]) {
         ev_io_start(main_loop, ipc_io);
     }
 
-    int log_socket = create_socket(config.log_stream_socket_path, &current_log_stream_socket_path);
+    /* Chose a file name in /tmp/ based on the PID */
+    char *log_stream_socket_path = get_process_filename("log-stream-socket");
+    int log_socket = create_socket(log_stream_socket_path, &current_log_stream_socket_path);
+    free(log_stream_socket_path);
     if (log_socket == -1) {
         ELOG("Could not create the log socket, i3-dump-log -f will not work\n");
     } else {
```

orestisfl (changes requested):
Canceling `i3-dump-log -f` with ctrl-c crashes i3 for me. If you can't reproduce this tell me so that I can provide more details.

orestisfl:
To avoid exiting with an error, here is a ~~simple~~ patch:

```diff
diff --git a/src/log.c b/src/log.c
index 8ef9914dc..31f913ab0 100644
--- a/src/log.c
+++ b/src/log.c
@@ -406,11 +406,17 @@ void log_new_client(EV_P_ struct ev_io *w, int revents) {
 }
 
 void log_broadcast_to_clients(const char *message, size_t len) {
-    log_client *current;
-    TAILQ_FOREACH (current, &log_clients, clients) {
-        /* XXX: In case slow connections turn out to be a problem here */
-        /* (unlikely as long as i3-dump-log is the only consumer), introduce */
-        /* buffering, similar to the IPC interface. */
-        swrite(current->fd, message, len);
+    log_client *current = TAILQ_FIRST(&log_clients);
+    while(current != TAILQ_END(&log_clients)){
+        /* XXX: In case slow connections turn out to be a problem here
+         * (unlikely as long as i3-dump-log is the only consumer), introduce
+         * buffering, similar to the IPC interface. */
+        ssize_t n = writeall(current->fd, message, len);
+        log_client *previous = current;
+        current = TAILQ_NEXT(current, clients);
+        if (n < 0) {
+            TAILQ_REMOVE(&log_clients, previous, clients);
+            free(previous);
+        }
     }
 }
```

It doesn't make sense to crash i3 because some `i3-dump-log` client is misbehaving anyway.

src/main.c
stapelberg: Done.

stapelberg:
> To avoid exiting with an error, here is a ~simple~ patch:

Thanks, merged!

orestisfl (approved):
SGTM!

orestisfl:
Oh one last thing: update the release notes

stapelberg:
> Oh one last thing: update the release notes

Done!

orestisfl:
Ok, feel free to squash merge (to 2 commits I guess) :)

PR #4331, opened 2021-01-27
 Properly quote rofi call in i3 config

ImranVirani:


orestisfl:
What error do you get? It works fine for me.

Airblader:
Seems fair, though, rofi's manpage also lists it as a comma separated list.

orestisfl:
> Seems fair, though, rofi's manpage also lists it as a comma separated list.

Replacing the comma makes the `run` menu not show up in my machine.

Airblader:
Maybe the command needs to be quoted?

ImranVirani:
Exact error is:

ERROR: Received ConfigureNotify for unknown window 0x00600000
ERROR: Received ExposeEvent for unknown window 0x00600000
ERROR: Received ConfigureNotify for unknown window 0x00600000
ERROR: Expected one of these tokens: <end>, '[', 'move', 'exec', 'restart', 'reload', 'shmlog', 'debuglog', 'border', 'layout', 'append_layout', 'workspace', 'focus', 'kill', 'open', 'fullscreen', 'stick', 'split', 'floating', 'mark', 'unmark', 'resize', 'rename', 'nop', 'scratchpad', 'swap', 'title_format', 'mode', 'bar', 'gaps'

ERROR: Your command: exec rofi -modi drun run,-show drun ( There is an underline under -show drun

It also only worked when I changed bindcode to bindsym, and removed the comma. If the comma stays and bindcode is in place of bindsym no errors are shown.

ImranVirani:
It might be distro dependant? I'm using Arctix Linux, which is arch based if that helps.

orestisfl:
> Maybe the command needs to be quoted?

In the i3 config file? Yes, and that's how I use it in my config.

ImranVirani:
@orestisfl That might be it, the command is not commented by default in the keycodes file, so it wasn't commented in mine. I just installed i3-gaps today.

etc/config.keycodes
```diff
@@ -48,7 +48,7 @@ bindcode $mod+Shift+24 kill
 # start dmenu (a program launcher)
 bindcode $mod+40 exec --no-startup-id dmenu_run
 # A more modern dmenu replacement is rofi:
-# bindcode $mod+40 exec rofi -modi drun,run -show drun
+# bindcode $mod+40 exec rofi -modi drun run -show drun
```
orestisfl: That should work:

```suggestion
# bindcode $mod+40 exec "rofi -modi drun,run -show drun"
```

orestisfl (changes requested):
Also please update this file with the same changes: https://github.com/i3/i3/blob/next/etc/config

ImranVirani:
Updated. Thanks for the help.

orestisfl (approved):

PR #4308, opened 2021-01-03
Clear framebuffer before drawing borders

tbgiles:
The framebuffer clear was done too late, causing window borders to be cleared too.

This PR fixes this by clearing the framebuffer before the call to draw the borders, instead of every time a pixmap copy happens.

Should mean that PR #4307 is no longer required.

Closes #3481 

CC: @orestisfl

PR #4336, opened 2021-01-31
Clear framebuffer for containers with visible windows

### LINKED ISSUE #3481: x_draw_decoration should fill pixmap before starting to draw

tbgiles:
From #4308, all containers were being cleared, rather than only on ones that will be showing a window. This meant that decorations in tabbed container were being cleared.

This PR clears the framebuffer on containers that contain visible windows (from the docs "meaning they represent an actual window instead of a split container").

Closes #3481

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -31,3 +31,4 @@ strongly encouraged to upgrade.
   • i3bar: properly restart status command after config change
   • i3bar: exit with 1 when a wrong command line argument is used
   • fix commented-out rofi call in default i3 config
+  • clear pixmap before drawing to prevent visual garbage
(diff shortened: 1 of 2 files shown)


orestisfl (approved):
Run the patch for a bit, seems to be good now

PR #4338, opened 2021-02-03
Move workspaces to multiple outputs

### LINKED ISSUE #4337: `move workspace to output next` instead of directional instruction for non-linear multi-monitor layouts
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
`move workspace to output right` cycles fine through my usual dual-monitor layout when the monitors are configured side-by-side.

I have recently started running a three-monitor layout with two monitors on top of each other and one to their left. Full xrandr configuration call is `xrandr --output DVI-D-0 --rotate right --mode 1600x1200 --primary --output HDMI-A-0 --right-of DVI-D-0 --mode 1920x1080 --output DisplayPort-0 --below HDMI-A-0 --mode 1920x1080`.

`move workspace to output right` or `left` will cycle between DVI-D-0 and DisplayPort-0, ignoring HDMI-A-0. `move workspace to output up` or `down` cycles between DisplayPort-0 and HDMI-A-0, ignoring DVI-D-0. There seems to be no way to cycle through *all three* outputs with just a single keybind in this layout.

## Desired Behavior
An additional argument, `next`, so `move workspace to output next` which simply cycles to the next output, rather than taking a directional cue and deciding based on that. I don't know whether this can be implemented, I imagine it would require deciding upon a cyclic relation of the enabled outputs that is rebuilt whenever the outputs are changed.

## Impact
<pre>
</pre>

## Environment
<pre>
Binary i3 version:  4.19 (2020-11-15) © 2009 Michael Stapelberg and contributors

Running i3 version: 4.19 (2020-11-15) (pid 1482)
Loaded i3 config: /home/pol/.config/i3/config (Last modified: Tue 02 Feb 2021 22:09:55 CET, 11655 seconds ago)

</pre>

<pre>
- Linux Distribution & Version: Arch Linux, kernel 5.10.11-arch1-1, Xorg 1.20.10-3, last full upgrade January 31st, 2021.
</pre>

orestisfl:
Fixes #4337

TODO:
- [x] userguide
- [x] `move container` should be updated as well for symmetry
- [x] Fix new warnings in `./GENERATED_command_call.h`

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -2406,15 +2406,18 @@ To move a container to another RandR output (addressed by names like +LVDS1+ or
 
 *Syntax*:
 ------------------------------------------------------------
-move container to output left|right|down|up|current|primary|<output>
-move workspace to output left|right|down|up|current|primary|<output>
-------------------------------------------------------------
+move container to output left|right|down|up|current|primary|next|<output1> [output2]…
+move workspace to output left|right|down|up|current|primary|next|<output1> [output2]…
+-------------------------------------------------------------------------------------
 
 *Examples*:
 --------------------------------------------------------
 # Move the current workspace to the next output
 # (effectively toggles when you only have two outputs)
-bindsym $mod+x move workspace to output right
+bindsym $mod+x move workspace to output next
+
+# Cycle this workspace between outputs VGA1 and LVDS1 but not DVI0
+bindsym $mod+x move workspace to output VGA1 LVDS1
 
 # Put this window on the presentation output.
 bindsym $mod+x move container to output VGA1
@@ -2422,6 +2425,11 @@ bindsym $mod+x move container to output VGA1
 # Put this window on the primary output.
 bindsym $mod+x move container to output primary
 --------------------------------------------------------
+
+If you specify more than one output, the container/workspace is cycled through
+them: If it is already in one of the outputs of the list, it will move to the
+next output in the list. If it is in an output not in the list, it will move to
+the first specified output. Non-existing outputs are skipped.
 
 Note that you might not have a primary output configured yet. To do so, run:
 -------------------------
--- a/src/commands.c
+++ b/src/commands.c
@@ -1024,23 +1024,113 @@ void cmd_mode(I3_CMD, const char *mode) {
 }
 
 /*
- * Implementation of 'move [window|container] [to] output <str>'.
+ * Implementation of 'move [window|container|workspace] [to] output <strings>'.
  *
  */
-void cmd_move_con_to_output(I3_CMD, const char *name) {
-    DLOG("Should move window to output \"%s\".\n", name);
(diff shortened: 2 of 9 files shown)


docs/userguide
```diff
@@ -2406,15 +2406,18 @@ To move a container to another RandR output (addressed by names like +LVDS1+ or
 
 *Syntax*:
 ------------------------------------------------------------
-move container to output left|right|down|up|current|primary|<output>
-move workspace to output left|right|down|up|current|primary|<output>
+move container to output left|right|down|up|current|primary|next|<output1> [output2]…
+move workspace to output left|right|down|up|current|primary|next|<output1> [output2]…
 ------------------------------------------------------------
```
Airblader: nit: align the dashes with the new length of the syntax?

docs/userguide
```diff
@@ -2423,6 +2426,11 @@ bindsym $mod+x move container to output VGA1
 bindsym $mod+x move container to output primary
 --------------------------------------------------------
 
+If you specify more than one outputs, the container/workspace is cycled through
```
Airblader: "more than one outputs" → "more than one output"

src/commands.c
```diff
@@ -1024,23 +1024,113 @@ void cmd_mode(I3_CMD, const char *mode) {
...
+        }
+
+        Output *current_output = get_output_for_con(ws);
+
+        Output *target_output = NULL;
+        TAILQ_FOREACH (uo, &user_output_names, user_output_names) {
+            if (strcasecmp(output_primary_name(current_output), uo->name) == 0) {
+                /* The current output is in the user list */
+                while (true) {
+                    /* This corrupts the outter loop but it is ok since we are
```
Airblader: nit: outter → outer

testcases/t/543-move-workspace-to-multiple-outputs.t
```diff
@@ -0,0 +1,93 @@
+#!perl
```
Airblader, line 1: Do we already have test cases for a single output in the list, i.e. to make sure that this works as before?

docs/userguide
Airblader: We should also mention here that non-existing outputs are skipped.

Airblader (changes requested):

testcases/t/543-move-workspace-to-multiple-outputs.t
orestisfl: Yes!

Airblader (approved):

Airblader:
@orestisfl Maybe squash once before merging?

vincentbernat:
It would be useful to also be able to use `focus output next`.

orestisfl:
Vincent, please open a new issue to track this

PR #4349, opened 2021-02-16
make vim keybindings actually be vim keybindings

smeagolthellama:
Sadly, this means $mod+h can no longer mean split horizontally, but I moved that to $mod+g, since that is nearby. It should probably be changed to be more consistent.

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -37,10 +37,10 @@ bindsym XF86AudioMicMute exec --no-startup-id pactl set-source-mute @DEFAULT_SOU
 
 # use these keys for focus, movement, and resize directions when reaching for
 # the arrows is not convenient
-set $up l
-set $down k
-set $left j
-set $right semicolon
+set $up k
+set $down j
+set $left h
+set $right l
 
 # use Mouse+Mod1 to drag floating windows to their wanted position
 floating_modifier Mod1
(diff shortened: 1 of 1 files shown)


Airblader:
Hi,

thanks for the PR! Unfortunately I can't see where this change would be coming from (is there an issue for this?). I don't think we're going to want to change these long-standing defaults, so I'm going to close this.

gerrywastaken:
Keywords: i3wm default key bindings vim "hjkl" "jkl;"  hjkl jkl; vim key bindings.

This confused me for so long. I didn't realise they were shifted over one key and just assumed that internally i3 was representing the stack in a different way to how it was visually. :laughing: :facepalm: 

For others, these are the three sections you will want to update:

```
# change focus
bindsym $mod+h focus left
bindsym $mod+j focus down
bindsym $mod+k focus up
bindsym $mod+l focus right
```

```
# move focused window
bindsym $mod+Shift+h move left
bindsym $mod+Shift+j move down
bindsym $mod+Shift+k move up
bindsym $mod+Shift+l move right
```

```
# split in horizontal orientation
bindsym $mod+g split h
```

> I don't think we're going to want to change these long-standing defaults, so I'm going to close this.

Well you say that, but the only time these defaults kick in is when people are first generating their config and after that it's set in stone until they change it. So I believe this really only affects new users. Most of those new users who also want to use vim directions, probably wouldn't expect this:

```
down to go left
up to go down
right to go up
and left to not work
```

This is baked into the muscle memory of most vim users and this "default" is messing with it. :)

Either way, at least it's simple to change when people realise what's actually happening. Maybe most notice the shift when they first look at the docs. I didn't. :facepalm:

PR #4360, opened 2021-03-05
Remove unnecessary calls to xcb_flush()

psychon:
I did a little exorcism and removed some calls to `xcb_flush()`:
```
 18 files changed, 5 insertions(+), 58 deletions(-)
```

My general strategy is:
- Before blocking (e.g. explicitly via `xcb_wait_for_event()` or implicitly by returning to the event loop), pending requests should be sent / `xcb_flush()` should always be called. The code was mostly already doing this.
- Since the above ensures that requests will be flushed, all other calls to `xcb_flush()` are redundant and can be removed.

I started with many small commits but then got lazy when I reached the "main code base".

TODO:
- [ ] ] I saw the small paragraph in `docs/hacking-howto` about calling `xcb_flush()`, but I am not sure what to do with it.
- [ ] I am not sure I got the IPC stuff right. I *think* your code is racy, but I am not entirely sure:
    - I had a test suite failure where the `_NET_CURRENT_DESKTOP` property was outdated. What I think happened: Previously, the code first `xcb_flush()`d and then replied to the ipc request from the test. However, this is racy. Nothing guarantees that the X11 server already handled i3's `ChangeProperty` request before the test suite managed to query the property. However, I guess it is very, very unlikely for i3 to loose this race.
    - To paper over this bug, I added a call to `xcb_aux_sync()` and left a TODO comment. Someone who knows this IPC code better than me should try to explain why the race I described above is not possible.
    - It does not help that I failed to find any documentation for the `sync` IPC command not do I really understand the test suite's `sync_with_i3`. Either I really managed to miss a required `xcb_flush()` or the syncing here is somehow broken...
- [ ] I did not dare to touch the sequence `xcb_ungrab_server(conn); xcb_flush(conn);`. I think that flush is not necessary, but I do not dare to actually remove it. Feel free to just leave them there. (The same for `xcb_ungrab_pointer()` and `xcb_ungrab_keyboard()` followed by a flush.)
- [x] I did not touch `testcases/` (nor do I understand the code in there)
- [ ] Some calls to `xcb_flush()` look fishy to me, but a comment claims that they are there for a reason. I doubt that comment, but okay... https://github.com/i3/i3/blob/7da136dca494073a5de9a38df9eb1ee220cf6006/src/x.c#L992-L1001

### DIFF
--- a/i3-config-wizard/main.c
+++ b/i3-config-wizard/main.c
@@ -926,7 +926,6 @@ int main(int argc, char *argv[]) {
     draw_util_surface_init(conn, &surface, win, get_visualtype(root_screen), WIN_WIDTH, WIN_HEIGHT);
 
     /* Grab the keyboard to get all input */
-    xcb_flush(conn);
 
     /* Try (repeatedly, if necessary) to grab the keyboard. We might not
      * get the keyboard at the first attempt because of the keybinding
--- a/i3-config-wizard/main.c
+++ b/i3-config-wizard/main.c
@@ -535,8 +535,6 @@ static int handle_expose(void) {
         txt(5, 10, "<ESC>", red, black);
     }
 
-    xcb_flush(conn);
-
     return 1;
 }
 
@@ -565,7 +563,6 @@ static int handle_key_press(void *ignored, xcb_connection_t *conn, xcb_key_press
                                 8,
                                 strlen("i3: generate config"),
                                 "i3: generate config");
-            xcb_flush(conn);
         } else
             finish();
     }
@@ -986,6 +983,7 @@ int main(int argc, char *argv[]) {
         }
 
         free(event);
+        xcb_flush(conn);
     }
 
     /* Dismiss drawable surface */
--- a/i3-input/main.c
+++ b/i3-input/main.c
@@ -138,8 +138,6 @@ static int handle_expose(void *data, xcb_connection_t *conn, xcb_expose_event_t 
         i3string_free(input);
     }
 
-    xcb_flush(conn);
-
     return 1;
 }
 
@@ -532,6 +530,7 @@ int main(int argc, char *argv[]) {
         }
 
         free(event);
+        xcb_flush(conn);
     }
 
     draw_util_surface_free(conn, &surface);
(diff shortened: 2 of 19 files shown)


psychon:
> I had a test suite failure where the _NET_CURRENT_DESKTOP property was outdated. 

Well... this one thing is definitely a broken test:
https://github.com/i3/i3/blob/7da136dca494073a5de9a38df9eb1ee220cf6006/testcases/t/294-update-ewmh-atoms.t#L101-L104
This asks i3 to close a window via IPC and then talks to the X11 server to check the result. Nothing here guarantees that the X11 server already handled "all the stuff" that happened to close the window. No idea if the window in question supports `WM_DELETE_WINDOW`, but if it does, then not even `sync_with_i3` would be enough here since the client can take arbitrarily long deal with the `WM_DELETE_WINDOW` message. (well, there is a call to `sync_with_i3` in `get_current_desktop` already.)

psychon:
This now failed again on Travis:
```
#   Failed test 'Received 1 workspace::init event'
#   at /usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/533-randr15.t line 141.
#          got: '0'
#     expected: '1'
#   Failed test 'Workspace initialized in up2414q'
#   at /usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/533-randr15.t line 142.
#          got: undef
#     expected: 'up2414q'
```
My guess here is that the X11 server did not yet handle i3's "please send me RandR events"-event...? (And I think nothing in there guarantees that, doesn't `launch_with_config` or `activate_i3` need a `sync_with_i3` call to ensure that startup is really complete? Right now it only sends an IPC request, but that does not sync with the X11 server.  Hm, but it does so in this PR since I added a call to `xcb_aux_sync()` in IPC handling. Oh, wow, but the code uses `get_tree` to sync with i3 here and I did not add a call to `xcb_aux_sync()` there. I think your test suite needs some fixing...)

psychon:
Closing this since this is quite a can of worms that I opened. The test suite needs to be fixed before one can remove `xcb_flush()` calls (or rather: I have to understand how its synchronisation works, see #4365).

Also, this is high risk and low gain. This only gets rid of some syscalls and a lock+unlock of a mutex. The gains will not be measurable.

stapelberg:
> Also, this is high risk and low gain. This only gets rid of some syscalls and a lock+unlock of a mutex. The gains will not be measurable.

I agree with your assessment here. Thanks for taking a stab at it!

PR #4361, opened 2021-03-05
No temporary cairo surface for font drawing

psychon:
Instead of creating a temporary cairo surface in `draw_text()`, this now requires a cairo surface from the caller. That seems more sane to me than what the code currently does.

-------

CC @mychris I did not try reproducing your problem, but if I understood it correctly, then this might fix it. The idea here is that cairo now actually knows the "real" size of the drawable and is not lied to with a smaller size.

(Regardless of whether this fixes the issue from #4357 I think this is worth merging)

### DIFF
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -452,13 +452,6 @@ void draw_text(i3String *text, xcb_drawable_t drawable, xcb_gcontext_t gc,
                xcb_visualtype_t *visual, int x, int y, int max_width);
 
 /**
- * ASCII version of draw_text to print static strings.
- *
- */
-void draw_text_ascii(const char *text, xcb_drawable_t drawable,
-                     xcb_gcontext_t gc, int x, int y, int max_width);
-
-/**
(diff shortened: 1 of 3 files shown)


mychris:
Yes, this fixes my the problem in #4357. I think the fix does what I meant by saying:

> I start to believe that the best solution would be, to always draw the text last, so it can be drawn above the border/spacer area, if the font has such a layout that the ink extent is outside the logical extent.

Thanks for looking into this.

mychris:
Just noticed one little thing. The right most item in the status bar is still clipped. I guess this might be because for the [Caveat](https://fonts.google.com/specimen/Caveat?preview.text_type=custom) font, the ink extent is partially outside of the logical extent?

psychon:
I don't really know, but I would expect that this is the "desired behaviour" for when a letter "reaches outside" of its logical extents (so that the ink extents are larger than the logical extents, but the logical ones are still used for layout).

Airblader (approved):
Thanks!

PR #4362, opened 2021-03-05
Some follow-up to 

psychon:
This is some follow-up to #4361 (and includes the two commits from that PR, hence the draft status).

I'm not really sure if the "stuff" around the calls to `cairo_surface_flush()` are really worth it. Missing a call to `cairo_surface_flush()` can lead to weird bugs that no one normally sees, but then someone has a weird X11 server that does not support the RENDER extension (or only an old version of it) and things break. One can test stuff with `CAIRO_DEBUG="xrender-version=-1.-1"`, but meh... I don't know.

If deemed too risky, I can get rid of the commits around `cairo_surface_flush()`. This part seemed like a better idea when I started it...

### DIFF
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -566,8 +566,6 @@ typedef struct surface_t {
     /* A classic XCB graphics context. */
     xcb_gcontext_t gc;
 
-    xcb_visualtype_t *visual_type;
-
     int width;
     int height;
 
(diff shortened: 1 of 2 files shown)


psychon:
Rebased on `next` and removed all the scary stuff around `cairo_surface_flush`. Previously this PR pointer to https://github.com/i3/i3/pull/4362/commits/88bb7350644d30fdd66a9ca91be8c993456bfde7 and its parents. Now I only left the one simple commit in this PR. The rest is too complicated and has too much potential for breakage.

Airblader (approved):
Thanks!

PR #4367, opened 2021-03-07
Batch updating the keyboard state / regrabbing keys

### LINKED ISSUE #3924: ibus with input method not shared among windows causes massive re-grabbing churn when switching windows
Hello. When using **ibus** launched by `ibus-daemon -drx` from the `.xinitrc`, with the option to **"Share the same input method among all applications" disabled**, switching windows causes noticable lag. This is especially annoying on my low-powered laptop where it can take up to half a second to switch windows.

The cause is apparently i3 re-grabbing all bindings _many_ times over on every switch.
Attached is a log file capturing the situation. Note it is over 11 000 lines long, generated by a single window switch.

[i3log.txt](https://github.com/i3/i3/files/4131497/i3log.txt) | [https://clbin.com/ML9na](https://clbin.com/ML9na)

I have reproduced this issue both with the default config and my own, with the `i3-gaps` package currently in archlinux repos, the `gaps-next` master branch, the `i3` master branch and have been experiencing it for over a year on multiple machines. As such, I believe the specific version is irrelevant to the problem. I don't know whether any particular version of i3 first introduced the issue, as it was already present when I started using ibus. Nevertheless, this is the version of i3 which generated the attached log:
<pre>Binary i3 version:  4.17.1-135-g07d5fc76 (2020-01-18, branch "gaps-next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.17.1-135-g07d5fc76 (2020-01-18, branch "gaps-next") (pid 2154)
Loaded i3 config: /home/sammko/.config/i3/config (Last modified: Tue 28 Jan 2020 21:32:32 CET, 98729 seconds ago)

</pre>

psychon:
This adds a test case that tests that i3 does something quickly. With current `next`, the test needs 13 seconds on my computer and fails. That's not quick. The following commits fix this by batching "the keyboard changed"-stuff: Instead of executing this code once per event we receive, this is now only done once per main loop iteration.

This might or might not fix #3924 (no idea - I only tested with my new test case), hence CC @sammko 

Also, I have no clue how to write test cases. I do not really know perl and I did not figure out the numbering scheme (thus went with 600). Perhaps this even has to handle the case "`xmodmap` is not installed" somehow? Dunno.

The test uses a time limit of two seconds because `time()` only provides second granularity. Limiting this to one second could cause failures where the current time is `x.99999` when the test starts and `(x+1).000001` when the test finishes. This would already look like a one second time difference. Allowing two seconds thus seemed the safest option to me.

Finally: IMO I am using way too many static variables for this, but let's see what you guys say about this.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -81,6 +81,24 @@ bool event_is_ignored(const int sequence, const int response_type) {
     return false;
 }
 
+typedef enum { RG_DEFAULT,
+               RG_TRANSLATE } regrab_mode_t;
+
+/*
+ * Called when the keyboard state changed and we need to re-grab keys.
+ */
+static void regrab_keys(regrab_mode_t mode) {
+    ungrab_all_keys(conn);
+    switch (mode) {
+    case RG_DEFAULT:
+        break;
+    case RG_TRANSLATE:
+        translate_keysyms();
+        break;
+    }
+    grab_all_keys(conn);
+}
+
 /*
  * Called with coordinates of an enter_notify event or motion_notify event
  * to check if the user crossed virtual screen boundaries and adjust the
@@ -246,9 +264,7 @@ static void handle_mapping_notify(xcb_mapping_notify_event_t *event) {
 
     xcb_numlock_mask = aio_get_mod_mask_for(XCB_NUM_LOCK, keysyms);
 
-    ungrab_all_keys(conn);
-    translate_keysyms();
-    grab_all_keys(conn);
+    regrab_keys(RG_TRANSLATE);
 }
 
 /*
@@ -1298,9 +1314,7 @@ void handle_event(int type, xcb_generic_event_t *event) {
             keysyms = xcb_key_symbols_alloc(conn);
             if (((xcb_xkb_new_keyboard_notify_event_t *)event)->changed & XCB_XKB_NKN_DETAIL_KEYCODES)
                 (void)load_keymap();
-            ungrab_all_keys(conn);
-            translate_keysyms();
-            grab_all_keys(conn);
+            regrab_keys(RG_TRANSLATE);
         } else if (state->xkbType == XCB_XKB_MAP_NOTIFY) {
             if (event_is_ignored(event->sequence, type)) {
                 DLOG("Ignoring map notify event for sequence %d.\n", state->sequence);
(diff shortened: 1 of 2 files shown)


psychon:
Hrm. The test definitely needs some more work. It just failed locally for me and I have no idea why:
```
#   Failed test 'Test finishes quickly'
#   at /home/psychon/projects/i3/testcases/t/600-slow-keyboard-update.t line 30.
```
Someone who knows perl & the test suite is invited to look at this and improve it somehow (a good first step would be to improve the error message to include the value of `$delay`). AFAIK you can just push new commits to the branch behind this PR.

Also, another test failed on Travis, but I cannot reproduce this locally.
```
#   Failed test 'Received precisely one event'
#   at /usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/290-keypress-numlock.t line 86.
#          got: '0'
#     expected: '1'
#   Failed test 'change is "run"'
#   at /usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/290-keypress-numlock.t line 86.
#          got: undef
#     expected: 'run'
#   Failed test 'triggered the "a" keybinding'
#   at /usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/290-keypress-numlock.t line 86.
#          got: undef
#     expected: 'a'
```
A random guess would be: The test uses `xtest_sync_with_i3` which seems to work by sending an `I3_SYNC` message to the root window. My use of an ev idle watcher seems to be breaking the synchronisation here (since an `I3_SYNC` message could be handled before the idle watcher fires).

Urgh. Another synchronisation issue that I stepped into. :-(

Edit: Actually, its worse than that. The test (de)activates num lock, which means that keys are regrabbed. What prevents the following sequence of events?

- the test suite activates num lock
- i3 does `ungrab_all_keys()`
- the test suite presses `a`, expecting a binding to trigger
- i3 re-grabs the keys it is interested in

The above is quite unlikely in the current tests since all the xtest requests should be sent with a single `write()` syscall, but this still seems potentially racy.

Perhaps this needs a `xtest_sync_with_i3` after activating num lock and hoping that this gives enough time to the idle watcher to fire? Hrm...

Issue #4365, opened 2021-03-06
The sync protocol could be described better / I do not understand it


psychon:
Hi,

yesterday I rambled a bit about the sync protocol and said that I did not find any docs on it: #4360
(Sorry for all that text)

Today, I found https://i3wm.org/docs/testsuite.html#i3_sync
However, I do not understand this image:
https://i3wm.org/docs/i3-sync-working.png

This shows the sync protocol working:
1. The tests send IPC "focus left" to i3
2. The tests send I3_SYNC to X11
3. The X11 server relays the I3_SYNC to i3
4. i3 handles the IPC "focus left"
5. i3 handles the i3_SYNC

However, what prevents the following sequence of events? (5 happening before 4 in the sequence above)
1. The tests send IPC "focus left" to i3
2. The tests send I3_SYNC to X11
3. The X11 server relays the I3_SYNC to i3
4. i3 handles the i3_SYNC
5. i3 handles the IPC "focus left"

i3 handles the I3_SYNC from the X11 server before it gets the IPC "focus left". As far as I can tell, there is nothing that prevents this kind of race.
(Or to try to make it more explicit: There is a race between i3 handles the IPC "focus left" and replying to the I3_SYNC message. Or yet another way: Nothing establishes a happens-before relationship between these events.)

To actually prevent it, a different kind of protocol would be needed: The tests send something to via IPC to i3, i3 sends that to X11 and X11 sends to the tests.

Is this description somewhat clear? Synchronisation problems are kind of a hobby of mine, but I might have some problems making my thoughts clear.

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

psychon:
In the code, I also see an IPC `sync` command. I am not quite sure how it works (I have no prior experience with yajl), but it seems like it does what is needed here: It is an IPC command (thus ensuring that all previous IPC commands have already been executed) that sends an `I3_SYNC` message to an X11 window (thus ensuring that the X11 server already handled all previous X11 requests from i3).

Is there a reason why this is not documented at the same place as the other I3_SYNC mechanism? I also did not find anything making use of this IPC command (but I am not good at looking for it).

- Also, does my reasoning about make sense?
- If so, should I try fixing the races in the test suite?
[...]

Issue #4327, opened 2021-01-23
enhancement: documentation: translations and versioned sections, move to markdown


stapelberg:
**Status:** https://github.com/i3/translation and https://github.com/stapelberg/i3/commits/markdown are where things stand. I’m not currently actively working on this.

---

I have been thinking about how to support community-contributed translations of i3 documentation. In particular, I see the following constraints:

* Translating a whole document can be a lot of work, so partial translations should be easily possible.
* Documents change, so translations must be marked as out of date, ideally on a per-section level.
* Additional tooling should be preferred over additional hosted services. While there are plenty of services for **string translation** within a program (e.g. weblate), I wasn’t able to find a service that would help with **document translation** anyway.

## Proposal

I propose we convert our current asciidoc documentation into markdown.

Using markdown will allow us to easily write tools that make translations easier to manage for our release/contribution workflow.

As a side benefit, over the last few years, markdown has strongly overtaken asciidoc in popularity, so contributors will have an easier time, too.

Once the documentation is in markdown, we will **annotate headings with the translation status**, for example:

```
### Mouse bindings {#_mouse_bindings introduced="4_16" translated="4_18"}

A mouse binding makes i3 execute a command upon pressing a specific mouse…
```

We then use a custom markdown renderer (built on the well-established [goldmark](https://pkg.go.dev/github.com/yuin/goldmark) package) which converts our custom attributes into the desired HTML output.

You can find an initial sketch of `tl8-render` in the https://github.com/i3/translation repository.

Other markdown tools will still be able to make sense of our documents, but translation status will only be rendered when using our tool of course.

For our tooling to work, **all** headings must have an id, and that id must be consistent across translations. For switching from asciidoc to markdown, we already must ensure that all headings have the same id as they had before, in order to not break any (external or internal) links.

Translations will live in the https://github.com/i3/i3.github.io website repository, because they are based on released documentation, not in-development documentation.

As a nice bonus, versioned documentation as discussed in issue #2563 will become trivial to implement with this approach.

### Example: starting a new translation

```
git cp _docs/userguide.markdown _docs/fr/userguide.markdown
tl8-new _docs/fr/userguide.markdown  # marks all sections as untranslated
```

### Example: new i3 release / out-of-date sections

When we do a new release, our release.sh automation would call `tl8-flag` like so:

```
tl8-flag _docs/userguide.markdown  # flags changed sections as out of date in _docs/*/userguide.markdown 
```

[...]

PR #4370, opened 2021-03-09
Improve sync docs in the testsuite

psychon:
Here is my attempt at fixing #4365. Please double check what I did here, because:

- I am describing something that I learnt in the last few days by reading source code (no tests were run!). Thus, it is possible that my understanding is wrong.
- I had to use `dia` to change images. I have no idea what I am doing here (but felt that it is necessary to add the IPC response to the images). (Please don't make me touch `dia` again...)
- I have no idea what format is used for these docs and was too lazy to figure this out. Thanks to this I noticed that I am way too used to markdown these days.


docs/testsuite
```diff
@@ -569,7 +569,35 @@ i3 will then reply with the same ClientMessage, sent to the window specified in
...
+This protocol is used, for example, in t/525-i3bar-mouse-bindings.t: A mouse
+button press on i3bar is triggered. i3bar reacts to this by sending IPC commands
+to i3.
+
+The necessary synchronisation is achieved by sending an I3_SYNC event to i3bar:
+Because i3bar reacts with a sync IPC command to i3, all previous IPC commands from
+i3bar will be handled first. Because i3 reacts via X11, all previous X11
+requests from i3 will be handled by the X11 server first.
+
+The actual test also has to sync with i3 first due to how X11 handling works. If
```
Airblader: Can we please phrase this in a less dismissive way? :-)

> For more details refer to the documentation for XAllowEvents with mode ReplayPointer

Airblader (commented):
The images were previously *.png, now *.dia – how come? GH can't open them, so I haven't yet looked at them.

psychon:
> The images were previously *.png, now *.dia – how come? 

The images were always `*.dia` plus a `*.png` exported from `dia`. But GH is clearly now showing this properly. Its claim of "empty file" does not help...

This is how `git show` shows the situation:
```
diff --git a/docs/i3-sync-working.dia b/docs/i3-sync-working.dia
index 9f1c3bc8..25803828 100644
Binary files a/docs/i3-sync-working.dia and b/docs/i3-sync-working.dia differ
diff --git a/docs/i3-sync-working.png b/docs/i3-sync-working.png
index dce44ac5..419334d0 100644
Binary files a/docs/i3-sync-working.png and b/docs/i3-sync-working.png differ
diff --git a/docs/i3-sync.dia b/docs/i3-sync.dia
index 0945ae26..c8fb8359 100644
Binary files a/docs/i3-sync.dia and b/docs/i3-sync.dia differ
diff --git a/docs/i3-sync.png b/docs/i3-sync.png
index b64cce25..c9a5a01c 100644
Binary files a/docs/i3-sync.png and b/docs/i3-sync.png differ
```

docs/testsuite
psychon: Thanks for the suggestion. Done. (In a new commit - dunno if you want me to amend the existing commit instead.)

Airblader (commented):
I actually just read GH wrong – red color for indicating reduced file size. OK then.

Wondering if @stapelberg wants to give this a read? It all sounds good to me but frankly I'm not very deeply invested in this sync protocol.

stapelberg (approved):

stapelberg:
Looks good, thanks for this!

> * I have no idea what format is used for these docs and was too lazy to figure this out. Thanks to this I noticed that I am way too used to markdown these days.

It’s asciidoc, but we’re switching to markdown in https://github.com/i3/i3/issues/4327 :)

PR #4372, opened 2021-03-10
Remove a useless call to draw_util_surface_set_size()

psychon:
There was just a call to draw_util_surface_init() with the same
width/height just some lines above. Re-setting the size does nothing.


orestisfl:
One seems to be about `->frame` and the other about `->frame_buffer`, did you notice that?

psychon:
And I thought I triple-checked this... Sigh.

Sorry for the noise!

PR #4373, opened 2021-03-10
Do not "set" the wallpaper during startup

### LINKED ISSUE #2869: X applications launched before i3 are buggy
```
Running i3 version: 4.13 (2016-11-08) (pid 8678)o abort…)
Loaded i3 config: /home/hadrien/.config/i3/config (Last modified: Mon 21 Aug 2017 10:24:21 AM CEST, 418 seconds ago)

```

https://files.catbox.moe/t13m4i

**What I did:**
Start applications like this:
```
> tail -n5 .xinitrc
st -g 80x32+50+320 -c "floating" -e mutt &
st -g 80x32+50+847 -c "floating" -e newsbeuter &
st -g 92x65+705+320 -c "floating" -e cmus &
sleep 3
exec i3 -V >> ~/i3log-$(date +'%F-%k-%M-%S') 2>&1
```
The `sleep 3` makes it reproducible. The first time launching X after booting the machine doesn't need this to get the bug.

**What I saw:**
https://files.catbox.moe/lp0wbk.webm
`xwininfo`/`xprop` can't detect those leftover windows.

**What I expected instead:**
Well, no workspace wide zombie windows.

### LINKED ISSUE #4371: i3 replaces my wallpaper with a screenshot
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

[I removed most of the rest of the template since it makes more sense to start with the reproduction instructions. I hope that's okay.]

## What I did

I read some source code and wondered "wait, what?!". Then I did the following just so to trigger this issue.
```sh
Xephyr -listen tcp :1  & DISPLAY=:1 sh -c '(sleep 1 ; xterm & sleep 1 ; (sleep 1 ; pkill xterm) & i3 -c /dev/null)'
```
Result:
foo

## What is wrong

That's not an xterm. That's a screenshot of an xterm. As my wallpaper.

## How to make this into a slightly less weird issue

Imagine I set my wallpaper not from the i3 config, but before starting i3. Together with some transparency, I bet I could turn this into an actual "bad rendering" problem...

## Why this happens

Edit: This goes back to 4f5e0e794 from 2013.

https://github.com/i3/i3/blob/af2171d01d40613ec8e6eddcc148722b0ec248f9/src/main.c#L979-L998

(The above is btw not the best way to set a wallpaper since one cannot query the background of a window. Hence, wallpaper setters are expected to set the `ESETROOT_PMAP_ID` and/or `XROOTPMAP_ID` properties of the root window. But doing this properly requires a separate X11 connection since you have to actually leak the pixmap so that others can still use it. Which means you need to `SetCloseDownMode` to prevent the X11 server from destroying the pixmap when you disconnect. It is complicated.)

## What I would do about this

Just remove that piece of code and tell people to properly set a wallpaper.

## Environment

<pre>
i3 version: 
Binary i3 version:  4.19.1-non-git © 2009 Michael Stapelberg and contributors
Running i3 version: 4.19.1-non-git (pid 709)
Loaded i3 config: /dev/null (Last modified: Sa 06 Mär 2021 08:41:19 CET, 361408 seconds ago)

</pre>
Why the heck does it claim that my config is `/dev/null`? That's the last config that I loaded, but not in this X11 server.... ARGH, it broke my IPC socket again. Unrelated to this issue, but: How can I start more than one i3 process without having my IPC socket break all the time? :-(

<details><summary>Config file</summary><pre>
I explicitly used /dev/null, just for this template.
</pre>
</details>

<pre>
Logfile URL:
</pre>

<pre>
- Linux Distribution & Version: Debian testing
</pre>

psychon:
This basically reverts 4f5e0e7: During startup, i3 would take a
screenshot and set that as the background pixmap of the root window.
This is the easy part of setting an X11 wallpaper.

The code in question was added because something either set the
background pixmap of the root window to NONE or the X11 server was
started with "-background none". I'm tempted to say "the people in
question get what they asked for", but perhaps telling them to "exec
--no-startup-id xsetroot -solid black" in their config is a better
reaction.

(I have not found any issue / bug report from "back then" which caused the code
in question to be added, so I cannot really say more.)

Fixes: https://github.com/i3/i3/issues/4371
Fixes: https://github.com/i3/i3/issues/2869

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -8,18 +8,17 @@ strongly encouraged to upgrade.
 
 Background/wallpaper workaround:
 
-Some login managers (e.g. gdm) start X11 with the -background none flag. When
-this flag is set, a background needs to be explicitly set later in the X11
-session, otherwise stale copies of closed windows remain visible on the X11 root
-window (symptom looks like “my terminal window is not closing”).
+Some login managers (e.g. gdm) start the X11 server with the -background none
+flag. When this flag is set, a background needs to be explicitly set later in
+the X11 session, otherwise stale copies of closed windows remain visible on the
+X11 root window (symptom looks like “my terminal window is not closing”).
 
-i3 works around this situation by setting a screenshot of the X11 root window as
-background when starting. Any background you set before starting i3 (e.g. in
-your Xsession) or after starting i3 (e.g. via exec statements in the i3 config)
-will be visible.
+i3 works around this situation by setting a screenshot as background when
+starting. Any background you set before starting i3 (e.g. in your Xsession) or
+after starting i3 (e.g. via exec statements in the i3 config) will be visible.
 
-A downside of this workaround is that if you have any windows open in your X11
-session, those will be part of the screenshot.
+A downside of this workaround is that if you have any windows already open in
+your X11 session, those will be part of the screenshot.
 
 To fix this issue, starting in v4.20, i3 detects whether the -background none
 option is enabled and only then sets a screenshot as background.
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -6,6 +6,13 @@
 This is i3 v4.19. This version is considered stable. All users of i3 are
 strongly encouraged to upgrade.
 
+i3 used to unconditionally take a screenshot and set it as the wallpaper when
+initially starting up. This was done to work around problems in case no
+wallpaper is set, but caused problems if any windows were visible in the
+screenshot. This work around is now only applied if no wallpaper is detected. If
+you not set a wallpaper, it is recommended to add exec --no-startup-id xsetroot
+-solid black to your config.
+
 
  ┌────────────────────────────┐
  │ Changes in i3 v4.20        │
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -6,12 +6,23 @@
 This is i3 v4.19. This version is considered stable. All users of i3 are
 strongly encouraged to upgrade.
 
-i3 used to unconditionally take a screenshot and set it as the wallpaper when
-initially starting up. This was done to work around problems in case no
-wallpaper is set, but caused problems if any windows were visible in the
-screenshot. This work around is now only applied if no wallpaper is detected. If
-you not set a wallpaper, it is recommended to add exec --no-startup-id xsetroot
--solid black to your config.
+Background/wallpaper workaround:
+
+Some login managers (e.g. gdm) start X11 with the -background none flag. When
+this flag is set, a background needs to be explicitly set later in the X11
+session, otherwise stale copies of closed windows remain visible on the X11 root
+window (symptom looks like “my terminal window is not closing”).
+
+i3 works around this situation by setting a screenshot of the X11 root window as
+background when starting. Any background you set before starting i3 (e.g. in
+your Xsession) or after starting i3 (e.g. via exec statements in the i3 config)
+will be visible.
+
+A downside of this workaround is that if you have any windows open in your X11
+session, those will be part of the screenshot.
+
+To fix this issue, starting in v4.20, i3 detects whether the -background none
+option is enabled and only then sets a screenshot as background.
 
 
  ┌────────────────────────────┐
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -655,3 +655,8 @@ int create_socket(const char *filename, char **out_socketpath);
  *
  */
 bool path_exists(const char *path);
+
+/**
+ * Grab a screenshot of the screen's root window and set it as the wallpaper.
+ */
+void set_screenshot_as_wallpaper(xcb_connection_t *conn, xcb_screen_t *screen);
--- a/include/libi3.h
+++ b/include/libi3.h
@@ -660,3 +660,11 @@ bool path_exists(const char *path);
  * Grab a screenshot of the screen's root window and set it as the wallpaper.
  */
 void set_screenshot_as_wallpaper(xcb_connection_t *conn, xcb_screen_t *screen);
+
+/**
+ * Test whether the screen's root window has a background set.
+ *
+ * This opens & closes a window and test whether the root window still shows the
+ * content of the window.
+ */
+bool is_background_set(xcb_connection_t *conn, xcb_screen_t *screen);
--- a/libi3/is_background_set.c
+++ b/libi3/is_background_set.c
@@ -1,0 +1,117 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3 - an improved dynamic tiling window manager
+ * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
+ *
+ */
+#include "libi3.h"
+
+#include <xcb/xcb_aux.h>
+
+/**
+ * Find the region in the given window that is not covered by a mapped child
+ * window.
+ */
+static cairo_region_t *unobscured_region(xcb_connection_t *conn, xcb_window_t window,
+                                         uint16_t window_width, uint16_t window_height) {
+    cairo_rectangle_int_t rectangle;
+    cairo_region_t *region;
+
+    rectangle.x = 0;
+    rectangle.y = 0;
+    rectangle.width = window_width;
+    rectangle.height = window_height;
+    region = cairo_region_create_rectangle(&rectangle);
+
+    xcb_query_tree_reply_t *tree = xcb_query_tree_reply(conn, xcb_query_tree_unchecked(conn, window), NULL);
+    if (!tree) {
+        return region;
+    }
+
+    /* Get information about children */
+    uint16_t n_children = tree->children_len;
+    xcb_window_t *children = xcb_query_tree_children(tree);
+
+    xcb_get_geometry_cookie_t geometries[n_children];
+    xcb_get_window_attributes_cookie_t attributes[n_children];
+
+    for (int i = 0; i < n_children; i++) {
+        geometries[i] = xcb_get_geometry_unchecked(conn, children[i]);
+        attributes[i] = xcb_get_window_attributes_unchecked(conn, children[i]);
+    }
+
+    /* Remove every visible child from the region */
+    for (int i = 0; i < n_children; i++) {
+        xcb_get_geometry_reply_t *geom = xcb_get_geometry_reply(conn, geometries[i], NULL);
+        xcb_get_window_attributes_reply_t *attr = xcb_get_window_attributes_reply(conn, attributes[i], NULL);
+
+        if (geom && attr && attr->map_state == XCB_MAP_STATE_VIEWABLE) {
+            rectangle.x = geom->x;
+            rectangle.y = geom->y;
+            rectangle.width = geom->width;
+            rectangle.height = geom->height;
+            cairo_region_subtract_rectangle(region, &rectangle);
+        }
+
+        free(geom);
+        free(attr);
+    }
+
+    free(tree);
+    return region;
+}
+
+static void find_unobscured_pixel(xcb_connection_t *conn, xcb_window_t window,
+                                  uint16_t window_width, uint16_t window_height,
+                                  uint16_t *x, uint16_t *y) {
+    cairo_region_t *region = unobscured_region(conn, window, window_width, window_height);
+    if (cairo_region_num_rectangles(region) > 0) {
+        /* Return the top left pixel of the first rectangle */
+        cairo_rectangle_int_t rect;
+        cairo_region_get_rectangle(region, 0, &rect);
+        *x = rect.x;
+        *y = rect.y;
+    } else {
+        /* No unobscured area found */
+        *x = 0;
+        *y = 0;
+    }
+    cairo_region_destroy(region);
+}
+
+static uint32_t flicker_window_at(xcb_connection_t *conn, xcb_screen_t *screen, int16_t x, int16_t y, xcb_window_t window,
+                                 uint32_t pixel) {
+    xcb_create_window(conn, XCB_COPY_FROM_PARENT, window, screen->root, x, y, 10, 10,
+            0, XCB_WINDOW_CLASS_INPUT_OUTPUT, XCB_COPY_FROM_PARENT,
+            XCB_CW_BACK_PIXEL | XCB_CW_OVERRIDE_REDIRECT, (uint32_t[]) { pixel, 1 });
+    xcb_map_window(conn, window);
+    xcb_clear_area(conn, 0, window, 0, 0, 0, 0);
+    xcb_aux_sync(conn);
+    xcb_destroy_window(conn, window);
+
+    xcb_get_image_reply_t *img = xcb_get_image_reply(conn,
+            xcb_get_image_unchecked(conn, XCB_IMAGE_FORMAT_Z_PIXMAP, screen->root, x, y, 1, 1, ~0),
+            NULL);
+    uint32_t result = 0;
+    if (img) {
+        uint8_t *data = xcb_get_image_data(img);
+        uint8_t depth = img->depth;
+        for (int i = 0; i < MIN(depth, 4); i++) {
+            result = (result << 8) | data[i];
+        }
+        free(img);
+    }
+    return result;
+}
+
+bool is_background_set(xcb_connection_t *conn, xcb_screen_t *screen) {
+    uint16_t x, y;
+    find_unobscured_pixel(conn, screen->root, screen->width_in_pixels, screen->height_in_pixels, &x, &y);
+
+    xcb_window_t window = xcb_generate_id(conn);
+
+    uint32_t pixel1 = flicker_window_at(conn, screen, x, y, window, screen->black_pixel);
+    uint32_t pixel2 = flicker_window_at(conn, screen, x, y, window, screen->white_pixel);
+    return pixel1 == pixel2;
+}
--- a/libi3/is_background_set.c
+++ b/libi3/is_background_set.c
@@ -81,18 +81,18 @@ static void find_unobscured_pixel(xcb_connection_t *conn, xcb_window_t window,
 }
 
 static uint32_t flicker_window_at(xcb_connection_t *conn, xcb_screen_t *screen, int16_t x, int16_t y, xcb_window_t window,
-                                 uint32_t pixel) {
+                                  uint32_t pixel) {
     xcb_create_window(conn, XCB_COPY_FROM_PARENT, window, screen->root, x, y, 10, 10,
-            0, XCB_WINDOW_CLASS_INPUT_OUTPUT, XCB_COPY_FROM_PARENT,
-            XCB_CW_BACK_PIXEL | XCB_CW_OVERRIDE_REDIRECT, (uint32_t[]) { pixel, 1 });
+                      0, XCB_WINDOW_CLASS_INPUT_OUTPUT, XCB_COPY_FROM_PARENT,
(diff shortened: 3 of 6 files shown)


Airblader:
It seems for years I was telling lies, saying that i3 does not manage/set the wallpaper. :-( As @stapelberg wrote in the commit message (albeit that was eight years ago *cough*)

> This commit has some chance of breakage, so I’m prepared to revert it unless we can figure out the issues and roll forward.

I'm OK with the change, but I think we'll need to document this somehow. At the very least, can you please update `RELEASE-NOTES-next` to add a note about this?

stapelberg:
AFAIR, this was a problem that was reported with certain login managers (lightdm perhaps? or slim?).

It might be good to check that this commit works with the most common login managers in use today.

psychon:
> At the very least, can you please update RELEASE-NOTES-next to add a note about this?

Sure. However, I am not quite sure about the usual file format of these files (neither of the two `RELEASE-NOTES` files has much content). What do you think about the following?

```diff
diff --git a/RELEASE-NOTES-next b/RELEASE-NOTES-next
index 928f276a..a5b4b75a 100644
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -21,6 +21,7 @@ strongly encouraged to upgrade.
   • Add %machine placeholder (WM_CLIENT_MACHINE) to title_format
   • Allow multiple output names in 'move container|workspace to output'
   • Add 'move container|workspace to output next'
+  • During startup the wallpaper was previously set to a screenshot
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
```

Alternative suggestion would be "No longer take a screenshot and set it as the wallpaper during startup", but that seems a bit long. "Stop setting the wallpaper to a screenshot during startup"?

> It might be good to check that this commit works with the most common login managers in use today.

To whoever will be testing this: Make sure your i3 config does not start anything that sets a wallpaper, since this basically defeats the purpose. As I understand it, the testing strategy would be:
- start i3
- start some window (e.g. a terminal)
- switch to another workspace

If your root window has its background pixmap set to `NONE`, the old "pixels" will stay visible: You get a terminal *without* a blinking cursor and without being able to interact with it.

(Last time I switched to another VT, Xorg no longer allowed me to switch back, so I pass. The error was something like `drmSetMaster() failed` according to `Xorg.0.log`. Sorry.)

Airblader:
> What do you think about the following?

The intention was never to "screenshot and copy it over"( even if that is what it did), so to avoid confusion I'd phrase it a bit more abstract, like

> Do not copy the pixmap of the root screen on startup

As for testing, we're looking at an almost fresh release cycle ahead of us, so I'd be willing to merge it without testing it extensively upfront.

psychon:
> Do not copy the pixmap of the root screen on startup

I went with the slightly longer "Do not copy the root window's content to its background pixmap on startup". I would like to write "to a new pixmap and use that as the root window's background pixmap", but that seems too long.

stapelberg:
> I went with the slightly longer "Do not copy the root window's content to its background pixmap on startup". I would like to write "to a new pixmap and use that as the root window's background pixmap", but that seems too long.

Optimizing for short lines is a secondary concern for our release notes.

Instead, I would encourage you to think from the perspective of a user. Repeatedly ask yourself “I have no idea what any of these words mean” and “but what does this mean for me?”.

In this particular case, when would people look for this commit? Likely if their background unexpectedly changes from the last version of i3 to the current version of i3. Can you update the message to briefly mention this case, so that users have something to go on when debugging?

Airblader:
That's why I raised the question of "where to document this". I don't expect any user to find it buried in the list of commits of the release notes, to be honest. It might rather be worth featuring this prominently in the summary of the release notes itself?

stapelberg:
> That's why I raised the question of "where to document this". I don't expect any user to find it buried in the list of commits of the release notes, to be honest. 

I have thus far considered the lists in the release notes to be directed at users, not merely a git commit dump. The longest bit about getting a new release out is (was?) re-writing commit messages to make sense for users.

> It might rather be worth featuring this prominently in the summary of the release notes itself?

Sounds like we need to run a few tests to gather more certainty as to whether this is a behavior-affecting change or not :)

Airblader:
It's not a commit dump, of course, but it is a long list of things that ultimately represent PRs/commits. It's not exactly a release blog post with shiny pictures and videos and marketing language of "what's new in i3 4.xyz". :-)

I have my doubts that most people update and carefully read the release notes (if at all), especially since we try to make compatible update paths exactly so that users don't have to worry about a changelog before updating. Of course I might be wrong.

stapelberg:
> I have my doubts that most people update and carefully read the release notes (if at all), especially since we try to make compatible update paths exactly so that users don't have to worry about a changelog before updating. Of course I might be wrong.

I agree with that, but at the same time, I think the release notes are the first place where people would start looking for details on what has changed if they encounter a problem after an upgrade. It’s certainly easier than to dig through actual git commits yourself :)

Airblader:
Right, and I'm totally on-board with phrasing it for them to find it easier. But in terms of offering a solution, we should probably feature that more permanently, or at least I think that would be quite helpful.

I've even considered whether we should add a commented-out xsetroot call in the default config that roughly says something along the lines of "if you don't use anything that sets a wallpaper, running something like this is a good idea".

psychon:
> I've even considered whether we should add a commented-out xsetroot call in the default config that roughly says something along the lines of "if you don't use anything that sets a wallpaper, running something like this is a good idea".

That  seems quite sensible to me. I can do that in this PR if wanted.

> Repeatedly ask yourself “I have no idea what any of these words mean” and “but what does this mean for me?”.

Okay... current version:

- Do not copy the root window's content to its background pixmap on startup

My proposal from the above:

- i3 used to take a screenshot and set it as the wallpaper when initially starting up. This was done some years ago to work around problems with some display manager, but caused problems when the screenshot was actually showing windows. Thus, this code was dropped in this release. If you do not set a wallpaper, it is recommended to add `exec --no-startup-id xsetroot -solid black` to your config.

Reason why I do not like the above: Way too verbose compared to the other lines in there.

stapelberg:
> * i3 used to take a screenshot and set it as the wallpaper when initially starting up. This was done some years ago to work around problems with some display manager, but caused problems when the screenshot was actually showing windows. Thus, this code was dropped in this release. If you do not set a wallpaper, it is recommended to add `exec --no-startup-id xsetroot -solid black` to your config.

This sounds good to me!

It’s the perfect length for the prose section above the lines. Depending on further testing, we should either put it into prose (if we find any breakage) or list (if we find no breakage). Either way, length is fine :)

psychon:
It is possible I finally understood what you meant. New version of the last commit pushed (I replaced the `RELEASE-NOTES` commit with a different one).

stapelberg:
The commits look good, but we still need to test the change, I think.

What does the test matrix look like (i.e. which display managers, in which configurations)? What is the test procedure?

psychon:
> What is the test procedure?

So far I only tested starting an `xterm` before `i3` in `Xephyr` and switching to another (empty) workspace. Before this change, I ended up with a non-working `xterm`. This is basically the test for "fixes #2869". (Oh and: I ran `i3 -c /dev/null` to make sure it does not set a wallpaper.)

With relation to display managers (=testing that this does not reintroduce https://github.com/i3/i3/commit/4f5e0e794f61cdcea2652f964ec25d0b7418f69b): If I understand the commit message correctly, this means:

- start `i3` (and make sure it does not set a wallpaper while starting)
- open some window
- switch to another (empty) workspace

If the background pixmap of the root window is set to `None`, the X11 server will not update the screen contents. Thus, you end up with a non-working "screenshot" of the window that you opened.

(Heh, before this I did not notice that both scenarios show basically the same problematic behaviour.)

> What does the test matrix look like (i.e. which display managers, in which configurations)? 

So far, I did no testing at all. (Well, only the testing I mentioned above.)

I just removed my wallpaper command from `~/.config/i3/config`, so somewhen in the next days I will be testing with lightdm.

https://github.com/i3/i3/commit/4f5e0e794f61cdcea2652f964ec25d0b7418f69b talks about GDM3. I do not have that installed and will not install it to test this, sorry.

Edit:

> I just removed my wallpaper command from ~/.config/i3/config, so somewhen in the next days I will be testing with lightdm.

I just ended up with the debian default wallpaper as my wallpaper. Nothing special happened and no breakage.

psychon:
Judging from https://github.com/GNOME/gdm/blob/fb319044165f2b4c6e194cf26380c29efc675f64/daemon/gdm-server.c#L221 and https://github.com/GNOME/gdm/blob/349e6af855a1c1dbee9af3e5881e00b66482d155/daemon/gdm-x-session.c#L275-L276, I would say that gdm most likely still starts the X11 server with `-background none`, so that definitely needs some testing.

Bonus points for `ps aux | grep background` and making sure the X11 server was really started with `-background none`.

Is anyone here able to easily do this test? Any volunteers?

Edit: With my current lightdm session, I get:
```
$ ps aux | grep Xorg
root         575  3.5  0.2 1003216 87672 tty7    Ssl+ 08:44   6:55 /usr/lib/xorg/Xorg :0 -seat seat0 -auth /var/run/lightdm/root/:0 -nolisten tcp vt7 -novtswitch

```

stapelberg:
> [4f5e0e7](https://github.com/i3/i3/commit/4f5e0e794f61cdcea2652f964ec25d0b7418f69b) talks about GDM3. I do not have that installed and will not install it to test this, sorry.

With gdm on Fedora 33, I indeed get the following Xorg command line which includes `-background none`:
```
michael         8983  20   0  0.1  0.1          \_ /usr/libexec/Xorg vt3 -displayfd 3 -auth /run/user/1000/gdm/Xauthority -nolisten tcp -background none -noreset -keeptty -novtswitch -verbose 3
```

I haven’t had time to test i3 with this pull request patched in yet.

stapelberg:
> If the background pixmap of the root window is set to `None`, the X11 server will not update the screen contents. Thus, you end up with a non-working "screenshot" of the window that you opened.

I just tested your pull request on Fedora 33 with gdm (default setup), and this is exactly the problem that I’m running into.

Hence, I don’t think just reverting the commit is good enough :-/.

What other options do we have?

psychon:
> I just tested your pull request on Fedora 33 with gdm (default setup),

Thanks for that!

> What other options do we have?

Uhm...
- Leave the status quo and accept the bugs it causes (pre-existing programs end up in the screenshot i3 creates)
- Add code to fully set the wallpaper. I do not want to go there, but did that already before for AwesomeWM, so I could do that. However, no idea where the pixels for the wallpaper should come from. Cairo can only load PNGs and that is unlikely to be enough, so would require new dependencies.
- Add code to set a black wallpaper. No idea how to tell people to enable this. Is "set wallpaper before i3 starts" a supported configuration? If not, i3 could just always set a black background.
- A variant of the above: Only set a black wallpaper (or alternatively a screenshot) if the root window does not have a `_XROOTPMAP_ID` or `ESETROOT_PMAP_ID` property. By convention, wallpaper setters are supposed to set these properties to the pixmap of the wallpaper, but since this is X11, this is only a convention. I bet `xsetroot` does not set these properties since it does not set a pixmap as background.

AFAIK there is no way to query the background of an X11 window. Meh. :-(

One could also determine the value at startup experimentally: Find an empty spot where the wallpaper would be visible. Map a window with a white background at this spot. Unmap it. Grab a "screenshot" and check if that pixel is white. (And then repeat the same with black.) Also ugly. (Edit: No window required. Just draw to the root window and use `ClearArea` to tell the X11 server to erase the drawing and re-apply the background. But this still needs an "empty space".)

By the way: `_XROOTPMAP_ID` and `ESETROOT_PMAP_ID` exists so that clients can query the wallpaper (see above: No way to query the background of a window). This is e.g. used by urxvt for fake transparency. I guess this is also currently broken with i3's screenshot since these properties are not set. Dunno if that is a bug or a feature. ;-)

Seems like the following WMs do not set a background when started (or my search for `BackPixmap` is too simplistic):
- Fluxbox: https://github.com/fluxbox/fluxbox/search?q=BackPixmap
- AwesomeWM's C code does not actually do anything unless Lua tells it to, but the default config sets a wallpaper.
- twm: https://github.com/freedesktop/twm/search?q=BackPixmap (well, okay, I am out of ideas where to search so do not take this line too seriously.)

psychon:
> One could also determine the value at startup experimentally: Find an empty spot where the wallpaper would be visible. Map a window with a white background at this spot. Unmap it. Grab a "screenshot" and check if that pixel is white.

I tried implementing this and tested with `Xephyr -background none`. However, it does not work. The screenshot always come back black.

However, I also cannot reproduce the original issue with `Xephyr` and this branch:

>> If the background pixmap of the root window is set to None, the X11 server will not update the screen contents. Thus, you end up with a non-working "screenshot" of the window that you opened.
>
> I just tested your pull request on Fedora 33 with gdm (default setup), and this is exactly the problem that I’m running into.

@stapelberg Can you perhaps try this program with your X11 server and without i3 / with i3 built from this PR? The desired outcome is two different values for "pixel 1" and "pixel 2", but I only ever get it to print 0.

https://gist.github.com/psychon/05d666fb123a019a636430e36c938b69

stapelberg:
> * Add code to set a black wallpaper. No idea how to tell people to enable this. Is "set wallpaper before i3 starts" a supported configuration? If not, i3 could just always set a black background.

Yeah, people (like me ;)) might be using `xsetroot` in their `~/.xsession` before starting i3.

I’m leaning towards auto-detecting (if possible) this issue and reacting accordingly, or, if that’s not possible, leave things at the status quo.

I ran the tests you asked on Fedora:

### Standard i3 without your pull request

When I run your `try_detect_root_background_none` from my `~/.xsession` before starting i3, I get:

```
There are 1 free/visible rectangles
Rect 0: top left 0x0 size 3840x2160
Pixel 1: 0; pixel 2: 255
```

If I run it after starting i3, I get:
```
There are 1 free/visible rectangles
Rect 0: top left 0x0 size 3840x2123
Pixel 1: 255; pixel 2: 255
```

Note that I can see a white rectangle at the top left of the screen.

### With this pull request applied

Before:
```
There are 1 free/visible rectangles
Rect 0: top left 0x0 size 3840x2160
Pixel 1: 0; pixel 2: 255
```

After:
```
There are 1 free/visible rectangles
Rect 0: top left 0x0 size 3840x2123
Pixel 1: 0; pixel 2: 255
```

White rectangle still visible.

psychon:
> I ran the tests you asked on Fedora:

Thanks a lot! Test result is "the test works". No idea why it does not work under `Xephyr`.

> I’m leaning towards auto-detecting (if possible)

Okay, I will update this PR accordingly, but likely not today. I will report back when I found the time.

> Note that I can see a white rectangle at the top left of the screen.

Yeah, that's basically exactly what the test does. Open a white rectangle and test if it stays white after the window was closed.

psychon:
I finally got around to updating this PR. It is completely redone. `git show --stat` of the result:
```
commit bf881c156f3e379e9f6f8556d9756b43d16428db
Author: Uli Schlachter <psychon@znc.in>
Date:   Tue May 18 17:19:00 2021 +0200

    "Set" the wallpaper during startup only sometimes
    
    Since commit 4f5e0e7, i3 would take a screenshot and set that as the
    background pixmap of the root window during startup. This is the easy
    part of setting a proper X11 wallpaper.
    
    The code in question was added because something either set the
    background pixmap of the root window to NONE or the X11 server was
    started with "-background none". This is apparently done by default by
    e.g. gdm to avoid some flickering while the X11 server starts up.
    
    This commit makes this code conditional: Only when no wallpaper is
    detected is a screenshot taken.
    
    Since I could not find any way to query the background of a window, a
    more direct approach is taken to detect this situation: First, we find
    some part of the root window that is not currently covered. Then we open
    a white window there, close it again and grab a screenshot. If a
    wallpaper is set, the X11 server will draw this wallpaper after the
    window is closed and something else will be visible in the screenshot.
    
    However, the wallpaper could have a white pixel at the tested position.
    Thus, this procedure is repeated with a black window.
    
    Only when this procedure produces two different pixel values is a
    screenshot taken and set as the wallpaper.
    
    Fixes: https://github.com/i3/i3/issues/4371
    Fixes: https://github.com/i3/i3/issues/2869

 RELEASE-NOTES-next        |   7 +++++++
 include/libi3.h           |   8 ++++++++
 libi3/is_background_set.c | 117 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 meson.build               |   1 +
 src/main.c                |  18 ++++++++++++++++--
 5 files changed, 149 insertions(+), 2 deletions(-)
```

RELEASE-NOTES-next
```diff
@@ -6,6 +6,13 @@
 This is i3 v4.19. This version is considered stable. All users of i3 are
 strongly encouraged to upgrade.
 
+i3 used to unconditionally take a screenshot and set it as the wallpaper when
```
stapelberg: I would like to make this a little more clear to users. How about:

---

Background/wallpaper workaround:

Some login managers (e.g. gdm) start X11 with the -background none flag. When this flag is set, a background needs to be explicitly set later in the X11 session, otherwise stale copies of closed windows remain visible on the X11 root window (symptom looks like “my terminal window is not closing”).

i3 works around this situation by setting a screenshot of the X11 root window as background when starting. Any background you set before starting i3 (e.g. in your Xsession) or after starting i3 (e.g. via exec statements in the i3 config) will be visible.

A downside of this workaround is that if you have any windows open in your X11 session, those will be part of the screenshot.

To fix this issue, starting in v4.20, i3 detects whether the -background none option is enabled and only then sets a screenshot as background.

---

Notably, the suggestion to add xsetroot shouldn’t be necessary, correct?

Also, in which situations would one typically encounter open X11 windows before starting i3?

psychon: > How about:

Thanks! I applied that in an extra commit did some small changes ontop of it. Since the line wrapping causes that commit to change "everything", here is `git show --word-diff` for that latest commit:
```
Some login managers (e.g. gdm) start {+the+} X11 {+server+} with the -background none
flag. When this flag is set, a background needs to be explicitly set later in
the X11 session, otherwise stale copies of closed windows remain visible on the
X11 root window (symptom looks like “my terminal window is not closing”).

i3 works around this situation by setting a screenshot[-of the X11 root window-] as background when
starting. Any background you set before starting i3 (e.g. in your Xsession) or
after starting i3 (e.g. via exec statements in the i3 config) will be visible.

A downside of this workaround is that if you have any windows {+already+} open in
your X11 session, those will be part of the screenshot.
```

> Notably, the suggestion to add xsetroot shouldn’t be necessary, correct?

Oh, right. That was totally a left over from the previous "just drop the screenshot". Thanks for catching that!

> Also, in which situations would one typically encounter open X11 windows before starting i3?

Uhm... Good question.

Hypothetical scenario: "I used to use fluxbox and for autostart I had things like `xterm& thunderbird& whatever & exec fluxbox` in my `~/.xsession`. I kept that stuff in there when switching to i3".

In the above situation, there is a race and e.g. xterm might be able to get its window visible before the WM starts.

Another idea would be `i3 --replace` replacing an already running WM, but it is very likely that the original WM exiting causes the whole X11 session do be brought down. (But this is only possible after `--replace` actually exists, which is #4374 which I still have to update again...)

stapelberg: Thanks for clarifying! I’m thinking the section is no longer as relevant as it used to be (no action required from users), so I might move it into its own section when preparing the next release. It’s fine to merge it as-is, though, I’d say.

stapelberg (approved):

Issue #536, opened 2011-10-10
Acquire the WM_Sx selection (enables --replace)


stapelberg:
A window manager should acquire ownership of the WM_Sx selection. Handling this properly allows us to replace existing window managers and be replaced by other window managers.

See also:
http://tronche.com/gui/x/icccm/sec-4.html#s-4.3

Airblader:
Just an FYI to anyone taking a shot at this, it is well worth looking into Openbox's source to see how this can be done.

psychon:
Another hint could be the following commits. (Note that I know that awesome is not confirming to the spec, I don't see the point in jumping through all the necessary hoops just to get a current timestamp that can be used for acquiring the selection, so I just used `XCB_CURRENT_TIME`):
https://github.com/awesomeWM/awesome/commit/5a17b3978c7aafa720197f670c961aa8cd794dd6 https://github.com/awesomeWM/awesome/commit/19111121ce51e555ba92d1f10e4735d9c217f7ad
https://github.com/awesomeWM/awesome/commit/e824f8f30d3f797374f6597baf4fb70f48f51b25

Issue #3009, opened 2017-10-01
enhancement: fix flaky tests


stapelberg:
I got the following tests to fail while running the testsuite in a loop while stress-ng is running:

* [ ] testcases/t/115-ipc-workspaces.t
* [ ] testcases/t/133-size-hints.t

If there are any others which fail (even just occasionally), please mention them in this issue so that we can work on a fix.

orestisf1993:
`t/525-i3bar-mouse-bindings.t`:

```
ok 1 - Connect to X11, verify XKB and XTEST are present (via Inline::C)
ok 2 - i3bar present
# i3bar window = 10485766
# Subtest: button 1 moves focus left
    ok 1 - tick nonce received
    ok 2 - button 1 moves focus left
    1..2
ok 3 - button 1 moves focus left
# Subtest: button 2 moves focus right
    ok 1 - tick nonce received
    ok 2 - button 2 moves focus right
    1..2
ok 4 - button 2 moves focus right
# Subtest: button 3 moves focus left
    ok 1 - tick nonce received
    not ok 2 - button 3 moves focus left

    #   Failed test 'button 3 moves focus left'
    #   at /home/orestis/Documents/programming/i3/build/../../i3/testcases/t/525-i3bar-mouse-bindings.t line 100.
    #     Structures begin differing at:
    #          $got->[0] = Does not exist
    #     $expected->[0] = '4194304'
    1..2
    # Looks like you failed 1 test of 2.
not ok 5 - button 3 moves focus left

#   Failed test 'button 3 moves focus left'
#   at /home/orestis/Documents/programming/i3/build/../../i3/testcases/t/525-i3bar-mouse-bindings.t line 127.
# Subtest: button 4 moves focus right
    ok 1 - tick nonce received
    ok 2 - button 4 moves focus right
    1..2
ok 6 - button 4 moves focus right
# Subtest: button 5 moves focus left
    ok 1 - tick nonce received
    ok 2 - button 5 moves focus left
    1..2
ok 7 - button 5 moves focus left
# Subtest: button 6 does not move focus while pressed
    ok 1 - tick nonce received
    ok 2 - button 6 does not move focus while pressed
    1..2
ok 8 - button 6 does not move focus while pressed
ok 9 - focus unchanged
# Subtest: button 6 release moves focus right
    ok 1 - tick nonce received
    ok 2 - button 6 release moves focus right
    1..2
ok 10 - button 6 release moves focus right
# Subtest: button 7 press moves focus left
[...]

PR #4374, opened 2021-03-10
Implement support for the WM_Sn selection

psychon:
This implements https://github.com/i3/i3/issues/536. At startup, i3 will now acquire the `WM_Sn` selection. If the selection is already owned, it will exit saying that another window manager is already running. When started with `--replace`, it will instead replace the existing WM (assuming it properly follows ICCCM, but that's always the case when working with X11).


src/handlers.c
```diff
@@ -1076,6 +1076,21 @@ static void handle_configure_notify(xcb_configure_notify_event_t *event) {
     randr_query_outputs();
 }
 
+/*
+ * Handles SelectionClear events for the root window, which are generated when
+ * we loose ownership of a selection.
```
Airblader: nit: s/loose/lose/

src/main.c
```diff
@@ -656,6 +685,95 @@ int main(int argc, char *argv[]) {
...
+                            8,
+                            (strlen("i3-WM_Sn") + 1) * 2,
+                            "i3-WM_Sn\0i3-WM_Sn\0");
+
+        xcb_set_selection_owner(conn, wm_sn_selection_owner, wm_sn, last_timestamp);
+
+        if (selection_reply && selection_reply->owner != XCB_NONE) {
+            xcb_get_geometry_reply_t *geom_reply = NULL;
+            DLOG("waiting for old WM_Sn selection owner to exit");
+            do {
```
Airblader, line 754: Do you think it'd be worth having some upper bound of how long we wait here to avoid busy-looping forever if the owner doesn't exit?

psychon: Dunno. I never saw anything doing anything clever here and I never heard about any complaints that this was going wrong.

Random example of something that uses a 15 seconds timeout and sleeps for 0.1 seconds between checks: https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/openbox/screen.c#L132-L153

psychon:
I just found https://github.com/i3/i3/blob/next/src/main.c#L426 which should be updated in this PR.

psychon:
The Travis build just failed. Somehow I feel like this has nothing to do with my PR and I have no idea what is going on.

https://travis-ci.com/github/psychon/i3/jobs/490045736

```
--- command ---
13:05:26 /usr/bin/perl complete-run.pl
--- stdout ---
Running tests under Xvfb display :99
Starting 4 Xephyr instances, starting at :100...
Writing logfile to 'testsuite-2021-03-11-13-05-26-4.15-1110-g6f05a5ca/complete-run.log'...
#   Failed test 'call to scratchpad failed'
#   at /usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/185-scratchpad.t line 492.
#          got: '1'
#     expected: '0'
Test Summary Report
-------------------
/usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/185-scratchpad.t (Wstat: 0 Tests: 100 Failed: 1)
  Failed test:  91
Files=252, Tests=3704, 22 wallclock secs ( 0.87 usr +  0.09 sys =  0.96 CPU)
Result: FAIL
The slowest tests are:
	/usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/301-shape.t with 6.33 seconds
	/usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/290-keypress-numlock.t with 5.57 seconds
	/usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/308-focus_wrapping.t with 4.07 seconds
	/usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/258-keypress-release.t with 3.79 seconds
	/usr/src/i3/distbuild/meson-private/dist-unpack/i3-4.19.2/testcases/t/286-root-window-mouse-binding.t with 3.60 seconds
--- stderr ---
Use of uninitialized value $timingsjson in numeric gt (>) at complete-run.pl line 206.
Use of uninitialized value in subtraction (-) at /usr/share/perl/5.32/TAP/Formatter/Base.pm line 386.
Negative repeat count does nothing at /usr/share/perl/5.32/TAP/Formatter/Base.pm line 386.
-------
```

Airblader (approved):
This looks good to me, and I'm also OK with not having the timeout for now (since this is an opt-in feature I don't see this as much of a risk). Wouldn't mind if someone had a look as well, in particular I don't have capacities at the moment to test this myself. @orestisfl perhaps?

orestisfl (changes requested):
Release notes need to be updated :)

SGTM but will check over the weekend

psychon:
I just found this in https://github.com/i3/i3/issues/3009:

> Sometimes i3 cannot change the root window’s event mask (X11 error BadAccess), implying that the old i3 process on that x server hasn’t yet been cleaned up (even though we wait for the i3 process to terminate). 

Two comments:
- Only because i3 terminates does not mean that the X11 server already handled the connection close
- This PR could help here by making the test suite use `--replace`. That should make the new i3 process wait until the old one is "really gone".

(I am writing this here so that #3009 doesn't get cluttered with comments)

psychon:
> SGTM but will check over the weekend

@orestisfl ping?

src/main.c
orestisfl: I am in favor of the openbox approach

orestisfl (approved):
Thanks for the ping, LGTM.

I'd also like to avoid the busy-loop. 50 requests with 0.1 sleep sound good (5 seconds in total should be enough even if the other wm is in a bad state).

src/main.c
psychon: I implemented a (completely untested) timeout of 15 seconds for the "wait for the old WM to exit" code. I hope that's what you meant.

I did not implement the "wait for a DestroyNotify event to indicate that the other WM went away"-part, since I prefer the current "use GetGeometry to check if the window went away". If you want, I can also switch from GetGeometry to waiting for a DestroyNotify event.

src/main.c
```diff
@@ -627,6 +643,22 @@ int main(int argc, char *argv[]) {
...
+        xcb_generic_event_t *event;
+        DLOG("waiting for PropertyNotify event\n");
+        while ((event = xcb_wait_for_event(conn)) != NULL) {
+            if (event->response_type == XCB_PROPERTY_NOTIFY) {
+                last_timestamp = ((xcb_property_notify_event_t *)event)->time;
+                free(event);
+                break;
+            }
+            free(event);
+        }
```
orestisfl, line 658: Is it possible to lose events that we might want to keep in this loop? Would it make sense to make this code more event-driven?

psychon: Apparently so and someone already wrote similar code later in the file: https://github.com/i3/i3/blob/4c93f61353ffa905a4f15cb98524280cd92c53cb/src/main.c#L958-L975

This just discards all pending events, but handles map requests.

Since we did not select SubstructureRedirect yet, there cannot be any pending MapRequests. All other events would be discarded "down there" anyway.

orestisfl (approved):

PR #4376, opened 2021-03-10
Share graphics context globally

### LINKED ISSUE #3478: Investigate using only two global GCs
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
i3 uses a GC per container.

## Expected Behavior
i3 could use a GC per depth (24, 32) instead to avoid creating unnecessarily many GCs. See https://github.com/i3/i3/issues/2742#issuecomment-433330426 for details.

## Environment

i3 4.15

psychon:
Instead of creating a graphics context for every surface_t, this commit
adds a cache that allows to "remember" up to two GCs. Thus, the code
uses less GCs. When a GC from the cache can be used, this also gets rid
of a round-trip to the X11 server. Both of these are tiny, insignificant
savings, but so what?

Fixes: https://github.com/i3/i3/issues/3478

-----

This adds a copy&paste duplicate of `get_visual_depth()` from `src/xcb.c` to `libi3. If you tell me what to do, I can clean this up again, but I was not really sure what the best option here is.

I tried testing the cache overflow by trying to create a window with a depth other than 24 or 32, but my X11 server does not offer any such visuals:
```
$ xdpyinfo| grep depth: | uniq
    depth:    24 planes
    depth:    32 planes
```
Also, initially I planned to use a `SLIST` for this and have the cache unlimited. After briefly staring at the macros, I went with a statically sized solution instead. Thus, there is now the possibility that a GC cannot be cached and is thus still owned by the `surface_t`.

### DIFF
--- a/libi3/draw_util.c
+++ b/libi3/draw_util.c
@@ -29,6 +29,76 @@ static bool surface_initialized(surface_t *surface) {
 }
 
 /*
+ * Get a GC for the given depth. The given drawable must have this depth.
+ *
+ * Per the X11 protocol manual for "CreateGC":
+ * > The gcontext can be used with any destination drawable having the same root
+ * > and depth as the specified drawable;
+ */
+static xcb_gcontext_t get_gc(xcb_connection_t *conn, uint8_t depth, xcb_drawable_t drawable, bool *should_free) {
+    static struct {
+        uint8_t depth;
+        xcb_gcontext_t gc;
+    } gc_cache[2] = {
+        0,
+    };
+
+    size_t index = 0;
+    bool cache = false;
+
+    *should_free = false;
+    for (; index < sizeof(gc_cache) / sizeof(gc_cache[0]); index++) {
+        if (gc_cache[index].depth == depth) {
+            return gc_cache[index].gc;
+        }
+        if (gc_cache[index].depth == 0) {
+            cache = true;
+            break;
+        }
+    }
+
+    xcb_gcontext_t gc = xcb_generate_id(conn);
+    xcb_void_cookie_t gc_cookie = xcb_create_gc_checked(conn, gc, drawable, 0, NULL);
+
+    xcb_generic_error_t *error = xcb_request_check(conn, gc_cookie);
+    if (error != NULL) {
+        ELOG("Could not create graphical context. Error code: %d. Please report this bug.\n", error->error_code);
+        free(error);
+        return gc;
+    }
+
+    if (cache) {
+        gc_cache[index].depth = depth;
+        gc_cache[index].gc = gc;
+    } else {
+        *should_free = true;
+    }
+
+    return gc;
+}
+
+/*
+ * Get depth of visual specified by visualid
+ *
+ */
+uint16_t get_visual_depth(xcb_visualid_t visual_id) {
+    xcb_depth_iterator_t depth_iter;
+
+    depth_iter = xcb_screen_allowed_depths_iterator(root_screen);
+    for (; depth_iter.rem; xcb_depth_next(&depth_iter)) {
+        xcb_visualtype_iterator_t visual_iter;
+
+        visual_iter = xcb_depth_visuals_iterator(depth_iter.data);
+        for (; visual_iter.rem; xcb_visualtype_next(&visual_iter)) {
+            if (visual_id == visual_iter.data->visual_id) {
+                return depth_iter.data->depth;
+            }
+        }
+    }
+    return 0;
+}
+
+/*
  * Initialize the surface to represent the given drawable.
  *
  */
@@ -41,15 +111,7 @@ void draw_util_surface_init(xcb_connection_t *conn, surface_t *surface, xcb_draw
     if (visual == NULL)
         visual = visual_type;
 
-    surface->gc = xcb_generate_id(conn);
-    xcb_void_cookie_t gc_cookie = xcb_create_gc_checked(conn, surface->gc, surface->id, 0, NULL);
-
-    xcb_generic_error_t *error = xcb_request_check(conn, gc_cookie);
-    if (error != NULL) {
-        ELOG("Could not create graphical context. Error code: %d. Please report this bug.\n", error->error_code);
-        free(error);
-    }
-
+    surface->gc = get_gc(conn, get_visual_depth(visual->visual_id), drawable, &surface->owns_gc);
     surface->surface = cairo_xcb_surface_create(conn, surface->id, visual, width, height);
     surface->cr = cairo_create(surface->surface);
 }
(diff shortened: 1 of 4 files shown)


libi3/draw_util.c
```diff
@@ -58,7 +116,9 @@ void draw_util_surface_init(xcb_connection_t *conn, surface_t *surface, xcb_draw
  *
  */
 void draw_util_surface_free(xcb_connection_t *conn, surface_t *surface) {
-    xcb_free_gc(conn, surface->gc);
+    if (surface->owns_gc) {
+        xcb_free_gc(conn, surface->gc);
```
orestisfl, line 136: What happens after you free the gc that is used in other surfaces?

psychon: Uhm... Sorry, but I don't quite get the question...?

I just re-read the diff and `owns_gc` is only set by `get_gc()`. That function only ever sets `owns_gc = true` when nothing else references the GC. Put differently: If the resulting GC is saved in the cache (either already before or during the call to `get_gc()`), `owns_gc = false`. If the resulting GC is only referenced by the `draw_util_surface`, then `owns_gc = true`.

So, to try to answer the question: If this were to free a GC that is also used in other `draw_util_surface`s, that would be bad. But I don't think that can possibly happen.

orestisfl: Correct, never mind my comment

libi3/draw_util.c
```diff
@@ -28,6 +35,64 @@ static void draw_util_set_source_color(surface_t *surface, color_t color);
...
+        gc_cache[index].depth = depth;
+        gc_cache[index].gc = gc;
+    } else {
+        *free = true;
+    }
+
+    return gc;
+}
+
+static uint16_t get_visual_depth(xcb_visualid_t visual_id) {
```
orestisfl: I would delete this function from xcb.c and move it to libi3 to not have duplicated code

libi3/draw_util.c
```diff
@@ -20,6 +20,13 @@ extern xcb_visualtype_t *visual_type;
 /* Forward declarations */
 static void draw_util_set_source_color(surface_t *surface, color_t color);
 
+static struct {
+    uint8_t depth;
+    xcb_gcontext_t gc;
+} gc_cache[2] = {
```
orestisfl: I feel like in i3 we are sticking too much in primitive data structures even though we can take advantage of more advanced ones, e.g. this could have been implemented more easily with [glib's hashtable](https://docs.gtk.org/glib/struct.HashTable.html) and glib is already a dependency.

I am not really requesting that you re-write this, just an observation. You can ignore this comment.

libi3/draw_util.c
```diff
@@ -20,6 +20,13 @@ extern xcb_visualtype_t *visual_type;
 /* Forward declarations */
 static void draw_util_set_source_color(surface_t *surface, color_t color);
 
+static struct {
```
orestisfl: This static struct can live entirely inside the `get_gc` function since it's only referenced there.

orestisfl (changes requested):
Thank you, and sorry for the delayed review

libi3/draw_util.c
psychon: Yeah, I didn't use much glib so far. I quickly looked at the link and then decided to go with `You can ignore this comment`. :sweat_smile:

psychon:
I rebased this on `next` (had some conflicts) and addressed the comments.

I am not entirely sure, but the following comments now seems outdated:
```c
    if (surface->owns_gc) {
        /* NOTE: This function is also called on uninitialised surface_t instances.
         * The x11 error from xcb_free_gc(conn, XCB_NONE) is silently ignored
         * elsewhere.
         */
        xcb_free_gc(conn, surface->gc);
    }
```
`uninitialised` cannot really mean "contains arbitrary garbage". That sounds like a recipe for crashes. So I guess this means zero-initialised. In that case, `owns_gc` should be `false` and I just made this error case go away.

However, this is just a guess and I don't have the time right now to dig deeper, so I am just mentioning this here and keeping things otherwise the way they already were.

libi3/draw_util.c
```diff
@@ -68,11 +130,13 @@ void draw_util_surface_free(xcb_connection_t *conn, surface_t *surface) {
...
-    /* NOTE: This function is also called on uninitialised surface_t instances.
-     * The x11 error from xcb_free_gc(conn, XCB_NONE) is silently ignored
-     * elsewhere.
-     */
-    xcb_free_gc(conn, surface->gc);
+    if (surface->owns_gc) {
+        /* NOTE: This function is also called on uninitialised surface_t instances.
+         * The x11 error from xcb_free_gc(conn, XCB_NONE) is silently ignored
+         * elsewhere.
+         */
```
orestisfl: > I rebased this on `next` (had some conflicts) and addressed the comments.
> 
> I am not entirely sure, but the following comments now seems outdated:
> 
> ```c
>     if (surface->owns_gc) {
>         /* NOTE: This function is also called on uninitialised surface_t instances.
>          * The x11 error from xcb_free_gc(conn, XCB_NONE) is silently ignored
>          * elsewhere.
>          */
>         xcb_free_gc(conn, surface->gc);
>     }
> ```
> 
> `uninitialised` cannot really mean "contains arbitrary garbage". That sounds like a recipe for crashes. So I guess this means zero-initialised. In that case, `owns_gc` should be `false` and I just made this error case go away.
> 
> However, this is just a guess and I don't have the time right now to dig deeper, so I am just mentioning this here and keeping things otherwise the way they already were.

The comment mentions `XCB_NONE` so that's what I imagine it means, not garbage. I've played around and we can still get errors on the free but they happen in `next` as well, we can continue ignoring them for now. I think we can remove the comment as it seems misleading now.

```suggestion
```

libi3/draw_util.c
```diff
@@ -28,6 +28,76 @@ static bool surface_initialized(surface_t *surface) {
...
+            return gc_cache[index].gc;
+        }
+        if (gc_cache[index].depth == 0) {
+            cache = true;
+            break;
+        }
+    }
+
+    xcb_gcontext_t gc = xcb_generate_id(conn);
+    xcb_void_cookie_t gc_cookie = xcb_create_gc_checked(conn, gc, drawable, 0, NULL);
```
orestisfl, line 63:
```suggestion
    /* The drawable is only used to get the root and depth, thus the gc is not
     * tied to the drawable and it can be re-used from different drawables. */
    xcb_void_cookie_t gc_cookie = xcb_create_gc_checked(conn, gc, drawable, 0, NULL);
```

I came up with this while trying to understand why this is safe to do e.g. when creating a gc from a drawable that is later freed. If my comment is correct, maybe we can include it for further context to readers?

orestisfl (changes requested):
Thanks, couple of last things

libi3/draw_util.c
psychon: Done.

"Definitive answer" seems to come from the X11 protocol reference manual: https://www.x.org/releases/X11R7.6/doc/xproto/x11protocol.html#requests:CreateGC

>  The gcontext can be used with any destination drawable having the same root and depth as the specified drawable [...]

orestisfl (approved):
Thank you

PR #4377, opened 2021-03-10
ipc: return signed int for container positions

lbonn:
If we use json as a language-agnostic representation, it makes sense to use proper signed integers here.

Before (with a floating window):

```
  "rect": {
    "x": 4294966757,
    "y": 273,
    "width": 804,
    "height": 602
  },
```

After:

```
  "rect": {
    "x": -316,
    "y": 228,
    "width": 804,
    "height": 602
  },
```

Note that I was not sure if this was intended like this as the behavior seems quite old, but it does looks like a bug: for example, the popular i3ipc-python also returns these large integers.

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -34,3 +34,5 @@ strongly encouraged to upgrade.
   • i3bar: exit with 1 when a wrong command line argument is used
   • fix commented-out rofi call in default i3 config
   • clear pixmap before drawing to prevent visual garbage
+  • ipc: return proper signed int for container positions: negative values were
+    returned as large 32 bits integers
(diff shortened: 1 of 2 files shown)


lbonn:
Any thoughts on that? The initial motivation is that it triggers crashes when using common i3ipc rust libraries: https://github.com/tmerr/i3ipc-rs and https://github.com/JayceFayne/swayipc-rs as they expect int32 when deserializing.

For example https://github.com/tmerr/i3ipc-rs/blob/master/src/reply.rs#L211

Airblader (commented):
Thanks for your PR – and sorry for the lack of replies! The change makes a lot of sense, and if you look at the definition of `Rect` (`include/data.h`) you'll actually even see a comment about this being the case.

Could you please just also add an entry to `RELEASE-NOTES-next`? Thanks!

lbonn:
No problem! Here it is with the release notes change.

Airblader (approved):
Thanks!

PR #4388, opened 2021-03-15
i3-dmenu-desktop: cmd-prefix

tonischranz:
I added the possibility to provide a command prefix as Option to i3-dmenu-desktop. 
I'm using it to launch apps on DISPLAY :20 instead of :0 with a second dmenu bindsym.

Since this is just a small enhancement on a handy side-script, I'm not sure if you generally accept pull requests for that, just thought maybe others could use that too, and they will probably never find my tweaked version unless it's in the main project.

To be honest, i never wrote Perl before, so maybe i missed something or did something in a uncool manner, sorry for that if so. I tested it to launch gimp and htop with and without prefix, worked as expected. Since it's not a code change i didn't run the testsuite, since never even built that whole thing.

### DIFF
--- a/i3-dmenu-desktop
+++ b/i3-dmenu-desktop
@@ -46,9 +46,11 @@ sub slurp {
 
 my @entry_types;
 my $dmenu_cmd = 'dmenu -i';
+my $cmd_prefix = '';
 my $result = GetOptions(
     'dmenu=s' => \$dmenu_cmd,
     'entry-type=s' => \@entry_types,
+    'cmd-prefix=s' => \$cmd_prefix,
     'version' => sub {
         say "dmenu-desktop 1.5 © 2012 Michael Stapelberg";
         exit 0;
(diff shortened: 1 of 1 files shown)


thiagowfx:
Is `DISPLAY=:20` because of chrome remote desktop?

tonischranz:
No, it‘s because i use ssh tunneling for X Forwarding (without needing sshd
on the main box)
So box:6020 is forwarded to rpi:6000 (display :0, connected to a TV, with
own KB/Mouse)

Does this help to clarify?

Thiago Perrotta ***@***.***> schrieb am Mo. 2. Aug. 2021 um
05:28:

i3-dmenu-desktop
```diff
@@ -456,6 +458,11 @@ if (exists($app->{Path}) && $app->{Path} ne '') {
 
 my $nosn = '';
 my $cmd;
+
+if (length $cmd_prefix) {
+    $cmd_prefix = "$cmd_prefix ";
```
thiagowfx, line 463: I would suggest not to change the user input, i's generally best not to.. Instead, add a space in L482 and L495.

thiagowfx (commented):
> Does this help to clarify?

Yes, makes sense. I don't have approval rights but I left a comment below.

orestisfl (changes requested):
Thank you for your PR.

More generally, I would advise to use something more mature than i3-dmenu-desktop for this purpose, like rofi.

That said, I think that your usecase makes sense and there isn't any way to pass environmental variables to executed commands. However, I don't like the prefix idea.

I would prefer if we went towards rofi's direction with these two flags:
```
-run-command cmd

Set command ({cmd}) to execute when running an application.  See PATTERN.

Default: {cmd}

-run-shell-command cmd

Set command to execute when running an application in a shell.  See PATTERN.

Default: {terminal} -e {cmd}
```

PR #4390, opened 2021-03-17
Improve IPC documentation

CyberShadow:
I wrote an IPC client from scratch today.

Here are some improvements to the IPC documentation that would have helped me do so. :)


stapelberg (approved):

stapelberg:
Thank you for taking the time to improve these docs!

PR #4397, opened 2021-04-07
Use mkdirp() in get_process_filename()

tomty89:
Avoids race condition in case multiple i3 instances are started in parallel with e.g. systemd user units for multiple X(vfb) servers.

### DIFF
--- a/libi3/get_process_filename.c
+++ b/libi3/get_process_filename.c
@@ -27,15 +27,14 @@ char *get_process_filename(const char *prefix) {
             char *tmp;
             sasprintf(&tmp, "%s/i3", dir);
             dir = tmp;
-            struct stat buf;
-            if (stat(dir, &buf) != 0) {
-                if (mkdir(dir, 0700) == -1) {
-                    warn("Could not mkdir(%s)", dir);
-                    errx(EXIT_FAILURE, "Check permissions of $XDG_RUNTIME_DIR = '%s'",
-                         getenv("XDG_RUNTIME_DIR"));
-                    perror("mkdir()");
-                    return NULL;
-                }
+            /* mkdirp() should prevent race between multiple i3 instances started
+             * in parallel from causing problem */
+            if (mkdirp(dir, 0700) == -1) {
+                warn("Could not mkdirp(%s)", dir);
+                errx(EXIT_FAILURE, "Check permissions of $XDG_RUNTIME_DIR = '%s'",
+                     getenv("XDG_RUNTIME_DIR"));
+                perror("mkdirp()");
+                return NULL;
             }
         } else {
             /* If not, we create a (secure) temp directory using the template


psychon:
Note to whoever merges/approves this: This PR causes a behaviour change. Dunno if you want that.

Before, running with `XDG_RUNTIME_DIR=/tmp/does/not/exist` did not do anything (well, caused i3 to exit with a failure). Now it will create the missing directories.

Perhaps it would be better to just remove the `stat()` call and add code to ignore `EEXIST` from `mkdir`?
```diff
diff --git a/libi3/get_process_filename.c b/libi3/get_process_filename.c
index 21429ec3..b846072a 100644
--- a/libi3/get_process_filename.c
+++ b/libi3/get_process_filename.c
@@ -28,15 +28,13 @@ char *get_process_filename(const char *prefix) {
             sasprintf(&tmp, "%s/i3", dir);
             dir = tmp;
             struct stat buf;
-            if (stat(dir, &buf) != 0) {
-                if (mkdir(dir, 0700) == -1) {
+                if (mkdir(dir, 0700) == -1 && errno != EEXIST) {
                     warn("Could not mkdir(%s)", dir);
                     errx(EXIT_FAILURE, "Check permissions of $XDG_RUNTIME_DIR = '%s'",
                          getenv("XDG_RUNTIME_DIR"));
                     perror("mkdir()");
                     return NULL;
                 }
-            }
         } else {
             /* If not, we create a (secure) temp directory using the template
              * /tmp/i3-<user>.XXXXXX */
```
The code in `mkdirp` also checks if "the thing" really is a directory in case of `EEXIST`. Dunno if that is necessary here. If so, it might make sense to factor that code out into some kind of `ensure_directory_exists` helper that does this part:
https://github.com/i3/i3/blob/fcf4c235d2a7e56fb2164a166ada2d3e7bc74726/libi3/mkdirp.c#L21-L38

tomty89:
More like we can/should do something like `if (!stat(dir = getenv("XDG_RUNTIME_DIR"), &buf))`?

> The code in mkdirp also checks if "the thing" really is a directory in case of EEXIST.

I don't see what's bad with that. Ultimately we need it to be a directory. And maybe a helper is good, so that we can use it for `XDG_RUNTIME_DIR` as well. (Or not necessary? Because the `mkdirp()` would do the job?)

P.S. ~I don't exactly see what's bad with creating `XDG_RUNTIME_DIR` either...hmm, maybe the fact that it won't be a tmpfs?~ Actually can a user even create a directory in `/run/user/` with usual/normal permission?

tomty89:
ping @stapelberg @orestisfl

libi3/get_process_filename.c
```diff
@@ -22,20 +22,21 @@ char *get_process_filename(const char *prefix) {
      * we only create a temporary directory once per i3 instance. */
     static char *dir = NULL;
     if (dir == NULL) {
-        /* Check if XDG_RUNTIME_DIR is set. If so, we use XDG_RUNTIME_DIR/i3 */
-        if ((dir = getenv("XDG_RUNTIME_DIR"))) {
+        /* Check if XDG_RUNTIME_DIR is set and exists.
+         * If so, we use XDG_RUNTIME_DIR/i3 */
+        struct stat buf;
+        if ((dir = getenv("XDG_RUNTIME_DIR")) && !stat(dir, &buf)) {
```
stapelberg: I think it’d be better to remove the `stat` check here. If $XDG_RUNTIME_DIR is set, it’s a grave misconfiguration if it doesn’t actually exist and can’t be created. Failing while creating it in `mkdirp` is an okay failure mode for that.

stapelberg:
Thanks for your PR!

> > The code in mkdirp also checks if "the thing" really is a directory in case of EEXIST.
> 
> I don't see what's bad with that. Ultimately we need it to be a directory.

Yeah, I don’t mind the check either. The error message will be more clear in case the entry is not a directory.

libi3/get_process_filename.c
tomty89: Yeah I think it's a bit of an overkill too. As I said, normally a user is not even allowed to create it anyway. Updated the code.

stapelberg (approved):

PR #4411, opened 2021-04-23
docs: fix broken link to Pango documentation

alarsyo:
Looks like `/stable/` no longer automatically points to the latest version, so I linked to the "homepage" instead. :/


Airblader (changes requested):
Thanks for noticing this and raising a PR! I think it'd be useful to still link to the correct place, so maybe we should just link to the (current) latest version. The exact version hopefully doesn't matter too much anyway. Can you change the PR that way? Thanks!

alarsyo:
Should be alright now :)

Airblader (approved):

Airblader:
Thank you!

Issue #4293, opened 2020-12-26
awesomewm like pop up help


bobblkabb:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Desired Behavior
I would like to see a binding overview help like in awesomewm. All keyboard shortcuts are displayed in a popup. You can call it up via keyboard shortcut.
The help is dynamic, as soon as you have changed bindings it is also reflected in the help.

Tiling managers in particular have a flat learning curve. Such a function is a gold feature.

orestisfl:
Not a bad idea IMO. ~~Can be a completely separate (perl) script that is included with i3~~.

orestisfl:
Continuation from https://github.com/i3/i3/pull/4415#issuecomment-829054360

> Are we even sure we want to make this an officially maintained tool which comes with a default key binding?

I believe that it is a good idea to provide *some* default official tool that provides dynamic help for beginner (and not only) users

> If so, is showing it in a terminal really the approach we should be going for?

No. I believe it's better to have a minimal UI client that can work out of the box without assumptions about the terminal

> The workaround required here for standalone mode is already something I'd be against, with the easy answer being "don't float the window".

Agree

> But really the question is whether an officially provided cheat sheet shouldn't rather be implemented as an actual client.

I agree that it should be implemented as an actual client. If it is a UI it will use the internal libi3 library.

> The question of a separate IPC command is independent of that.

Yes but I believe that we should split the responsibilities between client and server. The client parsing the config with regex is something I'd be against. We can either:
1. (I prefer) Extend the IPC to provide the needed information
2. Refactor the parser code into a library (#2858), use it in the client

> I'm a bit hesitant given that it simply returns a subset of what GET_CONFIG already returns, with the one difference that it returns structured information. The more generalized approach would be to have a structured representation of the entire configuration, not just the bindings.

That's a great idea. The UI can present two kinds of information: configured commands (=bindings) & active config directives (including defaults, if not present in the config).

ghost:
There are https://github.com/matclab/rofi-i3-shortcut-help or https://github.com/Zeioth/rofi-shortcuts

Have these been considered? I run the latter, heavily edited from default to reflect my config. Works like a charm.

orestisfl:
@poetician The first one seems to require a custom config format and the second one seems to read from a static file. We would like to read dynamically from the config and make it work with the default config.

[...]

Issue #2858, opened 2017-08-13
Refactor the config parser into a library


acrisci:
The current config parser is kind of a mess right now and could use some work to ease maintenance in the future.

To support a richer desktop environment experience for i3, I propose that we make this library available to users for their own projects. This will promote maintenance of the library by bringing in more parties who are interested in the proper functioning of the library.

Here are some possible external users of this library:

* Dock clients that interact with i3 such as xfce4-i3-workspaces-plugin.
* Config editors that want to display a representation of the current colorscheme (i3-style).
* Other window managers like sway that use config syntax related to i3 (that could be supported through built-in extensions).

There are probably other great use cases people will think of once this interface is exposed to users. If you build it, they will come.

If the library is written with GObject (included with GLib), we can use introspection to give the library to users as a python library to support user scripting.

cc @SirCmpwn

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow http://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

SirCmpwn:
>Other window managers like sway that use config syntax related to i3

We already have a pretty solid i3 config parser, I doubt the change overhead would be worth it for us.

>If the library is written with GObject (included with GLib)

Then we would *definitely* not use it :wink:

Airblader:
I support this idea. I even wonder whether we should move away from C for this. Rust is something often recommended these days, especially for things like parsers. I'm not sure @stapelberg would be open for this, however.

stapelberg:
I’m not opposed to Rust, provided you feel comfortable maintaining Rust code (my Rust knowledge is quite limited).

If the parser library was written in Rust, would it be usable from all languages which have a C FFI?

One thing which wasn’t mentioned so far is whether the library would live in a separate git repository. My intuition would be that it should live in the same repository as i3, much like the IPC type headers, to keep version skew to a minimum.

We should carefully review the interface we want to provide (regardless of implementation language), to make sure we don’t commit to maintaining a flawed interface for a long time.

Airblader:
I also don't really have knowledge of Rust as of yet. At the moment it's a mere proposal, I doubt the split if the parser will happen over night anyway. You're absolutely right that we need to review these decisions carefully.

[...]

PR #4415, opened 2021-04-28
Add a script to list keybindings

j-jzk:
Adds a script for listing all keybindings, as described in issue #4293.

I've used the system command `i3-msg` instead of `AnyEvent::I3` to not include any new runtime dependencies. I don't see any issues with `i3-msg ...`, but changing it shouldn't be a problem.

Also, the way the script makes its window floating is a little clumsy, but I couldn't find a better way. Is there one?

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -22,6 +22,7 @@ strongly encouraged to upgrade.
   • Add %machine placeholder (WM_CLIENT_MACHINE) to title_format
   • Allow multiple output names in 'move container|workspace to output'
   • Add 'move container|workspace to output next'
+  • Add a script for listing all keybindings and a keybinding for running it
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
--- a/etc/config.keycodes
+++ b/etc/config.keycodes
@@ -57,6 +57,9 @@ bindcode $mod+40 exec --no-startup-id dmenu_run
 # .desktop file. It is a wrapper around dmenu, so you need that installed.
 # bindcode $mod+40 exec --no-startup-id i3-dmenu-desktop
 
+# show list of keybindings
+bindcode $mod+56 exec i3-sensible-terminal -e "i3-list-bindings -s"
+
 # change focus
 bindcode $mod+44 focus left
 bindcode $mod+45 focus down
--- a/i3-list-bindings
+++ b/i3-list-bindings
@@ -1,0 +1,91 @@
+#!/usr/bin/env perl
+# vim:ts=4:sw=4:expandtab
+use strict;
+use warnings;
+use v5.10;
+
+# gets the loaded configuration file and parses it
+# the keybindings are grouped into blocks with an optional comment on top
+sub slurp {
+    # get the loaded config and add a few newlines at the end to make sure the last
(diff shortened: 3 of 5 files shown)


orestisfl:
I know that it is my fault since I made the comment in #4293 but I now feel that this is not a great way to implement this. I made a quick implementation adding a new IPC message: https://github.com/orestisfl/i3/commit/bdb62a6214485c35661dc7037fb4bf7a3c648c03. I vastly prefer to maintain 43 lines of C instead of 90 lines of perl.

We can then wrap around this message type to make a default user-friendly help window.

@Airblader  WDYT?

Airblader:
Are we even sure we want to make this an officially maintained tool which comes with a default key binding? If so, is showing it in a terminal really the approach we should be going for? The workaround required here for standalone mode is already something I'd be against, with the easy answer being "don't float the window". But really the question is whether an officially provided cheat sheet shouldn't rather be implemented as an actual client.

The question of a separate IPC command is independent of that. I'm a bit hesitant given that it simply returns a subset of what GET_CONFIG already returns, with the one difference that it returns structured information. The more generalized approach would be to have a structured representation of the entire configuration, not just the bindings.

orestisfl:
> Are we even sure we want to make this an officially maintained tool which comes with a default key binding?

I believe that it is a good idea to provide *some* default official tool that provides dynamic help for beginner (and not only) users

> If so, is showing it in a terminal really the approach we should be going for?

No. I believe it's better to have a minimal UI client that can work out of the box without assumptions about the terminal

> The workaround required here for standalone mode is already something I'd be against, with the easy answer being "don't float the window".

Agree

> But really the question is whether an officially provided cheat sheet shouldn't rather be implemented as an actual client.

I agree that it should be implemented as an actual client. If it is a UI it will use the internal libi3 library.

> The question of a separate IPC command is independent of that.

Yes but I believe that we should split the responsibilities between client and server. The client parsing the config with regex is something I'd be against. We can either:
1. (I prefer) Extend the IPC to provide the needed information
2. Refactor the parser code into a library (#2858), use it in the client

> I'm a bit hesitant given that it simply returns a subset of what GET_CONFIG already returns, with the one difference that it returns structured information. The more generalized approach would be to have a structured representation of the entire configuration, not just the bindings.

That's a great idea. The UI can present two kinds of information: configured commands (=bindings) & active config directives (including defaults, if not present in the config).

orestisfl:
Since we both disagree with the perl script here, let's move the discussion to #4293

orestisfl:
@j-jzk Irregardless, thank you very much for the contribution and your script. Since it wan work out-of-the box I am sure that many people in the i3 community will find it useful. Consider sharing it in /r/i3wm or the i3 mailing list.

Airblader:
I agree, thanks for the PR. Perhaps @tuxy wants to add it to https://github.com/i3db as well.

ghost:
@Airblader, I'm willing to add anything that i3 community finds useful to i3db. But requested project has to be well maintained and working with current version of i3wm. Otherwise its going to be removed.

EDIT-2: Also, I can add official members of i3wm to i3db organization if you like to contribute too.

EDIT-3: Added: https://github.com/i3db/scripts/blob/main/perl.md

PR #4416, opened 2021-04-28
Remove test timeout

j-jzk:
Meson has a 30-second timeout for tests, which may be too short on slower machines or VMs.

This PR sets the timeout to infinite (or 999 seconds on Meson < 0.57).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -708,6 +708,12 @@ anyevent_i3 = custom_target(
 )
 
 if meson.version().version_compare('>=0.46.0')
+  # meson < 0.57 doesn't support infinite (0) test timeouts
+  test_timeout = 999
+  if meson.version().version_compare('>=0.57')
+    test_timeout = 0
+  endif
+  
   test(
     'complete-run',
     perl,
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -716,6 +722,7 @@ if meson.version().version_compare('>=0.46.0')
       anyevent_i3,
       i3test_pm,
     ],
+    timeout: test_timeout,
```
orestisfl: I don't really think we need infinite timeout, we can just settle on a reasonable default since tests can hang.

Travis executes the tests in 20-30 seconds and it should be a reasonably slow machine. I think we can agree that after 120 seconds something is wrong.

```suggestion
    timeout: 120,  # Default of 30 seconds can cause timeouts on slower machines
```

meson.build
```diff
@@ -708,6 +708,12 @@ anyevent_i3 = custom_target(
 )
 
 if meson.version().version_compare('>=0.46.0')
+  # meson < 0.57 doesn't support infinite (0) test timeouts
+  test_timeout = 999
+  if meson.version().version_compare('>=0.57')
+    test_timeout = 0
+  endif
+
```
orestisfl:
```suggestion
```

See other comment

orestisfl (changes requested):
Thanks!

meson.build
j-jzk: Seems reasonable.

orestisfl (approved):

Issue #1197, opened 2014-03-11
Feature Request: config.d folder


i3bot:
[Originally reported by gustavnikolaj@…]
(Using i3 on multiple different machines poses a problem in terms of configuration. I wish to use the same basic configuration, across devices - but I don't wish to maintain more than one version of it.

A solution to this problem would be a config folder, that would go in the same folder as the config file.

```
- i3
| - config                 // regular config file
| - config.d               // config directory
  | - custom_bindings
  | - startup applications
...
```

The configuration file would be parsed with any additional files from the config.d in a sorted order. 

I understand, that an include directive in the config files have been rejected previously, but I believe this will solve the problem outlined in the questions below:

https://faq.i3wm.org/question/2402/split-configuration/
https://faq.i3wm.org/question/1367/anyway-to-include-in-config-file/

TheLQ:
Any reason this was closed? Has the logging of what config your using been added? Is this more possible in 2016?

Working around this with bash scripts isn't ideal and goes against what many other apps accomplish with the .d directories.

Airblader:
I don't see a reason either, not in this ticket anyway. I'll reopen for now. @stapelberg Has your view on this changed or is it still a valid request?

stapelberg:
commit 44367572c98bdb69730d86755dd3edb5207bed3c added the config file path (and its last modification time) to the output of `i3 --moreversion`.

The ticket was closed with the submit_to_cr.i3wm.org status because that was our process for marking suggestions as accepted before we moved to GitHub.

Airblader:
I'm not sure I'm interpreting that correctly. So is this still an accepted feature request?

stapelberg:
Yes.

Airblader:
Just to anyone taking a shot at this, I think the following things should be taken care of:

* Ensure that the used config folder is added to `i3 --moreversion`.
* Ensure we log all individual filenames we parse.
* Ensure we still log the "finished" config as we do today.
* Ensure a proper sorting of the files.

I'd also suggest that we append all files from the folder to the "main" config (rather than prepending).

Airblader:
One more thing is the location of `config.d`. I think easiest would be to just have it always be a folder in the same directory of the main config file? However, that also means that we always require a config file, even if it's empty. I think that's OK, though.

cshinaver:
Should the new "finished" config file be created, parsed and used by i3, and then discarded? 

-austrinus

Airblader:
[...]

PR #4420, opened 2021-05-13
Implement include config directive

### LINKED ISSUE #4192: Allow for including configuration partials
## I'm submitting a…
<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
Currently, i3 only "knows" about a single configuration file it reads at runtime and processes as a whole. It does allow for modifications coming from [`Xresources`](https://i3wm.org/docs/userguide.html#xresources), however, those are all included in the single, main configuration for it to read out once and then conclude its configuration phase at the end of it.

What the `Xresources` method is incapable of though is the modification of configuration *instructions* e.g. `for_window`, `assign` or similar objects. To change those you always have to go back and edit the configuration file and restart i3 afterwards (thankfully, i3 makes this process rather convenient!).

You cannot tell i3 to "look somewhere else" for other configuration instructions, it's a single file. Full stop.

## Desired Behavior

What I would love to have in i3 is the capability to **include other configuration partials**, or "snippets", which are then *merged* into a single, large configuration *tree*, interpreted and i3 then configured accordingly. Something along the lines of:

```
# /etc/i3/config
[...]
for_window [class=something] floating enable

include /home/user/.config/my_customizations
```
```
# .config/my_customizations
for_window [class=something] floating disable
```
If you're taking advise from popular configuration parsers it'll probably be along the lines of "last instructions wins" (e.g. whatever "overrides" the configuration in the tree last determines the behavior), but I'm not tied to it.

One could also ponder the ability to check for user-defined configurations automatically (e.g. in `~/.i3/config`) as i3 does it today, even when it's told to look somewhere else, but that might introduce more harm than good.

## What is the use case?

I'm a contributor for a i3-based Linux distribution called Regolith Linux, which is making it possible for folks to configure i3 to their liking to a certain extend, using `Xresources`. We are pretty far ahead with how many of the potential options you're able to change, however, for the more advanced customizations (e.g. `assign` or `for_window`) you have to follow [a process called "staging" for the configuration files](https://regolith-linux.org/docs/howto/stage-configs/#i3-config) in order to extend i3's functionality. This has the obvious disadvantage that it might potentially cause the staged configuration to become stale as soon as the system-wide i3 configuration file living in `/etc/regolith/i3/config` changes over time (e.g. we're fixing bugs, introducing new functionality etc.). This diversion not only causes pain for the users, as they have to manually keep track of how the global configuration changes over time, but it's also a pain in the butt for us contributors, since a lot of bug reports are essentially "staged" configurations that haven't been updated in a while.

The ability to partially override settings of i3 that aren't assignments static instructions, but actually provide instructions on your own would complement the already existing flexibility in the configuration system and allow for folks to rely on sane defaults from the get-go.

## Environment

This applies to all environments i3 is running in.

stapelberg:
The implementation [uses wordexp(3) just like sway](https://github.com/i3/i3/issues/1197#issuecomment-226844106).

Thanks to jajm [for their implementation](https://github.com/jajm/i3/commit/bb55709d0aa0731f7b3c641871731a992ababb1a).

This required refactoring the config parser to be re-entrant
(no more global state) and to return an error instead of dying.

In case a file cannot be opened, i3 reports an error but proceeds with the
remaining configuration.

All files that were successfully included are displayed in `i3 --moreversion`.

One caveat is i3 config file variable expansion, see the note in the userguide.

fixes #4192

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -319,6 +319,90 @@ include the following line in your config file:
 # i3 config file (v4)
 ---------------------
 
+[[include]]
+=== Include directive
+
+Since i3 v4.20, it is possible to include other configuration files from your i3
+configuration.
+
+*Syntax*:
+-----------------
+include <pattern>
+-----------------
+
+i3 expands `pattern` using shell-like word expansion, specifically using the
+https://manpages.debian.org/wordexp.3[`wordexp(3)` C standard library function].
+
+*Examples*:
+--------------------------------------------------------------------------------
+# Tilde expands to the user’s home directory:
+include ~/.config/i3/assignments.conf
+
+# Environment variables are expanded:
+include $HOME/.config/i3/assignments.conf
+
+# Wildcards are expanded:
+include ~/.config/i3/config.d/*.conf
+
+# Command substitution:
+include ~/.config/i3/`hostname`.conf
+
+# i3 loads each path only once, so including the i3 config will not result
+# in an endless loop, but in an error:
+include ~/.config/i3/config
+
+# i3 changes the working directory while parsing a config file
+# so that relative paths are interpreted relative to the directory
+# of the config file that contains the path:
+include assignments.conf
+--------------------------------------------------------------------------------
+
+If a specified file cannot be read, for example because of a lack of file
+permissions, or because of a dangling symlink, i3 will report an error and
+continue processing your remaining configuration.
+
+To list all loaded configuration files, run `i3 --moreversion`:
+
+--------------------------------------------------------------------------------
+% i3 --moreversion
+Binary i3 version:  4.19.2-87-gfcae64f7+ © 2009 Michael Stapelberg and contributors
+Running i3 version: 4.19.2-87-gfcae64f7+ (pid 963940)
+Loaded i3 config:
+  /tmp/i3.cfg (main) (last modified: 2021-05-13T16:42:31 CEST, 463 seconds ago)
+  /tmp/included.cfg (included) (last modified: 2021-05-13T16:42:43 CEST, 451 seconds ago)
+  /tmp/another.cfg (included) (last modified: 2021-05-13T16:42:46 CEST, 448 seconds ago)
+--------------------------------------------------------------------------------
+
+Variables are shared between all config files, but beware of the following limitation:
+
+* You can define a variable and use it within an included file.
+* You cannot use (in the parent file) a variable that was defined within an included file.
+
+This is a technical limitation: variable expansion happens in a separate stage
+before parsing include directives.
+
+Conceptually, included files can only add to the configuration, not undo the
+effects of already-processed configuration. For example, you can only add new
+key bindings, not overwrite or remove existing key bindings. This means:
+
+* The `include` directive is suitable for organizing large configurations into
+  separate files, possibly selecting files based on conditionals.
+
+* The `include` directive is not suitable for expressing “use the default
+  configuration with the following changes”. For that case, we still recommend
+  copying and modifying the default config.
+
+[NOTE]
+====
+Implementation-wise, i3 does not currently construct one big configuration from
+all `include` directives. Instead, i3’s config file parser interprets all
+configuration directives in its `parse_file()` function. When processing an
+`include` configuration directive, the parser recursively calls `parse_file()`.
+
+This means the evaluation order of files forms a tree, or one could say i3 uses
+depth-first traversal.
+====
+
 === Comments
 
 It is possible and recommended to use comments in your configuration file to
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -718,6 +718,40 @@ void reorder_bindings(void) {
 }
 
 /*
+ * Returns true if a is a key binding for the same key as b.
+ *
+ */
+static bool binding_same_key(Binding *a, Binding *b) {
+    /* Check if the input types are different */
+    if (a->input_type != b->input_type) {
+        return false;
+    }
+
+    /* Check if one is using keysym while the other is using bindsym. */
+    if ((a->symbol == NULL && b->symbol != NULL) ||
+        (a->symbol != NULL && b->symbol == NULL)) {
+        return false;
+    }
+
+    /* If a is NULL, b has to be NULL, too (see previous conditional).
+     * If the keycodes differ, it can't be a duplicate. */
+    if (a->symbol != NULL &&
+        strcasecmp(a->symbol, b->symbol) != 0) {
+        return false;
+    }
+
+    /* Check if the keycodes or modifiers are different. If so, they
+     * can't be duplicate */
+    if (a->keycode != b->keycode ||
+        a->event_state_mask != b->event_state_mask ||
+        a->release != b->release) {
+        return false;
+    }
+
+    return true;
+}
+
+/*
  * Checks for duplicate key bindings (the same keycode or keysym is configured
  * more than once). If a duplicate binding is found, a message is printed to
  * stderr and the has_errors variable is set to true, which will start
@@ -730,31 +764,13 @@ void check_for_duplicate_bindings(struct context *context) {
         TAILQ_FOREACH (bind, bindings, bindings) {
             /* Abort when we reach the current keybinding, only check the
              * bindings before */
-            if (bind == current)
+            if (bind == current) {
                 break;
+            }
 
-            /* Check if the input types are different */
-            if (bind->input_type != current->input_type)
+            if (!binding_same_key(bind, current)) {
                 continue;
-
-            /* Check if one is using keysym while the other is using bindsym.
-             * If so, skip. */
-            if ((bind->symbol == NULL && current->symbol != NULL) ||
-                (bind->symbol != NULL && current->symbol == NULL))
-                continue;
-
-            /* If bind is NULL, current has to be NULL, too (see above).
-             * If the keycodes differ, it can't be a duplicate. */
-            if (bind->symbol != NULL &&
-                strcasecmp(bind->symbol, current->symbol) != 0)
-                continue;
-
-            /* Check if the keycodes or modifiers are different. If so, they
-             * can't be duplicate */
-            if (bind->keycode != current->keycode ||
-                bind->event_state_mask != current->event_state_mask ||
-                bind->release != current->release)
-                continue;
+            }
 
             context->has_errors = true;
             if (current->keycode != 0) {
--- a/src/config.c
+++ b/src/config.c
@@ -10,6 +10,9 @@
  */
 #include "all.h"
 
+#include <libgen.h>
+#include <unistd.h>
+
 #include <xkbcommon/xkbcommon.h>
 
 char *current_configpath = NULL;
@@ -17,6 +20,7 @@ char *current_config = NULL;
 Config config;
 struct modes_head modes;
 struct barconfig_head barconfigs = TAILQ_HEAD_INITIALIZER(barconfigs);
+struct includedfiles_head included_files = TAILQ_HEAD_INITIALIZER(included_files);
 
 /*
  * Ungrabs all keys, to be called before re-grabbing the keys because of a
@@ -225,8 +229,42 @@ bool load_configuration(const char *override_configpath, config_load_t load_type
             "$XDG_CONFIG_HOME/i3/config, ~/.i3/config, $XDG_CONFIG_DIRS/i3/config "
             "and " SYSCONFDIR "/i3/config)");
     }
-    LOG("Parsing configfile %s\n", current_configpath);
-    const bool result = parse_file(current_configpath, load_type != C_VALIDATE);
+
+    IncludedFile *file;
+    while (!TAILQ_EMPTY(&included_files)) {
+        file = TAILQ_FIRST(&included_files);
+        FREE(file->path);
+        TAILQ_REMOVE(&included_files, file, files);
+        FREE(file);
+    }
+
+    char resolved_path[PATH_MAX] = {'\0'};
+    if (realpath(current_configpath, resolved_path) == NULL) {
+        die("realpath(%s): %s", current_configpath, strerror(errno));
+    }
+
+    file = scalloc(1, sizeof(IncludedFile));
+    file->path = sstrdup(resolved_path);
+    TAILQ_INSERT_TAIL(&included_files, file, files);
+
+    LOG("Parsing configfile %s\n", resolved_path);
+    struct stack stack;
+    memset(&stack, '\0', sizeof(struct stack));
+    struct parser_ctx ctx = {
+        .use_nagbar = (load_type != C_VALIDATE),
+        .assume_v4 = false,
+        .stack = &stack,
+    };
+    SLIST_INIT(&(ctx.variables));
+    FREE(current_config);
+    const int result = parse_file(&ctx, resolved_path);
+    free_variables(&ctx);
+    if (result == -1) {
+        die("Could not open configuration file: %s\n", strerror(errno));
+    }
+
+    extract_workspace_names_from_bindings();
+    reorder_bindings();
 
     if (config.font.type == FONT_TYPE_NONE && load_type != C_VALIDATE) {
         ELOG("You did not specify required configuration option \"font\"\n");
@@ -245,5 +283,5 @@ bool load_configuration(const char *override_configpath, config_load_t load_type
         xcb_flush(conn);
     }
 
-    return result;
+    return result == 0;
 }
--- a/src/config_directives.c
+++ b/src/config_directives.c
@@ -9,6 +9,85 @@
  */
 #include "all.h"
 
+#include <wordexp.h>
+
+/*******************************************************************************
+ * Include functions.
+ ******************************************************************************/
+
+CFGFUN(include, const char *pattern) {
+    DLOG("include %s\n", pattern);
+
+    wordexp_t p;
+    const int ret = wordexp(pattern, &p, 0);
+    if (ret != 0) {
+        ELOG("wordexp(%s): error %d\n", pattern, ret);
+        result->has_errors = true;
+        return;
+    }
+    char **w = p.we_wordv;
+    for (size_t i = 0; i < p.we_wordc; i++) {
+        char resolved_path[PATH_MAX] = {'\0'};
+        if (realpath(w[i], resolved_path) == NULL) {
+            ELOG("realpath(%s): %s\n", w[i], strerror(errno));
+            result->has_errors = true;
+            continue;
+        }
+
+        bool skip = false;
+        IncludedFile *file;
+        TAILQ_FOREACH (file, &included_files, files) {
+            if (strcmp(file->path, resolved_path) == 0) {
+                skip = true;
+                break;
+            }
+        }
+        if (skip) {
+            LOG("Skipping file %s (already included)\n", resolved_path);
+            continue;
+        }
+
+        LOG("Including config file %s\n", resolved_path);
+
+        file = scalloc(1, sizeof(IncludedFile));
+        file->path = sstrdup(resolved_path);
+        TAILQ_INSERT_TAIL(&included_files, file, files);
+
+        struct stack stack;
+        memset(&stack, '\0', sizeof(struct stack));
+        struct parser_ctx ctx = {
+            .use_nagbar = result->ctx->use_nagbar,
+            /* The include mechanism was added in v4, so we can skip the
+             * auto-detection and get rid of the risk of detecting the wrong
+             * version in potentially very short include fragments: */
+            .assume_v4 = true,
+            .stack = &stack,
+            .variables = result->ctx->variables,
+        };
+        switch (parse_file(&ctx, resolved_path)) {
+            case PARSE_FILE_SUCCESS:
+                break;
+
+            case PARSE_FILE_FAILED:
+                ELOG("including config file %s: %s\n", resolved_path, strerror(errno));
+                /* fallthrough */
+
+            case PARSE_FILE_CONFIG_ERRORS:
+                result->has_errors = true;
+                TAILQ_REMOVE(&included_files, file, files);
+                FREE(file->path);
+                FREE(file);
+                break;
+
+            default:
+                /* missing case statement */
+                assert(false);
+                break;
+        }
+    }
+    wordfree(&p);
+}
+
 /*******************************************************************************
  * Criteria functions.
  ******************************************************************************/
--- a/src/config_parser.c
+++ b/src/config_parser.c
@@ -35,18 +35,14 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <libgen.h>
 
 #include <xcb/xcb_xrm.h>
-
-// Macros to make the YAJL API a bit easier to use.
-#define y(x, ...) yajl_gen_##x(command_output.json_gen, ##__VA_ARGS__)
-#define ystr(str) yajl_gen_string(command_output.json_gen, (unsigned char *)str, strlen(str))
 
 xcb_xrm_database_t *database = NULL;
 
 #ifndef TEST_PARSER
 pid_t config_error_nagbar_pid = -1;
-static struct context *context;
 #endif
 
 /*******************************************************************************
@@ -76,46 +72,25 @@ typedef struct tokenptr {
 
 #include "GENERATED_config_tokens.h"
 
-/*******************************************************************************
- * The (small) stack where identified literals are stored during the parsing
- * of a single command (like $workspace).
- ******************************************************************************/
-
-struct stack_entry {
-    /* Just a pointer, not dynamically allocated. */
-    const char *identifier;
-    enum {
-        STACK_STR = 0,
-        STACK_LONG = 1,
-    } type;
-    union {
-        char *str;
-        long num;
-    } val;
-};
-
-/* 10 entries should be enough for everybody. */
-static struct stack_entry stack[10];
-
 /*
  * Pushes a string (identified by 'identifier') on the stack. We simply use a
  * single array, since the number of entries we have to store is very small.
  *
  */
-static void push_string(const char *identifier, const char *str) {
+static void push_string(struct stack *ctx, const char *identifier, const char *str) {
     for (int c = 0; c < 10; c++) {
-        if (stack[c].identifier != NULL &&
-            strcmp(stack[c].identifier, identifier) != 0)
+        if (ctx->stack[c].identifier != NULL &&
+            strcmp(ctx->stack[c].identifier, identifier) != 0)
             continue;
-        if (stack[c].identifier == NULL) {
+        if (ctx->stack[c].identifier == NULL) {
             /* Found a free slot, let’s store it here. */
-            stack[c].identifier = identifier;
-            stack[c].val.str = sstrdup(str);
-            stack[c].type = STACK_STR;
+            ctx->stack[c].identifier = identifier;
+            ctx->stack[c].val.str = sstrdup(str);
+            ctx->stack[c].type = STACK_STR;
         } else {
             /* Append the value. */
-            char *prev = stack[c].val.str;
-            sasprintf(&(stack[c].val.str), "%s,%s", prev, str);
+            char *prev = ctx->stack[c].val.str;
+            sasprintf(&(ctx->stack[c].val.str), "%s,%s", prev, str);
             free(prev);
         }
         return;
@@ -130,14 +105,15 @@ static void push_string(const char *identifier, const char *str) {
     exit(EXIT_FAILURE);
 }
 
-static void push_long(const char *identifier, long num) {
+static void push_long(struct stack *ctx, const char *identifier, long num) {
     for (int c = 0; c < 10; c++) {
-        if (stack[c].identifier != NULL)
+        if (ctx->stack[c].identifier != NULL) {
             continue;
+        }
         /* Found a free slot, let’s store it here. */
-        stack[c].identifier = identifier;
-        stack[c].val.num = num;
-        stack[c].type = STACK_LONG;
+        ctx->stack[c].identifier = identifier;
+        ctx->stack[c].val.num = num;
+        ctx->stack[c].type = STACK_LONG;
         return;
     }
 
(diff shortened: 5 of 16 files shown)


kgilmer:
Happy to see this <3.  Do you plan to ship an updated default config file with a wildcard matching scheme similar to [the Sway solution](https://git.sr.ht/~sircmpwn/dotfiles/tree/master/item/.config/sway/config#L175)?

docs/userguide
```diff
@@ -319,6 +319,68 @@ include the following line in your config file:
...
+Loaded i3 config:
+  /tmp/i3.cfg (main) (last modified: 2021-05-13T16:42:31 CEST, 463 seconds ago)
+  /tmp/included.cfg (included) (last modified: 2021-05-13T16:42:43 CEST, 451 seconds ago)
+  /tmp/another.cfg (included) (last modified: 2021-05-13T16:42:46 CEST, 448 seconds ago)
+--------------------------------------------------------------------------------
+
+Variables are shared between all config files, but beware of the following limitation:
+
+* You can define a variable and use it within an included file.
+* You cannot use (in the parent file) a variable that was defined within an included file.
```
kgilmer, line 379: Is it actually a parent/child relationship here or simply order of evaluation of the partials?

Airblader:
I'll hopefully get around to reviewing this tomorrow. I'll +1 adding wildcard includes for distributions like Regolith.

One concern coming to mind is duplicate binding handling. There's a case to be made that overrides should be allowed across files, but perhaps just not within the same file?

docs/userguide
stapelberg: Can you elaborate on which nuance you’re trying to get at?

Since the parent is always evaluated before any children, the parent always comes first in the order of evaluation.

Perhaps we should mention that variables defined in one file can be used in a sibling file (e.g. `include a*.conf` with a00.conf and a01.conf, you can set a variable in a00.conf and use it in a01..conf).

stapelberg:
> I'll +1 adding wildcard includes for distributions like Regolith.

What makes Regolith special in this regard?

How would these wildcard includes need to look like?

Just `include /etc/i3/config.d/*.conf` to handle system-level additions, or also `include config.d/*.conf` to handle user-level additions? Or would we leave the latter up to the user to set up?

> One concern coming to mind is duplicate binding handling. There's a case to be made that overrides should be allowed across files, but perhaps just not within the same file?

Perhaps we loosen the duplicate binding restriction in a separate step later on? It’s not clear to me that we necessarily need to, or why bindings are a special case compared to other directives?

Airblader:
> What makes Regolith special in this regard?

They currently need to ship a customized, self maintained default config AFAIK. With wildcards they could instead just ship additional files but leave the default config from i3 in place.

> Or would we leave the latter up to the user to set up?

I don't feel too strongly about those, but don't see a harm in adding that either. The system level ones will just make life for packagers easier who want to offer customized i3 packages.

> or why bindings are a special case compared to other directives?

If anything, they're a special case right now, since we don't check for duplicate directives for anything other than bindings. So overwriting any directives is totally fine, just bindings cause issues. 
Again, for something like Regolith that probably makes this feature less useful. But perhaps someone from Regolith can comment more on that, @kgilmer?

src/config_directives.c
```diff
@@ -9,6 +9,80 @@
...
+        file->path = sstrdup(resolved_path);
+        TAILQ_INSERT_TAIL(&included_files, file, files);
+
+        struct stack stack;
+        memset(&stack, '\0', sizeof(struct stack));
+        struct parser_ctx ctx = {
+            .use_nagbar = result->ctx->use_nagbar,
+            // The include mechanism was added in v4, so we can skip the
+            // auto-detection and get rid of the risk of detecting the wrong
+            // version in potentially very short include fragments:
```
orestisfl: Nit: `/* */` comments :)

src/config_directives.c
```diff
@@ -9,6 +9,80 @@
...
+            case -1:
+                ELOG("including config file %s: %s\n", resolved_path, strerror(errno));
+                // fallthrough
+
+            case 1:
+                result->has_errors = true;
+                TAILQ_REMOVE(&included_files, file, files);
+                FREE(file->path);
+                FREE(file);
+                break;
```
orestisfl, line 80: Since `parse_file` is an `int` we are missing a `default` case here, a compiler might complain

src/config_directives.c
```diff
@@ -9,6 +9,80 @@
...
+            .stack = &stack,
+            .variables = result->ctx->variables,
+        };
+        switch (parse_file(&ctx, resolved_path)) {
+            case 0:
+                break;
+
+            case -1:
+                ELOG("including config file %s: %s\n", resolved_path, strerror(errno));
+                // fallthrough
```
orestisfl:
```suggestion
                /* fallthrough */
```

src/config.c
```diff
@@ -225,8 +229,33 @@ bool load_configuration(const char *override_configpath, config_load_t load_type
...
+        file = TAILQ_FIRST(&included_files);
+        FREE(file->path);
+        TAILQ_REMOVE(&included_files, file, files);
+        FREE(file);
+    }
+
+    char resolved_path[PATH_MAX] = {'\0'};
+    if (realpath(current_configpath, resolved_path) == NULL) {
+        die("realpath(%s): %s", current_configpath, strerror(errno));
+    }
```
orestisfl, line 244: It seems that the root file can be included by an `include` directive. This seems to fix that:

```suggestion
    }
    
    file = scalloc(1, sizeof(IncludedFile));
    file->path = sstrdup(resolved_path);
    TAILQ_INSERT_TAIL(&included_files, file, files);
    
```

docs/userguide
```diff
@@ -319,6 +319,68 @@ include the following line in your config file:
...
+include ~/.config/i3/assignments.conf
+
+# Environment variables are expanded:
+include $HOME/.config/i3/assignments.conf
+
+# Wildcards are expanded:
+include ~/.config/i3/config.d/*.conf
+
+# Command substitution:
+include ~/.config/i3/`hostname`.conf
```
orestisfl, line 348: Which is the code that implements the backtick functionality? What happens (and what should happen?) when the command fails?

orestisfl (changes requested):

docs/userguide
orestisfl: Ok, it's [`wordexp`](https://man7.org/linux/man-pages/man3/wordexp.3.html)

src/config_directives.c
orestisfl: Btw, we should add `include` in here: https://github.com/i3/i3/blob/9db03797da3cea5dc6898adc79a68ba4523a409c/src/config_parser.c#L638-L644

stapelberg: Comment changed. We already have `include` in the code you link.

src/config_directives.c
stapelberg: Done.

src/config_directives.c
stapelberg: Done.

src/config.c
stapelberg: Good catch, done!

docs/userguide
stapelberg: Stdout of the command is relevant, the exit code is ignored.

If a command fails without producing any output, the backticks evaluate to an empty string.

stapelberg: Also added a test to exercise this behavior.

docs/userguide
kgilmer: Surely, I can elaborate..my question comes from reasoning about i3 partial load order and how, say, keybinding state in i3 may be mutated as a consequence of evaluation of one partial to another.  I've noticed in another thread in the PR that keybinding overrides may not be allowed (as-is i3 throws an error if the same key is bound to within the config file).  In the case that overrides are allowed from partial to partial, it becomes important to be able to understand partial evaluation order.  If internally it's evaluated as list this is easier than a tree structure, but I guess really I just need to know that the sort on config file partial is stable/consistent.

This all stems from the desire for Regolith to conform to baseline configuration as much as possible, as to not hinder users from being able to take advantage of all of the existing community work in customizations etc..  Regolith (by choice) does not conform to the standard i3 keybindings (in particular window navigation, dmenu/rofi launch).  In order for Regolith to keep it's variance we either need to be able to override default keybindings in a partial, or just launch i3 in the Xsession initialization with a non-standard config file as is done now.  Thinking through this, overriding keybindings in partials may end up being more trouble/confusion to end users, as they would need to understand this overriding behavior as well as partial loading heuristics (the root problem being that the entire config file merged would have conflicting declarations).

stapelberg: > If internally it's evaluated as list this is easier than a tree structure, but I guess really I just need to know that the sort on config file partial is stable/consistent.

There is no sorting going on. i3 is recursively parsing configuration files in the order in which you tell it to.

I guess that maps to a “tree” in your terminology.

stapelberg:
Sounds like we should tie the binding duplicate check to the file in which the binding is defined.

kgilmer:
> They currently need to ship a customized, self maintained default config AFAIK. With wildcards they could instead just ship additional files but leave the default config from i3 in place.

Yes,  this would be a good change if Regolith was able to use stock i3 config file(s).  A concern I have is possible difficult in users understanding why a given keybinding was mapped if there are multiple partials that all may set that keybinding.  For example, the stock i3 may map the app launcher to `<super> d` but Regolith would override this to be `<super> space`.  Generally speaking, if overrides are allowed, to be useful for Regolith we'd need Update and Delete in addition to Create.

> If anything, they're a special case right now, since we don't check for duplicate directives for anything other than bindings. So overwriting any directives is totally fine, just bindings cause issues.
> Again, for something like Regolith that probably makes this feature less useful. But perhaps someone from Regolith can comment more on that, @kgilmer?

I may be missing some simpler approaches but as I see it now, if we allow overrides across partials, I'm guessing that i3 will undergo a general regression relating to the complexity of the i3 config file.  Because before config partials, a keybinding is a keybinding.  But after partials, a keybinding is whatever keybinding expression was evaluated last across the ordered set of evaluated partials.  Partials as a feature us a huge benefit to Regolith, and using a custom config is relatively minor.  That said I may be missing some simple way that keybinding overrides can be implemented that does not increase general complexity in the i3 config file.

stapelberg:
> Generally speaking, if overrides are allowed, to be useful for Regolith we'd need Update and Delete in addition to Create.

Currently, every directive behaves like Upsert: update if present, and insert if not present yet.

Can you explain why you need Delete? 

----

Edit: actually, key bindings don’t upsert currently, but only ever append.

Airblader:
> Generally speaking, if overrides are allowed, to be useful for Regolith we'd need Update and Delete in addition to Create.

That's a good point.

I think changing the behavior for duplicate binding detection might not be a good idea. An alternative could be modularizing the default config to eg move bindings into a separate file. Then packagers could just not ship that one file and instead their own bindings, which allows all additions and deletions. I'm not sure we'd want to do that with the default config, though.

kgilmer:
Let me say all that again in another way: I presently don't see a way of being able to re-use a stock i3 config file in Regolith without having to introduce conflicting keybinding statements (plus either/or partial semantics or some kind of "unbind" operator) in partials. These conflicts I expect would confuse some users.

kgilmer:
> Can you explain why you need Delete?

A contrived example is that stock i3 uses `<super> d` for drun.  Regolith uses `<super> space`.  I would want` <super> d` to do nothing. (in practice `<super> d` in Regolith maps to gnome display settings, but I'm guessing there are actual cases like this presently).

kgilmer:
> I think changing the behavior for duplicate binding detection might not be a good idea. An alternative could be modularizing the default config to eg move bindings into a separate file. Then packagers could just not ship that one file and instead their own bindings, which allows all additions and deletions. I'm not sure we'd want to do that with the default config, though.

Yes, in my mind weighing the options i3 would want to maintain continuity of the existing monolithic config file (because it is simple, easy to reason about, and well known) rather than the top level file just being a skeleton and breaking partials out for each ["config category" as I have done in a PoC for Regolith 2.0](https://github.com/regolith-linux/regolith-i3-config/tree/master/partials).

stapelberg:
> A contrived example is that stock i3 uses `<super> d` for drun. Regolith uses `<super> space`. I would want` <super> d` to do nothing. (in practice `<super> d` in Regolith maps to gnome display settings, but I'm guessing there are actual cases like this presently).

Right, I think we’ll have to declare this use-case out of scope.

Config parsing with includes is “additive” or “overwrite-only”, no directive (neither key bindings nor others like e.g. focus_follows_mouse) supports expressing “reset this back to default”.

Airblader (approved):
LGTM! This will surely make a lot of people happy.

docs/userguide
stapelberg: Added a note in the userguide about this.

stapelberg:
Pushed an updated version where I also introduce the `--remove` flag for `bindsym` and `bindcode`.

These can be used to overwrite or remove key bindings, which is useful from included config files.

Duplicate key binding checking happens after each file was parsed, i.e. on the return path of the recursion. I don’t think we need any changes to duplicate key binding checking after all.

I added basic tests for this, but please give it a shot and see if it works as expected.

This should cover most use-cases, and I think implementing it for key bindings as a special case makes sense. For the other config directives, we can consider similar behavior on a case-by-case basis, and I expect most will work fine without modifications.

Airblader:
I'm a bit confused about that change. I thought we didn't want to support this use case? I'm not actually a fan of a flag to remove bindings.

stapelberg:
> I thought we didn't want to support this use case? 

Initially, I thought one could at least overwrite key bindings, but then I realized that’s not currently possible.

Hence, I came around on the need to change something about key bindings.

> I'm not actually a fan of a flag to remove bindings.

What’s your concern about this flag? The implementation is pretty straight-forward, and use-cases such as “use the standard i3 config, but overwrite that one key binding” become much easier to express.

Airblader:
I think removing bindings "somewhere down the line" adds significant mental complexity. For a configuration format something like "undo this configuration I just did above" feels awkward; configs are static, which is what sets config and commands apart.

Allowing to statically remove statically added bindings from within the config, but not allowing adding or removing for_window directives at runtime, for example, sounds illogical to me and hard to justify (yes I'm aware they're two different things). There's also no way to statically remove for_window directives, nor bar blocks etc. for that matter.

Such a flag feels less like a solution and more like a workaround to me.

Airblader:
Another note on this: removing previous bindings is sensitive to the inclusion order.

stapelberg:
> I think removing bindings "somewhere down the line" adds significant mental complexity.

I agree that `include` directives in general, and overwriting semantics in particular make configuration more complex.

I suppose there are ways to counter that complexity with easier debugging. Perhaps even switching from recursively evaluating config files to actually concatenating them first might be a change that helps in that regard, but I think that can be done in a separate pull request.

> For a configuration format something like "undo this configuration I just did above" feels awkward;

For multi-file configs, this is a standard feature. See for example systemd’s drop-in config files, where the `ExecStart=` directive is a list, and if you want to overwrite that list from a drop-in config file, you need to empty it (“remove” it in our terminology) explicitly first, e.g.:
```
% cat /etc/systemd/system/robustirc-bridge.service.d/override.conf 
[Service]
ExecStart=
ExecStart=/usr/bin/robustirc-bridge -socks= -listen=localhost:6637 -network=robustirc.net
```
```
% systemctl status robustirc-bridge
● robustirc-bridge.service - RobustIRC bridge
     Loaded: loaded (/usr/lib/systemd/system/robustirc-bridge.service; enabled; vendor preset: disabled)
    Drop-In: /etc/systemd/system/robustirc-bridge.service.d
             └─override.conf
```

> configs are static, which is what sets config and commands apart.

I don’t think of them in terms of “static” (and “dynamic”?), but rather in terms of evaluation time: the configuration is applied when loading/reloading it, whereas commands are applied/interpreted/executed whenever needed.

> Allowing to statically remove statically added bindings from within the config, but not allowing adding or removing for_window directives at runtime, for example, sounds illogical to me and hard to justify (yes I'm aware they're two different things).

Amending the config at runtime is a separate feature, I think.

> There's also no way to statically remove for_window directives, nor bar blocks etc. for that matter.

Yes, it might make sense to eventually add support to these, too, if that turns out to be a popular user request.

> Another note on this: removing previous bindings is sensitive to the inclusion order.

Yes, but so are many other directives that already effectively have overwrite semantics, such as changing the font with the `font` directive.

Also note that the variable `set` command (and `set_from_resource`) use upsert semantics already.

src/bindings.c
```diff
@@ -83,7 +83,15 @@ Binding *configure_binding(const char *bindtype, const char *modifiers, const ch
         new_binding->keycode = keycode;
         new_binding->input_type = B_KEYBOARD;
     }
-    new_binding->command = sstrdup(command);
+    if (remove != NULL && command == NULL) {
```
orestisfl: This `if` is unnecessary, the comment can stay

src/bindings.c
```diff
@@ -98,7 +106,37 @@ Binding *configure_binding(const char *bindtype, const char *modifiers, const ch
...
+            if (binding_same_key(bind, new_binding)) {
+                TAILQ_REPLACE(mode->bindings, bind, new_binding, bindings);
+                found = true;
+                break;
+            }
+        }
+    }
+    if (!found) {
+        TAILQ_INSERT_TAIL(mode->bindings, new_binding, bindings);
+    }
```
orestisfl: Not necessarily a strictly better way to write this but WDYT about this:

```suggestion

if (remove) {
    bool found = false;
    Binding *bind;
    TAILQ_FOREACH (bind, mode->bindings, bindings) {
        if (binding_same_key(bind, new_binding)) {
            if (command){
                /* --remove specified with command: find existing key binding
                 * and replace it, or append if no existing key binding found. */
                TAILQ_REPLACE(mode->bindings, bind, new_binding, bindings);
                found = true;
                break;
            } else {
                /* --remove specified without command: remove the existing key
                 * binding (if any) without replacement. */
                TAILQ_REMOVE(mode->bindings, bind, bindings);
                return NULL;
            }
        }
    }

    if (!command){
        return NULL;
    } else if (!found){
        TAILQ_INSERT_TAIL(mode->bindings, new_binding, bindings);
    }
}
```

src/bindings.c
```diff
@@ -717,6 +755,40 @@ void reorder_bindings(void) {
...
+        return false;
+    }
+
+    /* Check if one is using keysym while the other is using bindsym. */
+    if ((a->symbol == NULL && b->symbol != NULL) ||
+        (a->symbol != NULL && b->symbol == NULL)) {
+        return false;
+    }
+
+    /* If bind is NULL, current has to be NULL, too (see previous conditional).
```
orestisfl: I think here you are referring to old variables, `bind`, `current`

src/config_directives.c
```diff
@@ -9,6 +9,83 @@
...
+
+            case 1:
+                result->has_errors = true;
+                TAILQ_REMOVE(&included_files, file, files);
+                FREE(file->path);
+                FREE(file);
+                break;
+
+            default:
+                break;
```
orestisfl, line 85: I was thinking more in the lines of an assertion failure maybe? The code doesn't return something different.

Alternatively, I might prefer to introduce an enum for readability here instead.

testcases/t/313-include.t
```diff
@@ -0,0 +1,323 @@
...
+
+bindsym Mod4+b nop original
+
+include $key
+
+set \$vartest special title
+for_window [title="\$vartest"] border none
+EOT
+
+is(launch_get_border($config), 'none', 'no border');
```
orestisfl: Wouldn't this test work best if we checked that both cases of the `--remove` keybinding work as intended (in a different file -- we already have binding test code)? I.e., a. remove and test that the keybinding does nothing b. remove and replace and test that the new command is executed.

orestisfl (changes requested):

testcases/t/313-include.t
stapelberg: Good point, done.

src/config_directives.c
stapelberg: Done.

src/bindings.c
stapelberg: Done.

src/bindings.c
stapelberg: Thanks for the suggestion.

I think I prefer the current version because of its early return pattern, which keeps indentation and code path complexity at a minimum.

src/bindings.c
stapelberg: I think it’s cleaner to express all the different cases like this.

Airblader:
> For multi-file configs, this is a standard feature.

Even if we agree on that, we're not introducing a standard solution here but rather a different thing for every directive. For most you just overwrite the value, for bindings there's a special flag now. For bar blocks or for_window it isn't even clear what the syntax would look like.

All of this amounts to trying to special case something that probably doesn't warrant that complexity. For normal users, they can take care of this problem by more carefully designing their files. My original thought were packagers like Regolith, for which this would be useful, but for such a small use case I wouldn't want to add this level of complexity to the config.

Furthermore, we shouldn't treat the default config as a sacred thing that cannot be touched. Configuration exists to be changed. Users who want to remove a default binding can do so already.

> Amending the config at runtime is a separate feature, I think.

So is "remove a previously defined binding within the config". :-)

I really find it odd that we support CRUD in the config but not at runtime. As far as "if it's a popular user request" goes — the expectation that all config directives can be executed as commands at runtime is probably one of the top issues I see people run into. And wanting to change the runtime makes much more sense to me as well as it opens many scripting use cases. 

This flag, however, is just a bandaid to serve a single, and very small, use case, and doesn't even provide a full solution for other directives.

> Yes, but so are many other directives that already effectively have overwrite semantics

Overwriting requires no special syntax, though. If anything, the consistent way to handle this would then be to remove bindings if they're empty:

```
# unbind
bindsym $mod+r
```

But that'd still not provide a solution for things like for_window which by nature must be additive.

Overall I feel adding this flag (at this stage) is a mistake, because it is inconsistent, incomplete and its usecase doesn't justify added syntax, so as far as my vote goes I'd vote no.

stapelberg:
I don’t think it makes sense to introduce an `include` directive without a way to overwrite bindings. In my mind, bindings are the primary thing that most i3 users configure, and not a small niche use-case.

You don’t like the `--remove` flag and that’s fair. But do you have a suggestion for how to implement overwriting bindings instead?

If we don’t want additional syntax, we must remove the duplicate bindings check (because those then overwrite and aren’t duplicative).

Or are you saying you vote against the `include` directive altogether?

Airblader:
> Or are you saying you vote against the include directive altogether?

No, the directive is extremely useful and a big plus for users. I only vote against the flag to remove bindings.

> I don’t think it makes sense to introduce an include directive without a way to overwrite bindings.

I think most users will not be using this flag. They modularize their config into multiple files and load certain ones based on hostname etc. I don't think having to overwrite a previous binding is something most users will (have to) do, instead they just won't define the binding in that environment to begin with.

The only two scenarios where a flag to delete a binding is useful in my mind is when trying to keep and extend the default config (which just isn't necessary) or for packagers like Regolith (which aren't a strong enough of a use case to justify this as they can just ship a modified default config).

kgilmer:
Having had a chance to think about the ramifications of this change a bit more, perhaps the ideal for Regolith (and maybe others) would be to have no default config file at all.  Rather, in the [Xsession init](https://github.com/regolith-linux/regolith-gnome-flashback/blob/master/usr/bin/regolith-session-init#L54) have something like:

(assuming i3 has a capability to [read config from `stdin`](https://faq.i3wm.org/question/2566/is-there-a-way-to-read-the-i3-configuration-from-stdin/index.html)):
```bash
echo "include /etc/i3/conf.d/*; include /etc/regolith/i3/conf.d/*; include ~/regolith/i3/conf.d/*" | i3 -
```

or if config directory feature was exposed via commandline arguments, where `-D` represents one or more include statements:
```bash
i3 -D /etc/i3/conf.d/* -D /etc/regolith/i3/conf.d/* -D ~/regolith/i3/conf.d/*
```

My reasoning is to why this would be better than having an equivalent root config file w/ is:
1. A common hurdle for new users is knowing they have to restart i3 if they change the location of the config file.  Despite documentation this is probably one of the top 3 issues I've noticed Regolith users have.  (eg They copy the i3 into the user dir, make a change, refresh and don't see any difference.)
2. If there is no root file there is less of a change that an accidental change in the root breaks everything.  I think this is a weak reason but also the simplicity of "everything is a config partial".

(Please LMK if this not the right forum for me to discuss ideas like this, happy to move it somewhere else or keep to myself without any offense taken.)

stapelberg:
> I only vote against the flag to remove bindings.

I reverted that part of the pull request for now and added a section to the userguide about this restriction.

I’ll ask people over on the issue what they think, i.e. whether that works for their desired use-cases or not.

> My reasoning is to why this would be better than having an equivalent root config file w/ is:

Reading the config from stdin isn’t great ergonomics, I think, independent of whether it actually works or not (it does not work currently).

Having the config visible in a file is simpler to grasp.

> 1. A common hurdle for new users is knowing they have to restart i3 if they change the location of the config file.  Despite documentation this is probably one of the top 3 issues I've noticed Regolith users have.  (eg They copy the i3 into the user dir, make a change, refresh and don't see any difference.)

Unless overwrite the config path using the `-c` flag when starting i3, usually i3 should be discovering the config path from scratch on every `reload`. If that’s not what you’re seeing, please file a separate issue about that.

orestisfl (approved):

orestisfl (changes requested):
`i3-msg -t get_config` is broken since `char *current_config` in `config_parser.c:893` only keeps the latest read file.

src/config_parser.c
```diff
@@ -877,34 +837,61 @@ static char *get_resource(char *name) {
...
 
-    if ((fd = open(f, O_RDONLY)) == -1)
-        die("Could not open configuration file: %s\n", strerror(errno));
+    char *old_dir = get_current_dir_name();
+    char *dir = NULL;
+    /* dirname(3) might modify the buffer, so make a copy: */
+    char *dirbuf = sstrdup(f);
+    if ((dir = dirname(dirbuf)) != NULL) {
+        LOG("Changing working directory to config file directory %s\n", dir);
+        chdir(dir);
```
orestisfl: Here and in other `chdir`s, I get a warning: 

```
[78/139] Compiling C object i3.p/src_config_parser.c.o
../i3/src/config_parser.c: In function ‘parse_file’:
../i3/src/config_parser.c:873:9: warning: ignoring return value of ‘chdir’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
  873 |         chdir(dir);
      |         ^~~~~~~~~~
../i3/src/config_parser.c:1107:5: warning: ignoring return value of ‘chdir’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
 1107 |     chdir(old_dir);
```

orestisfl (changes requested):

src/config_parser.c
stapelberg: Fixed, thanks.

stapelberg:
> `i3-msg -t get_config` is broken since `char *current_config` in `config_parser.c:893` only keeps the latest read file.

Fixed, thanks!

stapelberg:
Looks like we’re not going to get much more feedback here, so I suggest we merge what we have now and rely on users reporting issues before the release.

@orestisfl anything left to change or can you approve please?

include/bindings.h
```diff
@@ -19,6 +19,12 @@ extern pid_t command_error_nagbar_pid;
  */
 extern const char *DEFAULT_BINDING_MODE;
 
+/**
+ * Returns true if a is a key binding for the same key as b.
+ *
+ */
+bool binding_same_key(Binding *a, Binding *b);
```
orestisfl: Does this need to be exported? It seems to be used only in `bindings.c`.

orestisfl (changes requested):
No strong opinion for me but what do you think about `GET_CONFIG` returning the full config, after includes instead of just the top level config?

src/config_parser.c
```diff
@@ -877,36 +837,64 @@ static char *get_resource(char *name) {
...
 
-    if ((fd = open(f, O_RDONLY)) == -1)
-        die("Could not open configuration file: %s\n", strerror(errno));
+    char *old_dir = get_current_dir_name();
+    char *dir = NULL;
+    /* dirname(3) might modify the buffer, so make a copy: */
+    char *dirbuf = sstrdup(f);
+    if ((dir = dirname(dirbuf)) != NULL) {
+        LOG("Changing working directory to config file directory %s\n", dir);
+        (void)chdir(dir);
```
orestisfl: Sorry, it is doing it again:

```
../i3/src/config_parser.c: In function ‘parse_file’:
../i3/src/config_parser.c:873:15: warning: ignoring return value of ‘chdir’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
  873 |         (void)chdir(dir);
      |               ^~~~~~~~~~
../i3/src/config_parser.c:1108:11: warning: ignoring return value of ‘chdir’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
 1108 |     (void)chdir(old_dir);
      |           ^~~~~~~~~~~~~~
```

Maybe since we are assuming the call succeeds, we should accompany it with an `assert`?

orestisfl (changes requested):

include/bindings.h
stapelberg: Done.

src/config_parser.c
stapelberg: Changed it to report an error when the call fails.

stapelberg:
> No strong opinion for me but what do you think about `GET_CONFIG` returning the full config, after includes instead of just the top level config?

Yeah, improving what `GET_CONFIG` returns is something I thought about, but I’d like to tackle that in a separate pull request. This PR gives us an implementation to use, the next PR can change the implementation to make possible better `GET_CONFIG` output without changing the behavior otherwise.

src/config_parser.c
```diff
@@ -877,36 +837,66 @@ static char *get_resource(char *name) {
...
-    if ((fd = open(f, O_RDONLY)) == -1)
-        die("Could not open configuration file: %s\n", strerror(errno));
+    char *old_dir = get_current_dir_name();
+    char *dir = NULL;
+    /* dirname(3) might modify the buffer, so make a copy: */
+    char *dirbuf = sstrdup(f);
+    if ((dir = dirname(dirbuf)) != NULL) {
+        LOG("Changing working directory to config file directory %s\n", dir);
+        if (chdir(dir) == -1) {
+            ELOG("chdir(%s) failed: %s\n", dir, strerror(errno));
```
orestisfl, line 874: Shouldn't we `return PARSE_FILE_FAILED;` here?

src/config_parser.c
```diff
@@ -1090,40 +1083,38 @@ bool parse_file(const char *f, bool use_nagbar) {
...
     free(buf);
 
-    while (!SLIST_EMPTY(&variables)) {
-        current = SLIST_FIRST(&variables);
-        FREE(current->key);
-        FREE(current->value);
-        SLIST_REMOVE_HEAD(&variables, variables);
-        FREE(current);
+    if (chdir(old_dir) == -1) {
+        ELOG("chdir(%s) failed: %s\n", old_dir, strerror(errno));
```
orestisfl, line 1112: I guess continuing here without a hard error breaks the expectations of other parts of the code but on the other hand we don't want i3 to exit with an error during a reload/restart.

This shouldn't really happen anyway.

orestisfl (changes requested):

src/config_parser.c
stapelberg: No, I don’t think this error is grave enough to warrant aborting.

src/config_parser.c
stapelberg: Yep. See above.

stapelberg:
@orestisfl can you approve so we can merge? We can always improve further later. Thanks :)

src/config_parser.c
orestisfl: It can in theory mean that i3 reads configuration from the wrong files if for some reason this happens but the `open` call does not fail (and since `chdir` failing should always lead to `open` failing, we are practically only aborting in cases where the `open` would lead to aborting as well).

Since the configuration parsing is one place where i3 can execute arbitrary code, I would prefer if we were overly cautious about it.

stapelberg: Done.

orestisfl (approved):
Thanks!

PR #4427, opened 2021-05-19
Add a new IPC message for requesting structured configuration

j-jzk:
Adds a new IPC message for requesting the configuration in JSON, as discussed in #4293.

I wanted to generate the JSON somewhat automatically, but didn't find any acceptable way of doing it - if we generated the JSON at the same time we parsed the config file, the result wouldn't be very structured and would have duplicit values if some directives were set and then overridden.

TODO:
- [ ] parts of the configuration:
    - [ ] keybindings
    - [ ] workspaces
    - [ ] bar
- [ ] tests
- [ ] documentation


orestisfl:
Thanks!

Let me take a look first before putting more time into this.

j-jzk:
@orestisfl What's up? :)

PR #4379, opened 2021-03-11
draw_util_surface_free(): Remove unnecessary argument

psychon:
Cairo allows querying the underlying xcb_connection_t* of a cairo
surface and thus this function does not really need the caller to pass
this in.

No idea if this change is really worth it. I am mostly doing this
because I can / because it seemed sensible (this means the caller cannot
pass different xcb_connection_t* to init and free, but that's likely a
very low risk).

Signed-off-by: Uli Schlachter <psychon@znc.in>

PR #4433, opened 2021-05-31
Check if cairo can work with the X11 server

psychon:
This commits adds code to create a cairo surface for the root window
during startup. If this check fails, i3 fails with an error message,
because it is very likely that all rendering will fail later on.

I only say "very likely", because for later drawing a visual is selected
by get_visualtype_by_id(get_visualid_by_depth(win_depth)), i.e.
depending on the depth of actual windows. This can still end up picking
a visual that cairo cannot work with. However, if cairo cannot even work
with the root window's visual, it is very likely that everything else
will also fail. Thus, this check should at least help somewhat.

----

This is a patch that I came up with while figuring out what went wrong in #4379. I fully expect this new check to make CI fail, because of the 256-colors-X11-server that is used in CI. I do not really know what to do about this, so this is kind of a bug report.(oh, hey, this means I accidentally bypassed @i3bot. That was not my intention.) Anyway, I still believe this to be more or less a good idea, but this new check also should not really ever fail. Hm, so this fixes a bug that no one ever runs into...?

Edit: In case anyone wonders, locally the test suite fails with the following error message, which is no completely helpful, but that's cairo's fault:
```
ERROR: Cairo does not support this X11 setup!
ERROR: Trying to draw to the root window resulted in error 16: invalid value for an input cairo_format_t.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ![Logo](docs/logo-30.png) i3: A tiling window manager
 =====================================================
 
-[![Build Status](https://travis-ci.org/i3/i3.svg?branch=next)](https://travis-ci.org/i3/i3)
+[![Build Status](https://github.com/i3/i3/actions/workflows/main.yml/badge.svg)](https://github.com/i3/i3/actions/workflows/main.yml)
 [![Issue Stats](https://img.shields.io/github/issues/i3/i3.svg)](https://github.com/i3/i3/issues)
 [![Pull Request Stats](https://img.shields.io/github/issues-pr/i3/i3.svg)](https://github.com/i3/i3/pulls)
 


stapelberg:
Erroring out early and clearly rather than later is a good idea in general :)

Can you change the testsuite to keep working in this same pull request please?

Also, can you check if the various VNC servers still work with this change? Those are generally the weirdest X11 servers that people tend to use in practice.

psychon:
> Can you change the testsuite to keep working in this same pull request please?

Done, but GitHub seems to be having problems. I needed four attempts before the git push was accepted and it is not showing up on the web page.

> Also, can you check if the various VNC servers still work with this change?

Do you have a specific list in mind? I found `gnome-remote-desktop`, but that seems to be gnome-specific. `x11vnc` uses an already existing X session and "just" exports that over VNC.

`tightvncserver` seems to be the only "interesting" one I found (version 1:1.3.10-3). I just ran `tightvncserver :1` and got a weird X11 server. Excerpt from `xdpyinfo` (only one visual??):
```
vendor string:    AT&T Laboratories Cambridge
[...]
screen #0:
  dimensions:    1024x768 pixels (347x260 millimeters)
  resolution:    75x75 dots per inch
  depths (1):    24
  root window id:    0x25
  depth of root window:    24 planes
[...]
  number of visuals:    1
  default visual id:  0x22
  visual:
    visual id:    0x22
    class:    TrueColor
    depth:    24 planes
    available colormap entries:    256 per subfield
    red, green, blue masks:    0xff0000, 0xff00, 0xff
    significant bits in color specification:    8 bits
```
Anyway, running `i3` against this server results in:
```
02.06.2021 10:36:08 - ERROR: Another window manager seems to be running (X error 10)
```
Since this check is after the "cairo check", I guess that means that cairo has no problems with this X11 server.

PR #4436, opened 2021-06-02
Check cairo status in draw_util_surface_free()

psychon:
Since my #4379 is not going to work, I thought that one could at least improve the logging a bit, hopefully making debugging easier in the future.

While looking at this code, I found a minor memory leak that I am also fixing in this PR.

### DIFF
--- a/libi3/draw_util.c
+++ b/libi3/draw_util.c
@@ -58,6 +58,19 @@ void draw_util_surface_init(xcb_connection_t *conn, surface_t *surface, xcb_draw
  *
  */
 void draw_util_surface_free(xcb_connection_t *conn, surface_t *surface) {
+    cairo_status_t status = CAIRO_STATUS_SUCCESS;
+    if (surface->cr) {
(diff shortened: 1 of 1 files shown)


psychon:
Sigh, and something in the tests hangs again... I'll investigate.

orestisfl (changes requested):
Was the hang caused by the segfault situation you mention in the commit? In that case I'd squash the faulty commit (also the clang-format one).

psychon:
> Was the hang caused by the segfault situation you mention in the commit? 

Uhm. My guess would be "yes", but it has been a quarter year, so I don't remember for sure. At least I vaguely remember being annoyed that I had some problem with crashes and the test suite and that it was quite hard to get `gdb` in the game. But I don't remember any details. 

>  In that case I'd squash the faulty commit

Done. This PR now only has two commits.

orestisfl (approved):
Thanks!

PR #4448, opened 2021-06-17
Made state consistent with window in remanage_window

### LINKED ISSUE #4075: Window rectangle isn't updated when swallowed.
> I get some very strange bugs when restoring multiple windows in a 4x4 grid by appending a layout and then opening terminal windows whose titles adjust are changed after initial window creation. When I use the old workaround of unmapping all the windows in the workspace, appending the layout, then remapping the windows, it works fine as it always has.
> 
> How it should look (and how it does look when I restore it in the old way):
> img-2020-04-25-1750
> 
> How it looks when the windows are swallowed after a title change:
> img-2020-04-25-1751
> 
> These are not mere graphical artifiacts, because if I keep pressing enter you can see that the prompt never goes below the top of the "artifact", so the window must actually be ending there, and the tiling has completely broken.
> 
> Also thanks @izzel for your work on this, it's a great improvement to have. In most situations it has worked perfectly for me, and I'm not sure if this weird bug is directly related to this change or if this change is bringing to light a bug somewhere else in the codebase.

_Originally posted by @JonnyHaystack in https://github.com/i3/i3/issues/1668#issuecomment-619390792_

I misunderstood the variable needs_reshape in x_reparent_window. x_reparent_window fairly assumes that the window rect won't change because it was designed with floating stickies in mind. Copying the window_rect state in x_reparent_window fixes the problem, but I worry that other data needs to be copied.
@stapelberg Would child_mapped need to be copied as well? Is there anything else I'm missing in con_state that would need to be copied?

izzel:
Fixes #4075.

i3 believes everything is fine because the con_state window_rect says it is the right size. This commit updates window_rect to the actual state as well as child_mapped.

Does this need a test? I have no idea how to make a test for a bug like this.

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -251,6 +251,34 @@ void x_move_win(Con *src, Con *dest) {
     }
 }
 
+/*
+ * Reparents child window from old con into new con. The reparenting happens in
+ * the next call of x_push_changes(). Unlike x_reparent_child & x_move_win, this
+ * function assumes that the old con is about to be destroyed and preserves
+ * window_rect and child_mapped.
+ *
+ */
+void x_reparent_child2(Con *new, Con *old) {
+    struct con_state *new_state, *old_state;
+
+    if ((new_state = state_for_frame(new->frame.id)) == NULL) {
+        ELOG("window state for new con not found\n");
+        return;
+    }
+
+    if ((old_state = state_for_frame(old->frame.id)) == NULL) {
+        ELOG("window state for old con not found\n");
+        return;
+    }
+
+    new_state->need_reparent = true;
+    new_state->old_frame = old->frame.id;
+
+    new_state->child_mapped = old_state->child_mapped;
+
+    memcpy(&(new_state->window_rect), &(old_state->window_rect), sizeof(Rect));
+}
+
 static void _x_con_kill(Con *con) {
     con_state *state;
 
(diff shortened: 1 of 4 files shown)


src/x.c
```diff
@@ -251,6 +251,34 @@ void x_move_win(Con *src, Con *dest) {
...
 }
 
+/*
+ * Reparents child window from old con into new con. The reparenting happens in
+ * the next call of x_push_changes(). Unlike x_reparent_child & x_move_win, this
+ * function assumes that the old con is about to be destroyed and preserves
+ * window_rect and child_mapped.
+ *
+ */
+void x_reparent_child2(Con *new, Con *old) {
```
orestisfl: Maybe a name like this would be a bit better?

```suggestion
void x_reparent_child_deep(Con *new, Con *old) {
```

orestisfl (changes requested):
LGTM overall

src/x.c
izzel: Agreed

orestisfl (changes requested):
Thanks! One last thing is to add a short line describing the change in the [release notes](https://github.com/i3/i3/blob/next/RELEASE-NOTES-next)

orestisfl (approved):
Thanks!

orestisfl:
@stapelberg might be worth a look before the next release?

stapelberg:
> @stapelberg might be worth a look before the next release?

This doesn’t look trivial enough to be safe to include in the next release. I’ll rather merge this after the release so that it has time to soak (looks like a subtle change).

PR #4451, opened 2021-06-28
When moving to next or prev workspace, don't skip colliding number

### LINKED ISSUE #4452: No skipping number-colliding workspaces with `workspace next`/`prev`
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

`workspace next` and `prev` skip workspaces that have the same number as current.

For example, having workspace 1, 2:foo, 2:bar, and 3, with 2:foo focused and doing `workspace next` takes us to workspace 3, skipping 2:bar. Being on workspace 2:bar and doing `workspace prev` takes us to 1, skipping 2:foo.

## Desired Behavior

To not skip them.

## Impact

When typically using more than 10 workspaces, this helps to organize them by using the numbers to categorize workspaces and group them. The digit keys can be used to quickly get to an area of workspaces and then select the correct workspace with `workspace next`/`prev` keybindings. This is similar to how one would quickly navigate a video on YouTube.

Skipping the number-colliding workspaces doesn't seem useful, so this might actually be a bug report.

<pre>
</pre>

## Environment

<pre>
i3 version:
Binary i3 version:  4.19.2 (2021-02-27) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.19.2 (2021-02-27) (pid 1413)
Loaded i3 config: /home/jol/.config/i3/config (Last modified: Mon 28 Jun 2021 02:00:47 AM PDT, 922 seconds ago)

</pre>

<pre>
- Linux Distribution & Version: Archlinux last updated a couple of hours ago.
</pre>

jolmg:
This fixes an issue where, having workspaces: 1, 2:foo, 2:bar, and 3,
moving from 2:foo to next moves us to workspace 3, and moving from 2:bar
to prev moves us to workspace 1.

With this change, moving from 2:foo to next would move us to 2:bar, and
from 2:bar to prev would move us to 2:foo.

Closes #4452

### DIFF
--- a/src/workspace.c
+++ b/src/workspace.c
@@ -702,6 +702,7 @@ Con *workspace_next_on_output(void) {
     Con *current = con_get_workspace(focused);
     Con *next = NULL;
     Con *output = con_get_output(focused);
+    bool found_current = false;
 
     if (current->num == -1) {
         /* If currently a named workspace, find next named workspace. */
@@ -713,6 +714,12 @@ Con *workspace_next_on_output(void) {
                 continue;
             if (child->num == -1)
                 break;
+            if (child == current) {
+                found_current = true;
+            } else if (child->num == current->num && found_current) {
+                next = child;
+                goto workspace_next_on_output_end;
+            }
             /* Need to check child against current and next because we are
              * traversing multiple lists and thus are not guaranteed the
              * relative order between the list of workspaces. */
@@ -723,7 +730,6 @@ Con *workspace_next_on_output(void) {
 
     /* Find next named workspace. */
     if (!next) {
-        bool found_current = false;
         NODES_FOREACH (output_get_content(output)) {
             if (child->type != CT_WORKSPACE)
                 continue;
@@ -757,6 +763,7 @@ Con *workspace_prev_on_output(void) {
     Con *current = con_get_workspace(focused);
     Con *prev = NULL;
     Con *output = con_get_output(focused);
+    bool found_current = false;
     DLOG("output = %s\n", output->name);
 
     if (current->num == -1) {
@@ -769,6 +776,12 @@ Con *workspace_prev_on_output(void) {
         NODES_FOREACH_REVERSE (output_get_content(output)) {
             if (child->type != CT_WORKSPACE || child->num == -1)
                 continue;
+            if (child == current) {
+                found_current = true;
+            } else if (child->num == current->num && found_current) {
+                prev = child;
+                goto workspace_prev_on_output_end;
+            }
             /* Need to check child against current and previous because we
              * are traversing multiple lists and thus are not guaranteed
              * the relative order between the list of workspaces. */
@@ -779,7 +792,6 @@ Con *workspace_prev_on_output(void) {
 
     /* Find previous named workspace. */
     if (!prev) {
-        bool found_current = false;
         NODES_FOREACH_REVERSE (output_get_content(output)) {
             if (child->type != CT_WORKSPACE)
                 continue;
--- a/src/workspace.c
+++ b/src/workspace.c
@@ -581,8 +581,7 @@ Con *workspace_next(void) {
                 if (child == current) {
                     found_current = true;
                 } else if (child->num == -1 && found_current) {
-                    next = child;
-                    return next;
+                    return child;
                 }
             }
         }
@@ -604,8 +603,7 @@ Con *workspace_next(void) {
                 if (child == current) {
                     found_current = true;
                 } else if (child->num == current->num && found_current) {
-                    next = child;
-                    return next;
+                    return child;
                 }
                 /* Need to check child against current and next because we are
                  * traversing multiple lists and thus are not guaranteed the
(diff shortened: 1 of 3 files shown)


jolmg:
Closes #4452

testcases/t/544-next-prev-dont-skip-number-collisions.t
```diff
@@ -0,0 +1,38 @@
...
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
+# • https://build.i3wm.org/docs/ipc.html
+#   (or docs/ipc)
+#
+# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+#   (unless you are already familiar with Perl)
+#
+# Tests that switching to next/prev workspace doesn't skip those that have
+# the same number as current.
```
orestisfl, line 18:
```suggestion
# the same number as current.
# See: #4452
```

jolmg:
I just noticed that besides `workspace_{next,prev}()`, there are also `workspace_{next,prev}_on_output()`. Those would also need the same change.

jolmg:
Change has been extended to `{next,prev}_on_output`.

src/workspace.c
```diff
@@ -701,6 +714,12 @@ Con *workspace_next_on_output(void) {
                 continue;
             if (child->num == -1)
                 break;
+            if (child == current) {
+                found_current = true;
+            } else if (child->num == current->num && found_current) {
+                next = child;
+                goto workspace_next_on_output_end;
```
orestisfl:
```suggestion
                return child;
```

src/workspace.c
```diff
@@ -667,6 +673,12 @@ Con *workspace_prev(void) {
                     first_opposite = child;
                 if (child->num == -1)
                     continue;
+                if (child == current) {
+                    found_current = true;
+                } else if (child->num == current->num && found_current) {
+                    prev = child;
+                    return prev;
```
orestisfl:
```suggestion
                    return child;
```

src/workspace.c
```diff
@@ -601,6 +601,12 @@ Con *workspace_next(void) {
                     first_opposite = child;
                 if (child->num == -1)
                     break;
+                if (child == current) {
+                    found_current = true;
+                } else if (child->num == current->num && found_current) {
+                    next = child;
+                    return next;
```
orestisfl:
```suggestion
                    return child;
```

src/workspace.c
```diff
@@ -757,6 +776,12 @@ Con *workspace_prev_on_output(void) {
         NODES_FOREACH_REVERSE (output_get_content(output)) {
             if (child->type != CT_WORKSPACE || child->num == -1)
                 continue;
+            if (child == current) {
+                found_current = true;
+            } else if (child->num == current->num && found_current) {
+                prev = child;
+                goto workspace_prev_on_output_end;
```
orestisfl:
```suggestion
                return child;
```

orestisfl (changes requested):
Thanks!

Made some nit picks which I think mildly simplify things, otherwise LGTM

orestisfl:
Also, the release notes need to be updated: https://github.com/i3/i3/blob/next/RELEASE-NOTES-next

src/workspace.c
jolmg: This was to be consistent with the style used 23 lines above. Should I change that as well?

src/workspace.c
jolmg: Ditto, 24 lines above.

src/workspace.c
jolmg: Ditto, 19 lines below. This function follows a single-return style. It seems the previous ones were also originally written in such a style.

Personally, I find the assignment clearer. It allows one to visually scan the function source for `next =` in order to determine where `next` is determined. With this change, one would have to pay a bit more attention that `return child` also determines what the `next` is. It breaks the overall consistency.

src/workspace.c
jolmg: Ditto, 20 lines below.

src/workspace.c
Airblader: Yes, please change them. Early returns are preferable.

jolmg:
I've updated the release notes.

src/workspace.c
jolmg: Done. 12bd3f8063cc15c3cd4db5f738f81e5dc3263f82

src/workspace.c
jolmg: Done. 12bd3f8063cc15c3cd4db5f738f81e5dc3263f82

src/workspace.c
jolmg: Done. 12bd3f8063cc15c3cd4db5f738f81e5dc3263f82

src/workspace.c
jolmg: Done. 12bd3f8063cc15c3cd4db5f738f81e5dc3263f82

PR #4453, opened 2021-06-30
Implement window icon positioning configuration

iscgar:
Enhance the work that landed in #4439 by allowing customisation of the icon position.

This feature defaults to following the title alignment, and can be tuned for individual windows to be on the left (which is the behaviour prior to this commit) or on the right.

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -36,7 +36,7 @@ option is enabled and only then sets a screenshot as background.
   • i3-dump-log -f now uses UNIX sockets instead of pthreads. The UNIX socket approach
     should be more reliable and also more portable.
   • Implement the include config directive
-  • Implement optionally showing window icons in titlebar
+  • Implement optionally showing window icons in titlebar and setting their position
   • Allow for_window to match against WM_CLIENT_MACHINE
   • Add %machine placeholder (WM_CLIENT_MACHINE) to title_format
   • Allow multiple output names in 'move container|workspace to output'
--- a/docs/userguide
+++ b/docs/userguide
@@ -2715,6 +2715,7 @@ specific windows or for all windows (using the <<for_window>> directive).
 -----------------------------
 title_window_icon <yes|no>
 title_window_icon padding <px>
+title_window_icon position <left|right|title>
 ------------------------------
 
 *Examples*:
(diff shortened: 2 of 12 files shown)


omertuc:
Some background for this PR -
#4439 was based on a patch that was posted to AUR.

When using the `title_align center` i3 configuration, the original behavior of the patch #4439 was based on looked like this in stacked layout:
image

But a recent revision of the patch "broke" the original behavior and started doing this:
image

Looks like #4439 copied the broken behavior from the latest revision of the patch, when porting it to the official i3 repo.

This PR (#4453) makes it configurable so the user can choose either behavior using configurations such as:
```
title_align center
for_window [class=".*"] title_window_icon yes
for_window [class=".*"] title_window_icon position title
```
image


```
title_align center
for_window [class=".*"] title_window_icon yes
for_window [class=".*"] title_window_icon position left
```
image


```
title_align left
for_window [class=".*"] title_window_icon yes
for_window [class=".*"] title_window_icon position left
```
image



```
title_align left
for_window [class=".*"] title_window_icon yes
for_window [class=".*"] title_window_icon position right
```
image

Airblader:
Thanks for the PR! I'm not yet convinced that this change is useful enough to warrant more configuration, but let's keep it here and revisit it in a while if we get more feedback.

omertuc:
> Thanks for the PR! I'm not yet convinced that this change is useful enough to warrant more configuration, but let's keep it here and revisit it in a while if we get more feedback.

Since `for_window [class=".*"] title_window_icon position title` was the original behavior of the patch that #4439 tried to introduce into this repo, and `for_window [class=".*"] title_window_icon position left` is actually buggy behavior that just happened to be in the latest revision of that patch (this buggy behavior was accidentally adopted by #4439), do you at-least agree we should fix the icons to behave like `for_window [class=".*"] title_window_icon position title`, even if it's not configurable?

It's a "breaking" change, but since icons are so new, I don't think too many people would mind this change in behavior.

I think `for_window [class=".*"] title_window_icon position title` is more "correct" - since the main purpose of icons, as I see it, is to be able to easily tell apart windows with a glimpse. When using centered text, and stacked mode, on a big screen, the icons drift all the way to the left, far away from the titles, making it hard to look at both the titles and the icons at the same time, thus making it harder to tell apart windows with the same icon (multiple chrome windows, for example)

Airblader:
Yeah, that makes sense to me and sounds reasonable. Icons haven't yet been released anyway.

orestisfl:
Thanks for bringing this to our attention. My understanding is that we should close this PR now and fix the buggy behavior before the next release. I opened #4464 for this purpose.

PR #4460, opened 2021-07-23
Implement 'all' matching criterion

### LINKED ISSUE #4440: Add match all window criterion
As discussed in https://github.com/i3/i3/pull/4439#issuecomment-853741405

This will allow users to write `for_window [all] command`

## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

<pre>
i3 version: i3 version 4.19.2-101-geaa5e636+ © 2009 Michael Stapelberg and contributors
</pre>

iscgar:
This criterion matches all open windows, as a more readable version of for the `class=".*"` criterion. However, if both `all` and `class` are specified, `class` takes over (regardless of the order in which they are specified) in order to avoid confusion, since the `class` criterion is likely more restrictive.

Fixes #4440

### DIFF
--- a/src/match.c
+++ b/src/match.c
@@ -53,7 +53,8 @@ bool match_is_empty(Match *match) {
             match->window_type == UINT32_MAX &&
             match->con_id == NULL &&
             match->dock == M_NODOCK &&
-            match->window_mode == WM_ANY);
+            match->window_mode == WM_ANY &&
+            match->match_all_windows == false);
 }
 
 /*
@@ -260,6 +261,10 @@ bool match_matches_window(Match *match, i3Window *window) {
         LOG("window_mode matches\n");
     }
 
+    /* NOTE: See the comment regarding 'all' in match_parse_property()
+     * for an explanation of why match_all_windows isn't explicitly
+     * checked. */
+
     return true;
 }
 
@@ -435,6 +440,17 @@ void match_parse_property(Match *match, const char *ctype, const char *cvalue) {
         cvalue != NULL &&
         strcmp(cvalue, "user") == 0) {
         match->window_mode = WM_FLOATING_USER;
+        return;
+    }
+
+    /* match_matches_window() only checks negatively, so match_all_windows
+     * won't actually be used there, but that's OK because if no negative
+     * match is found (e.g. because of a more restrictive criterion) the
+     * return value of match_matches_window() is true.
+     * Setting it here only serves to cause match_is_empty() to return false,
+     * otherwise empty criteria rules apply, and that's not what we want. */
+    if (strcmp(ctype, "all") == 0) {
+        match->match_all_windows = true;
         return;
     }
 
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -41,6 +41,7 @@ option is enabled and only then sets a screenshot as background.
   • Add %machine placeholder (WM_CLIENT_MACHINE) to title_format
   • Allow multiple output names in 'move container|workspace to output'
   • Add 'move container|workspace to output next'
+  • Add 'all' window matching criterion
 
  ┌────────────────────────────┐
  │ Bugfixes                   │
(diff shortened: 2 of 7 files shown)


src/match.c
```diff
@@ -438,5 +438,14 @@ void match_parse_property(Match *match, const char *ctype, const char *cvalue) {
         return;
     }
 
+    /* 'all' is simply a shorthand for class=".*", but it might be surprising
+     * to users if we override a more restrictive class spec as this is an
+     * implementation detail that they aren't aware of, so only set the class
+     * regex if one isn't already set */
+    if (strcmp(ctype, "all") == 0 && match->class == NULL) {
```
Airblader: I'd prefer if we didn't abuse the class criterion for this but rather just make it an explicit flag we check for. IIRC there's also clients which don't set WM_CLASS (in time), and in that case even `.*` actually wouldn't match it, which can be reasoned with for `.*`, but not for `all`.

Airblader (changes requested):
Thanks for the PR! Can you please make sure to add some tests for this as well as update RELEASE-NOTES-next to mention this feature?

src/match.c
iscgar: I thought about this route when I started implementing this, but there's a potential cognitive problem with an explicit flag: I can add a `match_all_windows` member to `Match` and update `match_is_empty()` to check if it's set to `false`, but because `match_matches_window()` is only doing negative checks to implement logical AND and because `all` takes effect only if it wasn't combined with other criteria (e.g. `[all floating]` is effectively the same as simply `[floating]`), it won't be used in `match_matches_window()`, and it can confuse readers of the code if there's a member that only gets set but seems to be unused.
I didn't know that there's a possibility for `WM_CLASS` to not be set when `match_matches_window()` is called, so I didn't think about that. If you find my concern above irrelevant, I'll implement it as an explicit flag.

Airblader: But `match_matches_window` could just check this flag before anything else and return `true` in that case, couldn't it? I don't think we need to be concerned with combining `all` with any other criteria, because that would never make sense: as soon as you add any other criterion, `all` can be omitted. We could just validate that if `all` was set, nothing else can be set (and return an error).

iscgar: Right, but the same can be said about e.g. `[class=".*" floating]`, where the class specification is redundant (except for the unlikely case where `WM_CLASS` wasn't yet set). If we start treating `all` differently, it would be a deviation from the rule that the criteria is evaluated as a logical AND of all of its elements, and that can be surprising to users who will assume that they can simply replace `class=".*"` with `all`.
I'm fine with treating `all` differently as long as we fail to parse a criteria where `all` is specified with other criteria, but I'm not sure how to change the parser spec to add a state that will either capture `all` or forward to `CRITERIA` without eating the token.

Airblader: Yeah this is why I'm saying we should validate it: if all is specified with any other criteria, we error out. I don't think we need to solve this in the parser necessarily (though that'd be an option if possible).

I'm also OK with your original suggestion to not add anything in match_matches, but only in match_is_empty, especially if the validation turns out to be difficult. I'm not worried about the readability of the code for that either, but if we really want to we can also document that in the method.

iscgar: It's not that difficult to add a validation. It's just that the only way we have to error out in `match.c` is with `ELOG()`, which is invisible to the user, and the parser won't bail because of it, so users will assume that their config/command is OK when in fact i3 isn't doing what they expected it to do.

The only way that I can think of to make the parser bail is by adding a new token `else` that doesn't eat anything and just forwards to the next state, and add a new state between the open square bracket and `CRITERIA` that only accepts `all` or forwards to `CRITERIA`. Would such a change be acceptable?

If that is too complicated, then I think that going with my original suggestion above about not checking in `match_window_matches()` is the best way to add the `all` criterion, and I'll add a comment there to explain why it isn't used.

Airblader: Let's go with your original suggestion. There's no reason to make this more complex than it needs to be. :-)

iscgar:
I made the required changes (and reworded the commit message to not mention the relationship between `all` and `class` that was true of the previous iteration). I also added a test and a line to RELEASE-NOTES-next.

Airblader (approved):
Thanks! I think the changes look good. Let's wait for the CI (apparently we need to approve running that first now).

PR #4469, opened 2021-08-01
handlers.c: send an "output" event on monitor configuration change

vincentbernat:
When adding/removing a monitor, the outputs are likely to be modified.
Send an IPC event "output", like when there is a screen configuration
change.

Signed-off-by: Vincent Bernat <vincent@bernat.ch>

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -62,3 +62,4 @@ option is enabled and only then sets a screenshot as background.
   • ipc: return proper signed int for container positions: negative values were
     returned as large 32 bits integers
   • fix crash with "layout default"
+  • send an "output" event on XRandR 1.5 monitor configuration change
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -3,7 +3,7 @@
  │ Release notes for i3 v4.20   │
  └──────────────────────────────┘
 
-This is i3 v4.19. This version is considered stable. All users of i3 are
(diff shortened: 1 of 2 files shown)


orestisfl (approved):
Seems to work in my case.

The only other "output" event seems to be in `handlers.c:handle_screen_change`.

@Airblader WDYT?

Airblader:
Fine with me, however we need to add this to the release notes, too. @vincentbernat Can you add an entry in `RELEASE-NOTES-next` for this?

vincentbernat:
OK, added!

RELEASE-NOTES-next
```diff
@@ -3,7 +3,7 @@
  │ Release notes for i3 v4.20   │
  └──────────────────────────────┘
 
-This is i3 v4.19. This version is considered stable. All users of i3 are
+This is i3 v4.20. This version is considered stable. All users of i3 are
```
Airblader, line 6: Thanks for catching this mistake. Could you maybe split this into a separate commit? This would be worth keeping even if we somehow had to revert the PR (as unlikely as it might be).

Airblader (changes requested):
Thank you! Also thanks for catching that mistake and fixing it, but could we have that as a separate commit? Would be appreciated!

RELEASE-NOTES-next
vincentbernat: Done.

Airblader (approved):
Thank you!

PR #4475, opened 2021-08-11
Use ink rectangle when calculating Pango text width

mahkoh:
Before: pango-before

After: pango-after

### DIFF
--- a/libi3/font.c
+++ b/libi3/font.c
@@ -127,7 +127,7 @@ static int predict_text_width_pango(const char *text, size_t text_len, bool pang
     PangoLayout *layout = create_layout_with_dpi(cr);
 
     /* Get the font width */
-    gint width;
+    PangoRectangle ink_rect;
     pango_layout_set_font_description(layout, savedFont->specific.pango_desc);
 
     if (pango_markup)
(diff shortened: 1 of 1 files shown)


psychon:
Which of your two changes fixes the problem? Only both of them? Or is e.g. the `x + width` change already enough?

You add a comment linking to a 20 years old bug report. Are you saying this bug was still not fixed after 20 years or what is that linked supposed to mean?

Also, I briefly looked at the code, but I did not find anything that would explain the clipping. Is the right part of the simply overdrawn by later code, or what? If so, what exactly draws over part of the glyphs? I'm confused about what is going on...

mahkoh:
> Which of your two changes fixes the problem? Only both of them? Or is e.g. the `x + width` change already enough?

I use the font `pango:DejaVu Sans Mono 8`. With this font `ink_rect.x` is always 1 whereas `logical_rect.x` is always 0. If I remember correctly, `logical_rect.width` and `ink_rect.width` are always identical.

The fix therefore requires both changes.

>You add a comment linking to a 20 years old bug report. Are you saying this bug was still not fixed after 20 years or what is that linked supposed to mean?

The linked mail is the only source I've found that describes the difference between logical_rect and ink_rect. I do not know if the current behavior of Pango is a bug or not.

>Also, I briefly looked at the code, but I did not find anything that would explain the clipping. Is the right part of the simply overdrawn by later code, or what?

No. In https://github.com/i3/i3/blob/d65a7ed250e723d4fad9850ace88bab9dfe6545d/i3bar/src/xcb.c#L307-L310 the width of the surface that we draw to is explicitly set to the expected width of the text (`render->width`). I tried simply setting this width to the width of the entire statusline_buffer, but then the last piece of text (the time in my screenshot) will still be cut off because its x offset is one pixel too far to the right.

Issue #4022, opened 2020-04-16
compile time analysis


stapelberg:
I recently stumbled upon:

* https://aras-p.info/blog/2019/01/16/time-trace-timeline-flame-chart-profiler-for-Clang/
* https://github.com/aras-p/ClangBuildAnalyzer

…and ran it on i3 because I was curious.

Here’s the output of the build analyzer, which aggregates the JSON profile files:
<details><summary>build_analyzer output</summary><pre>
**** Time summary:
Compilation (81 times):
  Parsing (frontend):           30.3 s
  Codegen & opts (backend):      3.3 s

**** Files that took longest to parse (compiler frontend):
   823 ms: src/i3-commands.o
   770 ms: src/i3-config_directives.o
   760 ms: src/i3-commands_parser.o
   700 ms: src/i3-display_version.o
   687 ms: src/i3-handlers.o
   643 ms: src/i3-floating.o
   612 ms: src/i3-con.o
   561 ms: src/i3-workspace.o
   560 ms: src/i3-xcb.o
   552 ms: src/i3-util.o

**** Files that took longest to codegen (compiler backend):
   245 ms: i3bar/src/i3bar-xcb.o
   226 ms: src/i3-con.o
   191 ms: src/i3-bindings.o
   177 ms: src/i3-commands.o
   128 ms: i3-nagbar/i3_nagbar-main.o
   123 ms: src/i3-x.o
   117 ms: i3-config-wizard/i3_config_wizard-main.o
   100 ms: src/i3-ipc.o
    89 ms: src/i3-manage.o
    84 ms: src/i3-config_parser.o

**** Functions that took longest to compile:
    62 ms: manage_window (../../i3/src/manage.c)
    54 ms: translate_keysyms (../../i3/src/bindings.c)
    53 ms: main (../../i3/i3-nagbar/main.c)
    49 ms: main (../../i3/src/main.c)
    31 ms: con_swap (../../i3/src/con.c)
    30 ms: startup_sequence_rename_workspace (../../i3/src/startup.c)
    24 ms: draw_bars (../../i3/i3bar/src/xcb.c)
    24 ms: main (../../i3/i3-config-wizard/main.c)
    23 ms: bindings_get_buttons_to_grab (../../i3/src/bindings.c)
    22 ms: insert_con_into (../../i3/src/move.c)
    22 ms: switch_mode (../../i3/src/bindings.c)
    21 ms: handle_expose (../../i3/i3-nagbar/main.c)
[...]

PR #4480, opened 2021-08-20
Include what you use

zybreak:
Simply a code refactorization, instead of importing an "all.h" header file in most c-files which contains all project header-files, we now only include those header files needed for the specific c-file in question.

The benefit is a better overview of whats being used and where as well as easier to spot unused header-files


orestisfl:
Thank you for this PR.

Related issue: #4022

I don't think it's worth it to begin practicing "header discipline". Unused header files are not a big problems and seeing where each file gets its definitions from doesn't seem that useful right now.

@stapelberg WDYT?

stapelberg:
Yeah, it looks like the time savings are very small:

Without this PR:
```
ninja  15,73s user 2,51s system 1282% cpu 1,422 total
ninja  15,87s user 2,26s system 1265% cpu 1,433 total
ninja  15,81s user 2,51s system 1291% cpu 1,419 total
```

With this PR:
```
ninja  15,23s user 2,53s system 1263% cpu 1,406 total
ninja  15,28s user 2,53s system 1255% cpu 1,418 total
ninja  15,62s user 2,30s system 1274% cpu 1,406 total
```

Does include-what-you-use (the tool) work without problems on the i3 code base and with little setup? If so, we could consider merging this. If no, it’s probably not worth the hassle.

orestisfl:
I would argue that maintaining the extra dev/CI dependency and having to manually include files are not really worth it.

zybreak:
I haven't looked into whether the include-what-you-use tool will work or not, but I could do that if you think that would be of any benefit.

My hope was that this would bring clarity to the code, as I was struggling a bit to get a good overview with the all.h header.

I hope you can evaluate the suggestion again from a code clarity aspect. If you feel this brings little or no added clarity and just a hassle of writing a lot more imports, I would argue that writing imports is not a big deal. And new developers can get a pretty good idea of whats in use by looking at them.

If you have an editor with good features, it will show you any imports not needed (thus removing the need to run the tool seperately) 

Anyways, enough ranting.

stapelberg:
> I haven't looked into whether the include-what-you-use tool will work or not, but I could do that if you think that would be of any benefit.

I think if we find that the IWYU tool is easy to set up and works well with the i3 code base, that would be the only case in which I could be convinced to accept this PR. 

> I hope you can evaluate the suggestion again from a code clarity aspect. If you feel this brings little or no added clarity and just a hassle of writing a lot more imports, I would argue that writing imports is not a big deal.

I agree that listing headers explicitly is good from a code clarity perspective. But at the same time, having to manually modify headers when moving code around is such a hassle, it overweighs the benefit of clarity.

Airblader:
Having to change includes in i3 is something that occurs very, very rarely these days anyway. 

Except for enums and data types, most functions also follow the convention of being prefixed by their file name, which makes it pretty clear where they're coming from.

orestisfl:
I feel the consensus here is to not merge this. Feel free to reopen otherwise.

PR #4482, opened 2021-08-22
"Contact us" turned into a link to /contact/

Jonta:
Linking to both IRC and mailing list separately seems a bit much to me


Airblader:
Thank you for this, @Jonta. I will close this as it is superseded by #4503, but I appreciate your proposed change here (and have incorporated it).

PR #4483, opened 2021-08-22
Meaningful Alt-texts for images

Jonta:
I haven't rendered this

Beware of quotes within

I guessed that escaping them with backslashes would work


Jonta:
Also: linebreaks in Alt-texts might create issues in your renderer

PR #4485, opened 2021-08-24
Fix bug where long commands crash i3

### LINKED ISSUE #2968: Memory leak with long commands (commands_parser stack full)

Binary i3 version:  4.14-93-ga3a7d04a (2017-09-18, branch "remotes/upstream/next") © 2009 Michael Stapelberg and contributors
Running i3 version: 4.14 (2017-09-04, branch "4.14") (pid 774)
Loaded i3 config: /home/cornerman/.i3/config (Last modified: Mon 18 Sep 2017 06:24:50 PM CEST, 799 seconds ago)

RUNNING BINARY DIFFERENT FROM BINARY ON DISK!

http://logs.i3wm.org/logs/5643059551272960.bz2

**What I did:**

Thanks for fixing https://github.com/i3/i3/issues/2903, I tried it out, but now run into another problem: Long commands with more than 10 words can crash i3.

Reproduction:
```sh
Xephyr :1 &
DISPLAY=:1 i3 &
i3-msg -s $(DISPLAY=:1 i3 --get-socketpath) "swap container container container container container container container container container container"
```

**What I saw:**

Receiving this command, i3 crashes and the leak sanitizer reports memory leaks:
```
BUG: commands_parser stack full. This means either a bug in the code, or a new command which contains more than 10 identified tokens.
=================================================================
==6978==ERROR: LeakSanitizer: detected memory leaks
```

Why did I try this command?

I am maintaining a script to generate a bash completion script for i3 (https://github.com/cornerman/i3-completion). It runs bogus command against the i3 ipc interface and tries to find valid commands from the hints of the command parser ("Expected one of these tokens: ...").

And with the `swap` command the `Expected tokens` are a bit misleading (similar to the `move ... to ...` syntax):
```
i3-msg "swap container foobar"
Expected one of these tokens: 'container', 'with', 'id', 'con_id', 'mark'
```

**What I expected instead:**

It should not crash.

ToxicGLaDOS:
Commands with more than 10 words would result in i3 crashing. The root cause is that the stack which holds arguments with identifiers is only 10 big, but `generate-command-parser.pl` was giving every word an identifier of "" (empty string). This behavior could have been correct, but everywhere this value was referenced compares it to `NULL` instead of empty string. [Here](https://github.com/i3/i3/blob/36ba1043d50d815b65b7f22619611a420726e8e3/src/commands_parser.c#L281) for example.

It could be argued that the command `move window container to window container to window container to left` is invalid and therefore shouldn't be tested for. If so, I can remove the testcase. But crashing on that kind of input should be fixed in any case.

Also, this bug could be fixed in the opposite way, where we change all of the checks against `NULL` to check against `'\0'`. This way was simpler and seemed like what the desired behavior was all along to me.

Another interesting component to this bug is that it doesn't effect the config parser, only the command parser. That's because identifiers in `config_parser.c` are handled a little differently. When we push an identified string on the stack it appends the values together [if they have the same identifier](https://github.com/i3/i3/blob/36ba1043d50d815b65b7f22619611a420726e8e3/src/config_parser.c#L90) which causes all of the words with empty string as an identifier to take up only one spot on the stack.

Fixes #2968

### DIFF
--- a/generate-command-parser.pl
+++ b/generate-command-parser.pl
@@ -228,8 +228,15 @@ for my $state (@keys) {
             ($call_identifier) = ($next_state =~ /^call ([0-9]+)$/);
             $next_state = '__CALL';
         }
-        my $identifier = $token->{identifier};
-        say $tokfh qq|    { "$token_name", "$identifier", $next_state, { $call_identifier } },|;
+        my $identifier;
+        # Set $identifier to NULL if there is no identifier
+        if ($token->{identifier} eq ""){
+            $identifier = "NULL"
+        }
+        else{
+            $identifier = qq|"$token->{identifier}"|;
+        }
+        say $tokfh qq|    { "$token_name", $identifier, $next_state, { $call_identifier } },|;
     }
     say $tokfh '};';
 }
--- a/testcases/t/315-long-commands.t
+++ b/testcases/t/315-long-commands.t
@@ -1,0 +1,46 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • https://build.i3wm.org/docs/testsuite.html
+#   (or docs/testsuite)
+#
+# • https://build.i3wm.org/docs/lib-i3test.html
+#   (alternatively: perldoc ./testcases/lib/i3test.pm)
+#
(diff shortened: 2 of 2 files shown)


Airblader (approved):
Thank you so much for the contribution!

Issue #4494, opened 2021-09-06
Config partial loading of optional user directories


kgilmer:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
A feature in development and slated for the 4.20 release is config partials  Users are able to specify include directories and files to be loaded as part of configuration.  As a desktop component integrator (DCI?) I would like the ability to specify user directories to load from that may not actually exist.  However when adding this, i3 generates an error (please disregard i3 version, this is a personal build from `next`)

![VirtualBox_bullseye-test-r2-2_05_09_2021_21_10_48](https://user-images.githubusercontent.com/49683/132159748-2e9eb88d-2c4a-4fd3-a4f6-e37dfccd1d6e.png)

## Desired Behavior

I would like the ability to tell i3 to not generate a user-facing error message for situations involving missing config directory includes. The primary use case is to allow users to drop config partials into their home directories and for it to *just work* without having to modify root-owned config files.  

I could also resolve this by putting an empty file into the possibly non-existent directory when initializing the desktop session but would prefer to avoid the hack.

## Impact

Without this feature i3 partials do not really offer "config partial discoverability" as a feature due to the need to explicitly configure all directories.

<pre>
[ ] This feature requires new configuration and/or commands
</pre>

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: 4.20 (pre-release, built manually)
</pre>

<pre>
- Linux Distribution & Version: Regolith on Debian Bullseye
- Are you using a compositor (e.g., xcompmgr or compton):  No
</pre>

i3bot:
Please note that new features which require additional configuration will usually not be considered. We are happy with the feature set of i3 and want to focus in fixing bugs instead. We do accept feature requests, however, and will evaluate whether the added benefit (clearly) outweighs the complexity it adds to i3.

Keep in mind that i3 provides a powerful way to interact with it through its IPC interface: https://i3wm.org/docs/ipc.html.

Airblader:
I agree, I think this shouldn't cause errors.

kgilmer:
Related to https://github.com/i3/i3/pull/4420

PR #4495, opened 2021-09-06
Lower severity of missing include path specified from error to info.

kgilmer:
Addresses:  https://github.com/i3/i3/issues/4494

NOTE: As a very minor change to an unreleased feature I did not think it needs docs, but am happy to add something if that's desired.

## Testing Done

1. Build and copy binary to system exhibiting error due to missing config directory.  
2. Log back into session
3. Observe no nagbar error message nor syslog entry relating to invalid directory


Airblader (approved):

Airblader:
Thanks!

PR #4497, opened 2021-09-07
Make window icons follow title alignment

### LINKED ISSUE #4464: Window icons should follow title alignment
As discussed in #4453

## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

Window icons are always left-aligned

## Expected Behavior

Window icons should follow the title's alignment. They don't need to support extra configuration options.

## Reproduction Instructions

Use `title_align center`, run `title_window_icon on` on any window.

## Environment

<pre>
Binary i3 version:  4.19.2-103-gfc65ca36 © 2009 Michael Stapelberg and contributors
[KRunning i3 version: 4.19.2-103-gfc65ca36 (pid 546)
Loaded i3 config:
  /home/orestis/.i3/config (main) (last modified: 2021-07-27T08:07:20 CEST, 103 seconds ago)

</pre>

orestisfl:
Fixes #4464
Related to #4453 

See commit message

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -614,33 +614,11 @@ void x_draw_decoration(Con *con) {
 
     /* 6: draw the icon and title */
     int text_offset_y = (con->deco_rect.height - config.font.height) / 2;
-    int text_offset_x = 0;
 
     struct Window *win = con->window;
 
-    const int title_padding = logical_px(2);
     const int deco_width = (int)con->deco_rect.width;
-
-    /* Draw the icon */
-    if (con->window_icon_padding > -1 && win && win->icon) {
-        /* icon_padding is applied horizontally only,
-         * the icon will always use all available vertical space. */
-        const int icon_padding = logical_px(1 + con->window_icon_padding);
-
-        const uint16_t icon_size = con->deco_rect.height - 2 * logical_px(1);
-
-        const int icon_offset_y = logical_px(1);
-
-        text_offset_x += icon_size + 2 * icon_padding;
-
-        draw_util_image(
-            win->icon,
-            &(parent->frame_buffer),
-            con->deco_rect.x + icon_padding,
-            con->deco_rect.y + icon_offset_y,
-            icon_size,
-            icon_size);
-    }
+    const int title_padding = logical_px(2);
 
     int mark_width = 0;
     if (config.show_marks && !TAILQ_EMPTY(&(con->marks_head))) {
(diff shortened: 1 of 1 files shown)


src/x.c
```diff
@@ -614,33 +614,13 @@ void x_draw_decoration(Con *con) {
 
     /* 6: draw the icon and title */
     int text_offset_y = (con->deco_rect.height - config.font.height) / 2;
-    int text_offset_x = 0;
 
     struct Window *win = con->window;
 
-    const int title_padding = logical_px(2);
+    /* icon_padding is applied horizontally only, the icon will always use all
```
Airblader: Is this comment placed correctly?

src/x.c
```diff
@@ -699,33 +679,60 @@ void x_draw_decoration(Con *con) {
         goto copy_pixmaps;
     }
 
+    int icon_size = max(0, con->deco_rect.height - logical_px(2));
+    int icon_padding = logical_px(1 + con->window_icon_padding);
```
Airblader: I know this was like this before, but why do we actually add `1` to the defined padding? I'm asking because we haven't released icons yet, so if there's no convincing reason for this we could still change it.

src/x.c
```diff
@@ -699,33 +679,60 @@ void x_draw_decoration(Con *con) {
...
+    int total_icon_space = icon_size + 2 * icon_padding;
+    const bool has_icon = con->window_icon_padding > -1 && win && win->icon && total_icon_space < deco_width;
```
Airblader: nit for readability

```suggestion
    const bool has_icon = (con->window_icon_padding > -1) && win && win->icon && (total_icon_space < deco_width);
```

src/x.c
```diff
@@ -699,33 +679,60 @@ void x_draw_decoration(Con *con) {
...
+    icon_size *= has_icon;
```
Airblader: Maybe we can refactor all of these calculations into a separate function which returns the effective size, padding and total space? It's quite lengthy and noisy in this function, and I think we could write it a bit nicer in that case (like not multiplying by a boolean).

Airblader: In fact, we can probably extract more than one function; one for these calculations, and another one for the switch-case block to calculate offsets.

src/x.c
orestisfl: I think it's meant to be a `min`? Replaced it.

src/x.c
orestisfl: Meant to replace the `*=` with an `if` but forgot.

The problem is that in C it's not that convenient to have functions with multiple arguments & returns. Do you still think it will improve readability?

orestisfl: Meant to replace the `*=` with an `if` but forgot, check now.

In C it's a bit messy to have a function with multiple arguments/returns, are you sure it's going to improve readability here?

Airblader: I was thinking we could just introduce some structs (local to this file, of course) for this. But I'll leave it up to you to.

orestisfl: Would you be fine with accepting the PR like it is?

src/x.c
```diff
@@ -699,33 +677,62 @@ void x_draw_decoration(Con *con) {
         goto copy_pixmaps;
     }
 
+    /* icon_padding is applied horizontally only, the icon will always use all
+     * available vertical space. */
+    int icon_size = max(0, con->deco_rect.height - logical_px(2));
+    int icon_padding = logical_px(min(1, con->window_icon_padding));
```
Airblader: I don't understand this being `min` here. Doesn't that mean the padding will never be anything other than 0 or 1?

orestisfl: Thanks, good catch. It's supposed to be `max()`. Makes sense now?

Airblader (approved):
LGTM, but with the min/max confusion my question would be whether you actually tested it out with different padding settings? :-)

orestisfl:
I did, when it was `1 + pad`. I just mix up `min`/`max` sometimes :sweat_smile:

PR #4503, opened 2021-09-10
Update docs to Discussions

Airblader:
closes #4482


docs/i3-pod2html
```diff
@@ -83,7 +83,7 @@ $parser->html_header_after_title(
 			<ul id="nav">
 				<li><a style="border-bottom: 2px solid #fff" href="/docs">Docs</a></li>
 				<li><a href="/screenshots">Screens</a></li>
-				<li><a href="https://www.reddit.com/r/i3wm/">FAQ</a></li>
+				<li><a href="https://www.github.com/i3/i3/discussions">Get Help</a></li>
```
Airblader, line 86: It's not really an "FAQ", nor has it ever been one. It's a Q&A. But I think "Get Help" is a bit more understandable as non-native speakers might not be familiar with the abbreviation.

release.sh
```diff
@@ -227,4 +227,4 @@ echo ""
 echo "Announce on:"
 echo "  twitter"
 echo "  #i3 topic"
-echo "  reddit /r/i3wm (link post to changelog)"
+echo "  GitHub Discussions → Announcements"
```
orestisfl, line 231: @stapelberg are you planning to  keep making the posts to reddit?

stapelberg (approved):

orestisfl (approved):

release.sh
stapelberg: Probably good to post in both places, at least for now

Airblader: I've added it back.

PR #4528, opened 2021-09-19
GET_CONFIG: add raw/variable-processed contents of all config files

stapelberg:
We do this by adding to included_files as i3 processes the configs.

This should allow for easy debugging, without having to change how i3 processes
config files.

related to #4192

### DIFF
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -156,6 +156,7 @@ int main(int argc, char *argv[]) {
     char *payload = NULL;
     bool quiet = false;
     bool monitor = false;
+    bool raw_reply = false;
 
     static struct option long_options[] = {
         {"socket", required_argument, 0, 's'},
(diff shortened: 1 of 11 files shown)


i3-msg/main.c
```diff
@@ -262,32 +266,39 @@ int main(int argc, char *argv[]) {
...
-        yajl_status state = yajl_parse(handle, (const unsigned char *)reply, reply_length);
-        yajl_free(handle);
-
-        switch (state) {
-            case yajl_status_ok:
-                break;
-            case yajl_status_client_canceled:
-            case yajl_status_error:
-                errx(EXIT_FAILURE, "IPC: Could not parse JSON reply.");
+        if (raw_reply) {
```
Airblader: Is there something missing here? This case is empty.

stapelberg: Nothing missing, line 284 is what will print the reply :)

Airblader: In that case, why not just `if (!raw_reply)`?

stapelberg: Done

Airblader (approved):
LGTM, however the entry in the release notes is missing.

stapelberg:
Done (plus the missing docs/ipc update, too).

Issue #4532, opened 2021-09-23
Partial loss of keyboard focus with krita


ubitux:
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

When switching workspaces and back using the keyboard, Krita somehow looses focus partially: many keyboard shortcuts do not work anymore.

## Expected Behavior

Focus should be preserved/restored so that the keyboard shortcuts still work.

## Reproduction Instructions

1. Open krita (tested with 4.4.8)
2. Create a new drawing area (File → New)
3. While maintaining the `SPACE` key down use the left-click of the mouse to move the drawing area (this should work)
4. Change work space using the keyboard, and come back to the workspace with krita
5. Krita does not register the `SPACE` key anymore: left-clicking with the mouse will draw on the area instead of moving it

It is interesting to note that while the `Ctrl` modifier also suffers the same fate as `SPACE`, it is possible to grab the focus back by doing for example `Alt+f` (to open the `File` menu) and closing it with `Esc` (need to press it twice). After doing this round trip the keyboard focus is completely back.

## Environment

```
Binary i3 version:  4.19.2 (2021-02-27) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.19.2 (2021-02-27) (pid 603)
Loaded i3 config: /home/ux/.config/i3/config (Last modified: Tue 21 Sep 2021 05:55:46 PM CEST, 150987 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

<pre>
Logfile URL: https://logs.i3wm.org/logs/5145410440003584.bz2
</pre>

The log file contains exactly the *Reproduction Instructions*, followed by the opening of a term querying the log dump. It is also using the default config.

<pre>
- Linux Distribution & Version: ArchLinux (rolling-release)
- Are you using a compositor (e.g., xcompmgr or compton): No
</pre>

Edit 1: added log file
Edit 2: FWIW, this is not reproducible with sway

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

ubitux:
> I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

[...]

PR #4538, opened 2021-09-25
Log FocusOut events

psychon:
Currently, i3 only logs FocusIn events. Thus, a debug log tells us when
some window gets the focus. However, we don't know when it loses the
focus. This commit remedies this by adding some log messages for this.

To decide what to log, I just copied some code from handle_focus_in().
No idea how relevant all these details are.

Idea-in-context-of: https://github.com/i3/i3/issues/4532

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -1059,6 +1059,76 @@ static void handle_focus_in(xcb_focus_in_event_t *event) {
 }
 
 /*
+ * Log FocusOut events.
+ *
+ */
+static void handle_focus_out(xcb_focus_in_event_t *event) {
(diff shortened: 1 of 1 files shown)


orestisfl (changes requested):
I would prefer if this function produced just one log line. For FocusIn it makes sense that there are mutliple because it helps follow the flow of the code in the logs.

My suggestion is to get all the information you need in const variables and make a single DLOG call in the end.

psychon:
Updated. This now just logs all the fields of the event.

orestisfl (approved):
Thanks!

PR #4596, opened 2021-10-17
Fix focus issue when moving windows across output boundary

### LINKED ISSUE #3518: Focus does not follow window with directional move when focus_follows_mouse is enabled
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Expected Behavior

I have the following bindings:
`bindsym $mod+Control+H move left`
`bindsym $mod+Control+L move right`

When I use these to move a window from e.g. workspace 1 on my laptop display to workspace 2 on my monitor, I would expect the focus to follow the window as it moves. It is the usual behaviour for focus to follow the window as you move it left and right within a workspace, so I think it's reasonable to expect this to apply when moving it across a workspace with "move left" or "move right" (or even up and down if you have that kind of monitor layout).

## Current Behavior

If in my example scenario, there is already one or more windows in workspace 2, the window focus ends up on one of them instead of the one I'm moving. This behaviour is quite annoying because I then have to pause to find out where the focus has moved to, then move it back to the window I was moving so I can continue moving it to the desired location. It's really not very intuitive.

## Reproduction Instructions
To replicate:
- 2 displays needed
- Open one window on a workspace in primary display
- Open a second window on another workspace in the secondary display
- Use "move <left|down|up|right>" to move the window on the primary display to the workspace on the secondary display
- Observe that focus has not followed the window being moved per the usual behaviour

## Environment

<pre>
i3 version: 4.16
</pre>

<pre>
i3 config: https://pastebin.com/mgE6BRwr
</pre>

<pre>
- Linux Distribution & Version: Ubuntu 18.04, Linux 4.15.0-38-generic
</pre>

I also see that this issue has been raised before in #2257 (but not #809 which is a completely separate issue), and it seems it is definitely not the intended behaviour https://github.com/i3/i3/issues/2257#issuecomment-214448093 and yet it still remains...

Many thanks,
Jonathan

algmyr:
Focusing the container before doing the workspace switching dance seems to do the trick to fix #3518. Running locally the tests seem ok with the change.

I don't exactly know why the existing code doesn't work, setting focus to the specific container seems like it should work since is explicitly calls out to focus a specific container. If I had to guess it's a weird race condition like sequence, something like
* i3 moves window
* i3 makes sure the switch workspace event is changed by switching back and forth
* this workspace switch to another output triggers a mouse move to the currently selected container on the other output
* i3 focuses the new container
* the mouse move from earlier (with `focus_follows_mouse`) focuses the container under the cursor, which is the one that had focus before

I'm able to reproduce similar issues when focusing. Scenario: I'm at my rightmost container in my left output, right output has a vertical split. Doing the keybinds for `move right` and `move down` in quick succession focuses the bottom container for a brief moment before jumping back to the container that the mouse jumped to. This would support the hypothesis of this being a race condition between i3 and Xorg.

If the above is the case this is just a hack that makes the behavior non-damaging for this particular case. The proper fix would be to not have the focus follows mouse happen in this case where i3 itself is the one moving the mouse (to a container that is focused already).

### DIFF
--- a/RELEASE-NOTES-next
+++ b/RELEASE-NOTES-next
@@ -6,23 +6,30 @@
 This is i3 v4.20. This version is considered stable. All users of i3 are
 strongly encouraged to upgrade.
 
-Background/wallpaper workaround:
+Two long-awaited features have been added:
 
-Some login managers (e.g. gdm) start the X11 server with the -background none
-flag. When this flag is set, a background needs to be explicitly set later in
-the X11 session, otherwise stale copies of closed windows remain visible on the
-X11 root window (symptom looks like “my terminal window is not closing”).
+ 1. You can now use an “include” directive in your i3 config:
+    https://i3wm.org/docs/userguide.html#include
 
-i3 works around this situation by setting a screenshot as background when
-starting. Any background you set before starting i3 (e.g. in your Xsession) or
-after starting i3 (e.g. via exec statements in the i3 config) will be visible.
+ 2. You can now enable showing window icons in window titlebars:
+    https://i3wm.org/docs/userguide.html#title_window_icon
 
-A downside of this workaround is that if you have any windows already open in
-your X11 session, those will be part of the screenshot.
+In case you notice any issues regarding your background/wallpaper, note:
 
-To fix this issue, starting in v4.20, i3 detects whether the -background none
-option is enabled and only then sets a screenshot as background.
+  Some login managers (e.g. gdm) start the X11 server with the -background none
+  flag. When this flag is set, a background needs to be explicitly set later in
+  the X11 session, otherwise stale copies of closed windows remain visible on
+  the X11 root window (symptom looks like “my terminal window is not closing”).
 
+  i3 works around this situation by setting a screenshot as background when
+  starting. Any background you set before starting i3 (e.g. in your Xsession) or
+  after starting i3 (e.g. via exec statements in the i3 config) will be visible.
+
+  A downside of this workaround is that if you have any windows already open in
+  your X11 session, those will be part of the screenshot.
+
+  To fix this issue, starting in v4.20, i3 detects whether the -background none
+  option is enabled and only then sets a screenshot as background.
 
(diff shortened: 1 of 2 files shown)


algmyr:
Can confirm that the mouse warping is causing the underlying issue, commenting out the mouse warp at https://github.com/i3/i3/blob/ab389e1d760585e758f343221da7758f7bc2c949/src/workspace.c#L537-L539 makes the issue go away in the move case. Presumably there is a similar mouse warp in the case of focusing.

psychon:
> The proper fix would be to not have the focus follows mouse happen in this case where i3 itself is the one moving the mouse 

<details><summary>Slightly off-topic description on how AwesomeWM deals with this problem</summary>

AwesomeWM does this via some ugly code:
https://github.com/awesomeWM/awesome/blob/1d55ae09aa487a82f2808e82a8884123e2400722/mouse.c#L257-L263
https://github.com/awesomeWM/awesome/blob/1d55ae09aa487a82f2808e82a8884123e2400722/objects/client.c#L1669-L1701

Once upon a time, the code literally changed the event mask of each window so that the X11 server does not generate any events for the cursor entering/leaving windows, then it warped the cursor, then it changed the event masks back to their original value.

These days, awesomeWM just sends a request, writes down its sequence number, warps the mouse cursor, then sends another request. Incoming events with a sequence number between these two values are then later ignored: https://github.com/awesomeWM/awesome/blob/1d55ae09aa487a82f2808e82a8884123e2400722/event.c#L1062-L1090

(The GrabServer is there to hopefully block "everyone else" from doing anything with the X11 server so that we don't accidentally also ignore other events. The old code IMO also would have needed this, but AFAIK it was not present.)
</details>

orestisfl:
Sorry for closing this: I force pushed the wrong thing (current `next`) to your branch, github automatically closed this PR and because of that I lost write access to your branch and can't revert it. Opening a new one.

PR #4604, opened 2021-10-18
Fix focus issue when moving windows across outputs

### LINKED ISSUE #3518: Focus does not follow window with directional move when focus_follows_mouse is enabled

orestisfl:
Fixes #3518
Continuation of #4596

Added a comment explaining the situation.

I've modified the release notes in case we want to sneak this one in before the next release.

TODO release note: `fix focus when moving container between outputs with mouse warp and
  focus_follows_mouse`

### DIFF
--- a/release-notes/bugfixes/2-fix-focus-wrap
+++ b/release-notes/bugfixes/2-fix-focus-wrap
@@ -1,0 +1,1 @@
+fix focus when moving container between outputs with mouse warp and focus_follows_mouse
--- a/src/move.c
+++ b/src/move.c
@@ -228,10 +228,19 @@ static void move_to_output_directed(Con *con, direction_t direction) {
          * the focused container, con, is now a child of ws. To work around this
          * and still produce the correct workspace focus events (see
          * 517-regress-move-direction-ipc.t) we need to temporarily set focused
-         * to the old workspace. */
+         * to the old workspace.
+         *
+         * The following happen:
+         * 1. Focus con to push it on the top of the focus stack in its new
+         * workspace
+         * 2. Set focused to the old workspace to force workspace_show to
+         * execute
+         * 3. workspace_show will descend focus and target our con for
+         * focusing. This also ensures that the mouse warps correctly.
+         * See: #3518. */
+        con_focus(con);
         focused = old_ws;
         workspace_show(ws);
-        con_focus(con);
     }
 
     /* force re-painting the indicators */


Airblader:
Focus issues tend to be delicate and tricky to predict, so I would vote against merging this right before a release.

Also, we should be able to write a test for this, right?

orestisfl:
I am able to write a test that confirms the desires behaviour but it won't
trigger the bug with the old code.

Which is baffling because I can trigger the bug manually in xephyr with
fake outputs.

Airblader:
Do we know why it doesn't fail in the test?

orestisfl:
Pushed a test that passes the checks (without the fix first): https://github.com/i3/i3/pull/4604/checks?check_run_id=3943704155

Any ideas welcome

algmyr:
I guess I didn't end up mentioning that anywhere. The behavior is different when you run `i3-msg move right` and when you run it as a keybind. With the keybind you get focus one to the right of the intended window (i.e. the issue in this PR). With the command you seem to get focus on the rightmost thing on the original output (or the container at the root if you move the last window). All in all it's pretty weird.

orestisfl:
@Airblader what should we do about this? I also tried triggering the bug in a test with key bindings but still couldn't make it to work

Airblader:
Let's proceed without the test then; but we should not add a test that doesn't actually test the change, so let's remove that.

orestisfl:
Up

Airblader (approved):

PR #4610, opened 2021-10-19
Keep each release note in a single file

orestisfl:
To avoid annoying merge conflicts.

~python~ perl script is provided for convenience and simple format checking in
PRs.

### DIFF
--- a/release-notes/changes/0-example
+++ b/release-notes/changes/0-example
@@ -1,0 +1,1 @@
+Acquire the WM_Sn selection when starting as required by ICCCM
--- a/release-notes/generator.pl
+++ b/release-notes/generator.pl
@@ -1,0 +1,89 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+use v5.10;
+use Getopt::Long;
+
+my @template = (
+'
+ ┌──────────────────────────────┐
+ │ Release notes for i3 v4.21   │
+ └──────────────────────────────┘
+
+This is i3 v4.21. This version is considered stable. All users of i3 are
+strongly encouraged to upgrade.
+
+
+ ┌────────────────────────────┐
+ │ Changes in i3 v4.21        │
+ └────────────────────────────┘
+
+',
+'
+ ┌────────────────────────────┐
+ │ Bugfixes                   │
+ └────────────────────────────┘
+
+');
+
+my $print_urls = 0;
+my $result = GetOptions('print-urls' => \$print_urls);
+
+sub read_changefiles {
+    my $dirpath = shift;
+    opendir my $dir, $dirpath or die "Cannot open directory $dirpath: $!";
+    my @files = readdir $dir;
+    closedir $dir;
+
+    my $s = '';
+    for my $filename (@files) {
+        next if $filename eq '.';
+        next if $filename eq '..';
+        next if $filename eq '0-example';
+
+        die "Filename $filename should start with a number (e.g. the pull request number)" unless ($filename =~ /^\d+/);
+
+        $filename = $dirpath . '/' . $filename;
+        open my $in, '<', $filename or die "can't open $filename: $!";
+        my @lines = <$in>;
+        close $in or die "can't close $filename: $!";
+
+        my $content = trim(join("\n  ", map { trim($_) } @lines));
+        die "$filename can't be empty" unless length($content) > 0;
+
+        my $commit = `git log --diff-filter=A --pretty=format:"%H" $filename`;
+        $commit = trim($commit) if defined($commit);
+        die "$filename: git log failed to find commit" if ($?) || (length($commit) == 0);
+
+        my $url = '';
+        if ($print_urls) {
+            my $pr = find_pr($commit);
+            my $url = 'https://github.com/i3/i3/commit/' . $commit;
+            $url = 'https://github.com/i3/i3/pull/' . $pr if defined($pr);
+            $url = $url . "\n";
+        }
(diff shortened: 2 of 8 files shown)


Airblader (commented):
Doesn't this require changes to release.sh so that there are steps to generate the changelog and also to delete the change files again?

orestisfl:
These usually need some final grooming by @stapelberg. Maybe I can add a reminder in `release.sh`?

stapelberg:
When preparing the release notes for v4.20, I actually appreciated the fact that I could use `git blame` to quickly navigate to each PR that was important enough to be mentioned.

If we split the release notes into separate files, that is no longer possible.

But, more importantly: if we use separate files, I don’t see why we not just use `git log` to build the release notes (possibly annotating commits that should not be mentioned, or possibly annotating commits with the specific message that they should produce in the release notes). Using separate files just seems duplicative to me in that regard?

Airblader:
If we go with commits and annotating them in some way, I would vote for just adopting a well-established standard like [conventional commits](https://www.conventionalcommits.org/).

stapelberg:
Conventional commits doesn’t cover release notes, though, or at least I couldn’t see anything about that?

I’m used to having a RELNOTES field in the commit message or similar.

Airblader:
With conventional commits, a changelog is usually generated directly from the commit message headline. If you want the release notes to be orthogonal to that, then yes, conventional commits AFAIK doesn't cover that.

stapelberg:
Yes, in general I’d like the release notes to be orthogonal, otherwise we’re back to square one where I manually compile the release notes based on git history.

orestisfl:
The script can probably call git-blame and find the PR if we want.

My problem with the commit-based approach is that it's a bit more cryptic and that there is no real way to edit the release notes during development, at best we can append release notes with empty commits.

release-notes/generator.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
```
stapelberg: I know it’s annoying, but can you rewrite this in Perl for consistency with the rest of the code base? I’d like to keep the number of languages involved at a minimum. Let me know if you get stuck and I can try to give a pointer.

stapelberg (changes requested):
https://about.gitlab.com/blog/2018/07/03/solving-gitlabs-changelog-conflict-crisis/ comes to a similar conclusion, so I’m willing to try the one-file-per-change approach.

orestisfl:
Switch to Perl

Updated to find the PR if possible. Not 100% robust but I doubt there is a better way that doesn't involve fetching information from github.

That could be done like this: 
``` shell
$ curl 'https://api.github.com/repos/i3/i3/commits/b7af69cba8027ab7c36678c921136d85f1e7242b/pulls' | jq -r .[0].html_url
https://github.com/i3/i3/pull/4617
```

release-notes/generator.pl
```diff
@@ -0,0 +1,80 @@
...
+'
+ ┌────────────────────────────┐
+ │ Bugfixes                   │
+ └────────────────────────────┘
+
+');
+
+sub read_changefiles {
+    my $dirpath = shift;
+    opendir my $dir, $dirpath or die "Cannot open directory $dirpath: $!";
```
stapelberg, line 34: It’s probably a little easier to use https://perldoc.perl.org/File::Find instead of readdir.

release-notes/changes/0-example
```diff
@@ -0,0 +1 @@
+Acquire the WM_Sn selection when starting as required by ICCCM
```
stapelberg, line 1: Can you put in some unicode here? Special characters and/or emoji perhaps. Just to make sure the generator is UTF-8 clean.

release-notes/generator.pl
```diff
@@ -0,0 +1,80 @@
...
+        die "$filename can't be empty" unless length($content) > 0;
+
+        my $commit = `git log --diff-filter=A --pretty=format:"%H" $filename`;
+        $commit = trim($commit) if defined($commit);
+        die "$filename: git log failed to find commit" if ($?) || (length($commit) == 0);
+        my $pr = find_pr($commit);
+        my $url = 'https://github.com/i3/i3/commit/' . $commit;
+        $url = 'https://github.com/i3/i3/pull/' . $pr if defined($pr);
+
+        $s = $s . '  • ' . $content . "\n" . $url . "\n";
```
stapelberg: I’m not sure if we want to always include the URL in the output.

It’s useful for me when compiling the release notes (for clarifications if needed),
but I’m not sure others will appreciate the URLs as much (or just see them as line noise).

Maybe we could use a flag for this?

stapelberg (changes requested):

release-notes/generator.pl
orestisfl: Is it easy to restrict only to the first directory level?

release-notes/changes/0-example
orestisfl: The examples are for contributors.

Copied it over with emojis for testing

release-notes/generator.pl
orestisfl: A bit overkill IMO since this script is only for us and the CI but added a flag

release-notes/generator.pl
stapelberg: Can’t see a way to do that easily. We can probably stick to readdir then.

stapelberg (approved):

Issue #4614, opened 2021-10-21
i3bar crashes on the second monitor


rvalieris:
_Note by maintainers:_ This issue is a regression in i3 4.20. It has been fixed, but not yet released. If you're encountering this issue, please give it a thumbs up so we can determine how many people are affected.

## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
i3bar crashes on the second monitor, on a dual monitor setup. the primary monitor bar runs fine. this was not happening before the 4.20 update.

## Reproduction Instructions
on dual monitor setup, starting/restarting i3 crashes the i3bar on the second monitor.

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
Binary i3 version:  4.20 (2021-10-19) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.20 (2021-10-19) (pid 1667)
Loaded i3 config:
  ~/.config/i3/config (main) (last modified: Wed 20 Oct 2021 07:00:36 PM -03, 73457 seconds ago)
  ~/.config/i3/0.config.local##hostname.muon (included) (last modified: Thu 21 Oct 2021 03:17:54 PM -03, 419 seconds ago)
  ~/.cache/i3-auto-floating.config (included) (last modified: Thu 21 Oct 2021 01:00:19 PM -03, 8674 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
</pre>

<pre>
- Linux Distribution & Version: Archlinux
- Are you using a compositor (e.g., xcompmgr or compton): no
</pre>

## Backtrace

<pre>
Core was generated by `./i3bar --bar_id=bar-1'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa11f7b2a0e in __strcmp_avx2 () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007fa11f7b2a0e in __strcmp_avx2 () at /usr/lib/libc.so.6
#1  0x000056312e90464e in strings_differ (a=0x0, b=0x0) at ../i3bar/src/ipc.c:176
#2  0x000056312e90478f in got_bar_config_update
[...]

PR #4617, opened 2021-10-22
check for both arguments NULL on strings_differ

rvalieris:
fix for https://github.com/i3/i3/issues/4614


orestisfl (approved):

orestisfl:
@stapelberg this will need a minor release, that's why I ignored the release notes.

I predict that quite a few people will hit this regression.

PR #4618, opened 2021-10-22
Docs: Testsuite: Grammar

Jonta:


docs/testsuite
```diff
@@ -4,7 +4,7 @@ Michael Stapelberg <michael@i3wm.org>
 September 2012
 
 This document explains how the i3 testsuite works, how to use it and extend it.
-It is targeted at developers who not necessarily have been doing testing before
+It is targeted at developers who have not necessarily been doing testing before
```
orestisfl: I think present perfect would be even simpler here: "who haven't necessarily done testing before or..."

orestisfl (changes requested):
Thanks ;)

docs/testsuite
Jonta: I agree

orestisfl: So, do you want to update your PR? :)

Jonta: @orestisfl Done

orestisfl (approved):

PR #4637, opened 2021-10-31
Correct short form of raw option in i3-msg manpage

lycurgus:
It looks like the recently-added `raw` option to i3-msg has a typo in the manpage :)

Apologies for the tiny PR - noticed this one while checking something else out and thought it was quicker to raise a PR than to raise an issue and make someone else fix it....


Airblader (approved):
Good catch, thanks a lot for fixing it!

Issue #4381, opened 2021-03-11
Developing i3 inside an i3 session breaks the IPC socket of the outer session


psychon:
Okay, before the discussion in https://github.com/i3/i3/issues/4380#issuecomment-796611526 ends up "stealing" the issue, let's move this here.

## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

```
$ LC_ALL=C i3 --moreversion 
Binary i3 version:  4.19.1-non-git © 2009 Michael Stapelberg and contributors
(Getting version from running i3, press ctrl-c to abort…)i3: Could not connect to i3 on socket /run/user/1000/i3/ipc-socket.709: No such file or directory
```

## Expected Behavior

```
Binary i3 version:  4.19.2-73-gaf2171d0+ © 2009 Michael Stapelberg and contributors
Running i3 version: 4.19.2-73-gaf2171d0+ (pid 360305)
Loaded i3 config: /home/psychon/.config/i3/config (Last modified: Mo 08 Mär 2021 16:11:41 CET, 240617 seconds ago)

The i3 binary you just called: /home/psychon/projects/i3/build/i3
The i3 binary you are running: ./build/ychon/projects/i3/build/i3
```

## Reproduction Instructions

Run `i3`. `git clone` this repo inside such a session and run it as `Xephyr :1 & sleep 1 ; DISPLAY=:1 ./build/i3`.

(Note: I did not actually test whether the above breaks the IPC socket, but I guess so a lot)

## Why this happens

The outer i3 session sets `I3SOCK` when starting processes: https://github.com/i3/i3/blob/af2171d01d40613ec8e6eddcc148722b0ec248f9/src/startup.c#L193

The inner i3 session gets this environment variable and unlinks the socket: https://github.com/i3/i3/blob/af2171d01d40613ec8e6eddcc148722b0ec248f9/src/main.c#L647-L653

https://github.com/i3/i3/blob/af2171d01d40613ec8e6eddcc148722b0ec248f9/libi3/create_socket.c#L37-L38

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: 
Heh. Does i3bot pick up the output above or will it complain about missing version information?
</pre>

<details><summary>Config file</summary><pre>
Doesn't matter, I bet.
</pre>
</details>

<pre>
Logfile URL:
Sorry, no
</pre>

<pre>
[...]

PR #4638, opened 2021-10-31
Do not replace existing IPC socket

psychon:
Imagine you are a happy i3 user and want to also write patches for i3.
You use "Xephyr :1" to get another X11 server and then start your newly
build i3 in it with "DISPLAY=:1 ./i3". You test your changes and
everything seems fine. You are happy. Later that day, you try to log
out, but the $mod+Shift+e key binding from the default config no longer
works. i3-msg cannot connect to the IPC socket because "No such file or
directory". What is going on?

The problem boils down to $I3SOCK having something like two meanings.
When i3 starts, it sets the environment variable $I3SOCK to the path of
its IPC socket. That way, any process started from i3 inherits this and
i3-msg knows how to talk to i3. However, when this variable is already
set when i3 starts, then i3 will replace the existing socket. Thus, in
the earlier experiments, the "separate i3" that was used for
experimenting stole the "main i3"'s socket and replaced it with its own.
When it exited, it deleted that socket.

This commit adds half a work around to this problem: When creating the
IPC socket, i3 will now first try to connect() to the socket. If this
succeeds, it will complain and refuse to use this socket. If the
connect() call fails, it will proceed as usual and create the socket.

Note that trying to connect() to a socket that no process listens on
will fail. Thus, this new code only really "triggers" when some process
is actively listening on this socket and accepting connections.

Example output for when the socket is already in use:

$ I3SOCK=/tmp/sdfdsf DISPLAY=:2 ./i3
31.10.2021 17:03:55 - [libi3] ERROR: Refusing to create UNIX socket at /tmp/sdfdsf: Socket is already in use
31.10.2021 17:03:55 - ERROR: Could not create the IPC socket, IPC disabled

This commit sadly only provides part of the solution. i3 will still
delete the socket when shutting down, even if it failed to create the
IPC socket. Thus, the ipc socket will still break, but now only later.
This will be fixed separately.

First-step-towards-fixing: https://github.com/i3/i3/issues/4381

----

Only unlink() IPC socket if we created it

When the IPC socket could not be created, i3 would still unlink() the
socket during shutdown. This might mean that "unrelated stuff" is
removed. In the issue below, the IPC socket of another i3 instance was
unlinked.

Together with the previous commit, this fixes
https://github.com/i3/i3/issues/4381

### DIFF
--- a/libi3/ipc_connect.c
+++ b/libi3/ipc_connect.c
@@ -13,6 +13,7 @@
 #include <string.h>
 #include <sys/socket.h>
 #include <sys/un.h>
+#include <unistd.h>
 
 /*
  * Connects to the i3 IPC socket and returns the file descriptor for the
@@ -39,6 +40,20 @@ int ipc_connect(const char *socket_path) {
         path = sstrdup("/tmp/i3-ipc.sock");
     }
 
+    int sockfd = ipc_connect_impl(path);
+    if (sockfd < 0) {
+        err(EXIT_FAILURE, "Could not connect to i3 on socket %s", path);
+    }
+    free(path);
+    return sockfd;
+}
+
+/**
+ * Connects to the socket at the given path with no fallback paths. Returns
+ * -1 if connect() fails and die()s for other errors.
+ *
+ */
+int ipc_connect_impl(const char *socket_path) {
     int sockfd = socket(AF_LOCAL, SOCK_STREAM, 0);
     if (sockfd == -1)
         err(EXIT_FAILURE, "Could not create socket");
@@ -48,9 +63,10 @@ int ipc_connect(const char *socket_path) {
     struct sockaddr_un addr;
     memset(&addr, 0, sizeof(struct sockaddr_un));
     addr.sun_family = AF_LOCAL;
-    strncpy(addr.sun_path, path, sizeof(addr.sun_path) - 1);
-    if (connect(sockfd, (const struct sockaddr *)&addr, sizeof(struct sockaddr_un)) < 0)
-        err(EXIT_FAILURE, "Could not connect to i3 on socket %s", path);
-    free(path);
+    strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path) - 1);
+    if (connect(sockfd, (const struct sockaddr *)&addr, sizeof(struct sockaddr_un)) < 0) {
+        close(sockfd);
+        return -1;
+    }
     return sockfd;
 }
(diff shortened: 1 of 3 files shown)


orestisfl:
Do we realistically ever want to run i3 without IPC? @Airblader @stapelberg  it's currently allowed in 

https://github.com/i3/i3/blob/e4d3ec92172380f0ee278541b6e7356307e124ed/src/main.c#L992-L994

but do we ever want to have this happen in this day and age?

Airblader:
I can't imagine how such an instance of i3 would be useful at all.

psychon:
Rebased on `next` for #4647. Dropped commit e225b2c27a306c911583995b334a5900732d9471 "Only unlink() IPC socket if we created it"  since it is no longer relevant.

I did not re-test this change since I expect nothing changed.

libi3/ipc_connect.c
```diff
@@ -39,6 +40,19 @@ int ipc_connect(const char *socket_path) {
         path = sstrdup("/tmp/i3-ipc.sock");
     }
 
+    int sockfd = ipc_connect_impl(path);
+    if (sockfd < 0)
+        err(EXIT_FAILURE, "Could not connect to i3 on socket %s", path);
```
orestisfl:
```suggestion
    if (sockfd < 0) {
        err(EXIT_FAILURE, "Could not connect to i3 on socket %s", path); }
```

orestisfl (changes requested):
LGTM, one nitpick

orestisfl (approved):

orestisfl:
Ok, please squash everything and I'll merge

stapelberg (approved):

stapelberg:
Thanks!

stapelberg:
This doesn’t build?
```
 ../libi3/create_socket.c: In function ‘create_socket’:
  ../libi3/create_socket.c:43:9: error: ‘errno’ undeclared (first use in this function); did you mean ‘errx’?
           errno = EEXIST;
           ^~~~~
           errx
  ../libi3/create_socket.c:43:9: note: each undeclared identifier is reported only once for each function it appears in
  ../libi3/create_socket.c:43:17: error: ‘EEXIST’ undeclared (first use in this function); did you mean ‘ATEXIT’?
           errno = EEXIST;
                   ^~~~~~
                   ATEXIT
```

Not sure why this only fails in CI for the full build, not for the PR build.

psychon:
Sorry for the breakage.

Cannot reproduce: Current `next` works for me also after the merge.

Random guess: Something else already includes `errno.h` for me but not for you? Could you try adding `#include <errno.h>`, checking if that fixes things and then just commit that to `next` yourself.

stapelberg:
For the record, the breakage is fixed with PR #4649

PR #4644, opened 2021-11-02
portability: add missing headers

bapt:
libi3.h defins a macros with the default modes to create a directory.
Those modes are defined in sys/stat.h, so include the necessary header


orestisfl (approved):
Thanks!

PR #4645, opened 2021-11-02
portability: use getwd() instead of get_current_dir_name()

bapt:
Is the ability to check for the PWD env var was not intented it provides the same functionnality

### DIFF
--- a/src/config_parser.c
+++ b/src/config_parser.c
@@ -864,7 +864,7 @@ parse_file_result_t parse_file(struct parser_ctx *ctx, const char *f, IncludedFi
     FILE *fstr;
     char buffer[4096], key[512], value[4096], *continuation = NULL;
 
-    char *old_dir = get_current_dir_name();
+    char *old_dir = getcwd(NULL, 0);
     char *dir = NULL;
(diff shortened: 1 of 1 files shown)


psychon:
From `man getcwd`:

>        As  an extension to the POSIX.1-2001 standard, glibc's getcwd() allocates the buffer dynamically using malloc(3) if buf is NULL.

That doesn't sound portable.

bapt:
On FreeBSD, NetBSD and OpenBSD

> The ability to specify a NULL pointer and have getcwd() allocate memory as necessary is an extension.

musl libc also allocates the same way,
posix speaks about this possibility of this extension and it seems everyone implemented it.

orestisfl (approved):
LGTM but maybe @stapelberg has an opinion

stapelberg:
> That doesn't sound portable.

Then again: 

> get_current_dir_name() is a GNU extension.

So it’s at least not a step back.

In practice, the BSDs have the `getcwd(NULL, 0)` behavior and Linux has it, and we don’t typically test on more platforms than those (for others we can accept patches).

PR #4648, opened 2021-11-02
Move clang-format and safe-wrappers check in new job

orestisfl:
Benefits:
- Faster feedback in case of error
- More checks run in parallel
- Removes stuff from Dockerfiles

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,8 +25,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - run: git fetch --prune --unshallow
-    - name: check & print release notes
-      run: ./release-notes/generator.pl
     - name: construct container name
       run: |
         echo "BASENAME=i3wm/travis-base:$(date +'%Y-%m')-$(./travis/ha.sh travis/travis-base.Dockerfile)" >> $GITHUB_ENV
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -92,3 +86,15 @@ jobs:
...
       run: |
         ./travis/skip-pkg.sh || travis/deploy-github-pages.sh
+  formatting:
+    name: Check formatting
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y clang-format
```
stapelberg: Previously, we pinned clang-format-9, and with this line, we just install whichever clang-format happens to be the default.

Can we stay on a defined version number? I think it’s better to change the version independently of what ubuntu does :)

orestisfl: Any reason to not use the current llvm version? Our local tools are constantly getting updated, why not the CD one as well? If we ever run in a problem we can pin a version.

Besides, I don't think that this tool is radically changing between versions anymore.

stapelberg: > Any reason to not use the current llvm version? Our local tools are constantly getting updated, why not the CD one as well? If we ever run in a problem we can pin a version.

Typically I have updated clang-format conservatively to increase the chances that random contributors have access to that version, because otherwise submitting code can be really frustrating.

We can definitely update to a newer version, but that would best be done in a separate change.

> Besides, I don't think that this tool is radically changing between versions anymore.

I have no idea. Maybe changes are indeed mostly a thing of the past? AFAIR, every update thus far has required changes.

orestisfl: > Typically I have updated clang-format conservatively to increase the chances that random contributors have access to that version, because otherwise submitting code can be really frustrating.

Ubuntu's default is pretty conservative as well. I think 10 right now.

> I have no idea. Maybe changes are indeed mostly a thing of the past? AFAIR, every update thus far has required changes.

I've been using the latest one for a while now, doesn't have any differences.

stapelberg: Take a look at https://packages.ubuntu.com/search?keywords=clang-format&searchon=names&suite=all&section=all

10 or 11 seems fine to me. We can probably go to 12 next year.

orestisfl: But it's already 10 by default if you don't specify a version :)

Do Ubuntu lts releases change the defaults ever? Maybe by the time they update the default Ubuntu version on GH actions, we'll have other minor fixes to correct.

stapelberg: > But it's already 10 by default if you don't specify a version :)

Yeah, but as I mentioned, I’d like to specify a version.

> Do Ubuntu lts releases change the defaults ever? Maybe by the time they update the default Ubuntu version on GH actions, we'll have other minor fixes to correct.

I think Ubuntu doesn’t change clang-format within each individual release.
But, note that we’re using `ubuntu-latest` in our CI, so that could mean that from one day to the other we get a different clang-format version (unless we specify an explicit clang-format version).

orestisfl: > But, note that we’re using ubuntu-latest in our CI, so that could mean that from one day to the other we get a different clang-format version (unless we specify an explicit clang-format version).

Yes, that's what I meant: that it's not unlikely that we have to do minor fixes anyway whenever GH actions updates their default.

Anyway, since you feel strongly about having a fixed number, i've updated my PR.

orestisfl:
Moved the release-notes check in this job as well

stapelberg (approved):

PR #4649, opened 2021-11-04
Fix compilation error on debian

orestisfl:


stapelberg (approved):

orestisfl:
https://github.com/orestisfl/i3/runs/4108858899

PR #4667, opened 2021-11-11
Fix transient_for endless loop

### LINKED ISSUE #4404: Possible endless loop in WM_TRANSIENT_FOR handling
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
No idea, this is all theoretical. I don't actually have time to try to reproduce this right now.

I *guess* there will be an endless loop.

## Expected Behavior

No endless loop / i3 does not freeze.

## Reproduction Instructions

I just came across the code for tracing `WM_TRANSIENT_FOR` (there are other places where the transient for chain is also followed, I think, so do not just fix this one place):
https://github.com/i3/i3/blob/9bf9cb9926402e73aa62e4c1bcc03d671c849c2e/src/manage.c#L495-L511
As far as I can tell, the algorithm follows the `WM_TRANSIENT_FOR` chains until
- a window without `transient_for` is reached
- a window that points at itself with its `transient_for` property is found (the latter is a fix for `ssh-askpass` according to a code comment)

Now imagine two windows that point at each other with their `transient_for` property:
- Window A has `WM_TRANSIENT_FOR` set to `B`
- Window B has `WM_TRANSIENT_FOR` set to `A`

I *think* (but did not yet try - sorry, not enough time) that the result will be an endless loop.

## How is this problem dealt with elsewhere?
A certain other WM that I happen to know about has direct pointers to the end of the `WM_TRANSIENT_FOR` chain in each window. Whenever a window opens/closes, these direct pointers are recomputed as necessary. To avoid endless loops, the code counts the number of steps it did. When the number of steps becomes larger than the number of known windows, the trace is aborted.
https://github.com/awesomeWM/awesome/blob/fda950d186efdc43a64c4300a27a5e19a55c2bd7/objects/client.c#L1485-L1513
(Yes, this algorithm is semi-non-deterministic since loops are broken arbitrarily / depending on which window is updated first. That is still better than an endless loop.)

## Environment

<pre>
i3 version: 
Binary i3 version:  4.19.1-non-git © 2009 Michael Stapelberg and contributors
Running i3 version: 4.19.1-non-git (pid 704)
Loaded i3 config: /home/psychon/.config/i3/config (Last modified: Fr 02 Apr 2021 12:14:29 CEST, 1286092 seconds ago)

</pre>

<details><summary>Config file</summary><pre>
https://github.com/i3/i3/blob/next/etc/config
</pre>
</details>

<pre>
Logfile URL:
</pre>

<pre>
- Linux Distribution & Version: Debian testing
</pre>

orestisfl:
Fixes #4404

CC @psychon

### DIFF
--- a/include/con.h
+++ b/include/con.h
@@ -209,6 +209,14 @@ Con *con_by_frame_id(xcb_window_t frame);
 Con *con_by_mark(const char *mark);
 
 /**
+ * Start from a container and traverse the transient_for linked list. Returns
+ * true if target window is found in the list. Protects againsts potential
+ * cycles.
+ *
+ */
+bool con_find_transient_for_window(Con *start, xcb_window_t target);
+
+/**
  * Returns true if and only if the given containers holds the mark.
  *
  */
--- a/release-notes/bugfixes/3-transient_for
+++ b/release-notes/bugfixes/3-transient_for
@@ -1,0 +1,1 @@
+Fix endless loop with transient_for windows
--- a/src/con.c
+++ b/src/con.c
@@ -731,6 +731,41 @@ Con *con_by_mark(const char *mark) {
 }
 
 /*
+ * Start from a container and traverse the transient_for linked list. Returns
+ * true if target window is found in the list. Protects againsts potential
+ * cycles.
+ *
+ */
+bool con_find_transient_for_window(Con *start, xcb_window_t target) {
+    Con *transient_con = start;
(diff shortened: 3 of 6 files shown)


psychon (approved):
Thanks a lot, this looks quiet nice.

Airblader (approved):
Thanks!

PR #4668, opened 2021-11-12
docs: change IPC window_type value

ctemplin:
from "undefined" to "unknown" as returned by the `get_tree` command.

### DIFF
--- a/docs/ipc
+++ b/docs/ipc
@@ -423,9 +423,9 @@ window_properties (map)::
 	following list: *title*, *instance*, *class*, *window_role*, *machine*
 	and *transient_for*.
 window_type (string)::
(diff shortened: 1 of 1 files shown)


Airblader (changes requested):
Thank you so much for the PR to keep the docs up to date! `undefined` is actually _also_ correct here as that is returned when a container has no window; however, this is a true `undefined`, not the string `"undefined"`, which is a bit misleading.

Of course you are correct that `"unknown"` also exists. Could you change this into mentioning that additionally instead of removing `undefined`? Feel free to change the wording to clarify this difference, too, if you want!

ctemplin:
Thanks for the clarification. I've re-added `undefined` and made what I think is the minimum change required to convey the distinction.

Airblader (approved):
Thanks again! :-)

PR #4684, opened 2021-11-19
Switch from pcre to pcre2

### LINKED ISSUE #4682: enhancement: switch from obsolete PCRE 8.x to current PCRE2 10.x
reported in http://bugs.debian.org/1000105

psychon:
The issue at [0] was opened and I just took a stab at it. I have no
prior experience with pcre and pcre2, but using [1,2] I hacked together
something that seems to work. Next, Michael told me to turn that
patch/hack into a PR, so here we are.

The dependency in meson.build still uses version:'>=8.10' even though
the name of the library changed. No idea what to do with this.

There was a while loop in regex_new() that dealt with an error when pcre
was not compiled with UTF-8 support. This loop uses a magic constant of
32 for the error code. I just dropped this loop, because I was just
writing a hack and did not intend to turn this into a PR. Also, a quick "grep
32 /usr/include/pcre.h" does not find anything useful, so... *shrug*

pcre_study() was removed without replacement, so the corresponding code
is also simply removed.

Testing done: The test suite passes for me. YMMV.

[0]: https://github.com/i3/i3/issues/4682
[1]: https://www.pcre.org/current/doc/html/pcre2api.html
[2]: https://www.pcre.org/current/doc/html/pcre2demo.html

Fixes: https://github.com/i3/i3/issues/4682


psychon:
Someone left some comments on the commit itself. No idea who this person is and I already spent my motivation for understanding pcre2, so I am just mentioning this here so that someone else can look at this and either update my PR directly or tell me what to change.

- The version checks for pcre2 can be dropped: https://github.com/psychon/i3/commit/2defec9c9c9a65025cabbcf673ca608f077835d3#r60592337
- Uhm... I'm not sure what the benefit of this suggestion is. At first I thought "avoids one temporary allocation", but apprently that is not the case...? https://github.com/psychon/i3/commit/2defec9c9c9a65025cabbcf673ca608f077835d3#r60596702

Edit: Their github contributions suggests some involvement with pcre2, i.e. "this person should know better than @psychon".

meson.build
```diff
@@ -316,7 +316,7 @@ xcb_util_xrm_dep = dependency('xcb-xrm', method: 'pkg-config')
 xkbcommon_dep = dependency('xkbcommon', method: 'pkg-config')
 xkbcommon_x11_dep = dependency('xkbcommon-x11', method: 'pkg-config')
 yajl_dep = dependency('yajl', method: 'pkg-config')
-libpcre_dep = dependency('libpcre', version: '>=8.10', method: 'pkg-config')
+libpcre_dep = dependency('libpcre2-8', version: '>=8.10', method: 'pkg-config')
```
stapelberg: This dependency check can be bumped to >10.0 or something like that as per https://github.com/i3/i3/issues/4682#issuecomment-981171773

stapelberg (approved):

stapelberg:
Sorry for the late review, I didn’t see this earlier

meson.build
psychon: Okay, I went with `>=10`

stapelberg (approved):

stapelberg:
Thanks again!

PR #4705, opened 2021-11-29
debian: bump to libpcre2-dev for autobuilds

stapelberg:
related to https://github.com/i3/i3/issues/4682

cc @psychon


psychon:
I don't really understand the setup here, but... how come my PR didn't fail?

Besides that: Looks fine to me.

You could mention in the commit message something like the following from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1000105, but that does not really matter, I guess.

> Historical reasons mean that old PCRE is packaged as pcre3 in Debian

orestisfl (approved):

stapelberg:
> I don't really understand the setup here, but... how come my PR didn't fail?

PRs don’t build the full debian package, our CI only builds Debian packages for commits to the next branch.

Issue #4709, opened 2021-12-02
Add support for window_title_icon toggle


pstray:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

Currently we have `window_title_icon on|off`

## Desired Behavior

I would like to have `window_title_icon toggle` too, so I can have one keybinding to toggle the display of window icons, such as:

`bindsym $mod+i title_window_icon toggle`

In line with the `toggle` parameter to `floating` and`fullscreen`.

i3bot:
Please note that new features which require additional configuration will usually not be considered. We are happy with the feature set of i3 and want to focus in fixing bugs instead. We do accept feature requests, however, and will evaluate whether the added benefit (clearly) outweighs the complexity it adds to i3.

Keep in mind that i3 provides a powerful way to interact with it through its IPC interface: https://i3wm.org/docs/ipc.html.

Airblader:
This is a logical, simple addition. PRs welcome.

orestisfl:
https://github.com/i3/i3/pull/4710

PR #4710, opened 2021-12-02
Implement title_window_icon toggle

pstray:
An implentation of the feature request in i3/i3#4709

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -2064,20 +2064,40 @@ void cmd_title_format(I3_CMD, const char *format) {
 }
 
 /*
- * Implementation of 'title_window_icon <yes|no>' and 'title_window_icon padding <px>'
+ * Implementation of 'title_window_icon <yes|no|toggle>' and 'title_window_icon padding <px>'
  *
  */
 void cmd_title_window_icon(I3_CMD, const char *enable, int padding) {
-    if (enable != NULL && !boolstr(enable)) {
-        padding = -1;
+    bool is_toggle = false;
+    if (enable != NULL) {
+        if (strcmp(enable, "toggle") == 0) {
+            is_toggle = true;
+        } else if (!boolstr(enable)) {
+            padding = -1;
+        }
     }
     DLOG("setting window_icon=%d\n", padding);
     HANDLE_EMPTY_MATCH;
 
     owindow *current;
     TAILQ_FOREACH (current, &owindows, owindows) {
-        DLOG("setting window_icon for %p / %s\n", current->con, current->con->name);
-        current->con->window_icon_padding = padding;
+        if (is_toggle) {
+            const int current_padding = current->con->window_icon_padding;
+            if (padding > 0) {
+                if (current_padding < 0) {
+                    current->con->window_icon_padding = padding;
+                } else {
+                    /* toggle off, but store padding given */
+                    current->con->window_icon_padding = -(padding + 1);
+                }
+            } else {
+                /* Set to negative of (current value+1) to keep old padding when toggling */
+                current->con->window_icon_padding = -(current_padding + 1);
+            }
+        } else {
+            current->con->window_icon_padding = padding;
+        }
+        DLOG("Set window_icon for %p / %s to %d\n", current->con, current->con->name, current->con->window_icon_padding);
 
         if (current->con->window != NULL) {
             /* Make sure the window title is redrawn immediately. */
--- a/testcases/t/314-window-icon-padding.t
+++ b/testcases/t/314-window-icon-padding.t
@@ -38,6 +38,36 @@ is(window_icon_padding($tmp), -1, 'window_icon_padding defaults to -1');
 cmd 'title_window_icon on';
 isnt(window_icon_padding($tmp), -1, 'window_icon_padding no longer -1');
 
+cmd 'title_window_icon toggle';
+is(window_icon_padding($tmp), -1, 'window_icon_padding back to -1');
+
+cmd 'title_window_icon toggle';
+isnt(window_icon_padding($tmp), -1, 'window_icon_padding no longer -1 again');
+
+cmd 'title_window_icon off';
+is(window_icon_padding($tmp), -1, 'window_icon_padding back to -1');
+
+cmd 'title_window_icon padding 3px';
+is(window_icon_padding($tmp), 3, 'window_icon_padding set to 3');
+
+cmd 'title_window_icon toggle';
+ok(window_icon_padding($tmp) < 0, 'window_icon_padding toggled off');
+
+cmd 'title_window_icon toggle';
+is(window_icon_padding($tmp), 3, 'window_icon_padding toggled back to 3');
+
+cmd 'title_window_icon toggle 5px';
+ok(window_icon_padding($tmp) < 0, 'window_icon_padding toggled off');
+
+cmd 'title_window_icon toggle 5px';
+is(window_icon_padding($tmp), 5, 'window_icon_padding toggled on to 5px');
+
+cmd 'title_window_icon toggle 5px';
+ok(window_icon_padding($tmp) < 0, 'window_icon_padding toggled off');
+
+cmd 'title_window_icon toggle 4px';
+is(window_icon_padding($tmp), 4, 'window_icon_padding toggled on to 4px');
+
 exit_gracefully($pid);
 
 ################################################################################
--- a/docs/userguide
+++ b/docs/userguide
@@ -2728,8 +2728,8 @@ specific windows or for all windows (using the <<for_window>> directive).
 
 *Syntax*:
 -----------------------------
-title_window_icon <yes|no>
-title_window_icon padding <px>
+title_window_icon <yes|no|toggle>
+title_window_icon <padding|toggle> <px>
 ------------------------------
 
 *Examples*:
--- a/include/commands.h
+++ b/include/commands.h
@@ -333,7 +333,7 @@ void cmd_shmlog(I3_CMD, const char *argument);
 void cmd_debuglog(I3_CMD, const char *argument);
 
 /**
- * Implementation of 'title_window_icon <yes|no>' and 'title_window_icon padding <px>'
+ * Implementation of 'title_window_icon <yes|no|toggle>' and 'title_window_icon <padding|toggle> <px>'
  *
  */
 void cmd_title_window_icon(I3_CMD, const char *enable, int padding);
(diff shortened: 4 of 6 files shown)


orestisfl (changes requested):
Thanks, LGTM

You'll also need to include:

- [x] A test case in `t/314-window-icon-padding.t`
- [x] A new line in a new file in `release-notes/changes`

pstray:
Not sure why those tests failed...

src/commands.c
```diff
@@ -2077,7 +2077,11 @@ void cmd_title_window_icon(I3_CMD, const char *enable, int padding) {
     owindow *current;
     TAILQ_FOREACH (current, &owindows, owindows) {
         DLOG("setting window_icon for %p / %s\n", current->con, current->con->name);
-        current->con->window_icon_padding = padding;
+        if (strcmp(enable, "toggle") == 0) {
```
orestisfl:
```suggestion
        if (enable != NULL && strcmp(enable, "toggle") == 0) {
```

src/commands.c
```diff
@@ -2077,7 +2077,11 @@ void cmd_title_window_icon(I3_CMD, const char *enable, int padding) {
...
+            current->con->window_icon_padding = current->con->window_icon_padding >= 0 ? -1 : padding;
```
orestisfl: `padding` can be `-1` here. In that case, you should use `0`.

orestisfl (changes requested):

src/commands.c
pstray: yes, it most probably will too... as boolstr("enable") return 0.  I'll fix

pstray:
Fixed it by actually saving the old value instead of setting it directly.

testcases/t/314-window-icon-padding.t
```diff
@@ -38,6 +38,15 @@ is(window_icon_padding($tmp), -1, 'window_icon_padding defaults to -1');
...
 
+cmd 'title_window_icon toggle';
+is(window_icon_padding($tmp), -1, 'window_icon_padding back to -1');
+
+cmd 'title_window_icon toggle';
+isnt(window_icon_padding($tmp), -1, 'window_icon_padding no longer -1 again');
+
+cmd 'title_window_icon off';
+is(window_icon_padding($tmp), -1, 'window_icon_padding back to -1');
+
```
orestisfl, line 49: Can you make sure that the toggle tests actually test the exact padding value? E.g. start with -1, set to 3, toggle to -1, toggle to 3

src/commands.c
```diff
@@ -2077,7 +2077,12 @@ void cmd_title_window_icon(I3_CMD, const char *enable, int padding) {
     owindow *current;
     TAILQ_FOREACH (current, &owindows, owindows) {
         DLOG("setting window_icon for %p / %s\n", current->con, current->con->name);
-        current->con->window_icon_padding = padding;
+        if (enable != NULL && strcmp(enable, "toggle") == 0) {
+            /* Set to negative of (current value+1) to keep old padding when toggling */
+            current->con->window_icon_padding = -(current->con->window_icon_padding + 1);
```
orestisfl: That's nice but I think that if `padding` is set it should have precedence. So, if enabling and `padding > 0`, use that value.

This will be more consistent with `border toggle <n>`

orestisfl (changes requested):

src/commands.c
pstray: There is currently no way of toggling _and_ setting the padding.  But if it is already set with `padding 3px`, then the toggle will keep that padding, and not reset it to 0 as the previous version did.

testcases/t/314-window-icon-padding.t
pstray: Added test for toggling after padding is set

src/commands.c
pstray: Thus, `on` sets to `0`, `off` sets to `-1`, `padding Npx` set to `N`, and `toggle` flips between padding and no padding by storing the the current padding as `-(N+1)`.  If the padding was `0`, it will become `-1`, if it was `-1`, it will become `0`, and `3` will become `-4`, and `-4` will become `3`.

orestisfl: It's not so difficult to adjust. Here is a MWE:

```diff
diff --git a/parser-specs/commands.spec b/parser-specs/commands.spec
index 89b258730..486f20dd3 100644
--- a/parser-specs/commands.spec
+++ b/parser-specs/commands.spec
@@ -468,7 +468,9 @@ state TITLE_FORMAT:
 state TITLE_WINDOW_ICON:
   'padding'
     -> TITLE_WINDOW_ICON_PADDING
-  enable = '1', 'yes', 'true', 'on', 'enable', 'active', '0', 'no', 'false', 'off', 'disable', 'inactive', 'toggle'
+  enable = 'toggle'
+    -> TITLE_WINDOW_ICON_PADDING
+  enable = '1', 'yes', 'true', 'on', 'enable', 'active', '0', 'no', 'false', 'off', 'disable', 'inactive'
     -> call cmd_title_window_icon($enable, 0)
 
 state TITLE_WINDOW_ICON_PADDING:
diff --git a/src/commands.c b/src/commands.c
index ad2b6cc20..5fc4440fd 100644
--- a/src/commands.c
+++ b/src/commands.c
@@ -2068,7 +2068,8 @@ void cmd_title_format(I3_CMD, const char *format) {
  *
  */
 void cmd_title_window_icon(I3_CMD, const char *enable, int padding) {
-    if (enable != NULL && !boolstr(enable)) {
+    const bool is_toggle = (enable && strcmp(enable, "toggle") == 0);
+    if (enable != NULL && !is_toggle && !boolstr(enable)) {
         padding = -1;
     }
     DLOG("setting window_icon=%d\n", padding);
@@ -2076,13 +2077,22 @@ void cmd_title_window_icon(I3_CMD, const char *enable, int padding) {
 
     owindow *current;
     TAILQ_FOREACH (current, &owindows, owindows) {
-        DLOG("setting window_icon for %p / %s\n", current->con, current->con->name);
-        if (enable != NULL && strcmp(enable, "toggle") == 0) {
-            /* Set to negative of (current value+1) to keep old padding when toggling */
-            current->con->window_icon_padding = -(current->con->window_icon_padding + 1);
+        if (is_toggle) {
+            const int current_padding = current->con->window_icon_padding;
+            if (padding > 0) {
+                if (current_padding < 0) {
+                    current->con->window_icon_padding = padding;
+                } else {
+                    current->con->window_icon_padding = -(padding + 1);
+                }
+            } else {
+                /* Set to negative of (current value+1) to keep old padding when toggling */
+                current->con->window_icon_padding = -(current_padding + 1);
+            }
         } else {
             current->con->window_icon_padding = padding;
         }
+        DLOG("Set window_icon for %p / %s to %d\n", current->con, current->con->name, current->con->window_icon_padding);
 
         if (current->con->window != NULL) {
             /* Make sure the window title is redrawn immediately. */
```

WDYT?

pstray: The C code seems good to me.

pstray: Not sure checking for enable != NULL is really needed.  Seems none of the other similar function do :)  But better safe than sorry i guess :grin:

src/commands.c
```diff
@@ -2064,20 +2064,40 @@ void cmd_title_format(I3_CMD, const char *format) {
...
 
 /*
- * Implementation of 'title_window_icon <yes|no>' and 'title_window_icon padding <px>'
+ * Implementation of 'title_window_icon <yes|no|toggle>' and 'title_window_icon padding <px>'
  *
  */
 void cmd_title_window_icon(I3_CMD, const char *enable, int padding) {
-    if (enable != NULL && !boolstr(enable)) {
-        padding = -1;
+    bool is_toggle = 0;
```
orestisfl:
```suggestion
    bool is_toggle = false;
```

src/commands.c
```diff
@@ -2064,20 +2064,40 @@ void cmd_title_format(I3_CMD, const char *format) {
...
+    if (enable != NULL) {
+        if (strcmp(enable, "toggle") == 0) {
+            is_toggle = 1;
```
orestisfl:
```suggestion
            is_toggle = true;
```

testcases/t/314-window-icon-padding.t
```diff
@@ -38,6 +38,30 @@ is(window_icon_padding($tmp), -1, 'window_icon_padding defaults to -1');
...
+cmd 'title_window_icon padding 3px';
+is(window_icon_padding($tmp), 3, 'window_icon_padding set to 3');
+
+cmd 'title_window_icon toggle';
+is(window_icon_padding($tmp), -4, 'window_icon_padding toggled to -4');
```
orestisfl: Let's not test for implementation details

```suggestion
ok(window_icon_padding($tmp) < 0, 'window_icon_padding toggled off');
```

testcases/t/314-window-icon-padding.t
```diff
@@ -38,6 +38,30 @@ is(window_icon_padding($tmp), -1, 'window_icon_padding defaults to -1');
...
+
+cmd 'title_window_icon toggle';
+is(window_icon_padding($tmp), 3, 'window_icon_padding toggled back to 3');
+
+cmd 'title_window_icon toggle 5px';
+is(window_icon_padding($tmp), -6, 'window_icon_padding toggled off, save 5px as -6');
```
orestisfl: Same

```suggestion
ok(window_icon_padding($tmp) < 0, 'window_icon_padding toggled off');
```

testcases/t/314-window-icon-padding.t
```diff
@@ -38,6 +38,30 @@ is(window_icon_padding($tmp), -1, 'window_icon_padding defaults to -1');
...
+
```
orestisfl, line 67: Let's toggle with 5px to make sure it toggles to the correct amount

```suggestion

cmd 'title_window_icon toggle 5px';
is(window_icon_padding($tmp), 5, 'window_icon_padding toggled on to 4px');

cmd 'title_window_icon toggle 5px';
ok(window_icon_padding($tmp) < 0, 'window_icon_padding toggled off');

```

orestisfl (changes requested):
Thanks! I left some final nitpicks

orestisfl:
Finally, please squash all commits into one before we can merge this.

pstray:
Added some fixes to comments and doc, and squashed it all.

orestisfl (approved):
Thank you!

PR #4745, opened 2021-12-24
change split layout to reasonable grid layout

neoedmund:
I'm sorry to pull this request. Because it maybe not be accepted directly. Because 1) It's a quick hacking to make things done. 2) code is formatted differently with other files.
But I still like to make pull, because I just want let people know I did such work and if someone interested, he can make use of it.
So what this change actually do ?
It changes the split layout algorithm to a reasonable grid layout. So when you press meta-E , you can see windows are arranged in a balanced grid. That's it, thanks for the reading.

### DIFF
--- a/src/render.c
+++ b/src/render.c
@@ -8,27 +8,27 @@
  *           various rects. Needs to be pushed to X11 (see x.c) to be visible.
  *
  */
-#include "all.h"
+# include "all.h"
 
-#include <math.h>
+# include <math.h>
 
 /* Forward declarations */
-static int *precalculate_sizes(Con *con, render_params *p);
-static void render_root(Con *con, Con *fullscreen);
-static void render_output(Con *con);
-static void render_con_split(Con *con, Con *child, render_params *p, int i);
-static void render_con_stacked(Con *con, Con *child, render_params *p, int i);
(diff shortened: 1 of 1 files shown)


neoedmund:
Also this change based on tag 4.20.1

Airblader:
Thanks for opening this PR. I'm going to close it partially for the reasons you mentioned, but also because it sounds like a big breaking change in behavior. Thanks for making it public for people to find if they're interested!

If you're looking to bring a change into i3, it'd be best to open an issue with a description of the intended change and we can evaluate its merit and think about a way of doing it in a non-breaking way.

neoedmund:
This change breaks the design aspect in source level, but not break anything in user level.

Airblader:
Either way, it is currently unreviewable. As I said, please feel free to open an issue to describe and discuss the change, maybe with some specific examples. Based in your PR description you aren't looking to actually get this merged, though (?)

neoedmund:
No, the message is sent and I'm done. I don't care that actually no one see this message except you in the world. 

BTW I found `i3` recently and like to use it very much, thanks for such good work! I found I wasted so much time to drag and resize the windows in Windows like systems. Aha.

PR #4795, opened 2022-01-17
Minor doc tweak to coach user after screen layout changes

jwhendy:
Initial attempt to add some language re. discussion #4784 .

@Airblader , following all the discussion, I was torn on modifying the section on reloading/restarting (with the goal that all users should refer to that one location on what to do)... or adding a note to 4.20 itself.

I went with the latter, as there was precedent for advising a reload in another section on screens, and if I saw note that relaunching might be necessary after layout changes, that would have been all I needed. Open to input if you prefer something else.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1041,6 +1041,9 @@ workspace 5 output VGA1 LVDS1
 workspace "2: vim" output VGA1
 ---------------------------
 
+When adjusting your layout in the i3 config, you may need to reload i3 or
+exit/re-launch i3 entirely to apply the changes.
+
 === Changing colors
 
 You can change all colors which i3 uses to draw the window decorations.


orestisfl (changes requested):
Thanks for your change!

I am a bit confused though:
- Why does your change refer to "layouts" when it's in the "workspace assignment" section?
- In which case would you need an actual restart? I'd consider that a bug unless if I am missing something
- All changes in the configuration file take effect after a reload. The userguide is clear enough* about this, why do we need to explicitly mention this here? 
 
*a quick statement could be written in the first paragraph of the 4th section as a minor improvement

jwhendy:
> Why does your change refer to "layouts" when it's in the "workspace assignment" section?

Probably because I still haven't wrapped my mind around the canonical terminology used. By "layout" I mean "the screen on which my workspaces are located in the physical world." I was trying to clarify how to apply changes to these locations.

> In which case would you need an actual restart? I'd consider that a bug unless if I am missing something

Take a look at the referenced issue ([this comment](https://github.com/i3/i3/discussions/4784#discussioncomment-1952622) and [this one](https://github.com/i3/i3/discussions/4784#discussioncomment-1963108)).

Restating my understanding of @Airblader , if the workspace(s) already exist and you change the assigned screen via the config, they will not move upon `reload` or `restart`, only when they are created.

> All changes in the configuration file take effect after a reload. The userguide is clear enough* about this, why do we need to explicitly mention this here?

This was [also covered](https://github.com/i3/i3/discussions/4784#discussioncomment-1956299) in the referenced issue, but I will repeat the flow of things:
- I changed the config and did `mod+shift+r` with no effect
- after re-checking the manual and config several times, I found that `exit + startx` worked
- I proposed that the docs could mention this
- I was told that exiting wasn't necessary, only `reload`, with the caveat in @Airblader 's first answer (which I didn't parse at the time)
- I revisited the docs finding that `restart` encompasses `reload`, and was again confused because I'd done something more extreme
- after another round of comments I properly understood that because ws1 and ws2 *already existed*, indeed an `exit + startx` was necessary (because location will only change upon creation)
- there's precedent for telling the user to apply changes, even though I completely get the goal of only saying this once generally and never saying it again (I also said this in the linked issue). From [Sec 5.3](https://i3wm.org/docs/userguide.html#_display_mode), which shouldn't be necessary, as the user already knows that all changes are applied/refreshed upon `reload`, as stated in 2.9 and 6.20:

> The mode option can be changed during runtime through the bar mode command. On reload the mode will be reverted to its configured value.

- the docs don't clarify what method (`reload`, `restart`, or `exit + startx`) is required when, and my understanding [from here](https://github.com/i3/i3/discussions/4784#discussioncomment-1960484) is that the approach was more or less "just keep doing more extreme things until something works" to avoid going on at length about it. I used the same approach in my proposed language, avoiding lengthy detail in favor of "if x doesn't work, do y."

> *a quick statement could be written in the first paragraph of the 4th section as a minor improvement

This? 

> 4. Configuring i3

I'm not sure what you're defining as a paragraph, but in reading the first entities with line breaks between them, I don't know where a quick statement would make sense.

---

**tl;dr:** this is just trying to add a very small note about this edge case in the relevant section. I had ws1 on my primary external monitor, ws2 on my laptop and wanted to reverse them. Adding this + `restart` didn't work:
```
### these were already in config; adding for context
set $ws1 "1"
set $ws2 "2"

### added these to get ws1 on left and ws2 on right, matching the physical world
workspace $ws1 output DP-6
workspace $ws2 output DP-0.1
```

If you think the manual lays out that `exit + restart` is needed here, please close. For my counterpoint: I ate up a lot of @Airblader 's time on a nuance that was obvious to him, but not to me. The resistance seems about adding a sentence to the documentation. One sentence in exchange for saving your future selves from one future me seems like a win.

You can also pick the sentence, the right terminology, and its location. I was just trying to do something to get the ball rolling and help out.

Airblader:
> The resistance seems about adding a sentence to the documentation.

It's not, but we need to make sure that it is appropriate in its context. Too much text can be an issue in itself. "Reload your config for changes to apply" is not specific to this section, and for obvious reasons we don't want to add that to every section.

The problem here is not that the config needs to be reloaded in any way – rather this is about the fact that workspace assignments are only evaluated when a workspace is being created. That's the fact we should document.

> You can also pick the sentence, the right terminology, and its location. 

How about something like "Note that workspace assignments are only evaluated when a new workspace is created, but not for existing workspaces."?

orestisfl:
> Note that workspace assignments are only evaluated when a new workspace is created, but not for existing workspaces.

If we append ", even upon +reload+ and +restart+."

Would that also be correct?

jwhendy:
> It's not, but we need to make sure that it is appropriate in its context. Too much text can be an issue in itself. "Reload your config for changes to apply" is not specific to this section, and for obvious reasons we don't want to add that to every section.

Indeed, I threw that in for generality's sake as "the normal thing to do," with the note about exiting as the real differentiator.

Ignoring that, this *is* specific to the section, no? What other sections entail a change in which `reload` or `restart` don't work?

Exiting is mentioned in 2.10  and 6.20, with nothing said about the utility of exiting/restarting, just how to do it.

> How about something like "Note that workspace assignments are only evaluated when a new workspace is created, but not for existing workspaces."?

I think one could add @orestisfl 's suggestion for further clarification, though this still leaves the user to connect the dots: "Hmm. Why mention workspace creation here? What are they trying to tell me? Oh! From section a.b and x.y, and c.d, I now see that workspace creation only occurs on startup or moving to a new workspace! Got it, I need to *exit*!"

Except, I ctrl+f'd the userguide for "create" and don't see any clarification that `reload` and `restart` don't [re-]create workspaces. Thus, I'm not sure if this sentence will trigger the user to know "applying *this* change is different from applying other changes."

In my opinion, this isn't sufficient to connect these dots:

>You can also restart i3 inplace with the restart command to get it out of some weird state (if that should ever happen) or to perform an upgrade without having to restart your X session.

It doesn't seem intuitive that the results of "restart i3 inplace" do not match the behavior of "starting i3".

To continue brainstorming, here's a change to my sentence, where I was attempting to be more direct about the required action for this fringe case (with ending parenthetical optional if you'd like to explain the reason):

> To apply workspace assignment changes to existing workspaces, you must exit and re-launch i3 (assignments are only evaluated upon workspace creation, and +reload+/+restart+ do not recreate existing workspaces).

?

Airblader:
There's a few things that are being consistently repeated, so I will clear them up one more (and one final) time.

> What other sections entail a change in which reload or restart don't work?

`reload` _does_ work. The command doesn't randomly decide that the particular change you made to the configuration isn't worth applying; the updated configuration will be in i3's memory afterwards. The problem is rather your expectation of how `assign` works.

`reload` will not re-execute any rules that have been executed already. `restart` will not re-execute any `exec` directives, reload will not re-execute any `exec_always` directives. 

> "[…] Got it, I need to exit!"

This conclusion is incorrect, as stated in the original Discussion. You need to (re-)create the workspace. Exiting entirely out of i3 is one way of achieving that, not _the_ (or the _only_) way.

---

The documentation of `reload` / `restart` is relatively sparse, to say the least, and certainly warrants going into more detail on the expectation of what those commands do.

I don't think there's any reasonable ambiguity on what "creating a workspace" means as this is covered in the userguide. And if it isn't clear enough, that is what warrants fixing. I've been supporting users for some seven years now, and cannot recall anyone ever not understanding what "create a workspace" means. So at the very least I'd claim this is not a common problem.

Explaining how workspaces work in the documentation for `assign` is the wrong place and just fights a symptom of insufficient explanation elsewhere. There's nothing wrong with users having to connect some dots, in fact I would very much expect that. The userguide has to strike a balance of being detailed enough to cover all relevant things, but concise enough to not fill a book.

All that said, this discussion is unnecessarily bloated in my opinion. If it is one sentence, uses the right terminology and is factually correct, I don't care too much about the exact sentence.

jwhendy:
> reload does work. The command doesn't randomly decide that the particular change you made to the configuration isn't worth applying; the updated configuration will be in i3's memory afterwards. The problem is rather your expectation of how assign works.

I didn't say it was random. I thought it would be obvious that "doesn't work" was purely in the context of what we've been discussing all along: "I want to change where my workspaces are. They exist. The typical mechanisms for applying changes don't work here."

The point here isn't what you as a maintainer can tell me about i3 and how my understanding is flawed. The documentation is your liaison to the outside world, so all you need to do is quote it or point me to section numbers that clarify the things you're saying:

- "The problem is rather your expectation of how assign works."
- "reload will not re-execute any rules that have been executed already. restart will not re-execute any exec directives, reload will not re-execute any exec_always directives."
- "This conclusion is incorrect, as stated in the original Discussion. You need to (re-)create the workspace. Exiting entirely out of i3 is one way of achieving that, not the (or the only) way."

*You* know these things; point out where *I* could have known these things.

>I've been supporting users for some seven years now, and cannot recall anyone ever not understanding what "create a workspace" means.

In what context would it have been confusing? `mod+shift+num` JustWorks. Starting i3 and auto-assigning to the monitors that physically exist JustWorks. Trying to get the auto-assigned workspaces to change where they are [edit: those that by default exist after starting i3, after editing the config, with the mechanisms the manual says apply changes]... *doesn't* work.

Airblader:
> "The problem is rather your expectation of how assign works."

This is not currently stated in the userguide, which is exactly why I made my above suggestion of the sentence to add here.

> "reload will not re-execute any rules that have been executed already. restart will not re-execute any exec directives, reload will not re-execute any exec_always directives."

> `exec` commands will not run when restarting i3, if you need a command to run also when restarting i3 you should use the `exec_always` keyword.
> – https://i3wm.org/docs/userguide.html#_automatically_starting_applications_on_i3_startup

jwhendy:
> All that said, this discussion is unnecessarily bloated in my opinion.

I'm sure we both feel this way.

>This is not currently stated in the userguide, which is exactly why I made my above suggestion of the sentence to add here.

But your proposed sentence only says that workspace assignments are only evaluated upon workspace creation. Namely this _also_ needs to be covered, which you chose not to respond to via a quote from the manual:

> This conclusion is incorrect, as stated in the original Discussion. You need to (re-)create the workspace. Exiting entirely out of i3 is one way of achieving that, not the (or the only) way.

Where is the behavior of `restart`/`reload` with respect workspace [re]creation stated, and where are the methods listed which [re]create workspaces (other than starting i3)?

> If it is one sentence, uses the right terminology and is factually correct, I don't care too much about the exact sentence.

Did my proposed sentence not accomplish this, combining the needed pieces of information (in content/sentiment, fully open to whatever word tweaks/phrasing/re-ordering of concepts are desired)?

> To apply workspace assignment changes to existing workspaces, you must exit and re-launch i3 (assignments are only evaluated upon workspace creation, and +reload+/+restart+ do not recreate existing workspaces).

So far, this has been guess and check. If I say reloading doesn't work (obviously meant in the context of _this_ situation), I'm told it _does_ work (but you mean "generally speaking," which isn't what we're talking about). If I say the user needs to exit, I'm told this is clearly false and the user just needs to "recreate the workspace"... but without citing where the manual says how to do this, and without elaborating on the many additional ways this can be accomplished (and are _these_ in the manual?).

P.S. and by "not the only way," is there really a better/easier way you'd propose, or is this a point being made purely as an academic exercise, like including things like kill i3_pid, kill X entirely, reboot, ...

jwhendy:
Probably foolish to add more words at this point, but I want to say: I set out to look for a suggestion on getting my workspace assignments to match my monitor arrangement, adding as a side comment that I ran into a hiccup in applying this change.

I've clearly frustrated and annoyed you (perhaps the both of you) but it was never my intention. Despite trying to be thorough, I made several inadvertent phrasing mistakes out of both ignorance and haste, which became a source of confusion. I apologize for not being more grounded/researched on i3 terminology in order to effectively communicate with you.

I also felt that despite my best efforts to explain the slight opportunity for improvement in the docs, the main responses felt like "you're just doing it wrong," "this is obvious," "the manual is perfectly clear," etc. Basically "there is no problem, just your mind."

In revisiting the very first response in the original discussion which you feel you are "consistently repeating," when I said the docs could mention that exiting is necessary, this was the response:

> Me: Request: could the docs be updated to clarify that a logout/startx cycle appears to be needed? I refreshed i3 after adding the above, restarted i3, and saw no changes. After exiting and running startx again, the change was indeed applied.

> Airblader: That wouldn't be accurate. :-) i3 reload should suffice – however, this only applies to when a workspace is created, not retroactively to existing workspaces.

In other words, it was accurate all along, as `reload` _doesn't_ suffice; everything after the "however..." precisely covers this [fringe] situation. I'd also explained that ws1 and ws2 were on respective monitors and this was about swapping their location, so it was clear that this involved moving existing workspaces.

Again, I doubt anything can be of much help at this point, but it confuses me to try and understand why this went sideways. I don't want to upset anyone and tried to maintain polite enthusiasm toward trying to be helpful in some way. I also felt beaten down more and more by these patterns and gave way to argumentative responses. I wish this had gone differently, and apologize for my part in wasting your time and frustrating you.

Thanks and good luck on the project, I really enjoy using i3.

Airblader:
I apologize in advance if this sounds a bit harsh, but you already closed the PR and deleted your branch. I think this has been discussed in all necessary detail, or at the very least to the amount of time invested I'm willing to invest. There are no hurt feelings on my side either way, and I'm glad you enjoy using i3!

jwhendy:
Not harsh at all, and none is meant in my reasoning: it didn't seem I'd gotten it right, and we spent more time discussing revisions to my attempt than it would have taken for either of you to write the sentence that made the most sense to you.

Thanks for the kind response and I will continue to enjoy i3 indeed :)

PR #4843, opened 2022-02-07
Use proper WM_Sx registration for i3

errge:
i3's WM ownership is registered with the X Atom WM_S_S0 (for screen0),
instead of the correct WM_S0.  The relevant xcb-util API didn't
change, this is simply a bug in i3 that has always been there:
https://gitlab.freedesktop.org/xorg/lib/libxcb-util/-/blob/0.4.0/src/atoms.c#L65

### DIFF
--- a/release-notes/bugfixes/5-fix-wm-registration
+++ b/release-notes/bugfixes/5-fix-wm-registration
@@ -1,0 +1,1 @@
+changed WM registration selection from WM_S_S<screen> to WM_S<screen>
--- a/src/main.c
+++ b/src/main.c
@@ -696,11 +696,11 @@ int main(int argc, char *argv[]) {
     /* Acquire the WM_Sn selection. */
     {
         /* Get the WM_Sn atom */
-        char *atom_name = xcb_atom_name_by_screen("WM_S", conn_screen);
+        char *atom_name = xcb_atom_name_by_screen("WM", conn_screen);
         wm_sn_selection_owner = xcb_generate_id(conn);
 
         if (atom_name == NULL) {
-            ELOG("xcb_atom_name_by_screen(\"WM_S\", %d) failed, exiting\n", conn_screen);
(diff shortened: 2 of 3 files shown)


Airblader (commented):
Thank you for the PR! The change certainly looks reasonable to me, but looping @psychon as the author of the original change introducing this just in case.

psychon:
https://github.com/awesomeWM/awesome/blob/2d6244f9445ffa22c8d53b3e4b4641e0e5bb6170/awesome.c#L288 :see_no_evil: 

The change looks reasonable to me. The link to the implementation in atoms.c is all the convincing that I need. But since @Airblader asked for more (at least I feel like it): https://tronche.com/gui/x/icccm/sec-4.html#s-4.3 definitely says:

> For each screen they manage, window managers will acquire ownership of a selection named WM_Sn, where n is the screen number,

psychon (approved):

errge:
@psychon interesting that this was already broken in awesome, I didn't think of you simply getting the code from there and checking what's up with them.  I don't personally use awesome, so I have little interest in fixing that myself, but thank you for pinging them with the bug report!

You are right about referencing ICCCM: the docs definitely say WM_Sx.  One more data point as a proof: I originally started debugging this, because xfce4-panel had a startup delay with a warning saying "waiting for WM", and the debugging led me here.  Once I fixed the i3, the xfce4-panel delay went away, so I'm quite sure, WM_Sx is the correct one.

To be fair, not registering your WM, doesn't cause too much harm, the worst I saw was startup delay in different parts of the desktop, but once booted, everything worked correctly.  This might be the reason, why people don't notice this bug neither in awesome, nor in i3.

Airblader:
> But since @Airblader asked for more (at least I feel like it)

I was actually just worried that this seemed so trivial that I must be missing something. :-D Thanks for confirming!

Airblader (approved):

PR #4943, opened 2022-04-08
Float WM_WINDOW_TYPE "_KDE_NEXT_WM_WINDOW_TYPE_OVERRIDE" for bad software

berbiche:
Zoom (a very popular videoconferencing app) shows popups messages like "X started screensharing" etc. without using any EWMH hints, WM_TRANSIENT or "override-redirect" behavior.
The only distinguishing property set on Zoom's popup windows is `_NET_WM_WINDOW_TYPE` with a proprietary `_KDE_NET_WM_WINDOW_TYPE_OVERRIDE` set.
Without this patch, Zoom popups show up as regular windows.
These windows can be incredibly overwhelming at times, opening 3 or more windows in an instant, disturbing work.

This PR resolves this issue for all people stuck with Zoom (probably a lot of school students).

The alternative to this simple patch is to use a `for_window` rule that matches _all_ Zoom windows.

This reddit post has additional context: https://old.reddit.com/r/i3wm/comments/mcwpur/how_do_i_make_zoom_notification_popups_floating/

edit: tested on release 4.20.1 + this patch, on NixOS 22.05.


Airblader:
I'm a Zoom user myself, and I very much understand the frustration with Zoom, in particular with recent 4.10 updates which made the situation even worse.

However, i3 is not here to fix broken closed source clients. The fact that Zoom offers no useful information on its windows is something Zoom must address, not i3. I don't see an argument for why this atom should _generally_ be used as a marker to float windows, and in fact the specs also state that applications must always include a fallback window type to one of the types in the spec. Zoom failing to do so is yet another one in the long list of its issues.

Please voice this, as I do, to Zoom, so that they can fix their software for all window managers to understand.

berbiche:
Thanks for the fast response.

I'll voice my concerns to Zoom and keep this patch locally until they fix their stuff.

PR #3949, opened 2020-02-19
Sanitize window dimensions and handle decoration overlap

xzfc:
This branch is based on #3948, so it should be merged after. Although it is ready for review (check last two commits).

### To reproduce the bug
1. Disable compositor.
2. Open a few windows in split mode.
3. Shrink all of them except the last one. It should look [like this](https://user-images.githubusercontent.com/5121426/74793594-b1e11600-52b8-11ea-915f-83b20b20a4dd.png) (this is still a split layout, not stack).

#### Old behavior
[screenshot](https://user-images.githubusercontent.com/5121426/74794580-801d7e80-52bb-11ea-87b9-a9ebb791e06b.png)
* Glitched window decorations (both rendering and mouse event handling).
* Some apps (like gnome-terminal) may become unresponsive.

#### New behavior
[screenshot](https://user-images.githubusercontent.com/5121426/74793595-b4437000-52b8-11ea-81a3-b7f0e5bbb3b4.png)

Issue #3544, opened 2018-12-05
Ensure *each client window* have a size of at least 1px after resize (x11 borders)


xzfc:
## I'm submitting a…
<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

This is partially fixed by #3389, but the bug still persists.

## Reproduction Instructions

```
font pango:sans
bar { }
floating_modifier Mod1
bindsym Mod1+1 workspace 1
exec --no-startup-id xterm -bw 100
exec --no-startup-id xterm -bw 100
```

Resize the left xterm wide enough using Mod1 + rightclick.

## Current Behavior

* The right xterm segfaults.
* Corresponding line in the i3 log:
   `05/12/18 10:55:33 - x.c:x_push_node:905 - setting window rect (2, 0, -8, 237)`
   Note the negative width.
* `xwininfo` of the right xterm before it segfaults: `-geometry 65528x237-0+20`.

[i3 log](http://dump.bitcheese.net/files/fafozas/log.txt.gz)

## Causes

As far as I can see, the following things can cause non-positive window size:

1. X11 borders (`xterm -bw 100`, described above)
2. container borders (`i3-msg border normal 100px`)
3. container titlebars
4. split layouts. Consider following example: `H1[w H2[w w w]]`.
   `H2` can not be narrower than 3 pixels as each of its children should be at least 1 pixel wide.
   See how `bspwm` handles this.

i3 version 4.16

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

PR #4955, opened 2022-04-22
1412 bspwm resize

cmprmsd:
Trying to rebase and fix @xzfc pull request https://github.com/i3/i3/pull/3545

original message:
bspwm-like tiling resize

* **Two-axis tiling resize**.  You can drag a 24-pixel window corner (purple) to perform a two-axis resize. Also, you can press Mod and drag a large corner inside the window (one-third of the window size).
  illustration

* **Live resize**: the tree is rendered during the resize process.

* You can still press **ESC to cancel** resizing and revert to the old state.

* Introduced the concept of the **minimum container size**.  E.g. the minimum width of leaf con with 10px borders is 21px (left border + right border + at least 1px for the window itself).  The minimum size is defined recursively, e.g. the minimum height of the vsplit is the sum of minimum sizes of its children.  The reason for this change: the live resize made it very easy to shrink containers as small as possible, so some restraints are required.

* ~Misc update: ignore the drag if i3 can't perform the resize.  Suppose the user tries to resize the rightmost window to the right using Mod+RMB drag.  Obliviously, i3 can't do this.  Before this PR, the drag has been propagated to the application (some apps would show a context menu).  After this PR, the drag would be ignored at all (and the cursor would be shown as ✕).  The reasoning behind this change: the user explicitly indicated their desire to resize the window by pressing the Mod key, and probably just miss clicked to the wrong part of the window.~ Done in [93e96f4](https://github.com/i3/i3/commit/93e96f4e6b28486c6b7be003e8db81bab30197ef).

* Bugfix in `tiling_resize()`: you can now resize the window with pixel border by dragging the top border.


src/render.c
```diff
@@ -131,6 +139,33 @@ void render_con(Con *con) {
                  child->rect.x, child->rect.y, child->rect.width, child->rect.height);
             x_raise_con(child);
             render_con(child);
+
+            /* Accumulate minimum size of the con based on child->min_size.
+             * We may use child->min_size only after calling render_con(child).
+             *
+             * TODO: this isn't perfect as it fails to handle some corner cases,
```
orestisfl, line 146: Can you please investigate if the discussion in https://github.com/i3/i3/pull/3545#discussion_r406882625 applies?

I'd rather not introduce less-than-optimal behaviour.

src/render.c
```diff
@@ -131,6 +139,33 @@ void render_con(Con *con) {
                  child->rect.x, child->rect.y, child->rect.width, child->rect.height);
             x_raise_con(child);
             render_con(child);
+
+            /* Accumulate minimum size of the con based on child->min_size.
```
orestisfl, line 143: If fixing https://github.com/i3/i3/issues/3544 is needed for this feature (ref: https://github.com/i3/i3/pull/3545#issuecomment-444440915) I would also prefer to extract the necessary bug fixes in a separate PR.

cmprmsd:
Thanks for your comments! I'll have to look into them. I clearly am no developer but trying my best to understand how this could work out. 😅 

Contributors are welcome. I guess that'll would mean inviting some people to this fork if there is anyone willing to try. :)

src/render.c
cmprmsd: Thanks for your review!
These corner cases are not unique and new to this pr. I tried them with a upstream i3 on v 4.19.2 and there you can resize the bottom window even through the stacked titles. 
image

also upstream with titlebar for the bottom tile you will see it disappear:
image

So these are not issues created by this pull request.
xzfc comment pointed to this edge case as well:
image

but as I verified it not being caused by this implementation it's fine for now. 
Maybe there's some better way in the future to calculate the minimum size in the future.

The way the minimum size is calculated in this version does stop the bottom tile from moving more than one decoration through the stacked top ones. So it's at least a little improvement. 

I work with this forked version for some days now and did not find any problems with it yet.

src/render.c
cmprmsd: #3544 was fixed by @xzfc and has been merged upstream two years ago. I think this has already been resolved in https://github.com/i3/i3/pull/3949 :)

cmprmsd: I think the comment was before this PR by the same person.

cmprmsd:
I was waiting on feedback what has to be done to get this merged. I think I replied to everything. 
Help 😅

cmprmsd:
Btw, I'm using this forked branch on 4.14 since a year now in virtual machines without any issues. 
Shall I rebase on the latest master branch to get this merged?

orestisfl:
Apologies for closing.

This branch does seem to have significant conflicts with `next` now though.

algmyr:
@cmprmsd Do you intend to continue the work here? I can share a branch with resolved conficts wrt current head if that helps. I could technically pick up work as well, but I don't know the changes well (other than what I had to look at to resolve conflicts).

https://github.com/algmyr/i3/tree/bspwm_resize

(also squashed some of the fixups while poking at the code anyway, presumably the second commit shouldn't be merged into here)

I left a thing to verify in https://github.com/algmyr/i3/blob/bspwm_resize/src/resize.c#L240 since some code that contained later bug fixes ended up being removed during this change.

FWIW, on top of being a neat feature, this also fixes some really bad mouse resize lag I experienced. :)

cmprmsd:
Hi @algmyr, 
Nice stuff, I was also just following up @xzfc work. I followed up on the issues @orestisfl mentioned but it never got past the current state. Feel free to take over. I'm using my outdated fork since than as I really like the feature. Will try your wip asap. 😬

PR #5007, opened 2022-06-11
Fix segfault with mode "default" key bindings

### LINKED ISSUE #5006: SIGSEGV in configure_binding with mode "default" bindsym
## I'm submitting a…
<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
Config with
<pre>mode "default" {
        bindsym X resize
}
</pre>
causes segfault

## Expected Behavior
It doesn't crash.

## Reproduction Instructions
Use the following config.

## Environment
<pre>
i3 version: 4.20-96-gce2665ca © 2009 Michael Stapelberg and contributors
</pre>

<details><summary>Config file</summary>
<pre>
# i3 config file (v4)
mode "default" {
        bindsym X resize
}
</pre>
</details>

<pre>
Logfile URL: https://logs.i3wm.org/logs/6264921952092160.bz2
Strace URL: https://github.com/i3/i3/files/8882278/sigsegv-strace.txt
Backtrace URL: https://github.com/i3/i3/files/8882279/sigsegv-backtrace.txt
</pre>

<pre>
</pre>

zhrvn:
Fixes #5006

### DIFF
--- a/src/config_directives.c
+++ b/src/config_directives.c
@@ -192,16 +192,14 @@ CFGFUN(mode_binding, const char *bindtype, const char *modifiers, const char *ke
 CFGFUN(enter_mode, const char *pango_markup, const char *modename) {
     if (strcmp(modename, DEFAULT_BINDING_MODE) == 0) {
         ELOG("You cannot use the name %s for your mode\n", DEFAULT_BINDING_MODE);
-        return;
-    }
-
-    struct Mode *mode;
-    SLIST_FOREACH (mode, &modes, modes) {
-        if (strcmp(mode->name, modename) == 0) {
-            ELOG("The binding mode with name \"%s\" is defined at least twice.\n", modename);
+    } else {
+        struct Mode *mode;
+        SLIST_FOREACH (mode, &modes, modes) {
+            if (strcmp(mode->name, modename) == 0) {
+                ELOG("The binding mode with name \"%s\" is defined at least twice.\n", modename);
+            }
         }
     }
-
     DLOG("\t now in mode %s\n", modename);
     FREE(current_mode);
     current_mode = sstrdup(modename);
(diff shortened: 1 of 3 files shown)


Airblader:
Thanks for the PR (and issue). However, fixing this by just removing the restriction is certainly not what we want to do. Configuring the default mode explicitly is intentionally disallowed, and it should remain that way.

zhrvn:
Thank you for your comment, but I don't really understand why should the explicit default mode be disallowed and what makes it different from any other mode that is defined more than once in config? Looking at the history it seems like this check is no longer needed after #2633. Can you clarify what should the parser do it this case?

Airblader:
The default binding mode is configured outside any `mode` block. It is not intended to explicitly open such a block for it, and allowing it would just raise question on the semantic differences.

zhrvn:
> The default binding mode is configured outside any `mode` block. It is not intended to explicitly open such a block for it, and allowing it would just raise question on the semantic differences.

Yes, I think you're right. `The binding mode with name default is defined at least twice.` can be confusing to the user, because it is implicitly defined, so we should leave the previous error message `You cannot use the name default for your mode`.

src/config_directives.c
```diff
@@ -192,16 +192,14 @@ CFGFUN(mode_binding, const char *bindtype, const char *modifiers, const char *ke
 CFGFUN(enter_mode, const char *pango_markup, const char *modename) {
     if (strcmp(modename, DEFAULT_BINDING_MODE) == 0) {
         ELOG("You cannot use the name %s for your mode\n", DEFAULT_BINDING_MODE);
-        return;
```
stapelberg: This still isn’t quite the right fix, I think.

Your PR changes it so that i3 logs an error, but then goes on to accept the key bindings (instead of crashing).

I think `enter_mode` should stay as-is, but `mode_binding` should return when `current_mode` is `NULL`, effectively ignoring the key bindings configured for the default mode.

stapelberg: I have updated the PR to ignore bindings when in an invalid mode.

stapelberg (approved):

PR #5035, opened 2022-06-28
Update per-workspace-layout.pl

### LINKED ISSUE #4186: per-workspace-layout.pl not working
## I'm submitting a…

<pre>
[ x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

On launching multiple windows in workspace 4 they split the workspace.

## Expected Behavior

The windows should get tabbed instead of splitting

## Reproduction Instructions

I downloaded the perl file and added the `execalways ~/per-workspace-layout.pl` to the `.i3/config`. I also ran it manually on the terminal and fixed module can't be located errors that were present. It runs without any such errors but the workspace isn't getting tabbed.

## Environment

<pre>
i3 version: 4.18.2
</pre>
I am using Manjaro i3

<pre>
Config URL: https://pastebin.com/pJVj403u
</pre>

<pre>
Logfile URL:
</pre>

<pre>
</pre>
I am using picom

joe-batt:
This changes the behaviour from  "change layout on focus of an empty workspace" to "change layout on first focus after init of an workspace".
The original behaviour works when manually moving to a workspace and opening an app, but it will not work when moving an window from a another workspace or when creating an window an a specfic workspace using a `for_window […] move container to workspace $ws` in the config.

So far it seems to work without any trouble. That said I never wrote a line of perl in my life before, so there might be a better way to do this.

This probably closes #4186

### DIFF
--- a/contrib/per-workspace-layout.pl
+++ b/contrib/per-workspace-layout.pl
@@ -21,6 +21,8 @@ my %layouts = (
     '5' => 'stacked',
 );
 
+my %new_ws;
+
 my $i3 = i3();
 
 die "Could not connect to i3: $!" unless $i3->connect->recv();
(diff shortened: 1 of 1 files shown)


orestisfl:
Side-effect of https://github.com/i3/i3/issues/3001, let's not include the workaround in the script but fix the original issue.

PR #5060, opened 2022-07-28
Fix Wbitwise-instead-of-logical warnings

orestisfl:
> error: use of bitwise '|' with boolean operands

### DIFF
--- a/src/randr.c
+++ b/src/randr.c
@@ -666,9 +666,9 @@ static bool randr_query_outputs_15(void) {
         new->primary = monitor_info->primary;
 
         new->changed =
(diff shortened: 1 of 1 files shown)


orestisfl:
@stapelberg hmm, spellchecking pipeline seems to fail

stapelberg:
Indeed. Sent https://github.com/i3/i3/pull/5061 to fix it

stapelberg:
This one is merged as https://github.com/i3/i3/commit/807e972330b011de6afd227283cd49ebcf0ce1e7 now

PR #5061, opened 2022-07-28
fix travis/check-spelling.pl for recent Lintian changes

stapelberg:


orestisfl (approved):
You can temporarily disable the pipeline and we can merge the other PR with the fix. Sucks to have two different fail reasons 😛

orestisfl:
Or just manually merge it

stapelberg:
I have cherry-picked your commit into this PR. Let’s see if we can get it merged easily this way :)

stapelberg:
I’m glad I didn’t just disable the pipeline, because I had to update your fix :)

Just changing from `|` to `||` isn’t enough, because that will change the short-circuit behavior. With `|`, all expressions are run, but with `||`, evaluation will abort after the first `true` result.

orestisfl:
Oops, you are right, should have paid more attention

Another way to work around this is to force the result into ints

PR #5083, opened 2022-08-12
Add new "mouse_warping" setting: remembered.

### LINKED ISSUE #1146: Remember mouse position when changing outputs
[Originally reported by pie3mn@…]
(Suggesting an option that when set remembers the position of the mouse cursor upon changing screens for each particular screen.

Currently if you jump to workspaces on different screens it will just center the cursor in the middle of the last active window on that screen.

Example: Two screens (A&B), You have some IM on A, some text on B, you want to skim through the text and copy&paste certain passages instantly into your IM, as you find them without having to readjust your mouse position every time you change screen.

This may sound more trivial on a small screen, but on a 1920x1080 screen (and resolutions increasing in the future), this will more and more become a usability problem when using the mouse.

### LINKED ISSUE #4981: add mouse_warping setting to warp to previous location on target output
## I'm submitting a…
<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
The two choices for mouse_warping are "none" and "output".  The former does not warp the mouse at all, the latter warps it to the center of the output.

## Desired Behavior
I would like a third option: to warp it to the location it had when it was last on that output.  Example: I have two outputs, one with documentation and one with code.  Both documentation and code live in the center of the output, so I move the mouse out of the way to see it clearly.  I use the keyboard to focus the other output and then move it back to the previous output.  In the current behavior, the mouse is either not warped at all or warped to the center, obstructing my view and forcing me to move the mouse again.  This gets worse if I'm making mouse selections on either output: I can't leave the mouse where I want to return to.

(Note that this is a revival of the long-since closed Issue #1146, in which the feature was welcome but never appeared.)

## Impact
<pre>
[X] This feature requires new configuration and/or commands
</pre>

At minimum: one new config parameter, two fields added to a struct, and a new if-statement.  I am happy to develop and submit a patch.  I have looked at the code and I have ideas how to make it happen, but (depending on the tolerance for ambition here) it might disrupt some stuff.  If you're amenable to the feature idea, I'll post my thoughts about a couple ways to make it happen, so that those with better understanding of the code can decide which (if any) direction makes the most sense.

## Environment
<pre>
i3 version: Binary i3 version:  4.20.1 (2021-11-03) © 2009 Michael Stapelberg and contributors
</pre>

<pre>
</pre>

pete-:
When the "mouse_warping" configuration is set to "remembered" and the user
switches focus to a different output, the mouse will be warped to the location
it had when last on that particular output.

This entails defining a new setting for the "mouse_warping" configuration
parameter; adding "remembered_x" and "remembered_y" fields to struct xoutput;
initializing those fields for the fake_output, xinerama, and xrandr cases; and
honoring the remembered values when the warping the mouse while the config
directive is active.

Most of those additions are straightforward; the most involved is the latter,
which comprises the changes in src/x.c.

The code that initializes the remembered_x and remembered_y fields lives in
three files (src/fake_outputs.c, src/xinerama.c, and two places in
src/xrandr.c).  I'm pretty sure I've covered all the functions where they need
to be initialized, but I'm not sure I've put it in the best place *within*
those functions.

I also noticed two potentially-uninuitive behaviors while this config setting
is active:

- When rapidly changing outputs (eg, by pounding "workspace_back_and_forth"),
  eventually the remembered x and y values are forgotten.  My guess is this
  due to how Xorg/xcb process events and may be out of the purview of i3, but
  I don't know enough to say for sure.

- Consider the case where I have two workspaces, 1 and 2, on different
  outputs; workspace 1 has two windows, A and B; the mouse is sitting on
  Window A but focus is on Window B.  Switch to workspace 2.  Switch back to
  Workspace 1.  The result is that the pointer is moved to its remembered
  location on Window A and Window B is focused, but then focus is immediately
  shifted to Window A.

  I'm not sure what the "correct" behavior ought to be in this case.  I can
  see an argument that the behavior described above is actually desired or
  resaonable.  I can also see how one would want Window B to remain focused,
  even though the pointer should lie within Window A after warping.  The
  latter might require adding a "which container was focused" field to the
  data structure representing each workspace.

I would greatly appreciate input/feedback on all aspects of this work.

Closes #1146
Closes #4981


asdf8dfafjk:
Hi, this work is a great idea and something much needed for people who constantly switch back and forth. I can't remember how many times I've neede it. 

One quick suggestion  , in your original comment, you should consider putting the conceptual part above the implementation part (window a window b stuff above remembered_x) , hopefully more people would be likely to comment. 

So i think window focus and mouse focus should both be remembered in your last example. That would be the ideal scenario IMHO. If window changes focus personally I'm not likely to use it

orestisfl (changes requested):
Good work overall :+1:. Sorry for the late response.

I agree that focus should be preserved (mouse should be able to be above non-focused window). This is usually done with `x_mask_event_mask(~XCB_EVENT_MASK_ENTER_WINDOW);` in other places.

Also, some debugging & understanding regarding the mouse position being lost after rapidly changing outputs would be appreciated.

pete-:
After two years of inaction, I have been able to make time to continue this.  And it seems like procrastination has paid off once again, because after rebasing to next/cfa4cf16 and performing a trivial merge, both the focus-preservation and mouse position lost problems have resolved themselves.

I'm not used to using github for development, so some guidance about next steps would be appreciated.  Is it just a matter of committing my rebased changes and pushing to github, which will automatically update this pull request with the HEAD commit of my dev branch?  And then a re-review?

Thanks.

stapelberg:
> I'm not used to using github for development, so some guidance about next steps would be appreciated. Is it just a matter of committing my rebased changes and pushing to github, which will automatically update this pull request with the HEAD commit of my dev branch? And then a re-review?

Yes, you would (force-)push to your dev branch and that updates the PR.

PR #5088, opened 2022-08-18
fix: move position cmd mode

ShimmerGlass:
The move position cmd was always behaving as absolute, even when the
absolute keyword wasn't used. We now make the coordinates relative to
the output when the absolute keyword is not set.


orestisfl:
This doesn't have enough context/underlying linked issue and is backwards-incompatible.

Discussion needs to be had in an issue first.

PR #5111, opened 2022-09-05
Sanitise LESS environment variable in i3-sensible-pager.

### LINKED ISSUE #5039: [PATCH] sanitise LESS environment variable in i3-sensible-pager
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
When an error is encountered such as "The configured command for this shortcut could not be run successfully", the "show errors" button on i3-nagbar doesn't work if $PAGER is less, and $LESS contains either the -E or -F flag (the window pops up, but immediately disappears).

## Expected Behavior
The window should stay visible.

## Reproduction Instructions
Add an invalid keybinding, say `bindsym` $mod+x invalid action`, then run i3 with the less environment containing "E" and/or "F", then input that key.  Click on "show errors".

```
LESS=FRXe /usr/bin/i3
```

## Environment
<pre>
i3 version: 
Running i3 version: 4.20.1 (2021-11-03) (pid 102426)
Loaded i3 config:
  /usr/local/google/home/bodea/.config/i3/config (main) (last modified: Thu 30 Jun 2022 18:06:45, 15 seconds ago)

The i3 binary you are running: /usr/bin/i3
</pre>

## Patch
<pre>
diff --git a/i3-sensible-pager b/i3-sensible-pager
index 386e2988..7e19d4cc 100755
--- a/i3-sensible-pager
+++ b/i3-sensible-pager
@@ -8,6 +8,14 @@
 # Distributions/packagers can enhance this script with a
 # distribution-specific mechanism to find the preferred pager.
 
+# The less -E and -F options exit immediately for short files, strip if present.
+while [[ $LESS = *[EF]* ]]; do
+  case "$LESS" in
+  *E*) LESS="${LESS%E*}${LESS##*E}";;
+  *F*) LESS="${LESS%F*}${LESS##*F}";;
+  esac
+done
+
 # Hopefully one of these is installed (no flamewars about preference please!):
 # We don't use 'more' because it will exit if the file is too short.
 # Worst case scenario we'll open the file in your editor.
</pre>

bodea:
When an error is encountered such as "The configured command for this shortcut
could not be run successfully", the "show errors" button on i3-nagbar doesn't
work if $PAGER is less, and $LESS contains either the -E or -F flag (the
window pops up, but immediately disappears).

Strip these flags from the LESS environment variable before invoking $pager.

Fixes: #5039

### DIFF
--- a/i3-sensible-pager
+++ b/i3-sensible-pager
@@ -8,6 +8,11 @@
 # Distributions/packagers can enhance this script with a
 # distribution-specific mechanism to find the preferred pager.
 
+# The less -E and -F options exit immediately for short files, strip if present.
+case "$LESS" in
+  *[EF]*) LESS=`echo "$LESS" | tr -d EF`
+esac
+
 # Hopefully one of these is installed (no flamewars about preference please!):
 # We don't use 'more' because it will exit if the file is too short.
 # Worst case scenario we'll open the file in your editor.


i3-sensible-pager
```diff
@@ -8,6 +8,14 @@
 # Distributions/packagers can enhance this script with a
 # distribution-specific mechanism to find the preferred pager.
 
+# The less -E and -F options exit immediately for short files, strip if present.
+while [[ $LESS = *[EF]* ]]; do
```
stapelberg: `[[` is a bashism, but I think standard shell does not support what you’re doing here.

It might be simplest to change the shebang at the top to #!/bin/bash, unless you know of a way to make this portable?

stapelberg (commented):
Thanks for your PR!

i3-sensible-pager
bodea: Sorry, I was *sure* that was included by now, but I see that SUSv3 says that it is "reserved, but unspecified".  Confirmed not to work with Debian's dash.  I've pushed a change to require bash.  If you would prefer to retain /bin/sh, I can amend to use the following, which is portable:

```
case "$LESS" in
  *[EF]*) LESS=`echo "$LESS" | tr -d EF`
esac
```

stapelberg: The portable variant doesn’t look too bad. Would you mind changing the PR to use the portable version? That should make things a little easier for the Debian packaging, for example. Thanks.

bodea: Done

stapelberg (approved):

PR #5112, opened 2022-09-06
Replaced the quadratic LCA algorithm with a linear one

kolayne:


stapelberg:
Thanks for your PR! Can you outline what the motivation for this change is? Did the move algorithm turn out to be too slow in practice somehow?

kolayne:
Hello! No, not really. In fact, back in April I was just going through the i3's source code, saw this function and rewrote it to check if there was a reason for it to be quadratic, and then forgot about that. A few days ago I've found that patch and created the PR in case it could be useful.

stapelberg:
Ah, thanks for the context! In that case, I think it’s best if I respectfully decline this PR.

Since the current implementation works, and there’s no strong need to replace it, I think it’s better to stick with it. It seems a little easier to understand to me.

PR #5117, opened 2022-09-09
config_parser: strip trailing whitespace in strings

### LINKED ISSUE #5064: Blank Space in config Causes Nonfunctionality of Bar Output
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
In the bar block in the config file; if you leave a space after the \<output\> in:
```
bar{
        output <output>
}
```
...It causes the bar not to show up at all.

## Expected Behavior
Expected behavior is that a single space shouldn't break the bar from appearing on the indicated monitor output.

## Reproduction Instructions
Add a space after the \<output\> to make the bar disappear

## Environment
FreeBSD 13.1-RELEASE
Running i3 version: 4.20.1 (2021-11-03) (pid 37001)
Loaded i3 config:
  ~/.config/i3/config (main) (last modified: Mon Aug  1 21:01:41 2022, 1896 seconds ago)

Are you using a compositor (e.g., xcompmgr or compton): **No**

stapelberg:
fixes https://github.com/i3/i3/issues/5064

### DIFF
--- a/parser-specs/config.spec
+++ b/parser-specs/config.spec
@@ -565,7 +565,7 @@ state BAR_POSITION:
       -> call cfg_bar_position($position); BAR
 
 state BAR_OUTPUT:
-  output = string
+  output = word
       -> call cfg_bar_output($output); BAR
 
 state BAR_TRAY_OUTPUT:
--- a/testcases/t/317-bar-output-trailing-space.t
+++ b/testcases/t/317-bar-output-trailing-space.t
@@ -1,0 +1,71 @@
+#!perl
+# vim:ts=4:sw=4:expandtab
+#
+# Please read the following documents before working on tests:
+# • https://build.i3wm.org/docs/testsuite.html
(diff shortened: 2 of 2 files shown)


Airblader (commented):
From the parser-spec:

```
state BAR_OUTPUT:
  output = string
      -> call cfg_bar_output($output); BAR

state BAR_TRAY_OUTPUT:
  output = word
      -> call cfg_bar_tray_output($output); BAR
```

What's the significance of "string" vs. something like "word" here? Is the issue rather that we should use "word" in the spec?

stapelberg:
> What's the significance of "string" vs. something like "word" here? Is the issue rather that we should use "word" in the spec?

That is a more minimal fix, thank you for the suggestion!

Done.

Airblader (approved):

PR #5129, opened 2022-09-17
tiling_drag: con_rect_plus_deco_height: Fix underflow

### LINKED ISSUE #5069: tiling drag: cannot drop onto single-window stacked containers
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

I have recorded a video to illustrate the issue: https://youtu.be/uX5BJbFug7A

In the video, I have a stacked container with 2 windows, then a splith container, and then a stacked container with just one window. 

1. First, I drag the splith container. You can see that the drop indicators don’t show when the mouse is over the stacked container with just one window. They do show when the mouse is over the stacked container with 2 windows.
2. When I drag the stacked-1-window, I see correct indicators over both splith and stacked-2-window.
3. When I drag the stacked-2-window, I also don’t see drop indicators over the stacked-1 container.

cc @orestisfl 

## Expected Behavior

I should be able to drop on a stacked container regardless of how many windows it has :)

## Environment

<pre>
i3 version: 4.20-102-gebcd1d43
</pre>

orestisfl:
Fixes #5069

This looks a bit hackish as a fix for the issue, maybe we should investigate for a better solution. But still, this improves on the previous situation and the check itself would stay after a better fix.

### DIFF
--- a/src/tiling_drag.c
+++ b/src/tiling_drag.c
@@ -19,7 +19,11 @@ static xcb_window_t create_drop_indicator(Rect rect);
 static Rect con_rect_plus_deco_height(Con *con) {
     Rect rect = con->rect;
     rect.height += con->deco_rect.height;
-    rect.y -= con->deco_rect.height;
+    if (rect.y < con->deco_rect.height) {
+        rect.y = 0;
+    } else {
+        rect.y -= con->deco_rect.height;
+    }
     return rect;
 }
 


stapelberg (approved):

stapelberg:
Thanks!

Issue #3678, opened 2019-04-05
Motif normal border style hints are ignored


shachaf:
## I'm submitting a…
<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
When a window requests (using _MOTIF_WM_HINTS) to be drawn without borders, or with pixel borders, and then requests to be drawn with normal borders, the second request is ignored.

## Expected Behavior
If the first request is respected, the second request should also be as well.

This issue was introduced in https://github.com/i3/i3/pull/2386 in response to https://github.com/i3/i3/issues/2385: A window had a "draw borders normally" motif hint, and it overrode the i3 configuration requesting a pixel border.

If the decision is that i3 configuration should take precedence in the case of BS_NORMAL hints, then BS_NORMAL should be taken to mean "what the border style would have been with no hints", not "what the border style currently is"

Alternatively there could be an option for border settings that take precedence over window hints. Or hints could be ignored entirely. But ignoring only hints requesting a normal border isn't a good solution.

## Reproduction Instructions
```c
#include <SDL2/SDL.h>
int main() {
  SDL_Init(SDL_INIT_VIDEO);
  SDL_Window *window = SDL_CreateWindow("test", 0, 0, 200, 200, 0);
  SDL_Event e;
  while (SDL_WaitEvent(&e) && e.type != SDL_QUIT) {
    if (e.type == SDL_KEYDOWN) {
      if (e.key.keysym.sym == 'a') SDL_SetWindowBordered(window, 0);
      if (e.key.keysym.sym == 'b') SDL_SetWindowBordered(window, 1);
    }
  }
}
```

Press 'a' and then 'b'; the window remains borderless. In Metacity (whose implementation window_update_motif_hints is designed to mimic), this is handled correctly.

## Environment
I'm using i3 4.14.1, but this issue still exists in master: https://github.com/i3/i3/blob/next/src/handlers.c#L1187

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
Sorry, we can only support the latest major version. Please upgrade from 4.14 to 4.16, verify the bug still exists, and re-open this issue.

shachaf:
This bug is still present in master -- I linked to the file and line. How do I reopen this issue?

[...]

PR #5135, opened 2022-09-18
motif: restore `BS_NORMAL` correctly

0x08088405:
related: #3678

It appears `_MOTIF_WM_HINTS` works for updating the decorations, but refuses to change back to `BS_NORMAL`, since that same value is also used to indicate that the hints were not found, making i3 think nothing changed at all. Well, it actually does work if you just set it to `MWM_DECOR_BORDER` as that's interpreted as `BS_PIXEL`, so this is extra weird. This causes weird behaviour such as godotengine/godot/issues/40037 where turning the borders off works, but not back on. I made it return a `bool` instead (not sure about the conventions, first time contributing, but I hope it's clear enough), also replaced the claim about the nonexistent documentation on Motif hints with a link to documentation on it.

Edit: I see the part about the bugfixes folder in CONTRIBUTING.md, should I add something there?

### DIFF
--- a/include/window.h
+++ b/include/window.h
@@ -94,7 +94,7 @@ void window_update_hints(i3Window *win, xcb_get_property_reply_t *prop, bool *ur
  * it is still in use by popular widget toolkits such as GTK+ and Java AWT.
  *
  */
-void window_update_motif_hints(i3Window *win, xcb_get_property_reply_t *prop, border_style_t *motif_border_style);
+bool window_update_motif_hints(i3Window *win, xcb_get_property_reply_t *prop, border_style_t *motif_border_style);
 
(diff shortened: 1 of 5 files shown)


orestisfl:
> Edit: I see the part about the bugfixes folder in CONTRIBUTING.md, should I add something there?

Yes, `echo 'Restore BS_NORMAL _MOTIF_WM_HINTS correctly' > release-notes/bugfixes/9-bs-normal'`

orestisfl (approved):
LGTM besides the release note

0x08088405:
Done!

orestisfl (approved):

Issue #5149, opened 2022-09-22
`default_border pixel` doesn't hide title bars after upgrading to 4.21


Juneezee:
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

Title bar is shown for a single window on a workspace, even when `default_border pixel 1` and `default_floating_border pixel 1` are set.

https://user-images.githubusercontent.com/20135478/191677123-17f55f28-652d-470e-a228-57b55f5cca3c.mp4

## Expected Behavior

Before upgrading to 4.21, opening up a single window on a workspace with `default_border pixel 1` and `default_floating_border pixel 1` will not show the title bar.

## Reproduction Instructions

With `default_border pixel 1` and `default_floating_border pixel 1`, open a single window (Chrome, VSCode, etc.) on a new workspace.

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
Binary i3 version:  4.21 (2022-09-21) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.21 (2022-09-21) (pid 1160)
Loaded i3 config:
  /home/jun/.config/i3/config (main) (last modified: Thu 22 Sep 2022 14:32:17 +08, 71 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
</pre>

<details><summary>Config file</summary>

```
set $mod Mod4

default_border pixel 1
default_floating_border pixel 1
hide_edge_borders smart

font pango:DejaVu Sans Mono Nerd Font 11

floating_modifier $mod

bindsym $mod+Shift+q kill

bindsym $mod+j focus left
bindsym $mod+k focus down
bindsym $mod+l focus up
bindsym $mod+semicolon focus right

bindsym $mod+Left focus left
bindsym $mod+Down focus down
bindsym $mod+Up focus up
bindsym $mod+Right focus right

bindsym $mod+Shift+j move left
bindsym $mod+Shift+k move down
bindsym $mod+Shift+l move up
bindsym $mod+Shift+semicolon move right

bindsym $mod+Shift+Left move left
bindsym $mod+Shift+Down move down
bindsym $mod+Shift+Up move up
bindsym $mod+Shift+Right move right

bindsym $mod+h split h

bindsym $mod+v split v

bindsym $mod+f fullscreen toggle

bindsym $mod+s layout stacking
bindsym $mod+w layout tabbed
[...]

PR #5150, opened 2022-09-22
Fix motif behavior according to spec

orestisfl:
See https://linux.die.net/man/3/vendorshell
The important part is:
> MWM_DECOR_ALL
> All decorations *except* those specified by other flag bits that are set

Related (but not fixing) #5149, we can merge this independently since it's an improvement on the current situation.

CC @viriuwu

### DIFF
--- a/release-notes/bugfixes/1-motif
+++ b/release-notes/bugfixes/1-motif
@@ -1,0 +1,1 @@
+Fix compliance to _MOTIF_WM_HINTS spec when all decorations are set
--- a/src/window.c
+++ b/src/window.c
@@ -403,6 +403,49 @@ void window_update_hints(i3Window *win, xcb_get_property_reply_t *prop, bool *ur
     free(prop);
 }
 
+/* See `man VendorShell' for more info, `XmNmwmDecorations' section:
+ * https://linux.die.net/man/3/vendorshell
+ * The following constants are adapted from <Xm/MwmUtil.h>.
+ */
+#define MWM_HINTS_FLAGS_FIELD 0
+#define MWM_HINTS_DECORATIONS_FIELD 2
+
+#define MWM_HINTS_DECORATIONS (1 << 1)
+#define MWM_DECOR_ALL (1 << 0)
+#define MWM_DECOR_BORDER (1 << 1)
+#define MWM_DECOR_TITLE (1 << 3)
+
+static border_style_t border_style_from_motif_value(uint32_t value) {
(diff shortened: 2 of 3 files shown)


Airblader (approved):

stapelberg:
Can you fix the failing tests? Should this be part of the release or can we punt it to until afterwards? See https://github.com/i3/i3/issues/5209 for timeline details. Thanks :)

orestisfl:
I don't really think this needs to be part of the bugfix release, I'll try to spend some time here but it doesn't need to block the release.

stapelberg:
I noticed the tests were hanging in our CI configuration. Turns out it was due to an AddressSanitizer warning:

```
=================================================================
==1340==ERROR: AddressSanitizer: heap-use-after-free on address 0x6060000037c0 at pc 0x555d805adba0 bp 0x7ffd995ca750 sp 0x7ffd995ca748
READ of size 4 at 0x6060000037c0 thread T0
    #0 0x555d805adb9f in window_update_motif_hints ../src/window.c:482
    #1 0x555d80579ca7 in manage_window ../src/manage.c:218
    #2 0x555d805511dd in handle_map_request ../src/handlers.c:276
    #3 0x555d8055a207 in handle_event ../src/handlers.c:1463
    #4 0x555d8056e2df in xcb_prepare_cb ../src/main.c:151
    #5 0x7f44ca4cf632 in ev_invoke_pending (/lib/x86_64-linux-gnu/libev.so.4+0x5632)
    #6 0x7f44ca4d30ea in ev_run (/lib/x86_64-linux-gnu/libev.so.4+0x90ea)
    #7 0x555d8056e076 in ev_loop /usr/include/ev.h:841
    #8 0x555d80577ef0 in main ../src/main.c:1200
    #9 0x7f44c9eef209  (/lib/x86_64-linux-gnu/libc.so.6+0x29209)
    #10 0x7f44c9eef2bb in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x292bb)
    #11 0x555d804dfd40 in _start (/usr/src/i3/build/i3+0x58d40)

```

The problem was that you were freeing `prop` too early, accessing it afterwards:

```c
uint32_t *motif_hints = (uint32_t *)xcb_get_property_value(prop);
FREE(prop);
if (motif_hints[MWM_HINTS_FLAGS_FIELD] & MWM_HINTS_DECORATIONS) { // this line accesses prop
```

I fixed this and updated the PR.

Issue #5156, opened 2022-09-23
[i3-dmenu-desktop] Error with emacsclient.desktop


ghost:
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
Emacs provides `emacsclient.desktop` to launch an Emacs daemon if one is not already running and then launch an Emacs client. The file's `Exec` line is a bit complex as you can see:
```
Exec=sh -c "if [ -n \\"\\$*\\" ]; then exec emacsclient --alternate-editor= --display=\\"\\$DISPLAY\\" \\"\\$@\\"; else exec emacsclient --alternate-editor= --create-frame; fi" placeholder %F
```
Trying to launch this desktop entry with `i3-dmenu-desktop` does not work, and running `i3-dmenu-desktop` from a terminal window and selecting `Emacs (Client)` from `dmenu` I get the following error message:
```
ERROR: Your command: exec  "sh -c \"if [ -n \\\\"\\$*\\\\" ]; then exec emacsclient --alternate-editor= --display=\\\\"\\$DISPLAY\\\\" \\\\"\\$@\\\\"; else exec emacsclient --alternate-editor= --create-frame; fi\" placeholder "
ERROR:                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR: Expected one of these tokens: <end>, '[', 'move', 'exec', 'exit', 'restart', 'reload', 'shmlog', 'debuglog', 'border', 'layout', 'append_layout', 'workspace', 'focus', 'kill', 'open', 'fullscreen', 'sticky', 'split', 'floating', 'mark', 'unmark', 'resize', 'rename', 'nop', 'scratchpad', 'swap', 'title_format', 'title_window_icon', 'mode', 'bar', 'gaps'
[...]

Issue #5152, opened 2022-09-22
PhpStorm not opening from i3-dmenu-desktop


mklein994:
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior

PhpStorm doesn't open after selecting "PhpStorm" with `i3-dmenu-desktop`.

## Expected Behavior

PhpStorm should open.

## Reproduction Instructions

Either install PhpStorm ([AUR](https://aur.archlinux.org/packages/phpstorm)), or at a bare minimum create this `.desktop` entry:

`/usr/share/applications/jetbrains-phpstorm.desktop`
```desktop
[Desktop Entry]
Version=1.0
Type=Application
Name=PhpStorm
Icon=phpstorm
Exec="/opt/phpstorm/bin/phpstorm.sh" %f
Comment=Lightning-smart PHP IDE
Categories=Development;IDE;
Terminal=false
StartupWMClass=jetbrains-phpstorm
```

Run `i3-dmenu-desktop` from the terminal, and select "PhpStorm":

```
$ i3-dmenu-desktop # select "PhpStorm"
ERROR: Your command: exec  ""/opt/phpstorm/bin/phpstorm.sh\" "
ERROR:                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR: Expected one of these tokens: '--no-startup-id', <string>
[{"success":false,"parse_error":true,"error":"Expected one of these tokens: '--no-startup-id', <string>","input":"exec  \"\"/opt/phpstorm/bin/phpstorm.sh\\\" \"","errorposition":"       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^"}]
Could not launch i3-msg: 512 at /usr/bin/i3-dmenu-desktop line 494, <GEN1> line 1.
```

Relevant code:

https://github.com/i3/i3/blob/5bc4280a48b44f768a4467e12d44ba9b1bae3953/i3-dmenu-desktop#L479-L494

## Environment

I'm using the [`i3-gaps` package](https://archlinux.org/packages/community/x86_64/i3-gaps/) from the Arch community repo.

Output of `i3 --moreversion 2>&-`:
<pre>
Binary i3 version:  4.21 (2022-09-21) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.21 (2022-09-21) (pid 4695)
Loaded i3 config:
[...]

Issue #4697, opened 2021-11-26
i3-dmenu-desktop can't launch electrum


532910:
## I'm submitting a…

<pre>
[x] Bug
</pre>

## Current Behavior

```
% i3-dmenu-desktop        
ERROR: Your command: exec  "sh -c \"PATH=\\"\\$HOME/.local/bin:\\$PATH\\"; electrum \""
ERROR:                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR: Expected one of these tokens: <end>, '[', 'move', 'exec', 'exit', 'restart', 'reload', 'shmlog', 'debuglog', 'border', 'layout', 'append_layout', 'workspace', 'focus', 'kill', 'open', 'fullscreen', 'sticky', 'split', 'floating', 'mark', 'unmark', 'resize', 'rename', 'nop', 'scratchpad', 'swap', 'title_format', 'title_window_icon', 'mode', 'bar'
[{"success":true},{"success":false,"parse_error":true,"error":"Expected one of these tokens: <end>, '[', 'move', 'exec', 'exit', 'restart', 'reload', 'shmlog', 'debuglog', 'border', 'layout', 'append_layout', 'workspace', 'focus', 'kill', 'open', 'fullscreen', 'sticky', 'split', 'floating', 'mark', 'unmark', 'resize', 'rename', 'nop', 'scratchpad', 'swap', 'title_format', 'title_window_icon', 'mode', 'bar'","input":"exec  \"sh -c \\\"PATH=\\\\\"\\\\$HOME/.local/bin:\\\\$PATH\\\\\"; electrum \\\"\"","errorposition":"                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^"}]
Could not launch i3-msg: 512 at /usr/bin/i3-dmenu-desktop line 493, <GEN1> line 1.
zsh: exit 2     i3-dmenu-desktop
```

## Expected Behavior

The same as `gtk-launch electrum` and wofi: should work

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: 
Binary i3 version:  4.20 (2021-10-19) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.20 (2021-10-19) (pid 7630)
Loaded i3 config:
[...]

PR #5162, opened 2022-09-25
Fix i3-dmenu-desktop quoting

stapelberg:
Commit 70f23caa9a18afc146f696fdf7d2481e5f7f0101 introduced new issues.

Instead of distinguishing " and \, as that commit attempted, let’s instead keep the level of escaping by escaping each backslash, just like each double quote.

I tested this with:

    # recommended way to quote $ and " in quoted arguments, not ambiguous
    Exec=/tmp/logargs "hello \\$PWD \\"and\\" more"

    # permitted way to quote $ and " in quoted arguments, but ambiguous
    Exec=/tmp/logargs "hello \$PWD \"and\" more"

    # permitted way to quote arguments, slightly unusual to quote first arg
    Exec="/tmp/logargs" hey

    # a complicated shell expression, not ambiguous
    Exec=sh -c "if [ -n \\"\\$*\\" ]; then exec /tmp/logargs --alternate-editor= --display=\\"\\$DISPLAY\\" \\"\\$@\\"; else exec /tmp/logargs --alternate-editor= --create-frame; fi" placeholder %F

related to https://github.com/i3/i3/issues/4697 (electrum, original) 

related to https://github.com/i3/i3/issues/5152 (phpstorm, breakage) 

related to https://github.com/i3/i3/issues/5156 (emacsclient, breakage)

### DIFF
--- a/i3-dmenu-desktop
+++ b/i3-dmenu-desktop
@@ -413,7 +413,7 @@ my $exec = $app->{Exec};
 my $location = $app->{_Location};
 
 # Quote as described by “The Exec key”:
-# https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s06.html
+# https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s07.html
 sub quote {
     my ($str) = @_;
     $str =~ s/("|`|\$|\\)/\\$1/g;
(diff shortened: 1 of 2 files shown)


stapelberg:
Depending on the timing of the bugfix release, I can try to add tests for i3-dmenu-desktop, so let’s not merge this right away.

I just wanted to send the PR so that affected folks can try it out :)

orestisfl (approved):
Tests would be nice :+1:

stapelberg:
Tests added and merged :)

PR #5168, opened 2022-09-27
Escape ~ to prevent interpretation as subscript.

bodea:
Escape first ~ in paragraph to prevent interpretation as subscript text.  Diff to generated XML:

```
 <title>DESCRIPTION</title>
-<simpara>i3-config-wizard is started by i3 in its default config, unless <subscript>/.i3/config
+<simpara>i3-config-wizard is started by i3 in its default config, unless ~/.i3/config
 exists. i3-config-wizard creates a keysym based i3 config file (based on
-/etc/i3/config.keycodes) in </subscript>/.i3/config.</simpara>
+/etc/i3/config.keycodes) in ~/.i3/config.</simpara>
 <simpara>The advantage of using keysyms is that the config file is easy to read,
```

### DIFF
--- a/man/i3-config-wizard.man
+++ b/man/i3-config-wizard.man
@@ -35,7 +35,7 @@ used.
 
 == DESCRIPTION
 
-i3-config-wizard is started by i3 in its default config, unless ~/.i3/config
+i3-config-wizard is started by i3 in its default config, unless \~/.i3/config
 exists. i3-config-wizard creates a keysym based i3 config file (based on
 /etc/i3/config.keycodes) in ~/.i3/config.
 


stapelberg:
Thanks for your PR! Do other files in man/ also need updating? At least man/i3.man also contains `~` in prose.

bodea:
I think that the other files are fine.  i3-config-wizard.man was only a problem because there was more than one ~ in a text paragraph.

   perl -00ne 'print "$ARGV:$_" if tr/\~// > 1' man/*.man

finds another couple of instances of more than one ~ in a paragraph, but they are all in contexts where the ~ is not being misinterpreted.

stapelberg (approved):

stapelberg:
Thanks!

PR #5192, opened 2022-10-08
using array to clean code

SimpleHumanI:


Airblader:
Thank you for the contribution. I don't believe this improves the code notably, and also the formatting looks quite odd and you lost the `$TERMINAL` variable in there. I'll be declining this change as I don't think there's much to be gained here.

nimaje:
additionally that would lose sh support

PR #5193, opened 2022-10-09
Fix border dimensions with `hide_edge_borders`

gnif:
If the borders are hidden by default, applications are incorrectly told they have borders, this PR corrects this.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -780,13 +780,24 @@ static void handle_client_message(xcb_client_message_event_t *event) {
          * accurate.
          */
         DLOG("_NET_REQUEST_FRAME_EXTENTS for window 0x%08x\n", event->window);
+        
+        int border_width = config.default_border_width;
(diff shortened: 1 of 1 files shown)


gnif:
I suspect this PR needs further work by someone that is more familiar with the i3 code base. There are some issues here still such as a floating window with borders will now report no borders. Also windows with a MOTIF hint for no decorations are told they have borders still.

While as per spec it's reasonable that the values are estimated, it's not reasonable that they are completely wrong when i3 knows if there should be borders or not.

orestisfl:
I don't think this is more correct, `hide_edge_borders` is conditional, it only applies when there are no other windows in the configured direction.

> While as per spec it's reasonable that the values are estimated, it's not reasonable that they are completely wrong when i3 knows if there should be borders or not.

Is there any real-life usecase where this matters?

Closing but let me know if I misunderstood something.

PR #5206, opened 2022-10-16
tiling drag: allow click immediately, to focus on decoration click

stapelberg:
With the introduction of tiling drag, i3’s click behavior in window decorations changed: before tiling drag, in a tabbed or stacked container, a window would be focused/raised on mouse down. After tiling drag, on mouse up.

This commit sends XCB_ALLOW_REPLAY_POINTER before running the tiling drag code, thereby restoring the focus/raise-on-mouse-down behavior without affecting the tiling drag operation.

### DIFF
--- a/release-notes/bugfixes/2-focus-click
+++ b/release-notes/bugfixes/2-focus-click
@@ -1,0 +1,1 @@
+tiling drag: allow click immediately, to focus on decoration click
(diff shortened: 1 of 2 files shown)


Airblader (approved):
LGTM, but probably needs an entry to the release notes still?

Issue #5185, opened 2022-10-02
title_click option


budRich:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## synopsis

I propose the introduction of a new configuration option:  
`title_click [--tiled|--floating] [--double] BUTTON ACTION`  

minimum version:  
`title_click [--tiled|--floating] BUTTON drag|resize`  

- ACTION is one of the following:  
  - drag (start tile drag or move drag depending on floating state)
  - resize
  - exec
  - kill
  - floating_toggle
- BUTTON is the name of a mousebutton, f.i. `button1` (left button) or `button3` (right button).  
- `--tiled` and `--floating` options are optional, if none (or both) are set the action applies to both types of windows.  An alternative to these options could be to use criterion..
- `--double` option means a *doubleclick* of BUTTON will start the action.

It is currently possible to do:  
`bindsym --release [tiling|floating] buttonN kill|floating toggle|exec ..` ,  
but not the `resize|drag` operations. And not `--double` clicking. And I believe it would be complicated and a bad idea, to introduce "drag" and "resize" as "commands".  

The main reason to introduce `title_click` is to ease the configuration of the tile-drag behavior. Without this option and with only the proposed solution from #5155 : `tiling_drag [enable|only_with_mod|disable]`,
it is not clear or possible to customize the action for the left mouse button. Furthermore this is something that some lefthanded users might want as an **accessibility** convenience (the ability to switch left/right mouse button actions).  

With `title_click` in place the `tiling_drag` option could be boolean and it would simply turn on/off invoking tiling-drag by holding the `floating_modifier` (`drag_modifier` would be better...). I imagine that tiling-drag would still be possible by dragging the title when `title_click --tiled button1 drag` is set. But this would also be confusing, especially if the current behavior is the default (both modifier and titlebar drag starts tile-dragging). I propose disabling the tile-draggin by titlebar by default, but keep it enabled with modifier.  

Pre 4.21 (title drag) the default config would have been declared like this:  

```
title_click --tiled button1 resize
title_click --floating button1 drag
title_click button3 resize

tile_drag off
```

currently the out-of-box behavior looks like this:  

```
title_click button1 drag
title_click button3 resize

tile_drag on
```

I propose this:  
*and having them "hidden" (not printed in the default config, but enabled)*

```
title_click --tiled button1 resize
title_click --floating button1 drag
title_click button3 resize

[...]

PR #5208, opened 2022-10-16
Tiling drag threshold tweaks, default change, configuration directive

stapelberg:
See the individual commits for details. The most important high-level changes:

1. The drag threshold is run through `logical_px()` for hi-dpi displays, and tweaked to match roughly the height of one window decoration. For users of very imprecise mouse movements, either stay within one window decoration height or change `tiling_drag` to `tiling_drag modifier` or `tiling_drag off` entirely.
2. The tiling drag feature is now configurable via the `tiling_drag` directive. It matches exactly the two places where we initiate a `tiling_drag()` call, which keeps things easy to explain and doesn’t explode complexity (like e.g. https://github.com/i3/i3/issues/5185 suggested).
3. The default behavior is changed from `tiling_drag modifier_or_titlebar` to `tiling_drag modifier`, which will get into fewer people’s ways. For new configs (i.e. users who have fewer expectations regarding existing behavior), we include `tiling_drag modifier_or_titlebar` in the config file. I think this is a good compromise to keep existing workflows working, while still making the feature available.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -196,6 +196,7 @@ provided by the i3 https://github.com/i3/i3/blob/next/etc/config.keycodes[defaul
 
 Floating windows are always on top of tiling windows.
 
+[[tiling_drag]]
 === Moving tiling containers with the mouse
 
 Since i3 4.21, it's possible to drag tiling containers using the mouse. The
@@ -1401,6 +1402,29 @@ fullscreen toggle
 # this line is not continued \
 bindsym Mod1+F fullscreen toggle
 -------------------
+
+[[config_tiling_drag]]
+=== Tiling drag
+
+You can configure how to initiate the tiling drag feature (see <<tiling_drag>>).
+
+*Syntax*:
+--------------------------------
+tiling_drag off
+tiling_drag modifier|titlebar [modifier|titlebar]
+--------------------------------
+
+*Examples*:
+--------------------------------
+# Only initiate a tiling drag when the modifier is held:
+tiling_drag modifier
+
+# Initiate a tiling drag on either titlebar click or held modifier:
+tiling_drag modifier titlebar
+
+# Disable tiling drag altogether
+tiling_drag off
+--------------------------------
 
 == Configuring i3bar
 
--- a/src/drag.c
+++ b/src/drag.c
@@ -42,7 +42,8 @@ struct drag_x11_cb {
 
 static bool threshold_exceeded(uint32_t x1, uint32_t y1,
                                uint32_t x2, uint32_t y2) {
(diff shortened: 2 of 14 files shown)


docs/userguide
```diff
@@ -1402,6 +1403,28 @@ fullscreen toggle
...
 -------------------
 
+[[config_tiling_drag]]
+=== Tiling drag
+
+You can configure how to initiate the tiling drag feature (see <<tiling_drag>>).
+
+*Syntax*:
+--------------------------------
+tiling_drag modifier|titlebar|modifier_or_titlebar|off
```
Airblader: One minor concern I have is that `modifier_or_titlebar` does not naturally extend if we were to add more options.

src/drag.c
```diff
@@ -42,7 +42,8 @@ struct drag_x11_cb {
 
 static bool threshold_exceeded(uint32_t x1, uint32_t y1,
                                uint32_t x2, uint32_t y2) {
-    const uint32_t threshold = 9;
+    /* The threshold is about the height of one window decoration. */
```
Airblader, line 45: Isn't the height of a window decoration strongly dependent on the font size?

docs/userguide
stapelberg: It was an intentional choice to use a name that does not extend to mean new things in the future :) 

I think that’s the better approach for carefully introducing new behavior without changing existing configs.

src/drag.c
stapelberg: Yes, you’re right. We could potentially actually calculate a decoration, but I figured it’s not worth the effort. As long as the threshold roughly matches what most people expect, that’s good enough for now.

docs/userguide
Airblader: I think I didn't describe my concern well enough. I rather mean that if we add a third option, this becomes very difficult to add in a natural way (`modifier_or_titlebar_or_x` and lots of other variations).

Listing the activated options would be more extensible (`tiling_drag titlebar modifier`).

src/drag.c
Airblader: sgtm

docs/userguide
Airblader: (I'm fine if we agree that this is unlikely, I just wanted to make sure we are aware of this)

Airblader (approved):
Approving pending the open discussion, will leave it to you to decide there.

docs/userguide
stapelberg: Fair point, modeling the option as a list does make it more extensible. I currently don’t foresee a new option, but you never know… :)

Done.

PR #5224, opened 2022-10-24
i3-dmenu-desktop test: Do not autostart i3

orestisfl:
This actually fixes a hang that happens on my machine for some reason. Regardless, starting i3 is not necessary for this test.

### DIFF
--- a/testcases/t/318-i3-dmenu-desktop.t
+++ b/testcases/t/318-i3-dmenu-desktop.t
@@ -18,7 +18,7 @@
 # and sends the command to i3 for execution.
 # Ticket: #5152, #5156
 # Bug still in: 4.21-17-g389d555d
-use i3test;
+use i3test i3_autostart => 0;
 use i3test::Util qw(slurp);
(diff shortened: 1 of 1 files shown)


stapelberg (approved):

stapelberg:
Thanks for fixing this! I only noticed that the test is flaky, but haven’t had a chance to look into it yet :)

stapelberg:
> Thanks for fixing this! I only noticed that the test is flaky, but haven’t had a chance to look into it yet :)

That said, something still seems to be wrong with the tests. So maybe this change doesn’t fully resolve the issue. But it shouldn’t hurt either.

orestisfl:
> > Thanks for fixing this! I only noticed that the test is flaky, but haven’t had a chance to look into it yet :)
> 
> That said, something still seems to be wrong with the tests. So maybe this change doesn’t fully resolve the issue. But it shouldn’t hurt either.

Indeed, it still happens

PR #5173, opened 2022-09-29
Fix segfault during config validation (#5167)

matugol:
A bug was introduced in #5118 in which configs with bar blocks will segfault during validation. They were copying the i3 font which is not set during validation. This PR simply checks that the `load_type` is not validate before copying the font. Fixes #5167.

PR #5229, opened 2022-10-25
Fix segfault when checking i3 configuration

spacekookie:
I've encountered a segfault regression since `4.20.1` where `i3 -c ...` will segfault because of a null string.

I don't know if this is the most correct way to fix this issue. I'm also not sure why the `pattern` string _is_ `0x0` (according to gdb)

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -282,7 +282,10 @@ bool load_configuration(const char *override_configpath, config_load_t load_type
             if (current->font != NULL) {
                 continue;
             }
-            current->font = sstrdup(config.font.pattern);
+
+            if (config.font.pattern != NULL) {
+                current->font = sstrdup(config.font.pattern);
+            }
         }
     }
 


Un1Gfn:
is #5173 related?

stapelberg:
Can you provide the config file for which validation fails please? (Ideally a minimal config file that triggers the bug)

Is the crash reproducible with the latest version of i3 from the `next` branch?

Thanks

wmerfalen:
> I've encountered a segfault regression since `4.20.1` where `i3 -c ...` will segfault because of a null string.

Can you re-run the command with added command line switches: `-d all -V`? It sounds like you didn't specify a `font` configuration in your i3 config. If you see the following string in the output, that would confirm that this indeed is the case:
`12/14/2022 10:06:27 AM - ERROR: You did not specify required configuration option "font"`

I suspect that `load_font()` in `libi3/font.c` is setting `font.pattern` to null.

In my own personal config, i have this in my i3 config file:
```
font pango:DejaVu Sans Mono 9
```

PR #5239, opened 2022-10-30
Implement i3bar padding directive, supporting the height directive from i3-gaps

stapelberg:
As discussed in issue https://github.com/i3/i3/issues/3721

I used separate commits (where one just imports what you have in the i3-gaps tree) to make the review easier.

I think it’ll be good to keep these commits in the i3 git history, so let’s not squash this commit, but rebase & merge when we’re happy.


Airblader (approved):
LGTM, and +1 on keeping the commits.

Issue #3724, opened 2019-06-17
[Umbrella] Support gaps


Airblader:
## Known issues / possible improvements

- [x] Ensure gaps work with titlebars without a compositor: https://github.com/i3/i3/issues/1966
- [x] Window size isn’t evenly distributed, in particular with large gaps: https://github.com/i3/i3/pull/5252
- [ ] #5253

---

Original issue text:

## Desired Behavior

After a recent discussion between @stapelberg and myself, we would like to reconsider bringing gaps into i3 itself, effectively deprecating i3-gaps. The main rationale here is that there clearly are thousands of users who want gaps.

In order to do so, we need to implement not only gaps, but also all of the current i3-gaps features as listed below. This issue serves as an umbrella issue, if necessary we can create separate issues for the individual features. Before work on this can be started, we need to state our expectations:

1. This is **not** about simply porting the gaps patch into i3. The current gaps patch is kept simple enough to make upstream maintenance possible; this means, however, that in many ways it is more of a hack and doesn't live up to our quality expectations. We should therefore consider – and discuss – how gaps can be brought into i3 in a clean way, particularly including support for titlebars.
2. At this point in time, there is no final decision on whether we would merge this. It very much depends on the discussion and the solution we can find.

# Gaps
This list tracks the core gaps features that need to be supported

- [x] Current i3-gaps configuration and command surface:
    - Configuration: `gaps [inner|outer|horizontal|vertical|top|left|bottom|right] <px>`
    - Configuration: `workspace <ws> gaps [inner|outer|horizontal|vertical|top|left|bottom|right] <px>`
    - Command: `gaps inner|outer|horizontal|vertical|top|right|bottom|left current|all set|plus|minus|toggle <px>`
- [x] Make gaps information properly available on the IPC (required for in-place restart)
- [x] Provide tests for gaps

We should overall strive to be compatible with i3-gaps to minimize friction, but where sensible we may deviate. For example, workspace-level gaps are currently solved a bit awkward in i3-gaps.

# Gaps-related features
The following list tracks other features that need to be supported in i3 as well¹.

- [x] `smart_gaps on | inverse_outer`
- [x] `smart_borders no_gaps`²
- [x] `hide_edge_borders smart_no_gaps`

¹ Note that we may choose to rename features if necessary.
² `smart_borders on` is an alias for `hide_edge_borders smart` and will thus be ignored here.

Linking related issues for visibility: #1474, #3611, #2890 

[...]

PR #5241, opened 2022-10-31
Initial gaps support

stapelberg:
I still have a few refactorings I want to give a try before we should merge this, but I think the current state is good enough for a first pull request review :) 


docs/userguide
```diff
@@ -1426,6 +1426,95 @@ tiling_drag modifier titlebar
...
+gaps outer|horizontal|vertical|top|left|bottom|right <gap_size_in_px>
+
+# Inner and outer gaps for all windows on a specific workspace.
+# <ws> can be a workspace number or name.
+workspace <ws> gaps inner <gap_size_in_px>
+workspace <ws> gaps outer|horizontal|vertical|top|left|bottom|right <gap_size_in_px>
+
+# Enabling “Smart Gaps” means no gaps will be shown when there is
+# precisely one window or split container on the workspace.
+smart_gaps on|off|inverse_outer
```
Airblader, line 1473: I think we need to explain the difference between on and inverse_outer here.

parser-specs/config.spec
```diff
@@ -65,6 +68,32 @@ state IGNORE_LINE:
...
+
+# smart_borders true|false
+# smart_borders no_gaps
+state SMART_BORDERS:
+  enabled = '1', 'yes', 'true', 'on', 'enable', 'active'
+      -> call cfg_smart_borders($enabled)
+  enabled = 'no_gaps'
+      -> call cfg_smart_borders($enabled)
+
+# smart_gaps on|off
```
Airblader: This comment needs to be updated, too.

docs/userguide
stapelberg: Done

parser-specs/config.spec
stapelberg: Done

stapelberg:
Pushed the other commits I had in mind: moving code to gaps.c where applicable, allowing optional “px” suffix, getting rid of `#define` in cmd_gaps and test cases for the `gaps` command.

Please take another look. If this looks reasonable, we could merge this state and do further cleanups/improvements in separate PRs :)

src/commands.c
```diff
@@ -2375,114 +2375,171 @@ void cmd_debuglog(I3_CMD, const char *argument) {
...
+        reset = true;
+    } else {
+        ELOG("Invalid mode %s when changing gaps", mode);
+        return false;
+    }
+
+    /* See https://github.com/Airblader/i3/issues/262 */
+    int min_value = 0;
+    const bool is_outer = get(&(config.gaps)) != gaps_inner(&(config.gaps));
+    // if (strcmp(type, "inner") != 0) {
```
Airblader: Leftover commented out code?

Airblader (approved):
LGTM % one nit. I have to admit that I find the define-based implementation easier to follow, but I might be biased there, am not really as experienced in C as my involvement in i3 would suggest, and it's not a big difference either way so happy to roll with this.

Thanks so much for taking care of this. We'll make a lot of people happy.

src/commands.c
stapelberg: Done

PR #5245, opened 2022-11-03
draw leaf window decorations on ->frame instead of ->parent->frame

stapelberg:
This fixes the graphical artifacts when gaps are enabled with title bars.

Not yet done (will update the PR):

* [x] documentation update (hacking-howto, userguide)
* [x] in `layout stacked` and `layout tabbed`, the terminal contents are one line short for some reason (wrong calculation somewhere?)

### DIFF
--- a/docs/hacking-howto
+++ b/docs/hacking-howto
@@ -564,9 +564,7 @@ split container.
 ==== Default layout
 
 In default layout, containers are placed horizontally or vertically next to
-each other (depending on the +con->orientation+). If a child is a leaf node (as
-opposed to a split container) and has border style "normal", appropriate space
-will be reserved for its window decoration.
+each other (depending on the +con->orientation+).
 
 ==== Stacked layout
 
--- a/docs/ipc
+++ b/docs/ipc
@@ -410,6 +410,14 @@ deco_rect (map)::
 	The coordinates of the *window decoration* inside its container. These
 	coordinates are relative to the container and do not include the actual
 	client window.
+actual_deco_rect (map)::
+	See +deco_rect+. i3 v4.22 changed the way title bars are rendered. Before
+	i3 v4.22, the deco_rect was always relative to the parent coordinates.
+	Starting with i3 v4.22, this remains true for tabbed/stacked containers
+	(actual_deco_rect is identical to deco_rect), but for normal-border leaf
+	containers within vertical/horizontal split containers, actual_deco_rect
+	is relative to the container itself. For more background, see
+	https://github.com/i3/i3/issues/1966
 geometry (map)::
 	The original geometry the window specified when i3 mapped it. Used when
 	switching a window to floating mode, for example.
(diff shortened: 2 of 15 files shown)


okraits:
> This fixes the graphical artifacts when gaps are enabled.

When gaps and titlebars are enabled, I guess, right?

stapelberg:
> When gaps and titlebars are enabled, I guess, right?

Yes

Airblader:
This closes #1966 I believe. I also wonder how you're feeling about https://github.com/i3/i3/issues/1966#issuecomment-158684486?

stapelberg:
> This closes #1966 I believe

Yep, the “fixes #1966” is already in the commit message (but not on the pull request level).

> I also wonder how you're feeling about [#1966 (comment)](https://github.com/i3/i3/issues/1966#issuecomment-158684486)?

I haven’t thought it through in detail.

On one hand, one could say that the deco_rect semantics are unchanged, only the rendering has changed — the deco_rect still specifies *where* the decoration is drawn, but now the drawing happens on a different X11 window.

I think the way we interpreted “retain semantics” in the past (at least in that issue) is to translate coordinates in such a way that the x/y position still matches the absolute position on screen where decorations appear today.

I wonder how widely used the deco_rect is by IPC interface users at all. The only usages I can find on GitHub Code Search that might be affected are https://github.com/cornerman/i3-easyfocus and https://github.com/kraftwerk28/dotfiles/blob/bd081348da1dfdc126db95563b534a7669e990bd/bin/grimvshot#L35

I’ll try and see if i3-easyfocus breaks with my change, and how hard it would be to do a translation.

stapelberg:
Okay, all done! Fixed the tabbed/stacked mis-calculation (within `con_border_style_rect()`), kept the existing `deco_rect` compatible (verified with i3-easyfocus that it works), updated the docs (ipc, hacking-howto, userguide).

Let me know if this looks good now :)

Airblader (approved):
🎉 In a way this change is even more meaningful than merging i3-gaps.

orestisfl:
I'll make a proper issue when I have time but this seems to break titlebars in my non-gaps setup: They are not showing anymore for tiled windows.

stapelberg:
> I'll make a proper issue when I have time but this seems to break titlebars in my non-gaps setup: They are not showing anymore for tiled windows.

Hmm, I can’t reproduce this unfortunately.

I just turned off gaps, but title bars still show up for me, both for a stacked/tabbed split container, and also in a splitv/splith normal border container. 

Can you test as a separate user to see if it also happens with the default config? 

Thanks

orestisfl:
It seems connected to `hide_edge_borders smart`. At least, if I delete the configuration and fallback to the default there is no issue.

stapelberg:
> It seems connected to `hide_edge_borders smart`. At least, if I delete the configuration and fallback to the default there is no issue.

I tried it with the default config and `hide_edge_borders smart` added to it, but everything seems to work as expected (or I’m trying to reproduce it incorrectly)

PR #5246, opened 2022-11-05
Allow text drawing to use the alpha channel.

stapelberg:
This is the last remaining diff from the i3-gaps tree.

Tested using the following config with picom:
```
bar {
	i3bar_command i3bar -t
	status_command i3status
	colors {
		# fully	transparent text on opaque background:
		statusline #ffffff00
		background #000000ff
	}
}
```

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1118,7 +1118,8 @@ client.background::
         which do not cover the whole area of this window expose the color. Note
         that this colorclass only takes a single color.
 
-Colors are in HTML hex format (#rrggbb), see the following example:
+Colors are in HTML hex format (#rrggbb, optionally #rrggbbaa), see the following
+example:
 
 *Examples (default colors)*:
 ----------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Airblader (approved):
LGTM. But I think this forces our hand on closing #3947, right?

Airblader (commented):
What about userguide and release notes?

stapelberg:
> LGTM. But I think this forces our hand on closing #3947, right?

No, I don’t think we necessarily have to close it — we could still accept the compromise of conditionally using either operator depending on the alpha channel. But it seems like (most of?) the PR’s motivation might no longer apply anyway: https://github.com/i3/i3/pull/3947#issuecomment-1304573171

stapelberg:
> What about userguide and release notes?

Updated. Were those the places / phrasing you had in mind? Not sure I 100% captured all effects of this change :)

Airblader (approved):

PR #5259, opened 2022-11-07
gaps: make workspace gap assignments order-independent

### LINKED ISSUE #5253: Potentially make the gaps configuration no longer order-sensitive

stapelberg:
This commit moves subtracting the global gaps from the workspace gaps: previously, this calculation was done while parsing the configuration (order dependent), now it’s done at workspace assignment evaluation time.

fixes https://github.com/i3/i3/issues/5253

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -1476,8 +1476,7 @@ between a screen edge and a window (or split container) will be the sum of outer
 and inner gaps.
 
 You can define gaps either globally or per workspace using the following
-syntax. Note that the gaps configurations should be ordered from least specific
-to most specific as some directives can overwrite others.
+syntax.
 
 *Syntax*:
 -----------------------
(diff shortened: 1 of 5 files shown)


Airblader (approved):
I can't help but point out that this is technically a breaking change in behavior. Sure it wasn't intended and discouraged to have them in the wrong order, but people may have done it anyway and adjusted the values to account for it. Probably still fine, but perhaps we should make a note?

The thing we definitely should do is remove the warning from the user guide that the configuration is order dependent.

stapelberg:
> I can't help but point out that this is technically a breaking change in behavior. Sure it wasn't intended and discouraged to have them in the wrong order, but people may have done it anyway and adjusted the values to account for it. Probably still fine, but perhaps we should make a note?

Agreed, added a release note. 

Also removed the note from the userguide.

Airblader (approved):

PR #5283, opened 2022-11-13
Use mask to determine workspace gaps assignments

### LINKED ISSUE #5282: `workspace X gaps outer|inner 0` does not work
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Reproduction Instructions

```
gaps outer 20
workspace 2 gaps outer 0
workspace 2 gaps outer 1
```

1. Apply config
2. Switch to workspace 2, outer gaps are 20
3. Switch to workspace 3, outer gaps are 1

## Environment

i3 4.21

orestisfl:
Fixes #5282

### DIFF
--- a/src/config_directives.c
+++ b/src/config_directives.c
@@ -231,143 +231,85 @@ CFGFUN(for_window, const char *command) {
     TAILQ_INSERT_TAIL(&assignments, assignment, assignments);
 }
 
-static void create_gaps_assignment(const char *workspace, const char *scope, gaps_t gaps) {
+static void apply_gaps(gaps_t *gaps, gaps_mask_t mask, int value) {
+    if (gaps == NULL) {
+        return;
+    }
+    if (mask & GAPS_INNER) {
+        gaps->inner = value;
+    }
+    if (mask & GAPS_TOP) {
+        gaps->top = value;
+    }
+    if (mask & GAPS_RIGHT) {
+        gaps->right = value;
+    }
+    if (mask & GAPS_BOTTOM) {
+        gaps->bottom = value;
+    }
+    if (mask & GAPS_LEFT) {
+        gaps->left = value;
+    }
+}
+
+static void create_gaps_assignment(const char *workspace, const gaps_mask_t mask, const int pixels) {
+    if (mask == 0) {
+        return;
+    }
+
     DLOG("Setting gaps for workspace %s", workspace);
 
+    bool found = false;
     struct Workspace_Assignment *assignment;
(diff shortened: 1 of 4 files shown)


src/config_directives.c
```diff
@@ -234,67 +234,53 @@ CFGFUN(for_window, const char *command) {
...
-            } else if (!strcmp(scope, "bottom")) {
-                assignment->gaps.bottom = gaps.bottom;
-            } else if (!strcmp(scope, "left")) {
-                assignment->gaps.left = gaps.left;
-            } else {
-                ELOG("Invalid command, cannot process scope %s", scope);
-            }
-
-            return;
+            break;
```
stapelberg, line 267: Here and in src/gaps.c, relying on the `assignment` loop variable and just `break`ing when finding a matching assignment looks incomplete/unsafe to me — wouldn’t the last checked (not necessarily matching!) assignment always be stored in `assignment` after the loop is over?

Can you instead use a separate variable to which matching assignments are explicitly assigned? You could also introduce a helper function for finding a workspace assignment by name.

stapelberg (commented):
Thanks, this looks good overall!

src/config_directives.c
orestisfl: Oops, that was silly

src/gaps.c
```diff
@@ -117,29 +117,35 @@ bool gaps_has_adjacent_container(Con *con, direction_t direction) {
...
+    bool found = false;
     struct Workspace_Assignment *assignment;
     TAILQ_FOREACH (assignment, &ws_assignments, ws_assignments) {
         if (strcmp(assignment->name, ws->name) == 0) {
             gaps = assignment->gaps;
+            found = true;
             break;
         } else if (ws->num != -1 && name_is_digits(assignment->name) && ws_name_to_number(assignment->name) == ws->num) {
             gaps = assignment->gaps;
+            found = true;
```
stapelberg: Should there also be a `break` here?

orestisfl: No, this was intentional. Exact named assignments take precedence over numerical assignments.

orestisfl:
Further simplified the configuration part

src/gaps.c
orestisfl: But there was a mistake here because the code after used `assignment` to get the mask

stapelberg (approved):

stapelberg:
Thanks!

PR #5299, opened 2022-11-21
userguide: variables: mention expansion support

asklow:
Variable expansion is possible through sub-shelling existing variables into a new variable.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -891,12 +891,17 @@ set $m Mod1
 bindsym $m+Shift+r restart
 ------------------------
 
-Variables are directly replaced in the file when parsing. Variables expansion
-is not recursive so it is not possible to define a variable with a value
-containing another variable. There is no fancy handling and there are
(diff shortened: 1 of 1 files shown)


Airblader:
There's actually some mix going on in your example between i3 variables and environment variables.

All in all that isn't something I would want to recommend to users. Our variable system is simple and "dumb", and there are better ways to achieve things.

Airblader:
> But the solution for that specific case is to just pass it through exec i3-msg

That has other side effects and also still won't work in all cases.

What you're describing here is shell behavior, not i3 behavior. We don't know or care what people let their shell do, but as far as i3 goes, there is nothing special that we support here.

As I mentioned above, I would actually not recommend people do this kind of thing at all (but we won't stop you if you do). Variables are for some limited sense if reusability in the config.

Thanks for the PR, but I'll have to decline this. I hope you understand.

PR #5307, opened 2022-12-01
meson: use sh

sewnie:
This pr will remove the need on `bash` shell for systems that do not need it, as this is only checked at build time for a *very* simple task and not needed at run time

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -610,15 +610,15 @@ install_headers(
 # because configure_file strips the backslash in e.g. \@display,
 # resulting in @display, breaking our Perl code:
 # https://github.com/mesonbuild/meson/issues/7165
-bash = find_program('bash')
+sh = find_program('sh')
 replace_dirs = [
-  bash, '-c',  # Use bash to capture output and mark as executable
+  sh, '-c',  # Use sh to capture output and mark as executable
   'sed -e \'s,@abs_top_builddir@,'
   + meson.current_build_dir()
   + ',g;s,@abs_top_srcdir@,'
   + meson.current_source_dir()+',g\''
   # Only mark files ending in .pl as executables
-  + ' "$0" > "$1" && { [[ "${1##*.}" == pl ]] && chmod +x "$1" || true; }',
+  + ' "$0" > "$1" && { [ "${1##*.}" = pl ] && chmod +x "$1" || true; }',
   '@INPUT0@',   # $0
   '@OUTPUT0@',  # $1
 ]


sewnie:
This PR has been marked as draft for the sole purpose of me not being able to build `i3`. as yajl is broken and for unknown reason.

sewnie:
seems this works, the 'checks' are failing because of a missing package:
```
 Reading state information...
E: Unable to locate package clang-format-10
Error: Process completed with exit code 100.
```

meson.build
```diff
@@ -610,15 +610,15 @@ install_headers(
...
 replace_dirs = [
-  bash, '-c',  # Use bash to capture output and mark as executable
+  sh, '-c',  # Use sh to capture output and mark as executable
   'sed -e \'s,@abs_top_builddir@,'
   + meson.current_build_dir()
   + ',g;s,@abs_top_srcdir@,'
   + meson.current_source_dir()+',g\''
   # Only mark files ending in .pl as executables
-  + ' "$0" > "$1" && { [[ "${1##*.}" == pl ]] && chmod +x "$1" || true; }',
+  + ' "$0" > "$1" && { [ "${1##*.}" = pl ] && chmod +x "$1" || true; }',
```
stapelberg, line 621: Is `"${1##*.}"` supported by all non-bash shells?

In general, I would prefer to just stick to bash. It’s widely available.

stapelberg:
> seems this works, the 'checks' are failing because of a missing package:

I’m addressing that with PR #5317

meson.build
sewnie: > Is "${1##*.}" supported by all non-bash shells?

It's POSIX. this PR aims to support `sh`, but not explicitly `POSIX`, as these lines are the only ones that conflict with the specification.

POSIX Shell is available on all UNIX systems. i'm not saying that now i3 supports computers like the Amiga 97 /s, but as rather for convenience for people who don't like the Bash shell or do not have it on their system.

stapelberg: > this PR aims to support sh, but not explicitly POSIX

Which shell implementation are you referring to when you say `sh`? Is it `dash`? Or a BSD `sh` perhaps?

In general, I think only mandating `bash` or mandating a POSIX-compliant shell makes sense. 

> but as rather for convenience for people who don't like the Bash shell or do not have it on their system.

i3 doesn’t need `bash` at runtime, so people who install i3 from their distribution (the preferred way) don’t need `bash` in any case. Only people who want to build i3 need `bash`, and, realistically, so many other programs need `bash` anyway that we might just depend on it. 

Supporting non-`bash` shells needs more than just clicking “merge” on this pull request — we would also need to test with different shells from now on to make sure we don’t regress on the support.

I don’t think this is worth our limited manpower, so I’m inclined to decline this PR and just stick to requiring `bash` for compilation.

eli-schwartz: Please no.

It's really simple to support non-bash shells, you just need to use https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html instead of `man bash`. Test with a system where `/bin/sh` is `dash` and you'll be totally fine.

The natural subset of bash that people tend to use is mostly POSIX compliant anyway, other than really trivial things that checkbashisms can report on.

```
$ cat /tmp/foo.sh
#!/bin/sh

sed -e \
    's,@abs_top_builddir@,current_build_dir,g;s,@abs_top_srcdir@,current_source_dir,g' \
    "$0" > "$1" && { [[ "${1##*.}" == pl ]] && chmod +x "$1" || true; }


$ checkbashisms /tmp/foo.sh
possible bashism in /tmp/foo.sh line 5 (alternative test command ([[ foo ]] should be [ foo ])):
sed -e \
    's,@abs_top_builddir@,current_build_dir,g;s,@abs_top_srcdir@,current_source_dir,g' \
    "$0" > "$1" && { [[ "${1##*.}" == pl ]] && chmod +x "$1" || true; }
possible bashism in /tmp/foo.sh line 5 (should be 'b = a'):
sed -e \
    's,@abs_top_builddir@,current_build_dir,g;s,@abs_top_srcdir@,current_source_dir,g' \
    "$0" > "$1" && { [[ "${1##*.}" == pl ]] && chmod +x "$1" || true; }
```

And checkbashisms agrees exactly with the changes in this PR. :)

eli-schwartz: It would be easier to use checkbashisms to ensure that shell scripts are actually POSIX compliant, if this was moved into something like `build-aux/replace-dirs.sh` instead of being inline.

stapelberg: I don’t mind factoring this out into a separate file.

Before the PR can get merged, it needs to be updated to run checkbashisms in GitHub Actions, otherwise we’ll inevitably regress on this over time.

Note that I’ll only commit to fixing checkbashisms warnings that are uncovered by GitHub Actions, but I won’t actually test with non-bash shells. I don’t have a system where /bin/sh is dash, and I’m not interested in having one.

sewnie:
uhhhhhhhhhhh is this supposed to be closed or merged or wat

stapelberg:
> uhhhhhhhhhhh is this supposed to be closed or merged or wat

From my perspective, this is the latest comment on my end which hasn’t been addressed:

> Before the PR can get merged, it needs to be updated to run checkbashisms in GitHub Actions, otherwise we’ll inevitably regress on this over time.
> 
> Note that I’ll only commit to fixing checkbashisms warnings that are uncovered by GitHub Actions, but I won’t actually test with non-bash shells. I don’t have a system where /bin/sh is dash, and I’m not interested in having one.

meson.build
sewnie: >  it needs to be updated to run checkbashisms in GitHub Actions

Is this meant to be added in the i3 .github directory beforehand or this pull request adds it?

stapelberg: The pull request should add it

PR #5378, opened 2023-01-16
Fix top border resizing registered on tiled windows

slyshot:
I noticed click+drag resize wasn't working on the top border of my tiled windows.
Unlike the other checks for where the mouse is clicked, it reference decoration. This commit makes it reference the borders like the other ones.

Thanks.

### DIFF
--- a/release-notes/bugfixes/1-top-border-resize
+++ b/release-notes/bugfixes/1-top-border-resize
@@ -1,0 +1,1 @@
+fix top border resizing on tiling windows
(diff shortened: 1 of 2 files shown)


orestisfl:
This will probably conflict with

```
tiling_drag titlebar
```

slyshot:
Ah, so does keeping the old code and putting the new code in an `else` (or `else if (dest == CLICK_BORDER)`) to deal with both cases make sense?

orestisfl (changes requested):
Sorry for the delay. It seems your change is fine as it is.

Can you add some release notes for this change?

```shell
echo 'fix top border resizing on tiling windows' > release-notes/bugfixes/1-top-border-resize
git add release-notes/bugfixes/1-top-border-resize
```

orestisfl (approved):

PR #5388, opened 2023-01-22
Push artifacts only on next branch

orestisfl:


.github/workflows/main.yml
```diff
@@ -50,26 +47,33 @@ jobs:
...
+    name: Push artifacts
+    runs-on: ubuntu-latest
+    needs: [build]
+    if: github.ref == 'refs/heads/next'
+    steps:
+    - uses: actions/checkout@v3
+    - name: fetch or build extra Docker containers
+      run: |
+        docker pull ${{ env.BASENAME_UBUNTU }} || ./travis/docker-build-and-push.sh ${{ env.BASENAME_UBUNTU }} travis/travis-base-ubuntu.Dockerfile
     - name: build dist tarball
```
stapelberg: The dist tarball should always be built, even on pull requests (not just on the next branch)

.github/workflows/main.yml
```diff
@@ -32,9 +27,6 @@ jobs:
     - name: fetch or build Docker container
       run: |
         docker pull ${{ env.BASENAME }} || ./travis/docker-build-and-push.sh ${{ env.BASENAME }} travis/travis-base.Dockerfile
```
stapelberg, line 32: docker-build-and-push.sh needs the DOCKER_ environment variables set so that it can push an updated version to the docker registry if needed. So please re-add the DOCKER_* variables to env: in line 19-21.

stapelberg (changes requested):

.github/workflows/main.yml
```diff
@@ -53,23 +45,36 @@ jobs:
...
+    name: Push artifacts
+    runs-on: ubuntu-latest
+    needs: [build]
+    env:
+      DOCKER_PASS: ${{ secrets.DOCKER_PASS }}
+      DOCKER_EMAIL: ${{ secrets.DOCKER_EMAIL }}
+      DOCKER_USER: ${{ secrets.DOCKER_USER }}
+      GH_TOKEN: ${{ secrets.GH_TOKEN }}
+      BALTO_TOKEN: ${{ secrets.BALTO_TOKEN }}
+    if: github.ref == 'refs/heads/next'
```
Airblader, line 61: Just double checking, this doesn't need to run for stable?

stapelberg: Correct, the autobuilder repositories only contain versions built from the `next` branch

stapelberg (approved):

PR #5390, opened 2023-01-23
Fix "default binding" mistake in userguide

nikitabobko:
Default binding to move window down is $mod+Shift+k, not $mod+Shift+j. Proof: https://github.com/i3/i3/blob/next/etc/config#L45 

This commit reverts this pull request https://github.com/i3/i3/pull/4146

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -317,7 +317,7 @@ single workspace on which you open three terminal windows. All these terminal
 windows are directly attached to one node inside i3’s layout tree, the
 workspace node. By default, the workspace node’s orientation is +horizontal+.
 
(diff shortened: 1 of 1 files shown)


orestisfl (approved):
Oops, my fault because I have a modified keybind set

orestisfl:
Can you please make the same change here so that it's immediately reflected on the website? https://github.com/i3/i3.github.io/blob/main/_docs/userguide

nikitabobko:
> Can you please make the same change here so that it's immediately reflected on the website?

Fixed https://github.com/i3/i3.github.io/pull/124

PR #5430, opened 2023-02-23
Use separate pixmaps for each edge of a container (#3479)

### LINKED ISSUE #3479: Use separate pixmaps for each edge of a container
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
i3 uses a single pixmap that covers the entire window.

## Expected Behavior
To reduce memory usage, we could use separate pixmaps for the edges (top, bottom, left right) to draw titlebar + borders, and leave the inside of the windows without a (useless) pixmap. See https://github.com/i3/i3/issues/2742#issuecomment-433330426 for details.

## Environment

i3 4.15

buchankn:
This pull request fixes https://github.com/i3/i3/issues/3479. 

i3 currently allocates one big pixmap for an entire window, but only draws on the edges of the pixmap. This code allocates four separate pixmaps for each edge of the window (a 'sparse surface'), thereby using less memory.

To ensure 'correctness', any drawing operation is performed on each of the four edge pixmaps, so that any operation that crosses pixmap boundaries will be drawn correctly. Note that any draws outside the pixmap boundary will just be clipped. This may not be the most efficient way to handle this, and could possibly be optimized, but I was going for correctness in this case. I'm open to suggestions on this...

### DIFF
--- a/src/sparse_surface.c
+++ b/src/sparse_surface.c
@@ -1,0 +1,303 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3 - an improved dynamic tiling window manager
+ * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
+ *
+ * sparse_surface.c: A 'sparse surface' represents the pixmap borders around
+ * a window. We allocate four separate pixmaps, one for each side of the
+ * window, which means we don't allocate any memory for the contents of
+ * the window, which we don't draw to.
+ */
+
+#include "sparse_surface.h"
+#include "all.h"
+
+/* The default visual_type to use if none is specified when creating the surface. Must be defined globally. */
+extern xcb_visualtype_t *visual_type;
+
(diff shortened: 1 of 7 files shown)


orestisfl:
Great attempt but this creates artifacts for me in the borders.

Before:
screenshot-2024-05-17T22:42:44 CEST

After:
screenshot-2024-05-17T22:42:21 CEST

src/sparse_surface.c
```diff
@@ -0,0 +1,307 @@
...
+        }
+    }
+}
+
+/*
+ * Copies a sparse surface onto another non-sparse surface.
+ *
+ */
+void sparse_surface_copy(sparse_surface_t *src_sparse, surface_t *dest, double src_x, double src_y,
+                         double dest_x, double dest_y, double width, double height) {
```
orestisfl, line 214: `width` and `height` ignored here.

orestisfl (changes requested):

PR #5444, opened 2023-03-05
Update i3-sensible-terminal: Adding Tess

Squitch1:
Tess is an electron based terminal emulator.


orestisfl:
Thank you for the suggestion.

I think terminals in this list should be stable software with a big following and active development, since we are somewhat "endorsing" them. Tess seems promising but it's not there yet so I'll have to close this PR

> As long as Tess is in beta, many bugs may occur and some of them have not yet been fixed. Below is a list of most of them

Users of course can pick their preferred terminal with the `$TERMINAL` environmental variable.

Issue #5462, opened 2023-03-31
Building i3 from source


serkan-bayram:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
There are no instructions for building the i3.

## Desired Behavior
For users that relatively new to Linux there should be some explanation. Even if it say "you gotta build it via meson" it would be very helpful.

## Impact

<pre>
[ ] This feature requires new configuration and/or commands
</pre>

## Environment

Output of `i3 --moreversion 2>&-`:
<pre>
i3 version: 4.22 (2023-01-02)
</pre>

<pre>
- Linux Distribution & Version: 
- Are you using a compositor (e.g., xcompmgr or compton):
</pre>

i3bot:
Please note that new features which require additional configuration will usually not be considered. We are happy with the feature set of i3 and want to focus in fixing bugs instead. We do accept feature requests, however, and will evaluate whether the added benefit (clearly) outweighs the complexity it adds to i3.

Keep in mind that i3 provides a powerful way to interact with it through its IPC interface: https://i3wm.org/docs/ipc.html.

Airblader:
You can find instructions here: https://i3wm.org/docs/hacking-howto.html#_building_i3

PR #5463, opened 2023-04-01
hacking-howto: Update build instructions for meson

orestisfl:
Related to #5462

### DIFF
--- a/docs/hacking-howto
+++ b/docs/hacking-howto
@@ -34,30 +34,18 @@ Guide → Compiling a Meson project]. In case you’re unfamiliar:
 
 === Build system features
 
-* We use the +AX_ENABLE_BUILDDIR+ macro to enforce builds happening in a separate
-  directory. This is a prerequisite for the +AX_EXTEND_SRCDIR+ macro and building
-  in a separate directory is common practice anyway. In case this causes any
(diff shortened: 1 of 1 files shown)


docs/hacking-howto
```diff
@@ -34,30 +34,18 @@ Guide → Compiling a Meson project]. In case you’re unfamiliar:
...
-  docs can be disabled with the +--disable-mans++ and ++--disable-docs++
-  configure options respectively.
-
-* non-release builds will enable address sanitizer by default. Use the
-  +--disable-sanitizers+ configure option to turn off all sanitizers, and see
-  +--help+ for available sanitizers.
-
-* Coverage reports are now generated using +make check-code-coverage+, which
-  requires specifying +--enable-code-coverage+ when calling configure.
+* +meson b_sanitize=address+ will enable the address sanitizer which is
```
stapelberg: I think you’re missing a -D here: it should be `meson -Db_sanitize=address`

stapelberg (changes requested):

docs/hacking-howto
orestisfl: Thanks!

stapelberg (approved):

PR #5467, opened 2023-04-01
Update ubuntu & clang-format in builds

orestisfl:


.github/workflows/main.yml
```diff
@@ -84,8 +84,8 @@ jobs:
       run: ./release-notes/generator.pl
     - name: Install dependencies
       run: |
-        sudo apt-get install -y clang-format-12
+        sudo apt-get install -y clang-format-15
```
stapelberg: We should probably stick to clang-format-14 for now, which seems more widely available (per packages.ubuntu.com)

stapelberg (changes requested):

stapelberg (approved):

PR #5469, opened 2023-04-02
Avoid creating redundant containers when switching between tabbed/stacked and split layouts

### LINKED ISSUE #3001: split + layout stacked / tabbed creates redundant containers

orestisfl:
Fixes #3001

### DIFF
--- a/release-notes/changes/3-redundant-containers
+++ b/release-notes/changes/3-redundant-containers
@@ -1,0 +1,1 @@
+avoid creating redundant containers when switching between tabbed/stacked and split layouts
--- a/src/con.c
+++ b/src/con.c
@@ -1942,6 +1942,31 @@ void con_set_layout(Con *con, layout_t layout) {
         }
     }
 
+    if (con->type != CT_WORKSPACE && con->parent->type != CT_WORKSPACE &&
+        con_num_children(con) == 1 && con_num_children(con->parent) == 1) {
+        /* Special case: Avoid creating redundant containers (#3001):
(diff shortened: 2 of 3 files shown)


stapelberg:
I don’t have time to look at this in detail right now, but can you outline why `tree_flatten` in src/tree.c is not sufficient? I would have expected a fix to land there.

orestisfl:
I don't think single-parent containers are inherently bad that's why it's hard to find a general rule that would fix the issue and keep `tree_flatten`'s behavior acceptable in all cases where it's called.

orestisfl:
@stapelberg Any chance we re-visit this PR? I feel it's one of the major causes of pain for users. I think including a fix in `tree_flatten` is not the way to go. The function cannot always be called except for "cleanup" operations after moves etc and I feel that it is much cleaner to prevent creating a situation rather than fixing it afterwards.

stapelberg:
OK, I’ll try to take another look in the coming days.

stapelberg (approved):

PR #5483, opened 2023-04-11
GH Actions: Check for gcc before deploying

### LINKED ISSUE #5482: GitHub Actions now deploy twice (once with gcc, once with clang)
Commit 6911c116e73f374a917f3281fc7c261a2a8cabc6 first removed the `[ "$CC" = "gcc" ]` check. I missed this during review. 

@orestisfl Can you send a PR to re-introduce the check? I’m getting warning emails on every commit now because of the double push of the auto-built documentation to GitHub pages (pushing twice in short succession results in cancellation of one run).

Thanks.

orestisfl:
Fixes #5482


stapelberg (approved):

stapelberg:
Thanks for the quick fix!

PR #5510, opened 2023-05-17
Remove the double forking in start_application

### LINKED ISSUE #5506: i3wm: avoid double `fork()` when starting applications to preserve parent information
## I'm submitting a…

<pre>
[~] Bug
[X] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

When an application is started from i3 (via `exec`), i3 uses two consecutive `fork()`s so that when the middle child exits, the `exec`ed application becomes an orphan (and is reparented to the closest alive reaper). This causes incorrect parent information for the processes started by i3.

## Desired Behavior

I would like the children of i3 to preserve the parent information. My particular use case is to terminate children of i3 when it exists (I was going to use a wrapper that would do `prctl(PR_SET_PDEATHSIG, SIGHUP)` before executing them, which would deliver the signal when i3 quit), but it doesn't make sense after child processes are reparented.

There might be other use cases for which the actual parent information would be desirable.

## Implementation

As I understood from the comments
https://github.com/i3/i3/blob/fde43a078b1fac5fc5842eb1cfb7e2a1a9f7bca8/src/startup.c#L121-L124
, this behavior is there to avoid zombie processes without having to implement the children collection logic. However, there is no need to implement it: according to `sigaction(2)`,
```
SA_NOCLDWAIT (since Linux 2.6)
       If signum is SIGCHLD, do not transform children into zombies when they terminate.   See  also  wait‐  
       pid(2).   This flag is meaningful only when establishing a handler for SIGCHLD, or when setting that
       signal's disposition to SIG_DFL.

       If the SA_NOCLDWAIT flag is set when establishing a handler for SIGCHLD, POSIX.1 leaves it  unspeci‐
       fied  whether  a  SIGCHLD  signal is generated when a child process terminates.  On Linux, a SIGCHLD
       signal is generated in this case; on some other implementations, it is not.
```

So, by running a single `sigaction` in the initialization stage, `i3` can avoid zombies caused by its `exec`s, and there is no need for double forking. This would preserve the actual parent information for its children.

I wonder if there is a reason I am missing why it wasn't implemented that way. If there isn't one, I would like to work on a PR for this.

## Impact

<pre>
</pre>

## Environment

<pre>
Running i3 version: 4.22 (2023-01-02) (pid 801)
Loaded i3 config:
  /home/nikolay/.config/i3/config (main) (last modified: Mon 08 May 2023 12:45:50 MSK, 461098 seconds ago)

</pre>

<pre>
- Are you using a compositor (e.g., xcompmgr or compton): yes, picom vgit-b700a
</pre>

kolayne:
Having just a single fork is beneficial, as it preserves the approprate parent information for the children of i3, which is useful in some scenarious e.g. when a child wants to do something on the wm's exit (possible via `prctl(PR_SET_PDEATHSIG, ...)`).

Moreover, this is a zero-cost benefit: i3 is already using libev with the default loop, which automatically reaps all the zombie children even when there is no corresponding event set.

Quote from the `ev(3)` man page (emphasis mine):
> **Libev grabs "SIGCHLD" as soon as the default event loop is initialised.** This is necessary to guarantee proper behaviour even if the first child watcher is started after the child exits. The occurrence of "SIGCHLD" is recorded asynchronously, but child reaping is done synchronously as part of the event loop processing. **Libev always reaps all children, even ones not watched.**

Resolves #5506

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -579,7 +579,11 @@ int main(int argc, char *argv[]) {
 
     /* Initialize the libev event loop. This needs to be done before loading
      * the config file because the parser will install an ev_child watcher
-     * for the nagbar when config errors are found. */
+     * for the nagbar when config errors are found.
+     *
+     * Main loop must be ev's default loop because (at the moment of writing)
+     * only the default loop can handle ev_child events and reap zombies
+     * (the start_application routine relies on that too). */
     main_loop = EV_DEFAULT;
     if (main_loop == NULL)
         die("Could not initialize libev. Bad LIBEV_FLAGS?\n");
--- a/src/startup.c
+++ b/src/startup.c
@@ -118,10 +118,9 @@ void startup_sequence_delete(struct Startup_Sequence *sequence) {
 }
 
 /*
- * Starts the given application by passing it through a shell. We use double
- * fork to avoid zombie processes. As the started application’s parent exits
- * (immediately), the application is reparented to init (process-id 1), which
- * correctly handles children, so we don’t have to do it :-).
+ * Starts the given application by passing it through a shell. Zombie processes
+ * will be collected by ev in the default loop, we don't have to manually
+ * deal with it.
  *
  * The shell used to start applications is the system's bourne shell (i.e.,
  * /bin/sh).
@@ -173,7 +172,8 @@ void start_application(const char *command, bool no_startup_id) {
 
     LOG("executing: %s\n", command);
     if (fork() == 0) {
-        /* Child process */
+        /* Child process.
+         * It will be reaped by ev, even though there is no corresponding ev_child */
         setsid();
         setrlimit(RLIMIT_CORE, &original_rlimit_core);
         /* Close all socket activation file descriptors explicitly, we disabled
@@ -186,18 +186,14 @@ void start_application(const char *command, bool no_startup_id) {
         unsetenv("LISTEN_PID");
         unsetenv("LISTEN_FDS");
         signal(SIGPIPE, SIG_DFL);
-        if (fork() == 0) {
-            /* Setup the environment variable(s) */
-            if (!no_startup_id)
-                sn_launcher_context_setup_child_process(context);
-            setenv("I3SOCK", current_socketpath, 1);
+        /* Setup the environment variable(s) */
+        if (!no_startup_id)
+            sn_launcher_context_setup_child_process(context);
+        setenv("I3SOCK", current_socketpath, 1);
 
-            execl(_PATH_BSHELL, _PATH_BSHELL, "-c", command, NULL);
-            /* not reached */
-        }
-        _exit(EXIT_SUCCESS);
+        execl(_PATH_BSHELL, _PATH_BSHELL, "-c", command, NULL);
+        /* not reached */
     }
-    wait(0);
 
     if (!no_startup_id) {
         /* Change the pointer of the root window to indicate progress */


orestisfl:
So, this would be a backwards-incompatible change (even if the new behavior is an improvement). I wonder how this is going to affect users and if there are usecases where keeping programs alive after i3's exit is beneficial.

Also, will changing this mean programs get killed after a restart? Because that would absolutely break users' expectations of i3 and can cause lost work.

kolayne:
Programs won't get killed neither on restart nor on i3 exit, unless the child process explicitly requests that via `prctl`. The behavior only changes in case the child application is watching the parent (either directly via `ppid` (it will now return i3's pid for as long as it's alive, rather than the closest reaper's pid, as it would with a double fork) or indirectly e.g. via `prctl` by requesting the kernel to deliver a signal when the parent dies - and if an application launched by a window manager requests that, I'd argue it wants to watch the window manager, not the closest reaper or PID1).

The restart should not involve this at all, as i3 `exec`s and it does not change the pid, but I did not test it, so I'll double check and come back with the description of scenarios and the changed behavior in a few hours.

kolayne:
## TL;DR

is in the next comment

# Clarification

To reiterate: the double `fork()` logic used in `start_application` (which this PR proposes to remove) has nothing to do with killing or not killing children: no signals are delivered in either way. It has do with _reaping_ children (that is, collecting zombie processes _after_ they exited or got killed). The double fork forces immediate reparenting of the child processes (child becomes an orphan and is reparented by the kernel to the closest alive subreaper ancestor, with PID1 being the global default), so `i3` doesn't have to deal with reaping. However, it turned out to be unnecessary: the `ev` library will reap all the zombies anyway.

And the benefit of the patch is that the valid parent information is preserved. Children can rely on it however they want, for example, _in my particular use case_, automatically die when the wm exits (but the patch does not enforce this behavior in any way).

# Behavior

For the tests I used simple programs of 3 types:
- careless - does not care about its parents at all, just doing it's thing
- parent_watcher - wants to do something parent-related, requests the parent pid from system for that
- pdeath_waiter - wants to do something on the event of parent's death

Here are their source codes
<details>
<summary>careless.c</summary>

```c
#include <stdio.h>
#include <unistd.h>

int main() {
    for(int i = 0; i < 6; ++i) {
        puts("I don't care who my parent is, I am running");
        fflush(stdout); 
        sleep(2);
    }   
    puts("Exiting gracefully");
}
```

</details>

<details>
<summary>parent_watcher.c</summary>

```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <string.h>

void print_cmd_name(pid_t pid) {
    char buf[512] = "/proc/";
    sprintf(buf + 6, "%ld", pid);
    size_t pos = strlen(buf);
    memcpy(buf + pos, "/comm", 6); 

    FILE *f = fopen(buf, "r");
    if(!f) {
        puts("couldn't get their executable name");
        return;
    }   

    fgets(buf, 512, f); 
    printf("which is %s", buf);
}

int main() {
    for(int i = 0; i < 6; ++i) {
        pid_t ppid = getppid();
        printf("My current parent is %ld, ", ppid);
        print_cmd_name(ppid);
        fflush(stdout);
        sleep(2);
    }   
    puts("Exiting gracefully");
}
```

</details>

<details>
<summary>pdeath_waiter.c</summary>

```c
#include <stdio.h>
#include <unistd.h>
#include <signal.h>
#include <sys/prctl.h>
#include <errno.h>
#include <assert.h>
    
void handle_hup(int _signal) {
    puts("Got a SIGHUP! Proceeding with the execution");
}   

int main() {
    prctl(PR_SET_PDEATHSIG, SIGHUP, 0, 0, 0);
    assert(!errno);
        
    struct sigaction sa;
    sa.sa_handler = handle_hup;
    sigemptyset(&sa.sa_mask);
    sa.sa_flags = 0;
    sigaction(SIGHUP, &sa, NULL);
    assert(!errno);

    for(int i = 0; i < 6; ++i) {
        puts("I'm waiting for a notification");
        fflush(stdout);
        sleep(2);
    }   
    puts("Exiting gracefully");
}
```

</details>

I ran each of them on both original and the patched versions and checked their behavior on i3's restart and exit (the lines starting with `<---` in the output are added by me).

## Output with upstream i3

**careless** does not care: it keeps running no matter what.
```
I don't care who my parent is, I am running
I don't care who my parent is, I am running
<--- i3 restarted here.
I don't care who my parent is, I am running
I don't care who my parent is, I am running
<--- i3 exited here.
I don't care who my parent is, I am running
I don't care who my parent is, I am running
Exiting gracefully
```

**parent_watcher** gets reparented immediately. On my system there is no subreapers, so it gets adopted by PID1:
```
My current parent is 1, which is systemd
My current parent is 1, which is systemd
<--- i3 restarted here.
My current parent is 1, which is systemd
My current parent is 1, which is systemd
<--- i3 exited here.
My current parent is 1, which is systemd
My current parent is 1, which is systemd
Exiting gracefully
```

Just like parent_watcher, **pdeath_waiter** becomes an orphan very early (in fact, so early that it doesn't have enough time to call `prctl` by that time: the call takes place only after reparenting), so it waits for its (new) parent to exit (which won't happen in case it's reparanted to PID1, but very well might happen if there's a subreaper on the way, which may lead to unintended behavior):
```
I'm waiting for a notification
I'm waiting for a notification
<--- i3 restarted here
I'm waiting for a notification
I'm waiting for a notification
<--- i3 exited here
I'm waiting for a notification
I'm waiting for a notification
Exiting gracefully
```

## Output with patched i3

For **careless** nothing changes, the behavior is exactly the same:
```
I don't care who my parent is, I am running
I don't care who my parent is, I am running
<--- i3 restarted here
I don't care who my parent is, I am running
I don't care who my parent is, I am running
<--- i3 exited here
I don't care who my parent is, I am running
I don't care who my parent is, I am running
Exiting gracefully
```

**parent_watcher** now correctly identifies `i3` as its parent for as long as it's alive; after i3 exits, the child gets reparented in the same way as before:
```
My current parent is 12232, which is i3
My current parent is 12232, which is i3
<--- i3 restarted here
My current parent is 12232, which is i3
My current parent is 12232, which is i3
<--- i3 exited here
My current parent is 1, which is systemd
My current parent is 1, which is systemd
Exiting gracefully
```

**pdeath_waiter** now can do arbitrary actions when i3 exits!
```
I'm waiting for a notification
I'm waiting for a notification
<--- i3 restarted here
I'm waiting for a notification
I'm waiting for a notification
<--- i3 exited here
Got a SIGHUP! Proceeding with the execution
I'm waiting for a notification
I'm waiting for a notification
Exiting gracefully
```

kolayne:
# TL;DR

- i3 restart does not cause any effects at all (both before and after the patch)
- i3 exit (whether graceful or abrupt) does not cause any effects on applications which don't request that explicitly (both before and after the patch)
- Applications which _do_ request something parent-related won't see i3 as their parent, as they are immediately reparented (and not necessarily to PID1, which makes the behavior less predictable), but _with_ the patch they will remain i3's children for as long as it's alive

kolayne:
@orestisfl, is there any action needed from my side? Should I explicitly request someone's review or just wait?

kolayne:
It seems that the double fork logic was added in f4f4d78. So, @stapelberg, I would appreciate your comments

stapelberg:
> It seems that the double fork logic was added in [f4f4d78](https://github.com/i3/i3/commit/f4f4d782bb590e7e03302586fa3028fe785ba95d). So, @stapelberg, I would appreciate your comments

No, that commit only moves the code around. It was originally added in https://github.com/i3/i3/commit/4589c265581021f77fbafa82ab1f5e984163ca5c, way before we even used libev.

The change seems reasonable to me, or at least reasonable enough to try it out :)

stapelberg (approved):

PR #5521, opened 2023-06-01
Fix multiple window movement issue

### LINKED ISSUE #5382: Can't move two focused windows to a different monitor without first toggling back to a different view mode.
## Current Behavior

I use multiple monitors with i3 with the following shortcut to spawn two terminal windows at the same time: 

bindsym Control+Mod1+t exec xfce4-terminal -e htop, exec xfce4-terminal

It makes two xfce4 terminal windows spawn at the same time, one is a just a plain terminal to be able to input commands and the other is htop.

If I focus both windows with this: 

bindsym $mod+a focus parent

And then try to move the two focused windows to another monitor using the arrow keys: 

move focused window
bindsym $mod+Shift+Left move left
bindsym $mod+Shift+Down move down
bindsym $mod+Shift+Up move up
bindsym $mod+Shift+Right move right

i3 doesn't do anything, no windows move to different monitor. There is also no error warning or anything similar.

I have to focus back to child and toggle between vertical and horizontal tiling using: 

bindsym $mod+s layout toggle split

After doing so, now using the arrow keys to move the two focused windows to another monitor works, even though the two windows were tiled in the original position before

## Expected Behavior

The two windows should just move between monitors like it does with just one window, or like it does after I toggle back between two different view modes.

## Reproduction Instructions

Use the commands described above.

Here is my config: https://pastebin.com/dAbG3e6X

## Environment

i3-wm version 4.22-3 from the official arch repositories.

- Linux Distribution & Version: Arch Linux X86_64, Kernel 6.1.7-arch1-1.

- Are you using a compositor (e.g., xcompmgr or compton): Yes, picom.

slyshot:
This solves #5382. The issue is caused because there's no container between the workspace and the windows, and when focus is moved up to the workspace, `tree_move` will refuse to move workspaces.

I was planning on making a change elsewhere(Making the split container if they'd otherwise attach to the workspace at all), but as I mentioned in my comment on the issue, the windows attaching directly to the workspace seems specifically intentional based on comments in code. For this reason, making the change at `tree_move` instead of earlier seems needed to respect that intention. If I'm mistaken in any way, or making the earlier change seems like a good idea anyways, let me know.

Thanks.

### DIFF
--- a/release-notes/bugfixes/3-moving-multiple
+++ b/release-notes/bugfixes/3-moving-multiple
@@ -1,0 +1,1 @@
+Fixed the multiple window movement issue from #5382
--- a/src/move.c
+++ b/src/move.c
@@ -265,8 +265,20 @@ void tree_move(Con *con, direction_t direction) {
     /* 1: get the first parent with the same orientation */
 
     if (con->type == CT_WORKSPACE) {
-        DLOG("Not moving workspace\n");
-        return;
+        /* If there is already a split container for the windows in the workspace, use that. */
+        /* Otherwise, make a split container for them and move that split container. This way, moving multiple windows at a time from a workspace works as expected. */
+        const bool moves_focus = (focused == con);
+        Con *first_child = TAILQ_FIRST(&(con->nodes_head));
+        if (first_child != NULL && !con_is_leaf(first_child) && con_num_children(con) == 1) {
+            con = first_child;
+        } else {
+            if ((con = workspace_encapsulate(con)) == NULL) {
+                return;
+            }
+        }
+        if (moves_focus) {
+            con_focus(con);
+        }
     }
 
     if (con->fullscreen_mode == CF_GLOBAL) {


orestisfl:
Converted to a draft since it depends on https://github.com/i3/i3/pull/5529

orestisfl:
@slyshot please rebase onto `next`

orestisfl (changes requested):
Thank you for your pull request.

Since this is a purely behavioral change, we will need some tests that cover this behavior. Let me know if you need help with them.

Samueru-sama:
Will this be merged? I've been using patch ever since it came out and everything performs like it should.

stapelberg:
> Will this be merged? I've been using patch ever since it came out and everything performs like it should.

This PR won’t be merged without tests. Currently, we’re waiting for tests. If anyone wants to pick this up with tests, we can review.

Samueru-sama:
> > Will this be merged? I've been using patch ever since it came out and everything performs like it should.
> 
> This PR won’t be merged without tests. Currently, we’re waiting for tests. If anyone wants to pick this up with tests, we can review.

Alright I can do the tests in that case 👀

PR #5529, opened 2023-06-18
Remove focus workaround

slyshot:
The way `cmd_move_direction` currently preserves focus is by saving it beforehand, and applying it after. As far as I can tell, the only reason this is required is because `move_to_output_directed` has unexpected behavior in the case of moving multiple windows otherwise. Since it uses `workspace_show` in an unusual way(That is, with focus on the other workspace), it ends up focusing on one of the internal windows of the container, instead of the container itself.

I kept `workspace_show` because the comment before it seems to indicate that it does some important work, even though it doesn't update focus correctly in this case. I'd like this to be looked at from a more experienced eye to tell me whether I'm essentially undoing that work using `con_focus`, or if the `workspace_show` is necessary there versus some alternative method of getting those things done. I'll come back on that if nobody else does for a while.

This pull request is a precursor to https://github.com/i3/i3/pull/5521, since that needs to change focus in `tree_move` to work as expected.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -1578,7 +1578,6 @@ void cmd_move_direction(I3_CMD, const char *direction_str, long amount, const ch
     owindow *current;
     HANDLE_EMPTY_MATCH;
 
-    Con *initially_focused = focused;
     direction_t direction = parse_direction(direction_str);
 
     const bool is_ppt = mode && strcmp(mode, "ppt") == 0;
@@ -1610,12 +1609,6 @@ void cmd_move_direction(I3_CMD, const char *direction_str, long amount, const ch
             tree_move(current->con, direction);
             cmd_output->needs_tree_render = true;
         }
-    }
-
-    /* The move command should not disturb focus. con_exists is called because
-     * tree_move calls tree_flatten. */
-    if (focused != initially_focused && con_exists(initially_focused)) {
-        con_activate(initially_focused);
     }
 
     // XXX: default reply for now, make this a better reply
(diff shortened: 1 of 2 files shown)


orestisfl (changes requested):
Can you please check if this re-introduces https://github.com/i3/i3/issues/3518?

slyshot:
Testing it, it does not. If I recall correctly, when I was making this, If I left this https://github.com/i3/i3/blob/6db5af5ee77a23c99c177327c220b36c0a784d1e/src/move.c#L241-L243

without adding the second `con_focus` afterwards, it did reintroduce that, which is why I added it. My memories a little fuzzy about this, but I think it was because `ws_show` would descend focus, but never get to `focused` because the code set it to `old_ws`. Thanks.

orestisfl (approved):
Great, thank you. I've quickly tested that indeed focus is not disturbed in any case by applying this patch:

```diff
diff --git a/src/commands.c b/src/commands.c
index 009e98bf..53185855 100644
--- a/src/commands.c
+++ b/src/commands.c
@@ -1578,6 +1578,7 @@ void cmd_move_direction(I3_CMD, const char *direction_str, long amount, const ch
     owindow *current;
     HANDLE_EMPTY_MATCH;
 
+    Con *initially_focused = focused;
     direction_t direction = parse_direction(direction_str);
 
     const bool is_ppt = mode && strcmp(mode, "ppt") == 0;
@@ -1611,6 +1612,10 @@ void cmd_move_direction(I3_CMD, const char *direction_str, long amount, const ch
         }
     }
 
+    if (focused != initially_focused && con_exists(initially_focused)) {
+        ELOG("GREPME: focus moved\n");
+    }
+
     // XXX: default reply for now, make this a better reply
     ysuccess(true);
 }
```

and then running the test suite.

PR #5538, opened 2023-06-24
Improve documentation regarding tiling drag

HarmtH:
I was confused by the documentation, as it stated that I could use tiling drag by dragging the title bar. Although that is the default in the example configuration, it's not the default if you already have a configuration file.

Only later I figured out there was an option to configure it. So I think it will be helpful to point to that configuration option at the documentation of the tiling drag feature. Also added the default of the `tiling_drag` option to the documentation (see https://github.com/i3/i3/blob/stable/src/config.c#L231).


orestisfl:
Thank you for your pull request, indeed the documentation needs improvement. However, you should target your pull request towards the `next` branch, not the `stable` one as you did. Closing this and you can open a new one.

PR #5559, opened 2023-06-30
move  x-terminal-emulator to last

yutent:
for some terminal emulators does't provide a wrapper. 
so that the name can be displayed correctly.

### DIFF
--- a/i3-sensible-terminal
+++ b/i3-sensible-terminal
@@ -14,7 +14,7 @@
 # 2. Distribution-specific mechanisms come next, e.g. x-terminal-emulator
 # 3. The terminal emulator with best accessibility comes first.
 # 4. No order is guaranteed/desired for the remaining terminal emulators.
-for terminal in "$TERMINAL" x-terminal-emulator mate-terminal gnome-terminal terminator xfce4-terminal urxvt rxvt termit Eterm aterm uxterm xterm roxterm termite lxterminal terminology st qterminal lilyterm tilix terminix konsole kitty guake tilda alacritty hyper wezterm; do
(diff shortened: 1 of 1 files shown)


i3-sensible-terminal
```diff
@@ -14,7 +14,7 @@
 # 2. Distribution-specific mechanisms come next, e.g. x-terminal-emulator
 # 3. The terminal emulator with best accessibility comes first.
 # 4. No order is guaranteed/desired for the remaining terminal emulators.
-for terminal in "$TERMINAL" x-terminal-emulator mate-terminal gnome-terminal terminator xfce4-terminal urxvt rxvt termit Eterm aterm uxterm xterm roxterm termite lxterminal terminology st qterminal lilyterm tilix terminix konsole kitty guake tilda alacritty hyper wezterm; do
+for terminal in "$TERMINAL" mate-terminal gnome-terminal terminator xfce4-terminal urxvt rxvt termit Eterm aterm uxterm xterm roxterm termite lxterminal terminology st qterminal lilyterm tilix terminix konsole kitty guake tilda alacritty hyper wezterm x-terminal-emulator; do
```
orestisfl, line 17: From the comment:

```
# 2. Distribution-specific mechanisms come next, e.g. x-terminal-emulator
```

Why do we want to change this?

yutent: just for some terminal emulator package (eg. `blabla` ) does't provide a wrapper.
so that neofetch print the terminal name is `x-terminal-emulator` not `blabla`.

move the 'x-terminal emulator' to the end can solve this problem without any impact on the `i3` itself.

orestisfl: There will be an impact, since this will be a backwards-incompatible change. People that depend on x-terminal-emulator will start potentially getting a different terminal.

As always, this script is best-effort only, prefer setting the `$TERMINAL` variable.

I'll close this PR but nonetheless, thank you for your contribution.

yutent: I see, thanks.

PR #5591, opened 2023-07-23
Start i3 with a systemd service, if systemd is running and the service is loaded.

### LINKED ISSUE #5186: i3 session does not activate systemd graphical-session.target
## I'm submitting a…
<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

Note: The report is targeting Debian but applies to i3 in general.

Forwarded from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1003816.

systemd defines a special (user) target called [graphical-session.target](https://www.freedesktop.org/software/systemd/man/systemd.special.html#graphical-session.target)
which should be active when the user is logged into a graphical session.
Currently, when I'm logged into a default i3 session (entered via GDM), this
target is inactive:

```
  $ systemctl --user status graphical-session.target
  ● graphical-session.target - Current graphical user session
       Loaded: loaded (/usr/lib/systemd/user/graphical-session.target; static)
       Active: inactive (dead)
         Docs: man:systemd.special(7)
```

If I understood the clarifications in a
[thread](https://systemd-devel.freedesktop.narkive.com/J9pItAjq/graphical-session-targets-as-standard#post10)
on systemd's mailing list and the linked
[slides](https://people.debian.org/~mpitt/systemd.conf-2016-graphical-session.pdf)
correctly, this target should be activated by specifying the session in question
as a systemd service with `BindsTo=graphical-session.target` and launching this
service from the relevant `/usr/share/xsessions/*.desktop` instead of launching
the session directly.

In i3's case, the `/usr/share/xsessions/i3.desktop` session definition is
provided by the i3-wm package and just launches i3 directly. This makes perfect
sense, as systemd isn't mandatory on Debian and the aforementioned session-as-service
mechanism wouldn't work at all for people using SysVinit. But perhaps there
could be two session definitions, one with and one without an associated
systemd service?

An objection I expect is that users should just write their own session files
if they need this "feature", but the same would apply to the default
`/usr/share/xsessions/i3.desktop` file, which is only included for compatibility
with common display/login managers. Ditto for the argument that this is out of
scope for a window manager and gets into desktop environment territory.
Besides, it's not merely a missing feature but a bug (not necessarily in the i3
package but somewhere in the Debian distribution): Promises that are made in
systemd's
[documentation](https://www.freedesktop.org/software/systemd/man/systemd.special.html#graphical-session.target)
are not kept when using distribution-provided graphical session definitions.

Unfortunately, the only precedent for this type of session file in other window
manager or desktop environment packages I could find is in the gnome-session-
common package (apt-file search session.target), so perhaps it would in fact be
better to provide this via a separate package? There are a bunch of packages
for the Openbox window manager that do nothing but provide sessions for running
it along with the non-WM processes of common desktop environments (openbox-
gnome-session, openbox-lxde-session, openbox-kde-session, ...), so a package
providing only a session file for running i3 in a systemd-compatible manner
doesn't sound totally unreasonable to me. This could then be suggested by the
i3 package.

davidsansome:
The systemd service activates `graphical-session.target` which in turn may activate other useful services such as the ssh-agent.

Fixes #5186

### DIFF
--- a/libexec/i3-session-start
+++ b/libexec/i3-session-start
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+# This script tries to start i3 via the i3-session systemd unit if it is loaded
+# and systemd is running, otherwise falls back to execing i3 directly.
+
+# Check if the i3-session systemd unit is loaded.
+UNIT_LOAD_STATE=$(systemctl show i3-session.service -P LoadState)
+if [[ $? != 0 || "${UNIT_LOAD_STATE}" != "loaded" ]]; then
+  # The unit is not loaded, or systemctl failed, so just start i3 normally.
+  exec i3
+fi
+
+# Start the i3 systemd session.
+systemctl start --user --wait i3-session.service
--- a/systemd/i3-session.service
+++ b/systemd/i3-session.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=i3 Session
+
+Requires=graphical-session-pre.target
+After=graphical-session-pre.target
+BindsTo=graphical-session.target
+Before=graphical-session.target
+
+[Service]
+Type=exec
+ExecStart=i3
--- a/meson.build
+++ b/meson.build
@@ -53,6 +53,9 @@ endif
 cdata.set('HAVE_STRNDUP', cc.has_function('strndup'))
 cdata.set('HAVE_MKDIRP', cc.has_function('mkdirp'))
 
+# Used in share/xsessions/i3.desktop.in to find i3-session-start.
+cdata.set('LIBEXECDIR', join_paths(get_option('prefix'), get_option('libexecdir')))
+
 # Instead of generating config.h directly, make vcs_tag generate it so that
 # @VCS_TAG@ is replaced.
 config_h_in = configure_file(
@@ -324,6 +327,8 @@ cairo_dep = dependency('cairo', version: '>=1.14.4', method: 'pkg-config')
 pangocairo_dep = dependency('pangocairo', method: 'pkg-config')
 glib_dep = dependency('glib-2.0', method: 'pkg-config')
 gobject_dep = dependency('gobject-2.0', method: 'pkg-config')
+
+systemd_dep = dependency('systemd', required: false)
 
 ev_dep = cc.find_library('ev')
 
@@ -597,11 +602,37 @@ install_subdir(
   install_dir: join_paths(get_option('sysconfdir'), 'i3'),
 )
 
-install_subdir(
-  'share/',
-  strip_directory: true,
+configure_file(
+  input: 'share/xsessions/i3.desktop.in',
+  output: 'i3.desktop',
+  configuration: cdata,
+  install: true,
+  install_dir: join_paths(get_option('datadir'), 'xsessions'),
+)
+
+install_data(
+  'share/applications/i3.desktop',
+  'share/xsessions/i3-with-shmlog.desktop',
+  rename: [
+    'applications/i3.desktop',
+    'xsessions/i3-with-shmlog.desktop',
+  ],
   install_dir: get_option('datadir'),
 )
+
+install_subdir(
+  'libexec/',
+  strip_directory: true,
+  install_dir: get_option('libexecdir'),
+)
+
+if systemd_dep.found()
+  install_subdir(
+    'systemd/',
+    strip_directory: true,
+    install_dir: systemd_dep.get_pkgconfig_variable('systemduserunitdir'),
+  )
+endif
 
 install_headers(
   'include/i3/ipc.h',
--- a/share/xsessions/i3.desktop.in
+++ b/share/xsessions/i3.desktop.in
@@ -1,8 +1,8 @@
 [Desktop Entry]
 Name=i3
 Comment=improved dynamic tiling window manager
-Exec=i3
-TryExec=i3
+Exec=@LIBEXECDIR@/i3-session-start
+TryExec=@LIBEXECDIR@/i3-session-start
 Type=Application
 X-LightDM-DesktopName=i3
 DesktopNames=i3


systemd/i3-session.service
```diff
@@ -0,0 +1,11 @@
...
+Description=i3 Session
+
+Requires=graphical-session-pre.target
+After=graphical-session-pre.target
+BindsTo=graphical-session.target
+Before=graphical-session.target
+
+[Service]
+Type=exec
+ExecStart=i3
```
Thaodan, line 11: Units usually use absolute paths for exec.
Should it be `@bindir@/i3`?

Thaodan, line 11:
```suggestion
ExecStart=i3
ExecReload=/usr/bin/i3-msg restart
```
In case the user want's to reload or restart i3.
I'm referencing restart instead of reload here since systemd's restart doesn't allow 
the user to specific it's on command for restarts, there is only the option to define
a reload, start or stop command.

libexec/i3-session-start
faivirol, line 1: If run a second time before the first is done, it will hang and wait for the same service. If not implementing support for multiple instances, a check and message for if the service is already running would at least show that the single instance limit is intended.

systemd/i3-session.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=i3 Session
+
+Requires=graphical-session-pre.target
+After=graphical-session-pre.target
+BindsTo=graphical-session.target
+Before=graphical-session.target
+
+[Service]
```
faivirol, line 9: `Slice=session.slice` should be set in this case according to [`systemd.special`](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#session.slice)

systemd/i3-session.service
faivirol, line 1: Where does the service get `DISPLAY` and `XAUTHORITY` from? I can get it to work using `systemctl --user import-environment DISPLAY XAUTHORITY` in `.xinitrc` or adding `ENVIRONMENT=...` to the service, but I don't see how the service can work as is, although I might be missing something.

Although I don't often have two instances open, it does happen. My suggestion is to make the unit into a template and use the instance variable like `Environment=DISPLAY=%i`. This will support multiple instances and nonstandard `DISPLAY` values, making `i3-session-start` a drop-in replacement for running `i3`.

As for `XAUTHORITY`, asking the user to run/script `systemctl --user import-environment XAUTHORITY` before `i3-session-start` seems like a good choice since it is global for a user/systemd user instance (unlike `DISPLAY`), although a fallback to `~/.Xauthority` would probably be enough for most. For reference, [Arch already sets `XAUTHORITY` this way](https://unix.stackexchange.com/a/653101), although it is only [suggested in a note](https://wiki.archlinux.org/title/Xinit#xinitrc) that you actually include this in your user's `.xinitrc`.

faivirol (commented):
With this PR, replacing autostart scripts with services becomes feasible. I would add documentation about something like this:

```
[Unit]
Description=picom
BindsTo=i3@%i.service
After=i3@%i.service

[Service]
Environment=DISPLAY=%i
ExecStart=picom <args>

[Install]
WantedBy=i3@.service
```

Placed at `$XDG_CONFIG_HOME/systemd/user/picom.service`, it could then be enabled with `systemctl --user enable picom@` .

Alternatively, if sticking with the current service in the PR and not implementing instances, a simpler service like this could be used:

```
[Unit]
Description=picom
BindsTo=i3.service
After=i3.service

[Service]
ExecStart=picom <args>

[Install]
WantedBy=i3.service
```

The enabling would then also be without the at sign.

systemd/i3-session.service
faivirol: Actually, this might not be the best since it seems like all processes i3 spawns, including applications, then fall under the session slice.

Thaodan:
I think applications should be started by the session-manager not by the
window-manager.

I don't think ***@***.*** is needed as there wouldn't be more than one
windoww-manager in the session.
Using `Install=windwmanager.servcice` in i3.service and
`After=windowmanager.service` in picom.service would make the unit more
universal for e.g. upstreaming of the picom service file.

faivirol:
> I think applications should be started by the session-manager not by the
window-manager.

What do you mean by this?

> I don't think \*\*\*@\*\*\*.\*\*\* is needed as there wouldn't be more than one
windoww-manager in the session.

After sleeping on it, this sounds more reasonable. The extra complexity is probably not justifiable for such a niche case. Using `systemd --user import-environment DISPLAY` would remove the need to provide it through an instance variable or hardcode it. If a user needs to run multiple instances, they can always fall back to executing the `i3` binary directly.

I am curious though about how you provide `DISPLAY` and `XAUTHORITY` to the service you posted in #5186, @Thaodan. Did you use `import-environment` as well?

> Using `Install=windwmanager.servcice` in i3.service and
`After=windowmanager.service` in picom.service would make the unit more
universal for e.g. upstreaming of the picom service file.

Yep, seems obvious in hindsight. Although `graphical-session.target` works as well (for `BindsTo`, `After` and `WantedBy`), so I think that's better than creating a new `windowmanager.service` standard.

Thaodan:
faivirol ***@***.***> writes:

>> I think applications should be started by the session-manager not by the
> window-manager.
>
> What do you mean by this?
>
>> I don't think \*\*\*@\*\*\*.\*\*\* is needed as there wouldn't be more than one
> windoww-manager in the session.
>
> After sleeping on it, this sounds more reasonable. The extra
> complexity is probably not justifiable for such a niche case. Using
> `systemd --user import-environment DISPLAY` would remove the need to
> provide it through an instance variable or hardcode it. If a user
> needs to run multiple instances, they can always fall back to
> executing the `i3` binary directly.
>
> I am curious though about how you provide `DISPLAY` and `XAUTHORITY`
> to the service you posted in #5186, @Thaodan. Did you use
> `import-environment` as well?

I only used the existing session started by KDE. The KDE session starter
first sets up it's own config, runs env scripts and then also imports the
systemd environment:
https://invent.kde.org/plasma/plasma-workspace/-/blob/master/startkde/startplasma-x11.cpp?ref_type=heads#L78


>> Using `Install=windwmanager.servcice` in i3.service and
> `After=windowmanager.service` in picom.service would make the unit more
> universal for e.g. upstreaming of the picom service file.
>
> Yep, seems obvious in hindsight. Although `graphical-session.target`
> works as well (for `BindsTo`, `After` and `WantedBy`), so I think
> that's better than creating a new `windowmanager.service` standard.
It's the same technique used to have multiple units to provide the same
service such as a display-manager.

systemd/i3-session.service
kjkent: > Where does the service get `DISPLAY` and `XAUTHORITY` from? 

`/etc/X11/xinit/xinitrc.d/50-systemd-user.sh` should be sourced by the user's `$XINITRC` (default `~/.xinitrc`, falling back to `/etc/X11/xinit/xinitrc` if neither is found), which is sourced by `startx`/`xinit`. The fallback `xinitrc` contains:

```Sh
# ...
# start some nice programs
if [ -d /etc/X11/xinit/xinitrc.d ] ; then
 for f in /etc/X11/xinit/xinitrc.d/?*.sh ; do
  [ -x "$f" ] && . "$f"
 done
 unset f
fi
```
As per the [Arch Wiki](https://wiki.archlinux.org/title/Systemd/User#Environment_variables):

>Units started by user instance of systemd do not inherit any of the [environment variables](https://wiki.archlinux.org/title/Environment_variables) set in places like .bashrc etc. There are several ways to set environment variables for them ...

There's a bit in _2.1.1_ talking about `DISPLAY` and `XAUTHORITY` in particular.

I hope some of this is useful and not just stuff you already know!

**Edit:** I've just read that `/etc/X11/xinit/xinitrc.d/50-systemd-user.sh` might be an Arch-implemented file (although it contains a license specifying it as part of systemd...), so here's its contents for convenience:

```Sh
#!/bin/sh
#  SPDX-License-Identifier: LGPL-2.1-or-later
#
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.

systemctl --user import-environment DISPLAY XAUTHORITY

if command -v dbus-update-activation-environment >/dev/null 2>&1; then
    dbus-update-activation-environment DISPLAY XAUTHORITY
fi
```

budRich:
i3-session-start is a **bash** script (`[[ $? != 0 || "${UNIT_LOAD_STATE}" != "loaded" ]]`) . Use appropriate shbang: `#!/usr/bin/env bash` . Also as in discussion about needing environment variables, before `systemctl start --user --wait i3-session.service` you could do: `dbus-update-activation-environment --systemd DISPLAY XAUTHORITY` .

Also the service type can be changed to `notify` since i3 implements the feature (#4454).


```
[Service]
Type=notify
ExecStart=i3
ExecStopPost=/bin/systemctl --user stop --no-block graphical-session.target
```

I use ExecStopPost to shut down other graphical programs started by systemd gracefully when i3 stops.

In my experience (i have been starting i3 with systemd for years), there are more environment variables one probably want. **PATH**, without it, relative paths in the i3 config file will not be found f.i. `exec --no-startup-id dex --autostart --environment i3` (from the default config) (dexplain to a new user why to init programs need to be involved to start i3...)  

`dbus-update-activation-environment --systemd DISPLAY XAUTHORITY` in my personal setup is actually: `dbus-update-activation-environment --systemd --all` , which will import **all** environment variables to **systemd**. But this is not proper to do security wise.  

`systemctl --user import-environment DISPLAY XAUTHORITY` <- this command and `dbus-update-activation-environment --systemd DISPLAY XAUTHORITY` are equivalent i think.

You will **break** lots of peoples configs if they start i3 without access to the PATH environment variable, but you will also upset lots of people for **compromising security** if you do import the PATH environment variable. In my opinion this whole systemd stuff is headaches that the package maintainers could customize for their distros if they wanted to. I wouldn't be surprised if a large chunk of i3 user base doesn't even have systemd..

libexec/i3-session-start
```diff
@@ -0,0 +1,14 @@
...
+
+# Check if the i3-session systemd unit is loaded.
+UNIT_LOAD_STATE=$(systemctl show i3-session.service -P LoadState)
+if [[ $? != 0 || "${UNIT_LOAD_STATE}" != "loaded" ]]; then
+  # The unit is not loaded, or systemctl failed, so just start i3 normally.
+  exec i3
+fi
+
+# Start the i3 systemd session.
+systemctl start --user --wait i3-session.service
```
stapelberg, line 14: Can you import the environment variables here please? Depending on the Linux distribution, this is either already done by xsession, or required for things to work. (Doing it multiple times is not an issue.)

```bash
systemctl --user import-environment DISPLAY XAUTHORITY

if command -v dbus-update-activation-environment >/dev/null 2>&1; then
    dbus-update-activation-environment DISPLAY XAUTHORITY
fi
```

systemd/i3-session.service
```diff
@@ -0,0 +1,11 @@
...
+Type=exec
```
stapelberg, line 10: Use `Type=notify` here please so that i3 is fully started by the time `graphical-session.target` activates.

systemd/i3-session.service
stapelberg: Thanks for digging this up! I added a comment on libexec/i3-session-start and am marking this one here resolved.

systemd/i3-session.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=i3 Session
+
+Requires=graphical-session-pre.target
+After=graphical-session-pre.target
+BindsTo=graphical-session.target
+Before=graphical-session.target
```
stapelberg, line 7: I think we should add `Wants=xdg-desktop-autostart.target` as well. Then, we’ll also need to update the default config to not start `dex` for this purpose if `xdg-desktop-autostart.target` is active.

systemd/i3-session.service
stapelberg, line 1: Why not just `i3.service`?

systemd/i3-session.service
stapelberg: Agreed, we should use `@BINDIR@/i3` here

stapelberg (changes requested):
This looks pretty good overall, but there are a few changes to make. Are you still active and interested in making these changes?

If not, I can take over this PR. Ideally, let me know with a short message please :)

Thaodan:
Michael Stapelberg ***@***.***> writes:

> @stapelberg requested changes on this pull request.
>
>
>> @@ -0,0 +1,11 @@
> +[Unit]
> +Description=i3 Session
> +
> +Requires=graphical-session-pre.target
> +After=graphical-session-pre.target
> +BindsTo=graphical-session.target
> +Before=graphical-session.target
> +
> +[Service]
> +Type=exec
>
> Use `Type=notify` here please so that i3 is fully started by the time `graphical-session.target` activates.

Does i3 tell systemd when its ready? As long as i3 doesn't do that the
notify type doesn't work.

>> @@ -0,0 +1,11 @@
> +[Unit]
> +Description=i3 Session
> +
> +Requires=graphical-session-pre.target
> +After=graphical-session-pre.target
> +BindsTo=graphical-session.target
> +Before=graphical-session.target
>
> I think we should add `Wants=xdg-desktop-autostart.target` as well. Then, we’ll also need to update the default config to not start `dex` for this purpose if `xdg-desktop-autostart.target` is active.

That's part of the session manager which would be something else.

> On systemd/i3-session.service:
>
> Why not just `i3.service`?

IMHO i3wm would be the best as i3 is kinda short and i3-session implies
something else. The window manager is only part of the puzzle the
session management starts after the window manager is active.

A good example how units for the sesson launch are split up are the KDE
systemd units I think.

stapelberg:
> Does i3 tell systemd when its ready? As long as i3 doesn't do that the notify type doesn't work.

Yes: https://github.com/i3/i3/pull/4454

> > I think we should add `Wants=xdg-desktop-autostart.target` as well. Then, we’ll also need to update the default config to not start `dex` for this purpose if `xdg-desktop-autostart.target` is active.
>
> That's part of the session manager which would be something else.

Tell me more about this please. Which other unit would that be part of? 

> > Why not just `i3.service`?
>
> IMHO i3wm would be the best as i3 is kinda short and i3-session implies something else. 

So what? :) Our binary is also called i3. I think i3.service is consistent.

> The window manager is only part of the puzzle the session management starts after the window manager is active. A good example how units for the sesson launch are split up are the KDE systemd units I think.

Which aspect of the KDE systemd unit organization specifically do you think we should make sure to have? Where can one find/inspect these units?

stapelberg:
I’m taking over this issue with PR https://github.com/i3/i3/pull/6547 and will close this one (thank you!).

Thaodan:
Michael Stapelberg ***@***.***> writes:

> stapelberg left a comment (i3/i3#5591)
>
>> Does i3 tell systemd when its ready? As long as i3 doesn't do that
>> the notify type doesn't work.
>
> Yes: https://github.com/i3/i3/pull/4454

Oh I didn't notice thanks.

>> > I think we should add `Wants=xdg-desktop-autostart.target` as
>> > well. Then, we’ll also need to update the default config to not
>> > start `dex` for this purpose if `xdg-desktop-autostart.target` is
>> > active.
>>
>> That's part of the session manager which would be something else.
>
> Tell me more about this please. Which other unit would that be part of? 
>
>> > Why not just `i3.service`?
>>
>> IMHO i3wm would be the best as i3 is kinda short and i3-session
>> implies something else.
>
> So what? :) Our binary is also called i3. I think i3.service is consistent.

>> The window manager is only part of the puzzle the session management
>> starts after the window manager is active. A good example how units
>> for the sesson launch are split up are the KDE systemd units I
>> think.
>
> Which aspect of the KDE systemd unit organization specifically do you
> think we should make sure to have? Where can one find/inspect these
> units?

I answer to the 2nd and last comment in one.

The main aspect I think would make sense here is that each individual
part of the session startup has is own unit so that the organization
of processes can grouped under their own cgroup and that the startup
process is reflected into the units as it actually happening.

For example the session restore waits for the window manager to finnish
starting. The unit for said session restore can then be started
after which means systemd will make sure that i3 is active
and ready before starting it. As for example can be seen
in the KDE Session Manager (KSM) systemd unit:
https://invent.kde.org/plasma/plasma-workspace/-/blob/master/ksmserver/plasma-ksmserver.service.in

Personally I have this unit for picom which uses just that condition.
First my i3 unit:
# /home/bidar/.local/etc/systemd/user/i3wm.service
[Unit]
Description=A tiling window manager for X11
After=plasma-kcminit.service
PartOf=graphical-session.target
Documentation=man:i3(1)

[Service]
ExecStart=/usr/bin/i3 --replace
ExecReload=/usr/bin/i3-msg restart
Slice=session.slice
Restart=on-failure

[Install]
Alias=plasma-kwin_x11.service
Alias=window-manager.service

Note: the first install alias is so I can replace kwin
in the plasma session, it has nothing to do with this discussion.
Site note: I think that i3 is part of the graphical-session but not
executed before it, since for example Xorg has to be up before i3 can start.

# /home/bidar/.local/etc/systemd/user/picom.service
[Unit]
Description=Picom compositor fork
Documentation=man:picom(1)
After=window-manager.service
PartOf=graphical-session.target
# Don't try to restart after 60 seconds
StartLimitIntervalSec=60

[Service]
ExecStart=/usr/bin/picom
ExecReload=/usr/bin/killall picom
# Try to restart no matter how the process ended
Restart=always
# Restart in 5 second intervals
RestartSec=5
# Try to restart up to 3 times
StartLimitBurst=3

[Install]
Alias=compositor.service
WantedBy=graphical-session.target

The After= condition in the unit section of the file above
will make sure that picom isn't started before i3 has finished
starting up. 

The main point is that each stage of the session startup can be split up
in its own unit so that in addition to cgroup settings being applied to
correctly that there can be also no race conditions.

The final startup order could be something like this:

1. start-i3session program initializes basic settings, checks that x11
 is active, reads startup config, set xdg variables etc.
 Check KDE's start kde program for reference:
 https://invent.kde.org/plasma/plasma-workspace/-/blob/master/startkde/startplasma-x11.cpp

2. The start session program then calls systemd to active the
   i3-session.target which requires all the services/targets required
   for the i3 session.
3. Systemd starts i3-session target.
   Check the plasma-workspace target for reference:
   https://invent.kde.org/plasma/plasma-workspace/-/blob/master/startkde/systemd/plasma-workspace.target
4. Inside that target the services for actual session to startup run.
   i.e.
   1. Services which apply settings such as xsettingsd start.
   2. i3 wm starts
   3. Any background service that needs a window manager or should start
   after the window manager i.e. such as a compositor start.
   4. The session starts, restarts previous applications. Systemds
      xdg-deskto-autostart.target starts.
      If installed a program to restore the previous sessions window
      layout could start.

5. start-i3session waits for the session to end and cleans
   the session up after the user requests to logout.

How these steps exactly are executed doesn't matter but a small
wrapper/script program is essential I think to ensure that the
environment is cleanly setup, the session target is executed and the
session is cleaned up.

The systemd website goes into some of this process:
https://systemd.io/DESKTOP_ENVIRONMENTS/


The systemd units can be found in their respective package,
that kinda makes them hard to navigate at once.

This Github shows them:
https://github.com/search?q=org%3AKDE+.service.in&type=code

This blogpost goes in detail about the session startup process in KDE
with systemd:
https://blog.davidedmundson.co.uk/blog/plasma-and-the-systemd-startup/

I hope I have explained everything detailed but also didn't go
overboard.

PR #5648, opened 2023-08-31
Fix dead link for Modern Perl book

### LINKED ISSUE #5523: testsuite doc. stale perl book URL
The following link is stale:

https://github.com/i3/i3/blob/a95870120ce30843d95bdc5d87d26973789ed947/docs/testsuite#L40

waterkip:
Closes: #5523

### DIFF
--- a/docs/testsuite
+++ b/docs/testsuite
@@ -36,8 +36,8 @@ that, but it will also be useful for every future change.
 
 Apart from this document, you should also have a look at:
 
-1. The "Modern Perl" book, which can be found at
-   http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+1. The "Modern Perl" book:
+   https://i3wm.org/downloads/modern_perl_a4.pdf
 2. The latest Perl documentation of the "i3test" (general testcase setup) and
    "i3test::Test" (additional test instructions) modules:
    https://build.i3wm.org/docs/lib-i3test.html respectively
--- a/testcases/new-test
+++ b/testcases/new-test
@@ -53,7 +53,7 @@ my $header = <<'EOF';
 # • https://build.i3wm.org/docs/ipc.html
 #   (or docs/ipc)
 #
-# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+# • https://i3wm.org/downloads/modern_perl_a4.pdf
 #   (unless you are already familiar with Perl)
 #
 # TODO: Description of this file.
--- a/testcases/t/001-tile.t
+++ b/testcases/t/001-tile.t
@@ -11,7 +11,7 @@
 # • https://build.i3wm.org/docs/ipc.html
 #   (or docs/ipc)
 #
-# • http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
+# • https://i3wm.org/downloads/modern_perl_a4.pdf
 #   (unless you are already familiar with Perl)
 
 use i3test;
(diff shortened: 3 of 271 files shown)


orestisfl (commented):
Given that this book is not available online and seems quite old, maybe it's better to just let test writers select their own Perl resources instead of directing them to potentially outdated info? WDYT @stapelberg

stapelberg:
> Given that this book is not available online and seems quite old, maybe it's better to just let test writers select their own Perl resources instead of directing them to potentially outdated info? WDYT @stapelberg

Hmm, not sure. I think the book is still valuable, though building it from source is not something that people will do, I think. Could we build it from source and distribute a static PDF version of it, or does the license not permit that?

orestisfl:
The book might be useful indeed but instructing contributors on how to learn a language is not really typical for an open source project. Besides, it might be discouraging people rather than helping them; it creates the impression that one has to read a full book before actually contributing to i3 ("Please read the following documents before working on tests"). I believe that for the scope of what we are doing, great knowledge of perl is not needed to contribute to i3. I've personally never read the book.

waterkip:
> Could we build it from source and distribute a static PDF version of it, or does the license not permit that?

You can, the works has a Creative Commons Attribution-Noncommercial-Share
Alike 3.0 United States License.

Although I agree with @orestisfl's statement: The book might be useful indeed but instructing contributors on how to learn a language is not really typical for an open source project.

stapelberg:
> You can, the works has a Creative Commons Attribution-Noncommercial-Share Alike 3.0 United States License.

Okay. I have saved a copy of the PDF, you can refer to it as https://i3wm.org/downloads/modern_perl_a4.pdf

> Although I agree with @orestisfl's statement: The book might be useful indeed but instructing contributors on how to learn a language is not really typical for an open source project.

Yes, but Perl isn’t really a typical language for an open source project either (these days). I’m open to changing the wording to seem less daunting, but I do think it is good to provide a few pointers regarding known-good resources to get into Perl for people who prefer that approach. Of course, everyone is free to disregard the advice and just jump into the code :)

waterkip:
> > You can, the works has a Creative Commons Attribution-Noncommercial-Share Alike 3.0 United States License.
> 
> Okay. I have saved a copy of the PDF, you can refer to it as https://i3wm.org/downloads/modern_perl_a4.pdf

I updated the link in the PR.

> Yes, but Perl isn’t really a typical language for an open source project either (these days). I’m open to changing the wording to seem less daunting, but I do think it is good to provide a few pointers regarding known-good resources to get into Perl for people who prefer that approach. Of course, everyone is free to disregard the advice and just jump into the code :)

Perl is a great language :)

docs/testsuite
```diff
@@ -36,8 +36,8 @@ that, but it will also be useful for every future change.
 
 Apart from this document, you should also have a look at:
 
-1. The "Modern Perl" book, which can be found at
-   http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf
```
orestisfl, line 40: This link is pasted in all testcases as well, can you replace it there as well?

You can easily do it with [sd](https://github.com/chmln/sd):

```
sd -s 'http://onyxneon.com/books/modern_perl/modern_perl_a4.pdf' 'https://i3wm.org/downloads/modern_perl_a4.pdf' $(git ls-files) 
```

or sed equivalent

waterkip: Done!

orestisfl (approved):
Thanks!

PR #5659, opened 2023-09-05
Remanage window after urgency flag change

### LINKED ISSUE #5658: `for_window [urgent=latest] ...` does not work
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Reproduction Instructions

1. Add this in your config
```
for_window [urgent=latest] focus
```
2. Make an existing window urgent
3. **Expected result:** focus to switch to the urgent window **Actual result:** nothing happens, rule is not triggered

## Environment
i3 4.22

Inspired by https://github.com/i3/i3/discussions/5656

orestisfl:
Fixes #5658

### DIFF
--- a/release-notes/bugfixes/5-for-window-urgent
+++ b/release-notes/bugfixes/5-for-window-urgent
@@ -1,0 +1,1 @@
+fix for_window not working with urgency flags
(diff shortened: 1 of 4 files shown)


stapelberg (commented):
Hmm, isn’t remanaging (which includes an X11 reparenting) somewhat disruptive? I’m concerned that this is either slow or has visible effects on screen. Have you verified the effect of the code in that regard?

orestisfl:
This is quite standard for other properties that trigger `for_window` rules. The remanaging does *not* happen except if there is a window that swallows. This should be a no-op in the common case.

I've pushed a commit that refactors `remanage_window` a bit to make it clear when the remanaging actually happens.

stapelberg (approved):
Great, thanks for making it clearer :)

PR #5663, opened 2023-09-09
Ignore Motif hints on window when swallowed by placeholder

### LINKED ISSUE #5438: Windows with _MOTIF_WM_HINTS disregard placeholder border styles when swallowed
## I'm submitting a…
<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
Windows with the property `_MOTIF_WM_HINTS` disregard the border styles of layout placeholders when swallowed.

## Expected Behavior
All windows should respect the border styles of placeholders when swallowed (as in v4.20).

## Reproduction Instructions
Using Chromium as an example application that sets `_MOTIF_WM_HINTS` and rxvt-unicode as one that does not:
```
tmp="$(mktemp)"
cat >"$tmp" <<EOF
{
	"layout": "splith",
	"nodes": [
		{
			"border": "none",
			"swallows": [
				{
					"class": "^Chromium$"
				}
			],
			"type": "con"
		},
		{
			"border": "none",
			"swallows": [
				{
					"class": "^URxvt$"
				}
			],
			"type": "con"
		}
	],
	"type": "con"
}
EOF

i3-msg append_layout "$tmp"
rm "$tmp"

chromium &
urxvt &
```

Chromium will have a border while urxvt will not.

## Environment
<pre>
Running i3 version: 4.22 (2023-01-02) (pid 803)
Loaded i3 config:
  /home/user/.config/i3/config (main) (last modified: Tue 28 Feb 2023 10:16:03 PM EST, 2059 seconds ago)

</pre>

<pre>
Config file URL: http://0x0.st/HzcS.txt
</pre>

<pre>
Log file URL: http://0x0.st/Hzcu.log
</pre>

<pre>
- Are you using a compositor (e.g., xcompmgr or compton): picom
</pre>

njhanley:
A layout placeholder's border style should be preserved when swallowing a matching window, regardless of any Motif hints.

Closes #5438

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -371,10 +371,12 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     }
 
     DLOG("new container = %p\n", nc);
+    bool swallowed_by_placeholder = false;
     if (nc->window != NULL && nc->window != cwindow) {
         if (!restore_kill_placeholder(nc->window->id)) {
             DLOG("Uh?! Container without a placeholder, but with a window, has swallowed this to-be-managed window?!\n");
         } else {
+            swallowed_by_placeholder = true;
             /* Remove remaining criteria, the first swallowed window wins. */
             _remove_matches(nc);
         }
@@ -518,7 +520,7 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
         }
     }
 
-    if (has_mwm_hints) {
+    if (has_mwm_hints && !swallowed_by_placeholder) {
         DLOG("MOTIF_WM_HINTS specifies decorations (border_style = %d)\n", motif_border_style);
         if (want_floating) {
             con_set_border_style(nc, motif_border_style, config.default_floating_border_width);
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -371,7 +371,9 @@ void manage_window(xcb_window_t window, xcb_get_window_attributes_cookie_t cooki
     }
 
     DLOG("new container = %p\n", nc);
+    bool swallowed_by_placeholder = false;
     if (nc->window != NULL && nc->window != cwindow) {
+        swallowed_by_placeholder = true;
```
orestisfl: Add this in the same block with the `_remove_matches` call.

orestisfl (commented):
LGTM

orestisfl:
Actually, considering this case:
1. window has Motif hints
2. placeholder has no border style

wouldn't it be valid to maintain the style dictated by Motif hints?

njhanley:
Could you give an example use case? I can't imagine wanting to restore a layout and not caring about the style.

There's currently no such thing as a placeholder with no border style; placeholders are containers and take the default style on creation. We could add the concept of implicit default style with a new `border_style_t` variant (`BS_UNDEFINED = -1` for example) or a placeholder-specific flag on `Con`, but I'm not convinced it's worth it.

PR #5158, opened 2022-09-24
Motif hints: Respect maximum border style configuration set by user

orestisfl:
See commit message

Waiting for user feedback, tests and source code comments will follow

Closes #3678
Fixes #5149

PR #5678, opened 2023-09-19
motif hints: respect maximum border style in append_layout

### LINKED ISSUE #5438: Windows with _MOTIF_WM_HINTS disregard placeholder border styles when swallowed

orestisfl:
Alternative approach to #5663 that just uses the existing `max_user_border_style` mechanism to ensure consistency with #5158.

@njhanley to answer your https://github.com/i3/i3/pull/5663#issuecomment-1724536318
> Could you give an example use case? I can't imagine wanting to restore a layout and not caring about the style.
> 
> There's currently no such thing as a placeholder with no border style; placeholders are containers and take the default style on creation. We could add the concept of implicit default style with a new `border_style_t` variant (`BS_UNDEFINED = -1` for example) or a placeholder-specific flag on `Con`, but I'm not convinced it's worth it.

A use case example would be:
```json
{
	"layout": "splith",
	"nodes": [
		{
			"swallows": [
				{
					"class": "^Chromium$"
				}
			],
			"type": "con"
		},
		{
			"border": "none",
			"swallows": [
				{
					"class": "^URxvt$"
				}
			],
			"type": "con"
		}
	],
	"type": "con"
}
```

where the user has not specified any border style for chromium, they merely said they want chromium left of urxvt. Forcing a different border style on Chromium compared to what the user would get if they launched chromium regularly is inconsistent.

Fixes #5438
Closes #5663

### DIFF
--- a/src/load_layout.c
+++ b/src/load_layout.c
@@ -40,7 +40,7 @@ static int num_marks;
 struct pending_marks {
     char *mark;
     Con *con_to_be_marked;
-} * marks;
+} *marks;
 
 /* This list is used for reordering the focus stack after parsing the 'focus'
  * array. */
--- a/src/load_layout.c
+++ b/src/load_layout.c
@@ -351,17 +351,18 @@ static int json_string(void *ctx, const unsigned char *val, size_t len) {
         } else if (strcasecmp(last_key, "border") == 0) {
             char *buf = NULL;
             sasprintf(&buf, "%.*s", (int)len, val);
-            if (strcasecmp(buf, "none") == 0)
-                json_node->border_style = BS_NONE;
-            else if (strcasecmp(buf, "1pixel") == 0) {
-                json_node->border_style = BS_PIXEL;
+            if (strcasecmp(buf, "none") == 0) {
+                json_node->max_user_border_style = json_node->border_style = BS_NONE;
+            } else if (strcasecmp(buf, "1pixel") == 0) {
+                json_node->max_user_border_style = json_node->border_style = BS_PIXEL;
                 json_node->current_border_width = 1;
-            } else if (strcasecmp(buf, "pixel") == 0)
-                json_node->border_style = BS_PIXEL;
-            else if (strcasecmp(buf, "normal") == 0)
-                json_node->border_style = BS_NORMAL;
-            else
+            } else if (strcasecmp(buf, "pixel") == 0) {
+                json_node->max_user_border_style = json_node->border_style = BS_PIXEL;
+            } else if (strcasecmp(buf, "normal") == 0) {
+                json_node->max_user_border_style = json_node->border_style = BS_NORMAL;
+            } else {
                 LOG("Unhandled \"border\": %s\n", buf);
+            }
             free(buf);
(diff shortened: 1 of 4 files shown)


njhanley:
This is a better approach than mine; I didn't understand exactly what `max_user_border_style` did. Thanks!

src/load_layout.c
```diff
@@ -351,17 +351,18 @@ static int json_string(void *ctx, const unsigned char *val, size_t len) {
...
                 json_node->current_border_width = 1;
-            } else if (strcasecmp(buf, "pixel") == 0)
-                json_node->border_style = BS_PIXEL;
-            else if (strcasecmp(buf, "normal") == 0)
-                json_node->border_style = BS_NORMAL;
-            else
+            } else if (strcasecmp(buf, "pixel") == 0) {
+                json_node->max_user_border_style = json_node->border_style = BS_PIXEL;
+            } else if (strcasecmp(buf, "normal") == 0) {
+                json_node->max_user_border_style = json_node->border_style = BS_NORMAL;
```
stapelberg, line 362: Would it be possible to just add `json_node->max_user_border_style = json_node->border_style` below the conditional to de-duplicate this code, or am I reading this wrong?

orestisfl: No, because of the `else {` condition. So we would either need a `goto` or a new variable to save the previous state, or I am missing some obvious alternative.

stapelberg: Fair enough. Let’s leave it as is.

stapelberg (approved):

PR #5679, opened 2023-09-20
Fix failing testcase when the font exists

zhmylove:
When the font from testcase's config exists on the system, load_configuration() does not fallback to a 'fixed' one resulting in a fail of this case.

The details are under another project's PR: https://github.com/stapelberg/X11-XCB/pull/5

### DIFF
--- a/testcases/t/317-bar-config-font-fallback.t
+++ b/testcases/t/317-bar-config-font-fallback.t
@@ -26,7 +26,8 @@ bar {
   # no font directive here, no i3-wide font configured (yet)
 }
 
-font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso10646-1
+# NOTE: iso99887 is invalid font specification, so it should always fallback
+font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso99887-9
 EOT
 
 my $i3 = i3(get_socket_path(0));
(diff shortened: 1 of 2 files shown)


testcases/t/317-bar-config-font-order.t
```diff
@@ -34,6 +34,12 @@ my $bars = $i3->get_bar_config()->recv;
 
 my $bar_id = shift @$bars;
 my $bar_config = $i3->get_bar_config($bar_id)->recv;
-is($bar_config->{font}, 'fixed', 'font ok');
```
stapelberg, line 37: Or maybe we could fix the test by specifying a font that does not exist, to always enforce the fallback?

zhmylove: Perhaps, it would be better to include both variants: this one triggering either valid font, or a fallback; and the second one with nonexistent font to trigger fallback anytime. I'll update the PR today

testcases/t/317-bar-config-font-fallback.t
```diff
@@ -0,0 +1,41 @@
...
+# Ticket: #5031
+# Bug still in: 4.20-105-g4db383e4
+use i3test i3_config => <<'EOT';
+# i3 config file (v4)
+
+bar {
+  # no font directive here, no i3-wide font configured (yet)
+}
+
+font -misc-fixed-medium-r-normal--13-120-75-75-C-70-iso99887-9
```
stapelberg, line 30: Can you add a comment above this line to point out that “iso99887-9” is invalid and hence this font won’t be found? It’s easy to miss otherwise when casually reading the test.

zhmylove: Sure, done

stapelberg (approved):

PR #5740, opened 2023-10-29
Grab the pointer asynchronously when executing a hotkey

### LINKED ISSUE #5735: The pointer is grabbed after executing a hotkey
Hello!

I have something like this in my .config/i3/config:

```
bindsym z exec --no-startup-id /tmp/a.out
```

a.out has the following source code:
also one needs to have notify-send installed

```c
#include <stdio.h>
#include <stdlib.h>
#include <X11/Xlib.h>

int main( int argc, char ** args ){
  Display * dpy = XOpenDisplay(NULL);
  char buf[100];
  int rc;

  rc = XGrabPointer(dpy, DefaultRootWindow(dpy), True, PointerMotionMask, GrabModeAsync, GrabModeAsync, None, None, CurrentTime);
  snprintf(buf, sizeof(buf), "notify-send 'rc: %i\nGrabSuccess: %i\nGrabFrozen: %i'", rc, GrabSuccess, GrabFrozen);
  system(buf);

  return 0;
}
```
It's compiled like so:
```sh
cc file.c $(pkg-config --cflags --libs x11)
```
And the the notify-send output goes like this:
```
rc: 4
GrabSuccess: 0
GrabFrozen: 4
```

What this amounts to is that shortly after executing a hotkey, the program being started can't use XGrabPointer, because it will return GrabFrozen because the pointer is already grabbed by i3. Despite the hotkey being bound to a key, it's the pointer that I'm unable to XGrabPointer.

A possible workaround is to set --release in the config, but that feels somewhat sluggish.
Someone over at [#836](https://github.com/gnif/LookingGlass/issues/836) has a similar issue. I just wanted to bring your attention to this, since I don't know how relevant that pointer grab for i3's internal logic.

I was working on some sort of ring menu, that when bound to a hotkey, opens a window that has like a ring with different segments. When you move the cursor far enough in one direction, one of the options executes. Just some gimmick I suppose. Is there any chance of dropping that pointer grab or some more elegant workaround? Many thanks

My version of i3:
i3 version 4.22 (2023-01-02) © 2009 Michael Stapelberg and contributors

MikuChan03:
Grabing the pointer synchronously makes it so no pointer events are sent to other applications and a XGrabPointer will fail with GrabFrozen. This makes very little sense when executing a hotkey.

This fixes #5735

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -126,9 +126,9 @@ static bool binding_in_current_group(const Binding *bind) {
 
 static void grab_keycode_for_binding(xcb_connection_t *conn, Binding *bind, uint32_t keycode) {
     /* Grab the key in all combinations */
-#define GRAB_KEY(modifier)                                                                       \
-    do {                                                                                         \
-        xcb_grab_key(conn, 0, root, modifier, keycode, XCB_GRAB_MODE_SYNC, XCB_GRAB_MODE_ASYNC); \
+#define GRAB_KEY(modifier)                                                                        \
+    do {                                                                                          \
+        xcb_grab_key(conn, 0, root, modifier, keycode, XCB_GRAB_MODE_ASYNC, XCB_GRAB_MODE_ASYNC); \
     } while (0)
     const int mods = (bind->event_state_mask & 0xFFFF);
     DLOG("Binding %p Grabbing keycode %d with event state mask 0x%x (mods 0x%x)\n",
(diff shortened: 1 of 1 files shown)


orestisfl:
It took me some time to figure this patch out, I think it makes sense to author the commit message, context and history and have the correct blame information for potential regressions in the pushed patch. I'll close this PR and open a new one tomorrow.

MikuChan03:
You guys take project management seriously, huh?
That's very admirable! Sorry for making the PR myself, I should work on my patience.

PR #5818, opened 2023-12-11
Shrinking on a per-block basis

### LINKED ISSUE #4113: i3bar: progressively shorten blocks when bar is too wide
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior
When the predicted bar width is too wide, i3bar will switch every block over to its `short_text` mode.

## Desired Behavior
When the predicted bar width is too wide, i3bar will progressively switch blocks over to `short_text` mode until the width requirement is satisfied. The idea is to use short_text only for the minimum of needed blocks.

Glancing over the code it seems doable, so I would be willing to give it a shot if this FR is accepted.

Arc676:
Closes #4113.

This is a prototype implementation of the feature described in #4113. Instead of a global "use short mode" flag, individual blocks can be shortened to fit the content of the status bar. The user can assign priority values to the blocks such that certain blocks are shrunk before others, i.e. prioritizing that certain other blocks retain their full length where possible.

When blocks are shrunk, all blocks of equal or higher priority[^1] are also shrunk. The default priority value is 0. This means that a config with no priority values results in the same behavior as before, i.e. either all blocks are shortened or none of them are.

The feature described in #4113 can be more faithfully implemented using the same strategy by assigning ascending priority values if none are provided by the user in the config. This behavior may be implemented either in `i3bar` itself, or in the bar generators.

This PR serves as a RFC of sorts such that the feature itself and its possible implementations can be discussed. Further cleanup and testing (and possibly optimizations) are required before any possibility of merging, but the idea is to provide a concrete implementation to discuss.

[^1]: This design choice is due to the fact that some bar generators do not create a single JSON object for each logical block, in order to provide features such as clickable icons in blocks. This implementation uses the priority value as a proxy for block names in order to identify logical blocks and ensure that the same logical block is either all in full form or all in short form, even if multiple JSON objects are output to represent the same block. An alternative is to use the `name` property, but this is optional at protocol level.

### DIFF
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -46,8 +46,6 @@ struct status_block {
     i3String *full_text;
     i3String *short_text;
 
-    bool use_short_text;
-
     char *color;
     char *background;
     char *border;
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -46,6 +46,8 @@ struct status_block {
     i3String *full_text;
     i3String *short_text;
 
+    uint32_t length_priority;
+
     char *color;
     char *background;
     char *border;
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -47,6 +47,7 @@ struct status_block {
     i3String *short_text;
 
     bool use_short;
+    uint32_t render_length;
 
     char *color;
     char *background;
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -47,6 +47,7 @@ struct status_block {
     i3String *short_text;
 
     uint32_t length_priority;
+    bool use_short;
 
     char *color;
     char *background;
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -82,7 +82,7 @@ struct status_block {
     TAILQ_ENTRY(status_block) blocks;
 };
 
-extern TAILQ_HEAD(statusline_head, status_block) statusline_head;
+extern TAILQ_HEAD(statusline_head, status_block) statusline_head, statusline_sorted;
 
 #include "child.h"
 #include "ipc.h"
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -84,7 +84,6 @@ struct status_block {
 };
 
 extern TAILQ_HEAD(statusline_head, status_block) statusline_head;
-extern size_t block_count;
 
 #include "child.h"
 #include "ipc.h"
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -46,7 +46,6 @@ struct status_block {
     i3String *full_text;
     i3String *short_text;
 
-    uint32_t length_priority;
     bool use_short;
 
     char *color;
@@ -85,7 +84,6 @@ struct status_block {
 
 extern TAILQ_HEAD(statusline_head, status_block) statusline_head;
 extern size_t block_count;
-extern struct status_block **statusline_sorted;
 
 #include "child.h"
 #include "ipc.h"
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -46,6 +46,8 @@ struct status_block {
     i3String *full_text;
     i3String *short_text;
 
+    bool use_short_text;
+
     char *color;
     char *background;
     char *border;
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -83,7 +83,9 @@ struct status_block {
     TAILQ_ENTRY(status_block) blocks;
 };
 
-extern TAILQ_HEAD(statusline_head, status_block) statusline_head, statusline_sorted;
+extern TAILQ_HEAD(statusline_head, status_block) statusline_head;
+extern size_t block_count;
+extern struct status_block **statusline_sorted;
 
 #include "child.h"
 #include "ipc.h"
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -21,6 +21,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
@@ -211,6 +212,9 @@ static int stdin_start_map(void *context) {
     ctx->block.border_bottom = 1;
     ctx->block.border_left = 1;
 
+    /* By default, no block takes priority for short mode */
+    ctx->block.length_priority = 0;
+
     return 1;
 }
 
@@ -326,6 +330,10 @@ static int stdin_integer(void *context, long long val) {
     }
     if (strcasecmp(ctx->last_map_key, "border_left") == 0) {
         ctx->block.border_left = (uint32_t)val;
+        return 1;
+    }
+    if (strcasecmp(ctx->last_map_key, "length_priority") == 0) {
+        ctx->block.length_priority = (uint32_t)val;
         return 1;
     }
 
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -76,6 +76,7 @@ parser_ctx parser_context;
 struct statusline_head statusline_head = TAILQ_HEAD_INITIALIZER(statusline_head);
 /* Used temporarily while reading a statusline */
 struct statusline_head statusline_buffer = TAILQ_HEAD_INITIALIZER(statusline_buffer);
+size_t block_count;
 
 int child_stdin;
 
@@ -103,13 +104,16 @@ void clear_statusline(struct statusline_head *head, bool free_resources) {
     }
 }
 
-static void copy_statusline(struct statusline_head *from, struct statusline_head *to) {
+static size_t copy_statusline(struct statusline_head *from, struct statusline_head *to) {
     struct status_block *current;
+    size_t count = 0;
     TAILQ_FOREACH (current, from, blocks) {
         struct status_block *new_block = smalloc(sizeof(struct status_block));
         memcpy(new_block, current, sizeof(struct status_block));
         TAILQ_INSERT_TAIL(to, new_block, blocks);
+        count++;
     }
+    return count;
 }
 
 /*
@@ -369,7 +373,7 @@ static int stdin_end_map(void *context) {
 static int stdin_end_array(void *context) {
     DLOG("copying statusline_buffer to statusline_head\n");
     clear_statusline(&statusline_head, true);
-    copy_statusline(&statusline_buffer, &statusline_head);
+    block_count = copy_statusline(&statusline_buffer, &statusline_head);
 
     DLOG("dumping statusline:\n");
     struct status_block *current;
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -379,7 +379,7 @@ static int stdin_end_map(void *context) {
     return 1;
 }
 
-int descending_length_priority(const void *a, const void *b) {
+static int descending_length_priority(const void *a, const void *b) {
     return (*(struct status_block **)b)->length_priority - (*(struct status_block **)a)->length_priority;
 }
 
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -8,6 +8,7 @@
  *
  */
 #include "common.h"
+#include "queue.h"
 #include "yajl_utils.h"
 
 #include <ctype.h> /* isspace */
@@ -78,6 +79,8 @@ struct statusline_head statusline_head = TAILQ_HEAD_INITIALIZER(statusline_head)
 /* Used temporarily while reading a statusline */
 struct statusline_head statusline_buffer = TAILQ_HEAD_INITIALIZER(statusline_buffer);
 size_t block_count;
+/* Blocks sorted by length priority */
+struct statusline_head statusline_sorted = TAILQ_HEAD_INITIALIZER(statusline_sorted);
 
 int child_stdin;
 
@@ -188,6 +191,7 @@ static int stdin_start_array(void *context) {
     // the blocks are still used by statusline_head, so we won't free the
     // resources here.
     clear_statusline(&statusline_buffer, false);
+    clear_statusline(&statusline_sorted, false);
     return 1;
 }
 
@@ -371,6 +375,18 @@ static int stdin_end_map(void *context) {
     }
 
     TAILQ_INSERT_TAIL(&statusline_buffer, new_block, blocks);
+
+    if (new_block->length_priority > TAILQ_FIRST(&statusline_sorted)->length_priority) {
+        TAILQ_INSERT_HEAD(&statusline_sorted, new_block, blocks);
+    } else {
+        struct status_block *block;
+        TAILQ_FOREACH(block, &statusline_sorted, blocks) {
+            if (new_block->length_priority > TAILQ_NEXT(block, blocks)->length_priority) {
+                TAILQ_INSERT_AFTER(&statusline_sorted, block, new_block, blocks);
+                break;
+            }
+        }
+    }
     return 1;
 }
 
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -78,9 +78,9 @@ parser_ctx parser_context;
 struct statusline_head statusline_head = TAILQ_HEAD_INITIALIZER(statusline_head);
 /* Used temporarily while reading a statusline */
 struct statusline_head statusline_buffer = TAILQ_HEAD_INITIALIZER(statusline_buffer);
-size_t block_count;
+size_t block_count = 0;
 /* Blocks sorted by length priority */
-struct status_block **statusline_sorted;
+struct status_block **statusline_sorted = NULL;
 
 int child_stdin;
 
@@ -191,7 +191,6 @@ static int stdin_start_array(void *context) {
     // the blocks are still used by statusline_head, so we won't free the
     // resources here.
     clear_statusline(&statusline_buffer, false);
-    FREE(statusline_sorted);
     return 1;
 }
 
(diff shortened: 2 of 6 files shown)


i3bar/src/child.c
```diff
@@ -351,24 +366,42 @@ static int stdin_end_map(void *context) {
...
+    new_block->use_short = false;
     if (new_block->short_text != NULL)
         i3string_set_markup(new_block->short_text, new_block->pango_markup);
 
     TAILQ_INSERT_TAIL(&statusline_buffer, new_block, blocks);
+
     return 1;
 }
 
+int descending_length_priority(const void *a, const void *b) {
```
orestisfl:
```suggestion
static int descending_length_priority(const void *a, const void *b) {
```

to fix the pipeline

orestisfl (changes requested):
Thanks, I will give a better look later at some point

However, I would start already with no user configuration and no protocol changes, just the minimal approach that solves the core issue.

Arc676:
Branch has been updated to just shorten blocks with the same name instead of requiring a length priority.

i3bar/src/xcb.c
```diff
@@ -233,10 +234,42 @@ static uint32_t predict_statusline_length(bool use_short_text) {
...
+    struct status_block *block;
+    TAILQ_FOREACH (block, &statusline_head, blocks) {
+        /* If this block has no short form, there is no point in checking if we need to switch;
+         * however, we do have to compute the width at least once */
+        if (block->short_text == NULL && width > 0) {
+            continue;
+        }
+        width = predict_statusline_length();
+        if (width < max_length) {
+            break;
```
orestisfl: That's okay but I wonder if it would be easy to not call this function every time and just substract from the width instead.

i3bar/src/xcb.c
```diff
@@ -2101,27 +2131,30 @@ void draw_bars(bool unhide) {
...
             x_dest -= logical_px(config.padding.width);
 
-            draw_statusline(outputs_walk, clip_left, use_focus_colors, use_short_text);
+            draw_statusline(outputs_walk, clip_left, use_focus_colors);
             draw_util_copy_surface(&outputs_walk->statusline_buffer, &outputs_walk->buffer, 0, 0,
                                    x_dest, 0, visible_statusline_width, (int16_t)bar_height);
 
             outputs_walk->statusline_width = statusline_width;
-            outputs_walk->statusline_short_text = use_short_text;
+            outputs_walk->statusline_short_text = used_short_text;
```
orestisfl: If I am reading the code correctly `->statusline_short_text` is not needed anymore

i3bar/src/xcb.c
```diff
@@ -233,10 +234,42 @@ static uint32_t predict_statusline_length(bool use_short_text) {
...
+            *used_short_text = true;
+
+            if (block->name) {
+                struct status_block *other;
+                TAILQ_FOREACH (other, &statusline_head, blocks) {
+                    if (other->name && !strcmp(other->name, block->name)) {
+                        other->use_short = true;
+                    }
+                }
+            }
```
orestisfl: What's the benefit of doing that? Isn't it much more simpler and predictable to just put each block one-by-one in short mode?

i3status.conf
```diff
@@ -0,0 +1,86 @@
+general {
```
orestisfl: I assume this file was accidentally commited.

orestisfl (changes requested):
Some comments

i3status.conf
Arc676: Oops; yes this was created during testing and shouldn't be here.

i3bar/src/xcb.c
Arc676: This is to provide support for bar generators that don't output a single JSON object for each logical block, as described in the original MR post. If just the individual blocks are shortened, any blocks that are represented as multiple JSON objects might only be partially shortened.

I'm not sure to what extent `i3bar` should be aware of this. On the one hand, I can see the argument that this shouldn't be i3's responsibility. On the other hand, this change would otherwise break some environments with external bar generators (I'm not sure how common this practice is among i3-compatible bar generators).

At least for `i3status-rs`, which I use in my own config and have been using for testing, the idea behind the separation of single status bar blocks into multiple JSON objects is to make it possible to create clickable buttons that behave as single blocks and to add support for short form formats that don't necessarily match the long form formats. If `i3bar` doesn't provide some kind of support for this design, this feature will effectively be incompatible with bar generators that do this, unless their formatting options are restricted to allow single block outputs.

i3bar/src/xcb.c
Arc676: If I'm honest I didn't really understand what this variable was used for at all, I just tried to retain the same original meaning. As shown in the commit history, there was a long break between when I made most of the changes to the code and when I opened the MR; by now I've forgotten if I had somewhat figured out this variable back then.

I'll remove the member if it's no longer needed.

Arc676: Removed in [65f4cb1](https://github.com/i3/i3/pull/5818/commits/65f4cb1129a5f3bf63f9016a0b0192bafff2d481)

i3bar/src/xcb.c
Arc676: Addressed in [a2bd001](https://github.com/i3/i3/pull/5818/commits/a2bd001e12f1c10225e6bdfdf7a23b615c114f3f) and  [059f03f](https://github.com/i3/i3/pull/5818/commits/059f03f89f672c90d74cfce9ff6bac35546ac5d8) using a caching approach; the last render width is stored and this is used to determine how much the block is shortened and then recompute the bar width.

Arc676:
What is the i3 team's general attitude towards configurable shrinking? I find it to be a quite useful feature; I would expand my media block to use more screen space when available but this should also be the first one to shrink even though it's not the first block. Would you say that a PR including this feature would be rejected? I only ask since I'm mainly keeping the PR in draft state since I consider this feature part of the desired changes of the PR.

orestisfl:
> What is the i3 team's general attitude towards configurable shrinking? I find it to be a quite useful feature; I would expand my media block to use more screen space when available but this should also be the first one to shrink even though it's not the first block. Would you say that a PR including this feature would be rejected? I only ask since I'm mainly keeping the PR in draft state since I consider this feature part of the desired changes of the PR.

In general, we try to minimize any configuration options for i3 as it increases complexity. This feature seems beneficial as a default behavior without any configuration. Just the simplest logic of "if I shrink this block then text will fit so the next one doesn't need to be shrinked" should be a good ratio of value / effort. If many people want to configure this behavior further, we can open tickets for that and track support for that new feature but I don't want to merge something complicated straight away.

Arc676:
Makes sense I guess. I'll mark this as finished then and leave the customizable shrinking for a new issue/PR; I'll leave in the shrinking of blocks with the same name, since omitting this check will create issues for environments where this is relevant.

i3bar/src/xcb.c
```diff
@@ -184,46 +185,55 @@ static void draw_separator(i3_output *output, uint32_t x, struct status_block *b
...
-                case ALIGN_CENTER:
-                    render->x_offset = padding_width / 2;
-                    render->x_append = padding_width / 2 + padding_width % 2;
-                    break;
-            }
-        }
+    TAILQ_FOREACH (block, &statusline_head, blocks) {
+        uint32_t block_width = predict_block_length(block);
+        if (block_width == 0)
+            continue;
```
orestisfl: nit: use curly brackets
```suggestion
        if (block_width == 0) {
            continue;
        }
```

i3bar/src/xcb.c
```diff
@@ -184,46 +185,55 @@ static void draw_separator(i3_output *output, uint32_t x, struct status_block *b
...
-        struct status_block_render_desc *render = &block->full_render;
-        if (use_short_text && block->short_text != NULL) {
-            text = block->short_text;
-            render = &block->short_render;
+    if (i3string_get_num_bytes(text) == 0)
+        return 0;
+
+    render->width = predict_text_width(text);
+    if (block->border)
+        render->width += logical_px(block->border_left + block->border_right);
```
orestisfl:
```suggestion
    if (i3string_get_num_bytes(text) == 0) {
        return 0;
    }

    render->width = predict_text_width(text);
    if (block->border) {
        render->width += logical_px(block->border_left + block->border_right);
    }
```

i3bar/src/xcb.c
```diff
@@ -184,46 +185,55 @@ static void draw_separator(i3_output *output, uint32_t x, struct status_block *b
...
+                render->x_offset = padding_width / 2;
+                render->x_append = padding_width / 2 + padding_width % 2;
+                break;
         }
+    }
 
-        if (i3string_get_num_bytes(text) == 0)
-            continue;
+    block->render_length = render->width + render->x_offset + render->x_append;
+    return block->render_length;
```
orestisfl: If you need to store the length anyway, no need to return it. Change the function to `void` which will make it more clear that it modifies its arguments.

```suggestion
    block->render_length = render->width + render->x_offset + render->x_append;
```

i3bar/src/xcb.c
```diff
@@ -233,10 +243,44 @@ static uint32_t predict_statusline_length(bool use_short_text) {
...
+                struct status_block *other;
+                TAILQ_FOREACH (other, &statusline_head, blocks) {
+                    if (other->name && !strcmp(other->name, block->name)) {
+                        uint32_t full = other->render_length;
+                        other->use_short = true;
+                        uint32_t diff = full - predict_block_length(other);
+                        width -= diff;
+                    }
+                }
+            }
```
orestisfl: Even though I am still not convinced that we should include this snippet (I'll play around with the code later), I would start with adding a comment with an explanation on why this is here, i.e. the intention.

i3bar/src/xcb.c
```diff
@@ -233,10 +243,44 @@ static uint32_t predict_statusline_length(bool use_short_text) {
...
+
+    /* Progressively switch the blocks to short mode */
+    struct status_block *block;
+    TAILQ_FOREACH (block, &statusline_head, blocks) {
+        if (block->short_text == NULL) {
+            continue;
+        }
+        if (width < max_length) {
+            break;
+        } else if (!block->use_short) {
```
orestisfl:
```suggestion
        if (width < max_length) {
            break;
        }
        if (block->short_text == NULL || !block->use_short) {
            continue;
        }
        // rest of code without curly brackets
```

One less indentation level.

orestisfl (changes requested):
Thanks, this is looking good. I'll do another pass and run the patch but I like the code in its current form :+1: .

You also need to update the release notes:
```shell
echo 'i3bar: shrink text on a per-block basis' > release-notes/changes/3-shrink-i3bar-blocks
```

i3bar/src/xcb.c
Arc676: Good call; addressed in 3a4348404d193adbe3d6e08c23a02fc9cbb16112

i3bar/src/xcb.c
Arc676: Addressed in a563cbd811284f4acb4ae5f7439c57b091d250b9

i3bar/src/xcb.c
Arc676: Resolved in 65e9333644a382debe2a5a52802399785f1c5b87.

i3bar/src/xcb.c
Arc676: Resolved in 65e9333644a382debe2a5a52802399785f1c5b87.

i3bar/src/xcb.c
Arc676: Updated in 65e9333644a382debe2a5a52802399785f1c5b87. This doesn't seem to be consistently enforced, though. Perhaps the `.clang-format` rules could be updated to require use of braces even for single line blocks, e.g. with [`InsertBraces`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces).

i3bar/src/xcb.c
Arc676: Marking this thread as resolved since it's the same topic as [another thread](https://github.com/i3/i3/pull/5818#discussion_r1427400300).

i3bar/src/xcb.c
orestisfl: Addressed in #5882

i3bar/src/xcb.c
```diff
@@ -184,50 +185,103 @@ static void draw_separator(i3_output *output, uint32_t x, struct status_block *b
...
+            TAILQ_FOREACH (other, &statusline_head, blocks) {
+                if (other->name && !strcmp(other->name, block->name)) {
+                    uint32_t full = other->render_length;
+                    other->use_short = true;
+                    predict_block_length(other);
+                    uint32_t diff = full - other->render_length;
+                    width -= diff;
+                }
+            }
+        }
```
orestisfl:
```suggestion
static uint32_t switch_block_to_short(struct status_block *block) {
    /* Skip blocks that have no short form or are already in short form */
    if (block->short_text == NULL || block->use_short) {
        return 0;
    }
    uint32_t full = block->render_length;
    block->use_short = true;
    predict_block_length(block);
    return full - block->render_length;
}

static uint32_t adjust_statusline_length(uint32_t max_length) {
    uint32_t width = predict_statusline_length();

    /* Progressively switch the blocks to short mode */
    struct status_block *block;
    TAILQ_FOREACH (block, &statusline_head, blocks) {
        if (width < max_length) {
            break;
        }
        width -= switch_block_to_short(block);

        /* Provide support for representing a single logical block using multiple
         * JSON blocks: if one block is shortened, ensure that all other blocks
         * with the same name are also shortened such that the entire logical block uses
         * the short form text. */
        if (block->name) {
            struct status_block *other;
            TAILQ_FOREACH (other, &statusline_head, blocks) {
                if (other->name && !strcmp(other->name, block->name)) {
                    width -= switch_block_to_short(other);
                }
            }
        }
    }
```

Changes:
1. Re-use common function
2. In the second loop, avoid reducing twice in the case `block == other`
3. In the second loop, check that `other` has a short form

Diff:
```diff
diff --git a/i3bar/src/xcb.c b/i3bar/src/xcb.c
index 22dbb80a..68353d88 100644
--- a/i3bar/src/xcb.c
+++ b/i3bar/src/xcb.c
@@ -248,6 +248,17 @@ static uint32_t predict_statusline_length(void) {
     return width;
 }
 
+static uint32_t switch_block_to_short(struct status_block *block) {
+    /* Skip blocks that have no short form or are already in short form */
+    if (block->short_text == NULL || block->use_short) {
+        return 0;
+    }
+    uint32_t full = block->render_length;
+    block->use_short = true;
+    predict_block_length(block);
+    return full - block->render_length;
+}
+
 static uint32_t adjust_statusline_length(uint32_t max_length) {
     uint32_t width = predict_statusline_length();
 
@@ -257,28 +268,17 @@ static uint32_t adjust_statusline_length(uint32_t max_length) {
         if (width < max_length) {
             break;
         }
-        /* Skip blocks that have no short form or are already in short form */
-        if (block->short_text == NULL || block->use_short) {
-            continue;
-        }
-        uint32_t full = block->render_length;
-        block->use_short = true;
-        predict_block_length(block);
-        uint32_t diff = full - block->render_length;
-        width -= diff;
+        width -= switch_block_to_short(block);
 
-        /* Provide support for representing a single logical block using multiple JSON blocks:
-           if one block is shortened, ensure that all other blocks with the same name are also
-           shortened such that the entire logical block uses the short form text. */
+        /* Provide support for representing a single logical block using multiple
+         * JSON blocks: if one block is shortened, ensure that all other blocks
+         * with the same name are also shortened such that the entire logical block uses
+         * the short form text. */
         if (block->name) {
             struct status_block *other;
             TAILQ_FOREACH (other, &statusline_head, blocks) {
                 if (other->name && !strcmp(other->name, block->name)) {
-                    uint32_t full = other->render_length;
-                    other->use_short = true;
-                    predict_block_length(other);
-                    uint32_t diff = full - other->render_length;
-                    width -= diff;
+                    width -= switch_block_to_short(other);
                 }
             }
         }
```

orestisfl (changes requested):
Found a small bug

orestisfl:
To quickly resolve the conflicts, just pick your own version in the conflict and then re-run clang-format

i3bar/include/common.h
```diff
@@ -81,6 +84,7 @@ struct status_block {
 };
 
 extern TAILQ_HEAD(statusline_head, status_block) statusline_head;
+extern size_t block_count;
```
orestisfl:
```suggestion
```

Sorry, just noticed this is actually not used. Let's reverse it:

```diff
diff --git a/i3bar/include/common.h b/i3bar/include/common.h
index 84a56cf4..9c13d02f 100644
--- a/i3bar/include/common.h
+++ b/i3bar/include/common.h
@@ -84,7 +84,6 @@ struct status_block {
 };
 
 extern TAILQ_HEAD(statusline_head, status_block) statusline_head;
-extern size_t block_count;
 
 #include "child.h"
 #include "ipc.h"
diff --git a/i3bar/src/child.c b/i3bar/src/child.c
index 844d1a2b..e4c97072 100644
--- a/i3bar/src/child.c
+++ b/i3bar/src/child.c
@@ -78,7 +78,6 @@ parser_ctx parser_context;
 struct statusline_head statusline_head = TAILQ_HEAD_INITIALIZER(statusline_head);
 /* Used temporarily while reading a statusline */
 struct statusline_head statusline_buffer = TAILQ_HEAD_INITIALIZER(statusline_buffer);
-size_t block_count = 0;
 
 int child_stdin;
 
@@ -106,16 +105,13 @@ void clear_statusline(struct statusline_head *head, bool free_resources) {
     }
 }
 
-static size_t copy_statusline(struct statusline_head *from, struct statusline_head *to) {
+static void copy_statusline(struct statusline_head *from, struct statusline_head *to) {
     struct status_block *current;
-    size_t count = 0;
     TAILQ_FOREACH (current, from, blocks) {
         struct status_block *new_block = smalloc(sizeof(struct status_block));
         memcpy(new_block, current, sizeof(struct status_block));
         TAILQ_INSERT_TAIL(to, new_block, blocks);
-        count++;
     }
-    return count;
 }
 
 /*
@@ -377,7 +373,7 @@ static int stdin_end_map(void *context) {
 static int stdin_end_array(void *context) {
     DLOG("copying statusline_buffer to statusline_head\n");
     clear_statusline(&statusline_head, true);
-    block_count = copy_statusline(&statusline_buffer, &statusline_head);
+    copy_statusline(&statusline_buffer, &statusline_head);
 
     DLOG("dumping statusline:\n");
     struct status_block *current;
```

orestisfl (changes requested):
Sorry, I know it's annoying to fix these things one-by-one but every time I try to approve this, I notice something new.

i3bar/include/common.h
Arc676: No worries, that's the point of reviews. Better to catch them now than to find out later there are unused variables. It's on me as well since I added these for the customizable shrinking feature but didn't remove them afterwards.

orestisfl (approved):
Great, that's a nice improvement. Thank you for your persistence.

PR #5880, opened 2024-01-25
con_is_maximized: Fix case where parent is workspace

orestisfl:
See added test for simple example.

CC @sethpollen


sethpollen:
Thanks for fixing this!

stapelberg (approved):

PR #5889, opened 2024-01-28
smart_borders: Deprecate option

### LINKED ISSUE #5406: "smart_borders no_gaps" option is broken
<pre>
[x] Bug
</pre>

orestisfl:
This had pretty much identical behaviour to hide_edge_borders which made it confusing. The `hide_edge_borders smart_no_gaps` implementation has an extra check which fixes #5406.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -789,6 +789,10 @@ The "smart_no_gaps" setting hides edge-specific borders of a container if the
 container is the only container on its workspace and the gaps to the screen edge
 are +0+.
 
+[[_smart_borders]]
++hide_edge_borders+ has replaced the old +smart_borders+ syntax. Use the former
+instead of the latter.
+
 *Syntax*:
 -----------------------------------------------
 hide_edge_borders none|vertical|horizontal|both|smart|smart_no_gaps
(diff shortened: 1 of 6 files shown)


docs/userguide
```diff
@@ -782,6 +782,7 @@ You can hide container borders adjacent to the screen edges using
 +hide_edge_borders+ (the default is +none+). Hiding borders is useful if you are
 using scrollbars, or do not want to waste even two pixels in displayspace.
 
+[[_smart_borders]]
```
stapelberg: Could you mention smart_borders in the Syntax or Example with a comment stating that it’s an old syntax for “hide_edge_borders smart”? I’d like to make sure that people can Ctrl+f this page for every line of their config file and find details on it.

src/config_directives.c
```diff
@@ -326,10 +326,12 @@ CFGFUN(gaps, const char *workspace, const char *scope, const long value) {
 }
 
 CFGFUN(smart_borders, const char *enable) {
+    ELOG("smart_borders is deprecated. Please use hide_edge_borders instead.\n");
```
stapelberg: Why raise an error? We could just remove this line and translate smart_borders into hide_edge_borders == smart indefinitely. I don’t think it’s necessary to bother users to update their config file for this.

stapelberg (changes requested):

src/config_directives.c
orestisfl: Moved the warning only when there is a conflict.

docs/userguide
orestisfl: I feel if we mention the old syntax as an example it's going to lead to people copying it. Added a simple mention of it.

stapelberg (approved):

PR #5899, opened 2024-02-04
doc: update meson build instruction

HarimbolaSantatra:
The [official build instruction][1] are deprecated on Meson 1.3.1.
These command:

    mkdir -p build && cd build
    meson ..
    ninja

... work but will yield the following warning:

> WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.

Here's the correct way, according to the [meson documentation][2]:

    mkdir -p build
    meson setup build
    meson compile -C build
    meson install -C build

[1]: https://i3wm.org/docs/hacking-howto.html#_building_i3
[2]: https://mesonbuild.com/Quick-guide.html#compiling-a-meson-project

### DIFF
--- a/docs/hacking-howto
+++ b/docs/hacking-howto
@@ -31,8 +31,6 @@ Guide → Compiling a Meson project]. In case you’re unfamiliar:
     mkdir -p build
     meson setup build
     meson compile -C build
-    meson install -C build
-
 
 === Build system features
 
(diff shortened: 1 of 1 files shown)


docs/hacking-howto
```diff
@@ -28,9 +28,11 @@ https://mesonbuild.com/[The Meson Build system]; see
...
 Guide → Compiling a Meson project]. In case you’re unfamiliar:
 
-    $ mkdir -p build && cd build
-    $ meson ..
-    $ ninja
+    mkdir -p build
+    meson setup build
+    meson compile -C build
+    meson install -C build
+
```
orestisfl: We don't want to recommend users to install a locally built binary to their system, potentially overriding their distro's package manager, this will cause more problems than it will solve.

```suggestion
```

orestisfl (changes requested):
Great, thank you.

orestisfl (approved):

PR #5900, opened 2024-02-04
Use new GitHub issue templates

orestisfl:
Biggest benefit is that the users are presented with a set of editable fields instead of having to edit markdown directly in a text box, which is less friendly. Links to userguide and IPC docs are clickable in the description.

Test in https://github.com/orestisfl/i3/issues/new/choose

See https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository


stapelberg (approved):

stapelberg:
Thanks!

PR #5384, opened 2023-01-21
Set `_NET_FRAME_EXTENTS` based on border width

SimonBrandner:
Fixes https://github.com/i3/i3/issues/4292

<hr>

209577166-bf5a934e-62c2-44ee-842f-aa90c3aff352
209577183-7475224e-bac0-41bf-ba7e-611da3cacdc5
209569688-53c8bed3-0023-4b28-baf9-499501261e68
209569716-877721a2-4977-4746-b248-12bb1e20ab1d

<hr>

While this doesn't work perfectly in the tabbed layout, I think it still does much better than the broken thing we have now

I attempted to make the tabbed layout look nicer but I don't know how to actually do that, to be honest

<hr>

Would this solution be acceptable?

PR #5944, opened 2024-03-09
Set `_NET_FRAME_EXTENTS` according to the actual decoration size

### LINKED ISSUE #4292: Set _NET_FRAME_EXTENTS
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Documentation Request
</pre>

## Current Behavior

_NET_FRAME_EXTENTS property is ~set to all 0s~ not set, even if i3 border size is not 0.

## Desired Behavior

_NET_FRAME_EXTENTS being set properly

## Environment

<pre>
Binary i3 version:  4.19 (2020-11-15) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.19 (2020-11-15) (pid 30702)
Loaded i3 config: /home/shui/.config/i3/config (Last modified: Mon 08 Jun 2020 02:41:57 PM BST, 17240827 seconds ago)

</pre>

sigprof:
Inspired by #5384, but instead of just using the border width, this PR reports the actual frame extents for the window, which may also include the title bar (for floating windows and tiled windows in plain split containers, but not for tiled windows in stacked/tabbed containers).

The existing `con_border_style_rect()` function should already handle all configuration options which can affect the decoration sizes (if it does not, that would also show up in other places); its result just needs to be converted into the format used by the `_NET_FRAME_EXTENTS` property.

This PR fixes #4292 probably in the best way possible (the reported `_NET_FRAME_EXTENTS` values should always match the actual sizes of window frame elements which are actually drawn into the X11 frame window into which the client window is reparented).  The only really problematic case is with the stacked/tabbed containers, for which the title bar is actually drawn into a completely separate window, therefore the title bar size cannot be reported in `_NET_FRAME_EXTENTS` (actually I tried to calculate the size of those decorations and add it to the top decoration size, but that did not change the behavior of `picom`).

<details><summary>Large screenshots here (3840×2160)</summary>

Example of configuration with `hide_edge_borders smart` — a single window does not have borders, so only the top frame size is non-zero:

but multiple windows have borders:

Changing border width works too (although with `border normal 8` you can see that the top border overlaps the title text, because on the i3 side that border does not really exists, and `picom` just draws it over; also the pixel sizes reported by `xprop` and `xwininfo` are not identical to what is specified in i3, because I use 168 dpi on this system, therefore 4 px in the i3 config = 7 dpx):

Handling of tabbed containers is less perfect though.  Here is a single tabbed container with `hide_edge_borders smart`, so it does not really have a border — note that all frame extents are zero, and the titlebar is rounded separately (although it could easily be excluded from rounding, that does not really help much):

Once the border actually appears, you may notice that the top part of the `picom` border actually gets drawn over the top part of the window, partially obscuring the top line in this terminal (`picom` does not mind that the top frame size is reported as 0):

Some examples of floating windows (no major problems there):

Options like `hide_edge_borders both` work too when gaps are removed (although the resulting behavior with `picom` is probably not very useful — the rounded border gets drawn only if all of the left, bottom and right borders are present):

The same with `border pixel 8` (note that windows with only the top border hidden still get the rounded border treatment by `picom`, but the border overlaps the top part of the window):

</details>


SolsticeSpectrum:
Can we accept this? This is actually perfect.

dimkauzh:
This looks good to me, I would love to see this merged!

ghost (approved):
i tried this and it finally fixes my picom and i3 border problem.

waiting for this to get merged...

ahfriedman:
This appears to fix the issue with smart borders I'd noticed in the prior PR https://github.com/i3/i3/pull/5384#issuecomment-1872232905. 

With respect to tabbed/stacked in my opinion, this all makes me think that the solution is just getting picom to allow one to specify the radius of each corner (that way, we could just specify that the I3 frame only gets rounded on the top) such as in these PRs/Issues https://github.com/yshui/picom/discussions/918 https://github.com/yshui/picom/issues/564. That said, I'm also using a fork of i3 where I only have indicators rendered on the bottom, so having indicators work perfectly is less of a concern for me, personally
image 
image

stickyautism:
This would be so great if it gets merged.

orestisfl (commented):
Not perfect for stacked/tabbed but we are technically not lying so I am willing to merge this. However, it still needs tests.

bsdice:
I have been using this patch for some months now to correct display issues of unmanaged windows in a Java software called "Trader Workstation" by Interactive Brokers. It doesn't fix all corner cases, compare the handling in https://github.com/qtile/qtile/blob/master/libqtile/layout/floating.py#L255 If there is an unofficial patch for that, too, I would gladly test. Thank you.

orestisfl:
Started adding some tests. @sigprof feel free to contribute to them otherwise I'll add more when I have the time.

orestisfl (approved):

stapelberg:
What about this code? https://github.com/i3/i3/blob/b9339988786346f44983e846770021844e244a1c/src/handlers.c#L817-L847

Should it be updated to match the extents calculation introduced in this PR?

sigprof:
The problem with the `_NET_REQUEST_FRAME_EXTENTS` message is that it gets sent before the window is actually mapped (that's the whole point of that message — the application can get the WM decoration size and calculate the requested window position and size according to that), so `con_by_window_id(event->window)` would probably be `NULL`, and the calculations based on `con_border_style_rect(con)` won't work.

In order to handle `_NET_REQUEST_FRAME_EXTENTS` 100% correctly i3 would need to pretend that the window is getting mapped, calculate the new layout according to the config and various window properties, extract the resulting decoration size, and then discard the result of layout calculations (basically do a part of `manage_window()` which includes at least `run_assignments()` to apply rules like `for_window [foo] border <bar>`, but in a way that does not actually affect the global state).  I don't see a realistic way to implement that with the current i3 architecture.

PR #5960, opened 2024-03-20
Bugfix posix signals

### LINKED ISSUE #5958: Copy'n'Paste Error for SIGUSR1 and SIGUSR2 signal handler
The `main.c` file uses 2 times the same signal for registering a signal handler. In my opinion, it seems like a copy-and-paste error and should be SIGUSR2 instead.

Locking into blame it seems that priviousely the SIGUSR2 was used as well.

https://github.com/i3/i3/blob/910e58585fb5675c157cddec2bcd2922045fcda5/src/main.c#L254

i3 version 4.23

systec-awe:
Since there is no separate error handling the `SIGUSR2` signal is registered to get the write return code after exiting the program.

Additionally, the `num_watchers` variable is made const since it depends on a fixed array size and should not be modifyable after creation. Optimization can also benefit from const value.

Fixes #5958


orestisfl (approved):
Thanks!

PR #5987, opened 2024-04-08
Use I3SOCK environment variable for path in AnyEvent::I3

waterkip:
Commit 3ae5f31d0 introduced the I3SOCK environment variable. This prevents us from having to call `i3 --get-socketpath'. In case the variable doesn't exist, fall back to the old ways.

### DIFF
--- a/AnyEvent-I3/lib/AnyEvent/I3.pm
+++ b/AnyEvent-I3/lib/AnyEvent/I3.pm
@@ -156,13 +156,19 @@ instance on the current DISPLAY which is almost always what you want.
 sub new {
     my ($class, $path) = @_;
 
-    $path = _call_i3('--get-socketpath') unless $path;
+    # We have I3SOCK now
+    $path ||= $ENV{I3SOCK};
+    $path ||= _call_i3('--get-socketpath');
 
     # Check if we need to resolve ~
     if ($path =~ /~/) {
         my $home = $ENV{HOME};
         confess "Could not get home directory" unless $home and -d $home;
         $path =~ s/~/$home/g;
+    }
+
+    if(!-S $path) {
+        die "$path is not a socket", $/;
     }
 
     bless { path => $path } => $class;


spwhitton:
Thanks, I was wondering why it still used `i3 --get-socketpath`.

AnyEvent-I3/lib/AnyEvent/I3.pm
```diff
@@ -182,6 +182,8 @@ instance on the current DISPLAY which is almost always what you want.
 sub new {
     my ($class, $path) = @_;
 
+    # We have I3SOCK now
+    $path ||= $ENV{I3SOCK};
```
stapelberg, line 160: Directly accessing this environment variable breaks tainted mode.

See `perl Makefile.PL && make test`, which fails with your change (we should probably add that to CI).

I don’t recall what we need tainted mode for, but it’s always been enabled in our tests (for the last 14 years), so either we remove it entirely or we don’t break it (that would be my preference).

stapelberg (changes requested):

AnyEvent-I3/lib/AnyEvent/I3.pm
waterkip: Right. So I tried untainting the value by testing to see if the path was a socket, but that didn't do much. I kept the change because I don't think we expect anything else there. Ended up with a regexp to see if it looks like a path. Which I later on changed to use File::Spec::Functions splitpath and catpath which are found in Perl core (https://perldoc.perl.org/File::Spec::Functions).

waterkip: BTW, it is just the test that is in taint mode, because I've been using the module without taint and it works fine. I guess this is to be compatible with scripts that use taint-mode.

stapelberg: I’m tempted to just rip out the tainted mode support, cut another release and see if anyone complains. My guess is that we just never thought too much about whether supporting tainted mode makes sense, but it’s a constant pain.

waterkip: I think you can. I think you wanted it because you call `i3 --getsocketpath` and there is `i3 --version`.

We can get rid of `--getsocketpath`. We have I3SOCK now. Granted, it's only available as of Oct 29th with 4.23, but people can use everything from this module up to 0.19.

I'm a bit confused at 64fddbe41 where the following comment has been added:

> Gets the i3 version via IPC, with a fall-back that parses the output of i3 --version (for i3 < v4.3).

This is from 2012, and it mentions v4.3. We are now at 4.23. I think this might be a typo for 4.13, so that fallback could be gone as well.

So, when we don't need to call the i3 binary, we don't need to have taint mode. We don't do anything special with environment variables (PATH, etc.). There is now a check that the path is a socket, and that should fix a whole bunch of weird things.

stapelberg: > I think you can. I think you wanted it because you call `i3 --getsocketpath` and there is `i3 --version`.

What makes you say that? How does taint mode help for `i3 --get-socketpath`? 

The other way around I understand: we needed special handling to make `i3 --get-socketpath` work in taint mode. But I don’t think whether we have taint mode or not plays a role for `i3 --get-socketpath` itself. 

Maybe I’m misunderstanding something, but I thought the point of taint mode is to sanitize/recognize untrusted user input. We don’t pass any user input to the `--get-socketpath` call. 

> We can get rid of `--getsocketpath`. We have I3SOCK now. Granted, it's only available as of Oct 29th with 4.23, but people can use everything from this module up to 0.19.

I3SOCK is exported since commit https://github.com/i3/i3/commit/aa203586202981836856d9cf9088620ae866185d, which was released with i3 v4.17. 

But, I3SOCK is only set for programs started by i3, so I don’t think we should get rid of the `--get-socketpath` code path. It’s still useful for scenarios where programs are started as part of a systemd user session or similar. Granted, `--get-socketpath` only works when the `DISPLAY` environment variable is set, but that is much easier to hard-code to `DISPLAY=:0` compared to figuring out the much more dynamic `I3SOCK`.

waterkip: Because the PATH is potentially set to something, most of the stuff is done in `_call_i3()`. This leads me to believe that when that was implemented, taint mode was implemented. So, if you can get rid of calling i3 from the module taint mode isn't needed anymore.

stapelberg: No, taint mode was enabled starting from the initial commit 14 years ago: https://github.com/i3/AnyEvent-I3/commit/3c448c470758e6fefa2b3cbaff4983d62d794f40

waterkip: Ah, k.

stapelberg:
Now that PR #5999 was merged, can you update your pull request please?

waterkip:
Done :)

stapelberg (approved):

PR #5999, opened 2024-04-12
AnyEvent::I3: rip out taint mode compatibility

stapelberg:
I suspect nobody actually uses Perl’s taint mode with AnyEvent::I3.

See https://github.com/i3/i3/pull/5987 for discussion.


stapelberg:
@waterkip Can you verify that AnyEvent::I3 still works as expected for you with this PR applied? Thanks.

waterkip:
It works as expected, my script keeps working.

stapelberg:
@orestisfl Could you approve this change when you have a second please? :)

orestisfl (approved):
Sorry, this slipped through

PR #6034, opened 2024-05-01
testsuite docs: Update instructions

orestisfl:
- Only recommend local perl library installation
- Update build instructions to match hacking-howto

### DIFF
--- a/docs/testsuite
+++ b/docs/testsuite
@@ -76,29 +76,19 @@ used to install the testsuite. Many users prefer to use the more modern
 The tests additionally require +Xephyr(1)+ to run a nested X server. Install
 +xserver-xephyr+ on Debian or +xorg-server-xephyr+ on Arch Linux.
 
-.Installing testsuite dependencies using cpanminus (preferred)
+.Installing testsuite dependencies using cpanminus
 --------------------------------------------------------------------------------
-$ cd ~/i3/testcases
-$ sudo apt-get install cpanminus
-$ sudo cpanm .
+# Install testsuite system-level dependencies. Xvfb is optional but recommended.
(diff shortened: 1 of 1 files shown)


docs/testsuite
```diff
@@ -119,48 +108,57 @@ tests are run under Xvfb.
...
 
-$ mkdir -p build && cd build
+$ mkdir -p build
 
-$ meson ..
+$ meson setup build
+$ cd build
 
-$ ninja
+$ meson compile -C build
```
stapelberg: The `-C build` here seems wrong because of the `cd build` above. 

Just `meson compile` should work, though

docs/testsuite
```diff
@@ -78,28 +78,17 @@ The tests additionally require +Xephyr(1)+ to run a nested X server. Install
 
 .Installing testsuite dependencies using cpanminus (preferred)
 --------------------------------------------------------------------------------
-$ cd ~/i3/testcases
 $ sudo apt-get install cpanminus
```
stapelberg: It occurs to me that we should probably:

1. recommend installing xcb-proto as well, as folks will need it to install X11-XCB successfully
2. recommend installing xserver-xephyr, which folks will need to run the testsuite. We mention it in prose, but it would be better to include it in the `apt-get install` command line.
3. recommend installing xvfb to speed up the testsuite

stapelberg (changes requested):
Thanks for updating the instructions!

docs/testsuite
orestisfl: Added packages to the list.

stapelberg (approved):

PR #6048, opened 2024-05-07
Remove pledge() support for OpenBSD

botovq:
While this initially worked fine, at some point these patches broke because libcairo started calling shmget(2) - a syscall not covered by any pledge promise - and a common pitfall when using pledge with graphics-oriented applications.

Various attempts were made to fix them, but at some time they were simply disabled in the OpenBSD port:

https://github.com/openbsd/ports/commit/a4a9f41dd79f8e9db6d58fb946d2a0c7151cde00 https://github.com/openbsd/ports/commit/5a03c386bab19dcc8e8d7a618693d557840068d1

This seems pointless and creates needless friction both for the i3 team who was willing to carry ugly code and for the OpenBSD ports maintainers who had to disable that code again.

Let's abandon this experiment.

### DIFF
--- a/i3-msg/main.c
+++ b/i3-msg/main.c
@@ -148,11 +148,6 @@ static yajl_callbacks config_callbacks = {
 };
 
 int main(int argc, char *argv[]) {
-#if defined(__OpenBSD__)
-    if (pledge("stdio rpath unix", NULL) == -1) {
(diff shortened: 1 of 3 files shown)


orestisfl (approved):

orestisfl:
Thank you for the PR & elaborate context.

PR #6054, opened 2024-05-14
docs: Fix "Focus the next output" example

bimlas:
The "Focus the next output" example was misleading, fixed the code and added another comment to the previous code.


bimlas:
Force pushed, because I just realized that the example of moving the workspace already documented in "6.10. Moving containers/workspaces to RandR outputs".

orestisfl (approved):

PR #6084, opened 2024-05-28
tiling_drag: Allow swapping containers

orestisfl:
This was originally mentioned in #3085 but left for a future PR.

One of the noticeable limitations is that pressing the modifier while the drag is already initiated, will not swap the containers but instead cancel the drag. This is because of how `drag_pointer()` is written and would be quite an involved case to handle it.

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -216,6 +216,10 @@ Drop on container::
     This happens when the mouse is relatively near the center of a container.
     If the mouse is released, the result is exactly as if you had run the
     +move container to mark+ command. See <<move_to_mark>>.
+    If the swap modifier is pressed before initiating the drag (+tiling_drag
+    swap_modifier+ set to Shift by default), the containers are swapped
+    instead. In that case, the result is exactly as if you had run the +swap
+    container with mark+ command. See <<swapping_containers>>.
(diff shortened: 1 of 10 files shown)


docs/userguide
```diff
@@ -1429,10 +1433,16 @@ You can configure how to initiate the tiling drag feature (see <<tiling_drag>>).
...
+Since i3 4.24, you can configure a modifier key which, when pressed, will swap
+instead of moving containers when dropping directly onto another container.
+Defaults to +Shift+. Note that you have to be pressing both the floating
+modifer and the swap modifier before the drag is initiated.
+
 *Syntax*:
 --------------------------------
 tiling_drag off
 tiling_drag modifier|titlebar [modifier|titlebar]
+tiling_drag swap modifier <modifier>
```
stapelberg: Can you make this `tiling_drag swap_modifier`? It’s not swapping modifiers around, it’s declaring the swap_modifier. This also makes it consistent with the existing `floating_modifier`.

stapelberg (commented):
Looks good overall and worked in a quick test 👍

stapelberg (approved):

PR #6139, opened 2024-07-07
 complete-run.pl: Un-set I3SOCK

orestisfl:
This causes test breakage when the developer/tester is running the tests in i3, as the actual active instance socket is used instead of the test instance.

Besides breaking tests, this is quite dangerous as tests like 319-gaps.t will replace the **actual** config of the user (i.e. ~/.config/i3/config) instead of the current test config.

Broken after #5987

### DIFF
--- a/testcases/complete-run.pl.in
+++ b/testcases/complete-run.pl.in
@@ -106,6 +106,8 @@ $ENV{PATH} = join(':',
     '@abs_top_builddir@',
     '@abs_top_srcdir@',
     $ENV{PATH});
+# Make sure we don't re-use the existing (developer's) i3 connection
+delete $ENV{I3SOCK};
 
 qx(Xephyr -help 2>&1);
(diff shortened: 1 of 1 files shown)


waterkip:
I would suggest setting `$ENV{I3SOCK}` to something else (or undef it) in `testcases/lib/i3test.pm.in`.

orestisfl:
> I would suggest setting `$ENV{I3SOCK}` to something else (or undef it) in `testcases/lib/i3test.pm.in`.

No objection but that can be left as a follow up, correctness over optimization

waterkip:
I think the testsuite (outside of the CPAN module) does something incorrect. I don't think the standalone CPAN module should cater that fix.

orestisfl:
@waterkip I've un-set the env-var in the testsuite, should work.

waterkip:
lgtm

stapelberg (approved):

PR #6154, opened 2024-07-13
i3-input: Re-work stdout

### LINKED ISSUE #6129: `i3-input`: Option to print to stdout instead of composing and running i3 commands
### Welcome

### Impact

- [ ] This feature requires new configuration and/or commands

### Current Behavior

Apparently, `i3-input` is strictly designed for composing and running i3 commands and there's no option to print the user's input to stdout. This is very limiting behavior, considering that not many i3 commands need composing, any way.

### Desired Behavior

An option like `-o`  or similar, to suppress all output, except for printing the user's input verbatim to stdout.

This can have many productive uses and get things done with less code, for example, reducing the need for defining  `mode` bindings that look like
```sh
bindsym a /path/to/my_script.sh a
bindsym b /path/to/my_script.sh b
# and even more bindysm...
```
since it can now be a single keybinding, triggering my_script.sh which invokes `i3-input`.

The alternative I've found is `dmenu < /dev/null`, but you must press Enter key after your input! `i3-input` has the advantage of `-l` option, which I think is more in flow with overall i3 experience. `dmenu` can be at top or bottom, so either it covers the titlebar or the status bar. `i3-input` wins again by being at the right position.

### i3 version

```text
Running i3 version: 4.20.1 (2021-11-03) (pid 1650)
Loaded i3 config:
  /home/$USER/.config/i3/config (main) (last modified: Sun 30 Jun 2024 09:49:46 AM +0330, 138620 seconds ago)
  /home/$USER/0.dotfiles/home/USER_NAME/.config/i3/config.d/keybinds-i3.conf (included) (last modified: Wed 26 Jun 2024 02:47:22 PM +0330, 466364 seconds ago)
  /home/$USER/0.dotfiles/home/USER_NAME/.config/i3/config.d/keybinds-user.conf (included) (last modified: Sun 30 Jun 2024 11:55:36 AM +0330, 131070 seconds ago)
  /home/$USER/0.dotfiles/home/USER_NAME/.config/i3/config.d/keybinds-mode.conf (included) (last modified: Thu 27 Jun 2024 12:39:09 PM +0330, 387657 seconds ago)

```

orestisfl:
Print two lines in stdout: The text entered by the user and the final i3 command, this way users can use i3-input in a more generic way in scripts.

- Use DLOG/ELOG for debugging
- Add -V/--verbose flag
- Update man page

Closes #6129

### DIFF
--- a/man/i3-input.man
+++ b/man/i3-input.man
@@ -9,7 +9,7 @@ i3-input - interactively take a command for i3 window manager
 
 == SYNOPSIS
 
-i3-input [-s <socket>] [-F <format>] [-l <limit>] [-P <prompt>] [-f <font>] [-v]
+i3-input [-s <socket>] [-F <format>] [-l <limit>] [-P <prompt>] [-f <font>] [-V] [-v]
 
 == DESCRIPTION
 
@@ -18,6 +18,10 @@ the user, and send it/them to i3. This is useful, for example, for the
 mark/goto command.
 
 You can press Escape to close i3-input without sending any commands.
+
+i3-input outputs two lines to stdout:
+The first one is the text typed by the user and the second one is the final i3
+command that is executed. You can re-use this output for scripting.
 
 == OPTIONS
 
@@ -42,6 +46,9 @@ The prompt string is not included in the user input/command.
 Use the specified X11 core font (use +xfontsel+ to choose a font).
 
 -v::
+Be verbose.
+
+-v::
 Show version and exit.
 
 == EXAMPLES
@@ -55,6 +62,11 @@ Go to the container marked with above example:
 -----------------------------------------------------
 i3-input -F '[con_mark="%s"] focus' -l 1 -P 'Go to: '
 -----------------------------------------------------
+
+Re-use text entered in i3-input
+------------------------------------------------------------
+i3-input -F 'nop %s' | head -n 1 | xargs -d '\n' notify-send
+------------------------------------------------------------
 
 == ENVIRONMENT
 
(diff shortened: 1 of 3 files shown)


man/i3-input.man
```diff
@@ -41,6 +45,9 @@ The prompt string is not included in the user input/command.
 -f <font>::
 Use the specified X11 core font (use +xfontsel+ to choose a font).
 
+-v::
+Be verbose.
+
 -v::
```
stapelberg, line 51: This line is in conflict with the -v you’re adding?

man/i3-input.man
```diff
@@ -56,6 +63,11 @@ Go to the container marked with above example:
 i3-input -F '[con_mark="%s"] focus' -l 1 -P 'Go to: '
 -----------------------------------------------------
 
+Re-use text entered in i3-input
+------------------------------------------------------------
+i3-input -F 'nop %s' | head -n 1 | xargs -d '\n' notify-send
```
stapelberg, line 68: What’s the rationale behind printing two pieces of information? Wouldn’t it be simpler if we only printed one, so that one can invoke i3-input in the shell using just e.g. `echo "some $(i3-input) is printed"`

a-pav: Also what's wrong with introducing a `-o` option to explicitly stop command composing and suppress all outputs except the user's input? This was what I suggested in #6129 under the **Desired Behavior** header.
It's preferable if `-F` and `nop` can be avoided altogether so there's less chance for the input to leak into logs or somewhere else.

a-pav: By the way, how is this change different from what is already possible to do like:
```sh
i3-input -F 'nop %s' | grep 'output =' | sed 's@output = @@'
```

PR #6160, opened 2024-07-16
provide I3_WINDOW_ID to processes called with exec when possible

### LINKED ISSUE #1729: [enhancement] Pass window id to exec in assignments
When calling external program in an assignment

```
for_window[class="(?i)firefox"] exec ~/scripts/do_some_stuff.sh
```

it would be helpful for the script to have information about the matched window so that it could make requests on i3's IPC for that window.

I propose setting an environment variable such as `$I3_MATCHED_WINDOW` to the window id so that the script could do a call like `i3-msg "[id=$I3_MATCHED_WINDOW] <command>"`.

Note: if #1697 gets merged, I'd like the equivalent for the clicked workspace button in those commands so that the custom script actually knows which button was clicked.

doy:
fixes #1729

### DIFF
--- a/src/startup.c
+++ b/src/startup.c
@@ -129,8 +129,11 @@ void startup_sequence_delete(struct Startup_Sequence *sequence) {
  * The no_startup_id flag determines whether a startup notification context
  * (and ID) should be created, which is the default and encouraged behavior.
  *
+ * The window_id, if given, will be provided to the child process via the
+ * I3_WINDOW_ID environment variable.
+ *
  */
-void start_application(const char *command, bool no_startup_id) {
+void start_application(const char *command, bool no_startup_id, xcb_window_t window_id) {
     SnLauncherContext *context = NULL;
 
     if (!no_startup_id) {
@@ -193,6 +196,13 @@ void start_application(const char *command, bool no_startup_id) {
             sn_launcher_context_setup_child_process(context);
         }
         setenv("I3SOCK", current_socketpath, 1);
+
+        if (window_id != XCB_NONE) {
+            char *window_id_str;
+            sasprintf(&window_id_str, "%d", window_id);
+            setenv("I3_WINDOW_ID", window_id_str, 1);
+            free(window_id_str);
+        }
 
         execl(_PATH_BSHELL, _PATH_BSHELL, "-c", command, NULL);
         err(EXIT_FAILURE, "execl return"); /* only reached on error */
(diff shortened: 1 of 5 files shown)


testcases/t/165-for_window.t
```diff
@@ -390,6 +398,25 @@ is_deeply($nodes[0]->{nodes}[0]->{marks}, [ 'triggered' ], "mark set for workspa
...
+
+$window = open_window(wm_class => 'exec');
+
+my $window_id;
+for (0..3) {
+    open my $fh, '<', $filename
+        or die "couldn't open tempfile $filename for reading: $!";
+    $window_id = do { local $/; <$fh> };
+    last if $window_id;
+    sleep(1);
```
stapelberg: This might seem pedantic, but we will not accept `sleep` calls into the testsuite. Our testsuite has been optimized to run within a second or two (on modern computers), so sleeping for a second increases the total testsuite runtime excessively.

Instead, maybe we could use a FIFO here. search for `mkfifo` in the `testcases` directory for a few examples.

testcases/t/165-for_window.t
```diff
@@ -390,6 +398,25 @@ is_deeply($nodes[0]->{nodes}[0]->{marks}, [ 'triggered' ], "mark set for workspa
...
 kill_all_windows;
 
+##############################################################
+# 13: check that we pass $I3_WINDOW_ID to exec'ed subprocesses
+##############################################################
+
+$window = open_window(wm_class => 'exec');
+
+my $window_id;
+for (0..3) {
```
stapelberg: Why do this multiple times instead of just once? Wouldn’t running just once give us the same guarantees?

src/startup.c
```diff
@@ -195,6 +198,13 @@ void start_application(const char *command, bool no_startup_id) {
         }
         setenv("I3SOCK", current_socketpath, 1);
 
+        if (window_id != 0) {
```
stapelberg: Here and elsewhere in the PR: always use `XCB_NONE` for comparing `xcb_window_t` types instead of `0`

stapelberg (changes requested):
Sorry for the slow review. Do you want to improve this PR before the upcoming release? (If not, that’s also fine, but if so, I can try to wait for this before the release.)

doy:
fixed!

stapelberg (approved):

PR #6415, opened 2025-03-21
Fix typos via codespell

algonell:
There are some typos in the documentation, messages, etc.
Fix them via codespell.


orestisfl (approved):
Thank you!

Issue #5456, opened 2023-03-17
[libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.


Earnestly:
* 4.22-24-ga5da4d54

For awhile now I've been witnessing a great deal of logging with the messages:

```
[libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.
```

I have not yet tried to bisect the problem but I did eventually discover this problem only occurs when I use `split h` or `split v`.

So to trigger this problem all I need to do is use my keybind to trigger `split h` or `split v` and immediately I will get six lines of this log message:

```
2023-03-17T20:42:00.70455 17/03/23 20:42:00 - [libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.
2023-03-17T20:42:00.78653 17/03/23 20:42:00 - [libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.
2023-03-17T20:42:01.32601 17/03/23 20:42:01 - [libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.
2023-03-17T20:42:01.32739 17/03/23 20:42:01 - [libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.
2023-03-17T20:42:01.41734 17/03/23 20:42:01 - [libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.
2023-03-17T20:42:02.84599 17/03/23 20:42:02 - [libi3] ERROR: Surface 0x55ab4795f468 is not initialized, skipping drawing.
```

While the split mode is active (i.e. the indicator on the window is visible) this message will print profusely whenever switching workspaces or interacting with the WM.

Is there anything I could do to try and nail down the reason why this might be happening?

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

i3bot:
I don’t see a version number. Could you please copy & paste the output of `i3 --version` into this issue?

orestisfl:
It's not hard to reproduce some of these messages (`layout tabbed` && `layout toggle split` with two windows) but I am not sure if there are any negative side effects.

Backtrace:

```
(gdb) bt
#0  surface_initialized (surface=0x6150000668a8) at ../libi3/draw_util.c:25
[...]

PR #6438, opened 2025-05-16
fix?: avoid annoying warning for "Surface 0x... is not initialized, skipping drawing"

thiswillbeyourgithub:
- **fix: Prevent drawing from uninitialized frame buffer in x_draw_decoration**
- **fix: Prevent drawing from uninitialized frame buffer in x_deco_recurse**

Hi,

This is a **possible** fix to #5456.

It was made purely by [aider-chat](https://github.com/Aider-AI/aider/) as I cannot read a single line of C.

I consider it bad practice *in general* to do this, but to the best of my prompting skills (which are lightyears away from my C skills), the LLMs acted sure of themselves and the fix was minimal. Additionally, I tested the fix and so far I can't reproduce the issue and have not noticed any regression.

I can provide the full chat logs of the LLMs if you're interested. The workflow I used is already detailed [here](https://github.com/i3/i3/issues/5456#issuecomment-2832313645). Both patch were done using gemini 2.5 and aider.

Can't wait to see what you have to say about this :/!

Edit: although I can't do C, I am still a dev, and can't help but notice that the fix *sounds like* it's just hiding the error. In the sense that obviously it just checks if the surface is not initialized before deciding to draw on it, so I'm willing to think there's a deep issue at play. But for my case: I have 230k useless lines of logs a month, and don't notice any regression so I don't plan on going back.


thiswillbeyourgithub:
Report after 3 weeks: still no issue whatsoever.

morgoth6:
Confirmed. This fixed the problem (or in worst case  the anoying log message)

mrdotx:
Works as expected, no crowded log anymore...

PR #6460, opened 2025-07-08
Update i3.desktop

FedGuy699:
This fixes issues with display link displays not working on i3 and should not affect any other features from my testing.


orestisfl:
Interestingly `XSession` is not part of the spec but it seems quite commonplace in other `.desktop` files. Do you have a source for it? We should at least mention this in the git commit message.

FedGuy699:
I don’t have an official source for this. In my testing, changing the type to XSession resolves issues with certain display drivers and window managers. With this change, sessions tend to work out of the box. While it’s not in the freedesktop spec, it seems to be a common convention used by many desktop environments.

stapelberg:
https://codesearch.debian.net/search?q=Type%3DXSession agrees that window managers / desktop environments usually set Type=XSession, so I think we should accept this change :)

stapelberg (approved):

PR #6482, opened 2025-08-10
Doc: add instructions to install required perl modules for i3-save-tree

HarimbolaSantatra:
I encounted the error:
```
$ i3-save-tree
Can't locate AnyEvent/I3.pm in @INC (you may need to install the AnyEvent::I3 module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.34.0 /usr/local/share/perl/5.34.0 /usr/lib/x86_64-linux-gnu/perl5/5.34 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.34 /usr/share/perl/5.34 /usr/local/lib/site_perl) at /usr/local/bin/i3-save-tree line 19.
BEGIN failed--compilation aborted at /usr/local/bin/i3-save-tree line 19.
```

The [`is-save-tree` file](https://github.com/i3/i3/blob/next/i3-save-tree) contains the line
```
# Requires perl ≥ v5.10, AnyEvent::I3 and JSON::XS
```

It is not documented in the official doc. This PR adds the instruction to install those modules.

I'm not sure if it's the correct way to add these instructions so feel free to improve it.


orestisfl:
How did you install i3? The package should provide the dependencies e.g. debian/ubuntu: https://github.com/i3/i3/blob/2711f427af4c9065d78f9423415c088af2c55eae/debian/control#L46, arch: https://archlinux.org/packages/extra/x86_64/i3-wm/

I'd prefer to not add these instructions when they could confuse people to install dependencies that are best installed via a package manager.

HarimbolaSantatra:
@orestisfl I build it from the source code.
If I did it that way it won't be installed automatically right ?

orestisfl:
Yes. For building from source you can update https://i3wm.org/docs/hacking-howto.html#_building_i3 instead.

HarimbolaSantatra:
okay. I will update the PR accordingly

PR #6507, opened 2025-10-06
main: debug build: Avoid buffer over-read and simplify

orestisfl:
In theory, if `/proc/sys/kernel/core_pattern` is 1024 or more bytes, the null character terminating the buffer can be overwritten.

Note: Found with [bugfinder](https://github.com/stanek-michal/bugfinder)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -551,30 +551,29 @@ int main(int argc, char *argv[]) {
     if (is_debug_build()) {
         struct rlimit limit = {RLIM_INFINITY, RLIM_INFINITY};
         setrlimit(RLIMIT_CORE, &limit);
+        LOG("CORE DUMPS: You are running a development version of i3, so coredumps were automatically enabled (ulimit -c unlimited).\n");
 
+#ifdef __linux__
         /* The following code is helpful, but not required. We thus don’t pay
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -551,30 +551,29 @@ int main(int argc, char *argv[]) {
...
-            cwd_size = cwd_size * 2;
-            cwd = srealloc(cwd, cwd_size);
-        }
-        if (cwd_ret != NULL) {
+        char *cwd = getcwd(NULL, 0);
+        if (cwd != NULL) {
             LOG("CORE DUMPS: Your current working directory is \"%s\".\n", cwd);
+            free(cwd);
         }
+        const size_t buffer_size = 1024;
```
stapelberg, line 564: Or maybe we just use the `slurp` function from `src/util.c`? That should be even simpler, and we don’t need to guard against adversarial input in the `core_pattern` setting.

stapelberg (changes requested):

src/main.c
orestisfl: `slurp` can't be use because `st_size` is set to 0 because it's a virtual file under /proc

stapelberg (approved):

PR #6514, opened 2025-10-24
Refactor: Extract parser stack implementation to parser_util

orestisfl:
The command and config parsers both used a similar stack implementation for handling tokens. This resulted in duplicated code.

This commit extracts the common stack implementation (push/get/clear functions for strings and longs) into a new `parser_util` module.

This resolves a long standing TODO comment.


src/parser_util.c
```diff
@@ -0,0 +1,101 @@
...
+ *
+ */
+void push_string(struct stack *stack, const char *identifier, const char *str) {
+    for (int c = 0; c < 10; c++) {
+        if (stack->stack[c].identifier != NULL &&
+            strcmp(stack->stack[c].identifier, identifier) != 0) {
+            continue;
+        }
+        if (stack->stack[c].identifier == NULL) {
+            /* Found a free slot, let		s store it here. */
```
stapelberg: The gap between “let” and “s” makes me suspect you used a tool that does not support unicode properly? :) 

(Also two more occurrences below.)

stapelberg (changes requested):

src/parser_util.c
orestisfl: Oops, guilty

include/parser_util.h
```diff
@@ -0,0 +1,59 @@
...
+    struct stack_entry stack[10];
+};
+
+/**
+ * Pushes a string (identified by 'identifier') on the stack.
+ * If a string with the same identifier is already on the stack, the new
+ * string will be appended, separated by a comma.
+ *
+ */
+void push_string(struct stack *stack, const char *identifier, const char *str);
```
stapelberg: Now that these names are no longer scoped to their compilation unit (e.g. commands_parser), I think we should prefix them as they are rather generic. How about `parser_push_string`?

stapelberg (commented):
Looks good overall!

stapelberg (approved):

PR #6516, opened 2025-10-26
Chore: Code cleanup and style improvements

orestisfl:
This PR is a pure linting and code cleanup effort with no functional changes, focusing on improving code quality and consistency.

The bulk of the changes involve:

*   Code Formatting: The entire codebase was reformatted after updating our `clang-format` version.
*   Compiler Warnings: Addressed `-Wsuggest-attribute` warnings from GCC by applying `pure`, `const`, and `format` attributes where appropriate. This helps the compiler with optimizations and bug detection.
*   Code Modernization:
    *   Variable declarations were moved closer to their first use or into tighter scopes.
    *   `memset` calls were replaced with C99 zero-initializers (`= {0}`).
    *   Redundant `struct` keywords, unnecessary type casts, and some superfluous `return` statements were removed.
*   CI Adjustments: The GitHub Actions workflow was updated to correctly apply compiler flags for both GCC and Clang and to use the newer `clang-format` version for checks.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -277,7 +277,7 @@ static int parse_restart_fd(void) {
 int main(int argc, char *argv[]) {
     /* Keep a symbol pointing to the I3_VERSION string constant so that we have
      * it in gdb backtraces. */
-    static const char *_i3_version = I3_VERSION;
+    static const char *_i3_version __attribute__((used)) = I3_VERSION;
     char *override_configpath = NULL;
     bool autostart = true;
     char *layout_path = NULL;
--- a/src/main.c
+++ b/src/main.c
@@ -762,7 +762,7 @@ int main(int argc, char *argv[]) {
         xcb_set_selection_owner(conn, wm_sn_selection_owner, wm_sn, last_timestamp);
 
         if (selection_reply && selection_reply->owner != XCB_NONE) {
-            int check_rounds = 150;            /* Wait for a maximum of 15 seconds */
+            int check_rounds = 150; /* Wait for a maximum of 15 seconds */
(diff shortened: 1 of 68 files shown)


stapelberg:
This PR is very hard to review. Can you please move the clang-format update into a separate PR? I’ll clone, run clang-format myself, and diff, to confirm that there are no unintended changes in such a large diff. Thanks.

orestisfl:
> This PR is very hard to review. Can you please move the clang-format update into a separate PR? I’ll clone, run clang-format myself, and diff, to confirm that there are no unintended changes in such a large diff. Thanks.

Moved clang-format changes to https://github.com/i3/i3/pull/6521

src/main.c
```diff
@@ -279,7 +277,7 @@ static int parse_restart_fd(void) {
 int main(int argc, char *argv[]) {
     /* Keep a symbol pointing to the I3_VERSION string constant so that we have
      * it in gdb backtraces. */
-    static const char *_i3_version __attribute__((used)) = I3_VERSION;
```
stapelberg, line 282: Don’t remove the `__attribute__((used))` please, we want this symbol to end up in gdb backtraces as the comment says :) 

I realize that you’re using the symbol in the log line below, but I don’t think a load-bearing log line is clean. I’d rather keep the explicit `__attribute__((used))` so that this line is standalone.

stapelberg (changes requested):
Seems fine overall. Please rebase after addressing comments

stapelberg (approved):

stapelberg:
Looks good, but can you rebase please?

Issue #4335, opened 2021-01-30
Assertion failure when appending layout


dominikh:
## I'm submitting a…

<pre>
[x] Bug
[ ] Feature Request
[ ] Documentation Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
Calling append_layout with `{"marks": ["foo"], "nodes": [{}]}` causes an assertion failure:

```
i3: ../src/ipc.c:434: dump_node: Assertion `false' failed.
```

```
Program terminated with signal SIGABRT, Aborted.
#0  0x00007faf0734633a in raise () from /nix/store/m0xa5bz7vw7p43wi0jppvvi3c9vgqvp7-glibc-2.32-25/lib/libc.so.6
(gdb) bt
#0  0x00007faf0734633a in raise () from /nix/store/m0xa5bz7vw7p43wi0jppvvi3c9vgqvp7-glibc-2.32-25/lib/libc.so.6
#1  0x00007faf073305d3 in abort () from /nix/store/m0xa5bz7vw7p43wi0jppvvi3c9vgqvp7-glibc-2.32-25/lib/libc.so.6
#2  0x00007faf0733041f in __assert_fail_base.cold.0 () from /nix/store/m0xa5bz7vw7p43wi0jppvvi3c9vgqvp7-glibc-2.32-25/lib/libc.so.6
#3  0x00007faf0733ed92 in __assert_fail () from /nix/store/m0xa5bz7vw7p43wi0jppvvi3c9vgqvp7-glibc-2.32-25/lib/libc.so.6
#4  0x0000000000436734 in dump_node ()
#5  0x0000000000436e17 in ipc_send_window_event ()
#6  0x000000000041fdd8 in con_mark ()
#7  0x0000000000437ca2 in json_end_map ()
#8  0x00007faf078ab6f0 in yajl_do_parse () from /nix/store/cr3l3crhnpcfbwqa3dq9izcx2qnmrsyp-yajl-2.1.0/lib/libyajl.so.2
#9  0x000000000043902f in tree_append_json ()
#10 0x00000000004131fa in cmd_append_layout ()
#11 0x000000000041dcac in next_state ()
#12 0x000000000041e7a3 in parse_command ()
#13 0x0000000000434374 in handle_run_command ()
#14 0x000000000043445d in ipc_receive_message ()
[...]

PR #6534, opened 2025-11-29
[WIP] Migrate yajl to yyjson

### LINKED ISSUE #6257: Migrate away from yajl
### Welcome

### Impact

- [ ] This feature requires new configuration and/or commands

### Current Behavior

i3 currently uses yajl for json serialization. Unfortunately, yajl is pretty much dead upstream now and according to the current maintainer in Fedora, it has collected quite a number of CVEs and they'd like to eventually stop maintaining it or remove it from the distribution entirely (see: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/YPFHPOKAND3RZR7ZKWTDHUQEESG6IUJ3/).

### Desired Behavior

Migrate to a maintained json library, e.g. json-c (this library is also used by sway).

I'd be willing to help out in such an effort, but only if there's any chance that this will land upstream in i3.

orestisfl:
Closes #6257

First full draft. Heavily assisted by LLMs, opening the PR for initial review. Significant reviewing is still needed both from me and reviewers.

Using the benchmark script from: https://github.com/i3/i3/issues/6257#issuecomment-2585732868

**before**:
```
i3 version 4.24-27-gb1e99d8f © 2009 Michael Stapelberg and contributors
nodes= 10 min= 0.07 median= 0.08 max= 0.28 avg= 0.08
nodes= 12 min= 0.09 median= 0.09 max= 0.34 avg= 0.10
nodes= 16 min= 0.11 median= 0.12 max= 0.36 avg= 0.13
nodes= 24 min= 0.16 median= 0.18 max= 0.58 avg= 0.20
nodes= 40 min= 0.25 median= 0.26 max= 0.53 avg= 0.27
nodes= 72 min= 0.46 median= 0.51 max= 0.83 avg= 0.53
nodes=136 min= 0.80 median= 0.91 max=16.96 avg= 1.18
nodes=264 min= 1.46 median= 1.65 max=15.32 avg= 1.97
python /tmp/i3-benchmark-get-tree.py 2> /dev/null  19,44s user 1,70s system 99% cpu 21,241 total
```

**after**:
```
i3 version 4.24-47-gbaedde18+ © 2009 Michael Stapelberg and contributors
nodes= 10 min= 0.04 median= 0.05 max= 0.27 avg= 0.06
nodes= 12 min= 0.05 median= 0.06 max= 0.76 avg= 0.09
nodes= 16 min= 0.06 median= 0.07 max= 0.25 avg= 0.07
nodes= 24 min= 0.08 median= 0.14 max= 0.69 avg= 0.16
nodes= 40 min= 0.13 median= 0.14 max= 0.42 avg= 0.15
nodes= 72 min= 0.21 median= 0.23 max= 0.64 avg= 0.26
nodes=136 min= 0.32 median= 0.46 max=14.75 avg= 0.74
nodes=264 min= 0.63 median= 0.79 max=17.16 avg= 1.14
python /tmp/i3-benchmark-get-tree.py 2> /dev/null  18,59s user 1,86s system 99% cpu 20,493 total
```

more benchmarks are needed.


orestisfl:
@stapelberg this is quite impossible to review like this. I can split it into multiple PRs and we could target a different branch before merging into next.

stapelberg:
Which LLM and prompts did you use for this (if you still have them)?

In general, just to be prudent, I wouldn’t want to merge an LLM-authored PR for something as intricate as the JSON parsing. 

However, we can still use LLMs to prototype this migration and get a better feel for things.

I notice that the PR contains patterns like:

```c
        yyjson_val *hr_val = yyjson_obj_get(root, "human_readable");
        if (hr_val && yyjson_is_str(hr_val)) {
            human_readable_version = sstrdup(unsafe_yyjson_get_str(hr_val));
        }
```

But the yyjson docs specifically states that the library performs null checks: https://ibireme.github.io/yyjson/doc/doxygen/html/api.html#null-check

So we could avoid the unsafe accessor and explicit type-checking:

```c
        const char* hr = yyjson_get_str(yyjson_obj_get(root, "human_readable"));
        if (hr) {
            human_readable_version = sstrdup(hr);
        }
```

Sometimes, the LLM seems to inline functions…?! See `ipc_marshal_workspace_event` for example, whose call was replaced with direct yyjson calls. 

Instead of having the LLM make the code changes, I would be more interested in having the LLM produce rewrite rules (like, `sed` commands, or a script, or something similar) that can be deterministically applied, reviewed and changed. Have you tried that approach?

orestisfl:
> Which LLM and prompts did you use for this (if you still have them)?

LLM is Claude Opus 4.5. It was quite a long back and forth.

> I notice that the PR contains patterns like:

Yup, I noticed that. It was probably my own mis-use of the libray in https://github.com/i3/i3/compare/next...orestisfl:i3:yyjson-tree_append_json. I used that as the "seed" of the refactor and copied my pattern. I'll be correcting the use if we plan to make this effort ready for review but I was running out of my time budget and wanted to push this draft for discussion.

> Sometimes, the LLM seems to inline functions…?! See ipc_marshal_workspace_event for example, whose call was replaced with direct yyjson calls.

Yes, noticed that :+1: 

> In general, just to be prudent, I wouldn’t want to merge an LLM-authored PR for something as intricate as the JSON parsing.
> ...
> Instead of having the LLM make the code changes, I would be more interested in having the LLM produce rewrite rules (like, sed commands, or a script, or something similar) that can be deterministically applied, reviewed and changed. Have you tried that approach?

Some thoughts:
1. yajl uses callbacks, yyjson uses DOM traversal. We can adapt yyjson to use callbacks but I feel that would be a missed opportunity as our callback-based code is quite difficult to understand and can lead to hard-to-fix/reason bugs: https://github.com/i3/i3/issues/4335. That said, my initial branch https://github.com/i3/i3/compare/next...orestisfl:i3:yyjson-tree_append_json did maintain the callback-based code so it's definitely possible to continue with that and have a smaller diff as a result.
2. The APIs for writing json are quite similar. We could adjust our macros `y`,`ystr` macros and leave the rest of the code largely intact apart from allocating/freeing.
3. Perfecting a script like that is extremely difficult because allocating and freeing is context dependent. We could still have a script that gives us 90% correct code, git commit the result to ease reviewing and then work on fixing the code base manually on subsequent commits.
4. This PR is not LLM-authored, I am still the main author here, the LLM is the tool. I steered the direction of the refactor, caught issues, gave directions to amend them, validated behavior and manually reviewed all files. The draft status of this PR indicates that the complete code is not ready to be reviewed. A ready-for-review PR should be reviewed like any human code.

## Addendum: optimization notes

For the DOM-based API, there are optimizations we could be using like instead of the "linear search times" [e.g. yyjson_obj_get](https://ibireme.github.io/yyjson/doc/doxygen/html/yyjson_8h.html#a1e8a4dea2e9e9248acde14c664ab702b) use [yyjson_obj_iter_get](https://ibireme.github.io/yyjson/doc/doxygen/html/yyjson_8h.html#a1f3b09c4f279287f8af93b3754a41e85)

stapelberg:
Thanks for the details!

> 1. yajl uses callbacks, yyjson uses DOM traversal. We can adapt yyjson to use callbacks but I feel that would be a missed opportunity as our callback-based code is quite difficult to understand and can lead to hard-to-fix/reason bugs: [Assertion failure when appending layout #4335](https://github.com/i3/i3/issues/4335). That said, my initial branch [next...orestisfl:i3:yyjson-tree_append_json](https://github.com/i3/i3/compare/next...orestisfl:i3:yyjson-tree_append_json) did maintain the callback-based code so it's definitely possible to continue with that and have a smaller diff as a result.

Agreed, we should switch to DOM-based traversal instead of callbacks.

> 2. The APIs for writing json are quite similar. We could adjust our macros `y`,`ystr` macros and leave the rest of the code largely intact apart from allocating/freeing.

Sounds very good. Starting with a minimal diff (e.g. changing only the write path with minimal changes) and then expanding from there is generally a good strategy.

> 3. Perfecting a script like that is extremely difficult because allocating and freeing is context dependent. We could still have a script that gives us 90% correct code, git commit the result to ease reviewing and then work on fixing the code base manually on subsequent commits.

Agreed. No need to make the translation perfect, as long as it removes most of the tedious steps of the refactoring.

Ideally, we could have git commits that are entirely deterministic tool applications, and then separate commits for where we manually fixed up things.

> 4. This PR is not LLM-authored, I am still the main author here, the LLM is the tool. I steered the direction of the refactor, caught issues, gave directions to amend them, validated behavior and manually reviewed all files. The draft status of this PR indicates that the complete code is not ready to be reviewed. A ready-for-review PR should be reviewed like any human code.

OK. As long as you review the entire diff you’re sending, that’s okay with me. 

That said, I think we should be transparent about LLM usage. Adding a line like “Assisted by Claude Opus 4.5, but every line is human-reviewed.” would be sufficient IMO.

> ## Addendum: optimization notes
> For the DOM-based API, there are optimizations we could be using like instead of the "linear search times" [e.g. yyjson_obj_get](https://ibireme.github.io/yyjson/doc/doxygen/html/yyjson_8h.html#a1e8a4dea2e9e9248acde14c664ab702b) use [yyjson_obj_iter_get](https://ibireme.github.io/yyjson/doc/doxygen/html/yyjson_8h.html#a1f3b09c4f279287f8af93b3754a41e85)

That’s fair, but we can move performance optimizations into a follow-up step to reduce scope of this already-large refactor :)

orestisfl:
Another point of incompatibility is between the write APIs: yajl_gen_string is used both for keys and values:
> Generate a string value or map key from str.

I've also tried to create a compatibility layer in C only. That could be something we can use as well as a base. Tests pass, no further manual testing done (e.g. i3bar is broken without handling streaming json): https://github.com/i3/i3/commit/8a08011b5ddd0cb7d1f08b2af6cebba93a831538

stapelberg:
Now that v4.25 is released, we can resume work on this.

> I've also tried to create a compatibility layer in C only.

Yeah, I like that idea. We could also change our existing code to be easier to migrate, e.g. by using two separate names for keys/values instead of directly using yajl_gen_string for both.

PR #6537, opened 2025-12-07
318-i3-dmenu-desktop: Use $^X for generated script shebang

CyberShadow:
The test creates a fake i3-msg Perl script at runtime. Previously it used `#!/usr/bin/env perl` as the shebang, but this can fail in certain build environments (e.g., Nix sandboxed builds) where the `perl` found via `/usr/bin/env` may not have the same module paths as the `perl` running the test suite.

Use `$^X` instead, which contains the path to the Perl interpreter currently running the test. This ensures the generated script uses the same Perl with the same `PERL5LIB` environment, guaranteeing that `JSON::XS` and other required modules are available.

### DIFF
--- a/testcases/t/318-i3-dmenu-desktop.t
+++ b/testcases/t/318-i3-dmenu-desktop.t
@@ -39,7 +39,7 @@ mkfifo("$tmpdir/fifo", 0600) or BAIL_OUT "Could not create FIFO: $!";
 
 open(my $i3msg_dump, '>', "$tmpdir/i3-msg");
 say $i3msg_dump <<EOT;
-#!/usr/bin/env perl
+#!/usr/bin/env $^X
 use strict;
 use warnings;
 use JSON::XS qw(encode_json);


testcases/t/318-i3-dmenu-desktop.t
```diff
@@ -38,8 +38,8 @@ $ENV{PATH} = "$tmpdir:" . $ENV{PATH};
 mkfifo("$tmpdir/fifo", 0600) or BAIL_OUT "Could not create FIFO: $!";
 
 open(my $i3msg_dump, '>', "$tmpdir/i3-msg");
+say $i3msg_dump "#!$^X";
 say $i3msg_dump <<EOT;
-#!/usr/bin/env perl
```
stapelberg, line 42: Just change this line to `#!$^X`, no need to add an extra `say` call.

stapelberg (changes requested):
Unfortunately, when changing the shebang to `#!$^X` (or just `#!perl`), the tests fail for me in a `nix develop nixpkgs#i3` environment:

```
[:100] /home/michael/i3/testcases/t/318-i3-dmenu-desktop.t: starting
[:100] /home/michael/i3/testcases/t/318-i3-dmenu-desktop.t: finished
output for /home/michael/i3/testcases/t/318-i3-dmenu-desktop.t:
i3-msg: Could not determine i3 socket path: Resource temporarily unavailable
Could not launch i3-msg: 256 at /home/michael/i3/i3-dmenu-desktop line 513, <GEN1> line 1.
not ok 1 - 'false' expected, at character offset 0 (before "failed with 1\n") at /home/michael/i3/testcases/t/318-i3-dmenu-desktop.t line 74.
# 

#   Failed test ''false' expected, at character offset 0 (before "failed with 1\n") at /home/michael/i3/testcases/t/318-i3-dmenu-desktop.t line 74.
# '
#   at /home/michael/i3/testcases/lib/TestWorker.pm line 53.
```

Any ideas?

CyberShadow:
> Unfortunately, when changing the shebang to `#!$^X` (or just `#!perl`), the tests fail for me in a `nix develop nixpkgs#i3` environment:

OK, wow, that's ironic!

First, I did not know that you could do that. Maybe I should do that instead.

Second, looking at nixpkgs:

https://github.com/NixOS/nixpkgs/blob/afc8b87762a92624ce3862a7c3bf7deeb534a966/pkgs/applications/window-managers/i3/default.nix#L91-L100

They actually patch this exact test because of the same problem!

I'm thinking that this is simultaneously more evidence that we should consider a more robust approach for this, while also vexing in that this attempt to fix it upstream breaks the downstream.

However, looking at the error you got, it doesn't look like the downstream patch is the problem - the failure was not due to `substituteInPlace` failing because it could no longer find the `#!/usr/bin/env perl` stream (though I think this would be an issue anyway), but because of this interesting series of events:

1. `$^X` in the nixpkgs derivation is just `perl`.
2. When constructed into a shebang, this is `#!perl`. Not a valid shebang, thus not a valid program
3. If `execve` finds a program in a `$PATH` directory but fails to execute it, it will keep looking.
4. After skipping the stub `i3-msg` with the invalid shebang, it finds the real `i3-msg`, which tries to connect to i3 and fails with the message above.

It can get more confusing because `perl` / `$^X` may refer to:
- an unrelated system perl
- `perl` before nixpkgs wrapping (`$PERL5LIB` is inherited)
- `perl` with nixpkgs wrapping (sets `$PERL5LIB`)

and to add to the confusion, nixpkgs' wrapped perl [doesn't seem to include](https://github.com/NixOS/nixpkgs/blob/b6fd4d089fb1c5315ee7f365de6480f7c3150c69/pkgs/applications/window-managers/i3/default.nix#L84-L88) `JSON::XS`.

However, for the problem at hand where `$^X` can be either a useful absolute path or just `perl`, I can see a simple solution - keep `#!/usr/bin/env` and just replace `perl` with `$^X`. That'll preserve compatibility when it's `perl`, and fix behavior when we have a good `$^X` and don't want to call `perl`. Pushed that patch, thoughts?

stapelberg (approved):

stapelberg:
> However, for the problem at hand where `$^X` can be either a useful absolute path or just `perl`, I can see a simple solution - keep `#!/usr/bin/env` and just replace `perl` with `$^X`

Did you not actually test this solution inside the Nix build sandbox? It doesn’t build there for me.

I sent https://github.com/i3/i3/pull/6548 to fix this both inside and outside the Nix build sandbox.

CyberShadow:
> Did you not actually test this solution inside the Nix build sandbox?

I did, but you'll have to clarify which Nix sandbox!

I think it would be out of scope for us to support the way nixpkgs does things, since that would be a bit backwards - it's nixpkgs who needs to support us. I believe this change should simplify or allow removing nixpkgs' downstream patch, but I did not experiment going as far.

CyberShadow:
My bad, looks like the Nix sandbox was actually not working on the machine I tested this with - the default best-effort behavior activated, so while I tested it inside a Nix derivation, I did not test inside the sandbox. It used the host `/usr/bin/env`. Apologies for this mess!

PR #6538, opened 2025-12-07
i3test: Add require statements before pragma imports

CyberShadow:
In the dynamically eval'd import code, `feature->import()`, `strict->import()`, and `warnings->import()` are called without first loading these modules. While these pragma modules may be coincidentally loaded by other modules in some Perl environments, this is not guaranteed.

Add explicit `require` statements before calling the import methods to ensure the modules are loaded. This follows correct Perl practice where a module must be loaded before calling its `->import` method.

Without this fix, tests may fail in some environments with:

```
Attempt to call undefined import method with arguments (":5.10") via package "feature"
```

### DIFF
--- a/testcases/lib/i3test.pm.in
+++ b/testcases/lib/i3test.pm.in
@@ -173,6 +173,9 @@ use Time::HiRes qw(sleep);
 use i3test::Test;
 __
     $tester->BAIL_OUT("$@") if $@;
+    require feature;
+    require strict;
+    require warnings;
     feature->import(":5.10");
     strict->import;
     warnings->import;


stapelberg:
> Without this fix, tests may fail in some environments with:

Can you tell me more about which environments the tests fail in without these lines?

(I’m hesitant to accept a plausible-sounding explanation from an LLM without fully understanding it.)

CyberShadow:
Sure! The environment is a Nix derivation sandbox - I tried to wrap the test suite into a Nix flake check so that I don't have to deal with any of the Perl stuff on my host machine. Two tests failed inside there, this being one of them - the explanation makes sense to me, but I'm by no means a Perl guru. Happy to post a reproducer if it helps.

stapelberg:
> Two tests failed inside there, this being one of them - the explanation makes sense to me, but I'm by no means a Perl guru. Happy to post a reproducer if it helps.

Yes, please post the reproduction steps.

CyberShadow:
OK, looking at this closer, it doesn't fix a test failure, only a warning:

     output for /build/067cvg5fy9n6hfdxikcw6qkmlbhymx1r-source/testcases/t/319-gaps.t:
     Attempt to call undefined import method with arguments (":5.10") via package "feature" (Perhaps you forgot to load the package?) at
     /build/067cvg5fy9n6hfdxikcw6qkmlbhymx1r-source/build/i3test.pm line 176, <$ipc_child> line 221.

We don't make warnings fatal, so this is benign (just confusing - the message doesn't make it clear it's non-fatal).

As for a reproducer, we can see the warning if we just run:

```console
$ perl -e 'feature->import(":5.10"); print "ok\n"'
Attempt to call undefined import method with arguments (":5.10") via package "feature" (Perhaps you forgot to load the package?) at -e line 1.
ok
```

I've amended the commit message to reflect this.

stapelberg (approved):

PR #6544, opened 2025-12-12
fix switching to/restoring from LC_NUMERIC

### LINKED ISSUE #6391: append_layout messes up containers sizes when container has a mark.
### Welcome

- [x] Yes, I'm using the latest major release or the current development version. These are the only supported versions.
- [x] Yes, I've searched similar issues and discussions on GitHub and didn't find any.

### Current Behavior

When restoring a layout using the `i3-msg 'workspace 1; append_layout ./workspace.json'` command, the container sizes indicated with the `percent` property are not respected when one container has a `mark` indicated. Just removing that mark in the `workspace.json` file makes everything behave normally.
This layout file produce the following containers with used with `i3-msg 'workspace 1; append_layout /home/thibaut/.config/i3/workspace-comms.json'`, which is not the desired behaviour (the main split should be 20%/80%).

[workspace-comms.json](https://github.com/user-attachments/files/18723718/workspace-comms.json)

Image

### Expected Behavior

Using this layout file (it's the same as the previous one, only the mark has been removed at line 20), produces the right result as shown.
[workspace-comms.json](https://github.com/user-attachments/files/18723723/workspace-comms.json)

Image

Thing is, I would like to have that mark, and to have a 20/80 split.

### Reproduction Instructions

Just restore the layouts using `i3-msg 'workspace 1; append_layout ./workspace.json'`

### i3 version

```text
Running i3 version: 4.24 (pid 819)
Loaded i3 config:
  /home/thibaut/.config/i3/config (main) (last modified: dim. 09 févr. 2025 08:47:54, 2019 seconds ago)

```

### Config file

```text
-
```

Archlinux rolling

picom

### Logfile

_No response_

stapelberg:
Before this commit, we used setlocale(LC_NUMERIC, "");,
but that is not correct because it doesn’t nest:
load_layout.c (sets LC_NUMERIC=C) calls con_mark(),
which calls ipc_send_window_event() (sets LC_NUMERIC=C),
which calls setlocale(LC_NUMERIC, ""); when returning,
but now load_layout has LC_NUMERIC set per the environment,
whereas the function expects to remain in LC_NUMERIC=C.

Using newlocale and uselocale just swaps handles,
which is a little cleaner than querying the locale with
strdup(setlocale(LC_NUMERIC, NULL)); and restoring it later.

The test only fails with certain locales, e.g. LC_NUMERIC=de_DE.
I don’t think it’s important to set a locale in our test runner,
(which locales are available is very system-dependent),
as I am personally regularly testing with LC_NUMERIC=de_DE ;)

fixes https://github.com/i3/i3/issues/6391

### DIFF
--- a/include/main.h
+++ b/include/main.h
@@ -10,6 +10,9 @@
 #pragma once
 
 #include <config.h>
+#include <locale.h>
+
+extern locale_t numericC;
 
 /**
  * Enable or disable the main X11 event handling function.
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -909,10 +909,10 @@ static void dump_bar_config(yajl_gen gen, Barconfig *config) {
 }
 
 IPC_HANDLER(tree) {
-    setlocale(LC_NUMERIC, "C");
+    locale_t prev_locale = uselocale(numericC);
     yajl_gen gen = ygenalloc();
     dump_node(gen, croot, false);
-    setlocale(LC_NUMERIC, "");
+    uselocale(prev_locale);
 
     const unsigned char *payload;
     ylength length;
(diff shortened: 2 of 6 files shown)


orestisfl (approved):
Hopefully something we won't have to mess with with yyjson

PR #6545, opened 2025-12-13
t/323-net-frame-extents.t: unflake by adding sync_with_i3

stapelberg:


orestisfl (approved):
Thanks! That was failing quite a bit

PR #6548, opened 2025-12-14
t/318-i3-dmenu-desktop.t: use $Config{perlpath} (absolute) over $^X

stapelberg:
I noticed that the Nix build does not actually work when using the “#!/usr/bin/env $^X” shebang because /usr/bin/env is not present in the Nix build sandbox.

$Config{perlpath} is always absolute, so let’s prefer that.

Tested both inside and outside a Nix build sandbox.

related to https://github.com/i3/i3/pull/6537

### DIFF
--- a/testcases/t/318-i3-dmenu-desktop.t
+++ b/testcases/t/318-i3-dmenu-desktop.t
@@ -23,6 +23,7 @@ use i3test::Util qw(slurp);
 use File::Temp qw(tempfile tempdir);
 use POSIX qw(mkfifo);
 use JSON::XS qw(decode_json);
+use Config;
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
> because /usr/bin/env is not present in the Nix build sandbox.

~~That can't possibly be right. `/usr/bin/env`'s presence should be guaranteed.~~

Edit: I was wrong - I stand corrected! `/usr/bin/env` is always present on NixOS, but not inside a build sandbox.

CyberShadow (approved):
> Tested both inside and outside a Nix build sandbox.

Confirmed. Thank you!

orestisfl (approved):

Issue #6372, opened 2025-01-14
Losing focus when moving workspace to another screen


dusatvoj:
### Welcome

- [X] Yes, I'm using the latest major release or the current development version. These are the only supported versions.
- [X] Yes, I've searched similar issues and discussions on GitHub and didn't find any.

### Current Behavior

When I press some of shortcuts described below, the workspace is moved between screens - but if mouse is at that screen (so I have focus on that window), mouse cursor stays at the same position at the screen - so I lose focus to window I'm moving.
So I have to move my workspace and then refocus it with mouse movement or another shortcut.
```
# move workspace to another screen
bindsym $mod+Control+Shift+Right move workspace to output right
bindsym $mod+Control+Shift+Up move workspace to output up
bindsym $mod+Control+Shift+Left move workspace to output left
bindsym $mod+Control+Shift+Down move workspace to output down
```

### Expected Behavior

Mouse is moved with the window I have focused

### Reproduction Instructions

1. Create multiple workspaces with windows in it
2. Move focused workspace to different output screen.
3. Mouse is not transfered to new output with the workspace I have moved.

### i3 version

```text
Binary i3 version:  4.24 © 2009 Michael Stapelberg and contributors
Running i3 version: 4.24 (pid 1058)
Loaded i3 config:
  /home/vojta/.config/i3/config (main) (last modified: Tue 18 Oct 2022 01:43:35 PM CEST, 70772556 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

### Config file

```text
Don't have any relevant config - only the config of shortcuts described above.
Not sure if I need any option to follow mouse when moving workspace between outputs.
```

### Linux distribution & Version

Arch linux - rolling

### Are you using a compositor?

I am sure I don't run any compositor

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

dustymabe:
looks similar to https://github.com/i3/i3/issues/6364

[...]

PR #6549, opened 2025-12-14
Add nixos build deps flake.nix file

AThilenius:
I had to make this while debugging #6372 if you want it! I won't be offended if you don't though.

If you're unfamiliar with Nix: this is much like a `Cargo.toml` and `Cargo.lock` or `package.json` and `package-lock.json` but for an entire operating system. In this case it overlays the deps you need to build i3, so anyone on a Nix system can run `nix develop` and within seconds have every dep they need to `mason build; cd build; ninja;` Once they leave that shell the deps vanish. That's normally managed by an `.envrc` file though, so when you `cd` into a project it sets up the deps for you.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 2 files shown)


stapelberg:
I use NixOS as well! See also https://michael.stapelberg.ch/posts/tags/nix/ :)

> In this case it overlays the deps you need to build i3, so anyone on a Nix system can run `nix develop` and within seconds have every dep they need to `mason build; cd build; ninja`

You can already use the existing nixpkgs flake to do that: `nix develop nixpkgs#i3`. 

Given that, I am inlined to not manage another flake in this repository. Or am I missing any advantage this approach would provide us?

AThilenius:
Woooooow. I had no idea you could do that! 🤯 Nope, that makes this totally redundant!

PR #6551, opened 2025-12-16
fix xss-lock in i3-session: import and specify $XDG_SESSION_ID

stapelberg:
I noticed that 'loginctl lock-session' no longer worked after switching to the i3.service user unit. xss-lock said: org.freedesktop.login1.NoSessionForPID: PID 26414 does not belong to any known session.

See also:
https://gitlab.archlinux.org/archlinux/packaging/packages/xss-lock/-/issues/1

### DIFF
--- a/etc/config
+++ b/etc/config
@@ -26,7 +26,7 @@ exec --no-startup-id dex --autostart --environment i3
 
 # xss-lock grabs a logind suspend inhibit lock and will use i3lock to lock the
 # screen before suspend. Use loginctl lock-session to lock your screen.
-exec --no-startup-id xss-lock --transfer-sleep-lock -- i3lock --nofork
+exec --no-startup-id xss-lock --transfer-sleep-lock --session=$XDG_SESSION_ID -- i3lock --nofork
 
 # NetworkManager is the most popular way to manage wireless networks on Linux,
 # and nm-applet is a desktop environment-independent system tray GUI for it.
--- a/etc/config.keycodes
+++ b/etc/config.keycodes
@@ -27,7 +27,7 @@ exec --no-startup-id dex --autostart --environment i3
 
 # xss-lock grabs a logind suspend inhibit lock and will use i3lock to lock the
 # screen before suspend. Use loginctl lock-session to lock your screen.
-exec --no-startup-id xss-lock --transfer-sleep-lock -- i3lock --nofork
+exec --no-startup-id xss-lock --transfer-sleep-lock --session=$XDG_SESSION_ID -- i3lock --nofork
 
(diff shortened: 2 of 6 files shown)


stapelberg:
(Updated the PR to contain an xss-lock wrapper script to keep existing setups working.)

orestisfl:
I am having some follow-up questions regarding https://github.com/i3/i3/pull/6547 because I fear some more workflows could be broken.

Before, when login managers executed share/xsessions/i3.desktop, the i3 executable would be directly called, inheriting the environment. With the systemd change (on a system with systemd), by default the environment is not inherited which means any existing configuration that depends on them will break. Taking another look at the previous PR, it looks like https://github.com/i3/i3/pull/5591#issuecomment-2239541265 brings up this concern.

Am I missing something?

stapelberg:
Yeah, that’s what the `systemctl --user import-environment` line is for. It’s a somewhat unfortunate situation, but on the other hand I’m not seeing any specific breakage in my testing so far (aside from the xss-lock issue I’m fixing here). 

Looking at the `import-environment` calls in Debian Code Search (https://codesearch.debian.net/search?q=import-environment&perpkg=1), I see some that go further (importing also `$PATH`, or even the entire environment, which the systemctl manpage explicitly discourages), but `$DISPLAY` seems to be the lowest common denominator.

Ultimately, if we want to avoid this problem, we’d need to keep running i3 outside of systemd, meaning we’d need to ship the NixOS workaround (activating a `nixos-fake-graphical-session.target` in the session) as i3-session. 

I’d rather attempt the clean solution (a systemd unit for i3) and respond to the breakage that comes in instead of jumping to a less clean solution just because we don’t know whether the change could break more use-cases. In the worst of cases, we publish a minor release which reverts the systemd changes altogether.

orestisfl:
I would much rather make this opt-in and disabled by default somehow (e.g. env var that's checked in libexec/i3-session). It's not worth breaking people's environments over our perceived clean implementation. Unless if I am missing something, this is not just theoretical, it's 100% certain that it's going to modify the environment of all users of the desktop file. I just tried with ly which automatically picks the share/xsessions/i3.desktop session and all env vars loaded via ~/.xprofile / zsh are lost, completely breaking my workflow. Reading through https://wiki.archlinux.org/title/LightDM#Environment_variables I believe the same is going to happen for LightDM as well.

orestisfl:
> e.g. env var that's checked in libexec/i3-session

Or simply provide two xsession .desktop files, one with and one without systemd

stapelberg:
I don’t see the point in making it optional. I’ll send a revert to get the release unblocked.

PR #6558, opened 2025-12-19
declare a development shell in flake.nix

stapelberg:
This makes it possible to build i3 (for development) on any system on which Nix can be installed (= most Linux systems).

For example, I start Emacs using `nix develop --command emacs` and that Emacs process is in an environment with all i3 build deps.

See also:
https://michael.stapelberg.ch/posts/2025-07-27-dev-shells-with-nix-4-quick-examples/

This file is provided best-effort, but PRs are welcome.

While users can already run `nix develop nixpkgs#i3`, for nix-direnv integration it is required to declare a flake.nix in the project directory (otherwise direnv cannot find it).

similar to https://github.com/i3/i3lock/pull/372

related to https://github.com/i3/i3/pull/6549

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1765838191,
+        "narHash": "sha256-m5KWt1nOm76ILk/JSCxBM4MfK3rYY7Wq9/TZIIeGnT8=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "c6f52ebd45e5925c188d1a20119978aa4ffd5ef6",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-25.11",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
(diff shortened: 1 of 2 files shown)


stapelberg:
@orestisfl Is this one good to merge? It would help me with my workflow (makes testing changes easier for me).

orestisfl (commented):
Sorry, this one slipped through the cracks. Can we move it to contrib/? Not super strict about that but given it isn't meant to be officially supported I would prefer to indicate that with the directory it's in. We could symlink it from the root directory (or ignore the file in root so that you can only symlink it on your machine).

If those options won't work, happy to merge it as is.

stapelberg:
> We could symlink it from the root directory (or ignore the file in root so that you can only symlink it on your machine).

I just tried, and unfortunately symlinks like that don’t seem to work with Nix flakes:

```
% nix develop
warning: Git tree '/home/michael/i3' is dirty
error:
       … while updating the lock file of flake 'git+file:///home/michael/i3'

       error: path '/nix/store/a89r7ac2jkki9c3jv4hvd5mx1v2zxvvq-source/flake.lock' is a symlink

```

I think we should merge this as-is for best compatibility.

orestisfl (approved):

PR #6559, opened 2025-12-19
x_push_changes: change root event mask to avoid EnterNotify

### LINKED ISSUE #6372: Losing focus when moving workspace to another screen

stapelberg:
Reconfiguring windows can cause EnterNotify events
when windows are mapped or configured under the mouse pointer.

To avoid handling such EnterNotify events like regular,
user-triggered EnterNotify events (through mouse movement),
we used to add the sequence to the ignore list, but under some
(unknown) circumstances, that does not seem to work reliably.

Instead of ignoring the events, we now change the event mask
of the root window to make the X server not send the events at all.

fixes https://github.com/i3/i3/issues/6372

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -1297,6 +1297,10 @@ void x_push_changes(Con *con) {
             xcb_change_window_attributes(conn, state->id, XCB_CW_EVENT_MASK, values);
         }
     }
+    /* Change the event mask so that we do not receive EnterNotify events
+     * as a result of reconfiguring windows. */
+    xcb_change_window_attributes(conn, root, XCB_CW_EVENT_MASK, (uint32_t[]){ROOT_EVENT_MASK & ~XCB_EVENT_MASK_ENTER_WINDOW});
+
     bool order_changed = false;
     bool stacking_changed = false;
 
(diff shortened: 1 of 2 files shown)


stapelberg:
@orestisfl Any concerns about this one or is this good to merge?

src/x.c
orestisfl, line 1386: I believe this line should also be:
```c
xcb_change_window_attributes(conn, root, XCB_CW_EVENT_MASK, (uint32_t[]){ROOT_EVENT_MASK & ~XCB_EVENT_MASK_ENTER_WINDOW});
```

PR #6565, opened 2025-12-22
Test fixes to reduce flakiness

stapelberg:
(Please don’t squash these commits, I want to rebase&merge this PR.)

### DIFF
--- a/testcases/t/133-size-hints.t
+++ b/testcases/t/133-size-hints.t
@@ -51,23 +51,27 @@ my $ar = $rect->width / $rect->height;
 cmp_float($ar, 2, 'Window set to floating with aspect ratio 2.0');
 
 cmd 'resize set 100';
+sync_with_i3;
 $rect = $win->rect;
 $ar = $rect->width / $rect->height;
 cmp_float($ar, 2, 'Window resized with aspect ratio kept to 2.0');
 
 cmd 'resize set 400 100';
+sync_with_i3;
 $rect = $win->rect;
 $ar = $rect->width / $rect->height;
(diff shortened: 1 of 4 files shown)


testcases/t/133-size-hints.t
```diff
@@ -51,23 +51,27 @@ my $ar = $rect->width / $rect->height;
 cmp_float($ar, 2, 'Window set to floating with aspect ratio 2.0');
 
 cmd 'resize set 100';
+sync_with_i3;
```
orestisfl, line 54: not sure why we need a sync after `cmd`. Shouldn't we be in sync with i3 at that point? Does i3 need to run something else in the next event loop after resize?

stapelberg: Maybe the name `sync_with_i3` gives the wrong idea. 

While `cmd` only returns after i3 has handled the command, the test process might schedule before the X server processed i3’s requests and hence the X server might process the test’s requests first. That results in `$win->rect` returning a result that doesn’t include the `resize set 100` command’s changes. 

What `sync_with_i3` does is it synchronizes with X11 *through* i3. See this part from its documentation:

> This can be used to ensure the results of a cmd 'focus left' are pushed to
X11 and that C<< $x->input_focus >> returns the correct value afterwards.
> 
> See also L<https://build.i3wm.org/docs/testsuite.html> for a longer explanation.

orestisfl: Thanks, I think my memory is flaky as well :D

orestisfl (approved):

PR #6573, opened 2025-12-25
Fix #6566, and fix a POSIX standard issue while here.

### LINKED ISSUE #6566: 4.25 unconditionally uses `uselocale()`
### Welcome

- [x] Yes, I'm using the latest major release or the current development version. These are the only supported versions.
- [x] Yes, I've searched similar issues and discussions on GitHub and didn't find any.

### Current Behavior

I'm trying to update i3wm in pkgsrc to 4.25, but the build fails with:
```
../src/ipc.c: In function 'handle_tree':
../src/ipc.c:912:28: error: implicit declaration of function 'uselocale'; did you mean 'setlocale'? [-Wimplicit-function-declaration]
  912 |     locale_t prev_locale = uselocale(numericC);
      |                            ^~~~~~~~~
      |                            setlocale
```
and many similar lines.

NetBSD does not provide `uselocale`.

### Expected Behavior

It would be good if the build checked for `uselocale` and only used it if the system provides it.

### Reproduction Instructions

Try building on NetBSD, in my case NetBSD 11.99.4/x86_64.

### i3 version

```text
From the last successfully building version:

Binary i3 version:  4.24 (2024-11-06) © 2009 Michael Stapelberg and contributors

but 4.25 doesn't build.
```

### Config file

```text
n/a
```

pkgsrc on NetBSD

Other

### Logfile

_No response_

0-wiz-0:
NetBSD is very picky about ctype(3) arguments. For details, see https://man.netbsd.org/ctype.3

### DIFF
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -909,11 +909,18 @@ static void dump_bar_config(yajl_gen gen, Barconfig *config) {
 }
 
 IPC_HANDLER(tree) {
+#if HAVE_USELOCALE
     locale_t prev_locale = uselocale(numericC);
+#else
+    setlocale(LC_NUMERIC, "C");
+#endif
     yajl_gen gen = ygenalloc();
     dump_node(gen, croot, false);
+#if HAVE_USELOCALE
     uselocale(prev_locale);
-
+#else
+    setlocale(LC_NUMERIC, "");
+#endif
     const unsigned char *payload;
     ylength length;
     y(get_buf, &payload, &length);
@@ -1585,7 +1592,11 @@ ipc_client *ipc_new_client_on_fd(EV_P_ int fd) {
  * generator. Free with yajl_gen_free().
  */
 yajl_gen ipc_marshal_workspace_event(const char *change, Con *current, Con *old) {
+#if HAVE_USELOCALE
     locale_t prev_locale = uselocale(numericC);
+#else
+    setlocale(LC_NUMERIC, "C");
+#endif
     yajl_gen gen = ygenalloc();
 
     y(map_open);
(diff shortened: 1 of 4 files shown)


src/ipc.c
```diff
@@ -909,11 +909,18 @@ static void dump_bar_config(yajl_gen gen, Barconfig *config) {
...
+#else
+    setlocale(LC_NUMERIC, "C");
+#endif
     yajl_gen gen = ygenalloc();
     dump_node(gen, croot, false);
+#if HAVE_USELOCALE
     uselocale(prev_locale);
-
+#else
+    setlocale(LC_NUMERIC, "");
```
stapelberg, line 922: Here and elsewhere, using `setlocale(LC_NUMERIC, "");` is not sufficient. Have a look at the PR description of https://github.com/i3/i3/pull/6544 for more context about the bug we fixed through the `uselocale` dance. We’ll need to query the current locale and restore that.

stapelberg (changes requested):
Would you mind splitting the ctype changes into a separate PR? I would like to have one PR per logical change, and these two seem separate to me. Thanks.

0-wiz-0:
I've split off the ctype change as request: https://github.com/i3/i3/pull/6576

Sorry, I'm not motivated to further improve the `uselocale` change. The commit as-is still improves the situation: on systems with `uselocale`, your bugfixed version is used, while on systems without, the bugfix is missing, but i3 still compiles and is as useful as it was before the bugfix.

stapelberg:
> Sorry, I'm not motivated to further improve the `uselocale` change. The commit as-is still improves the situation: on systems with `uselocale`, your bugfixed version is used, while on systems without, the bugfix is missing, but i3 still compiles and is as useful as it was before the bugfix.

Sorry, but while “As useful as before” is strictly better on NetBSD, it’s strictly worse for i3 as it’ll re-introduce a known bug, so we can’t merge this as-is.

I’m fixing this properly with https://github.com/i3/i3/pull/6577

PR #6595, opened 2026-01-22
Restore floating floating geometry after fullscreen

sabrehagen:
Hi,

I create floating contaners and apply position and geometry manually, to set up a workspace for work. When making a floating window fullscreen and returning it to focus, the  geometry and postion is not restored when leaving fullscreen.

This patch preserves the original floating rectangle so it comes back exactly where it started.

The change saves the floating rect when entering fullscreen and restores it via floating_reposition() when leaving, so geometry and coordinates are both recovered.

I'm opening this as it seems (to me) an improvement to core i3 functionality. Is there a reason this functionaity is not the default in i3?

Kind regards.

### DIFF
--- a/include/data.h
+++ b/include/data.h
@@ -688,6 +688,9 @@ struct Con {
     struct Rect deco_rect;
     /** the geometry this window requested when getting mapped */
     struct Rect geometry;
+    /** saved geometry used to restore floating containers after fullscreen */
+    struct Rect saved_floating_rect;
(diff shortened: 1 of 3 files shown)


orestisfl:
This **is** the default behavior, your test even passes on the next branch without the rest of the changes. Can we perhaps start with an issue first to trace down the problem you are experiencing?

sabrehagen:
Closed in light of @orestisfl's information.

PR #6601, opened 2026-02-08
testcases: make cmd call sync_with_i3 by default

stapelberg:
This commit reverses the default:

Previously, test case authors had to call sync_with_i3 explicitly whenever they wanted to verify the X11 effects of an i3 command.

With this commit, cmd calls sync_with_i3 by default. Test case authors no longer need to worry about this, only about the few cases where the opposite is needed: cmd_nosync.

Created with help by Claude Opus 4.6
(for prototyping this change, but all words in this commit are my own)


orestisfl (commented):
I am a bit sceptical about this, this moves test behaviour further from i3's actual behaviour in the real world, it could cause us to miss some races with xorg. This causes every command to have an extra delay.

I feel it's best if test authors are aware of the nuances.

If we want to avoid introducing flaky tests, we could have a script that runs on PRs that modify tests and it stress-tests (e.g. 100 executions) those modified tests.

stapelberg:
> I am a bit sceptical about this, this moves test behaviour further from i3's actual behaviour in the real world, it could cause us to miss some races with xorg. 

Did we ever catch such a race with test cases? I can’t remember a case personally.

> This causes every command to have an extra delay.

I couldn’t measure a slow-down on my machine. Can you measure a difference?

> If we want to avoid introducing flaky tests, we could have a script that runs on PRs that modify tests and it stress-tests (e.g. 100 executions) those modified tests.

That’s sort of how I started this whole effort :) 

I had Claude build a flaky test fixer and run that in a loop for a few hours. It produced a bunch of `sync_with_i3` calls that were missing. 

But even with all `sync_with_i3` out of the way, there are still other causes for flakiness in our test suite.

It would be great to get `sync_with_i3` out of the way as a cause and focus on the remaining, more tricky issues.

> I feel it's best if test authors are aware of the nuances.

I disagree. We have failed to get our test suite free of flaky tests in the last 9 years. 

When I cut the last release, I had to run the release script 5 times to get a working run. This isn’t a good state to be in, and if we can eliminate a whole class of test flakiness issues, my preference is we give it a try.

orestisfl (approved):

PR #6623, opened 2026-06-01
ui: allow stacked view to have title bars vertical

jcelerier:
Title says it all. I mostly use the stack view but it wastes a lot of space when I have 15 different windows stacked. 
This patch adds a toggle to switch the titlebars between horizontal (the default) and vertical:

a

vs

b

### DIFF
--- a/docs/userguide
+++ b/docs/userguide
@@ -731,6 +731,46 @@ Default is +left+
 title_align left|center|right
 ---------------------------------------------
 
+=== Stacked title column width
+
+The title bars of a +stacking+ container can optionally be laid out as a
+vertical column on the left-hand side of the container (similar to the vertical
+tabs found in some web browsers) instead of the classic horizontal bars across
+the top. This is toggled at runtime, per container, with the
++stack_title_position+ command (see <<stack_title_position>>).
+
+This option only controls how wide that column is: each title is clipped to
+roughly the given number of characters, while the focused window fills the
+remaining space to the right of the column. It has no effect while a container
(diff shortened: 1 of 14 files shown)


jcelerier:
it also works neatly with desktop widgets: 

image

cdlscpmv:
Is there a reason you don't want to use the tabbed layout?

jcelerier:
When I have 15+ tabs it becomes really unreadable (same reason I don't use horizontal tabs in browsers, vertical are the way to go)

orestisfl:
Thank you for the contribution. The core idea has value however I'll need to close the current PR on the following grounds:
1. I believe the PR is fully written by an LLM. If that's correct, it doesn't conform to the [AI contribution policy](https://github.com/i3/i3/blob/next/.github/CONTRIBUTING.md#ai-contribution-policy) of the project.
2. The testsuite fails which reduces my confidence on the correctness of the patch. The i3 rendering code is not always the easiest to modify and is not well-covered by tests. Such functionality usually need to consider various edge cases.
3. > We are generally happy with the current feature set of i3 and instead focus on maintenance such as stability and fixing bugs. New features will rarely be considered if they require additional configuration and/or commands, or if they add significant complexity (either through the exposed configuration or mental complexity) to the project.
https://github.com/i3/i3/blob/next/.github/CONTRIBUTING.md
4. If an exception is to be considered RE the above point: The proper process would be to: a. open an issue b. gather significant support c. let the community voice their opinion on the desired new behavior.

PR #6634, opened 2026-06-28
Fix titlebars for shaped windows

xzfc:
This PR fixes a bug: titlebars for shaped windows are not visible since d26ddcbf.

How to reproduce: run `xeyes`, make it floating.

Actual behavior: window titlebar is not visible. 

before

After the fix:

after

Created with help by Claude Opus 4.8.

### DIFF
--- a/src/x.c
+++ b/src/x.c
@@ -406,7 +406,9 @@ static void x_draw_decoration_after_title(Con *con, struct deco_render_params *p
  * number of rectangles.
  *
  */
-static size_t x_get_border_rectangles(Con *con, xcb_rectangle_t rectangles[4]) {
+static size_t x_get_border_rectangles(Con *con,
+                                      bool include_titlebar,
+                                      xcb_rectangle_t rectangles[4]) {
     size_t count = 0;
     int border_style = con_border_style(con);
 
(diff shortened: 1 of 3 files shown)


src/x.c
```diff
@@ -406,7 +406,9 @@ static void x_draw_decoration_after_title(Con *con, struct deco_render_params *p
  * number of rectangles.
  *
  */
-static size_t x_get_border_rectangles(Con *con, xcb_rectangle_t rectangles[4]) {
+static size_t x_get_border_rectangles(Con *con,
+                                      bool include_titlebar,
```
xzfc, line 410: Note: the observable behavior would be the same if we remove this argument and assume `include_titlebar == true` in both cases.

But, `include_titlebar == false` saves one redundant call to `draw_util_rectangle`\[1] which would be painted over\[2] anyway.

---

\[1]: https://github.com/i3/i3/blob/f7d5b8983f6caeab94585e5359d7de59e7041e9b/src/x.c#L575-L579

\[2]: https://github.com/i3/i3/blob/f7d5b8983f6caeab94585e5359d7de59e7041e9b/src/x.c#L631-L632

src/x.c
```diff
@@ -439,7 +441,10 @@ static size_t x_get_border_rectangles(Con *con, xcb_rectangle_t rectangles[4]) {
             };
         }
         /* pixel border have an additional line at the top */
-        if (border_style == BS_PIXEL && !(borders_to_hide & ADJ_UPPER_SCREEN_EDGE)) {
+        bool add_pixel =
+            border_style == BS_PIXEL && !(borders_to_hide & ADJ_UPPER_SCREEN_EDGE);
+        bool add_titlebar = include_titlebar && con_draw_decoration_into_frame(con);
```
xzfc, line 446: Note: check for `con_draw_decoration_into_frame()` is redundant _at the moment_ because `x_shape_frame()` is only called for leaf containers, so it's always true when `include_titlebar` is true.

I decided not to rely on this invariant.

Issue #6364, opened 2025-01-08
focus_follows_mouse is consistently failing and needs config reload


xstefank:
### Welcome

- [X] Yes, I'm using the latest major release or the current development version. These are the only supported versions.
- [X] Yes, I've searched similar issues and discussions on GitHub and didn't find any.

### Current Behavior

This is the same issue as described in https://github.com/i3/i3/issues/3518. I don't know in which version exactly, but the focus stopped following the moved workspace. It only happens from time to time, if I reload my config where I have `focus_follows_mouse yes` it works again for some time. So I don't understand what times out somewhere that I need to always reload my config?

### Expected Behavior

When I have `focus_follows_mouse yes` the focus follows mouse on window move to different screen.

### Reproduction Instructions

Just moving focus between screens. It times out everytime, after some delay which seems random.

### i3 version

```text
$ i3 --moreversion 2>&-
Binary i3 version:  4.24 (2024-11-06) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.24 (2024-11-06) (pid 3508)
Loaded i3 config:
  /home/mstefank/.config/i3/config (main) (last modified: Tue 07 Jan 2025 08:27:43 AM CET, 86958 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

### Config file

```text
https://github.com/xstefank/config/blob/main/home/config/i3/config
```

### Linux distribution & Version

Linux Set-FI 6.12.7-200.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Dec 27 17:05:33 UTC 2024 x86_64 GNU/Linux

### Are you using a compositor?

I am sure I don't run any compositor

i3bot:
I don’t see a link to logs.i3wm.org. Did you follow https://i3wm.org/docs/debugging.html? (In case you actually provided a link to a logfile, please ignore me.)

dustymabe:
added a relevant comment to https://github.com/i3/i3/issues/6372#issuecomment-2590489265

mutageneral:
I just had the same problem, `reload` of the config didn't fix it, but a `restart` of i3 did fix it, but I don't know for how long.
(Ok, it happened again, probably took a bit less than an hour.)

When moving a window around on the same monitor using `move left` or `move right` keybindings, it did not have any issues and focus was kept on the window being moved. But when I moved a window off to another monitor, the focus was kept on the originating monitor and the window lost focus.

[...]

PR #6641, opened 2026-07-24
[OBSOLETE] Enforce tracking of ignored events by 32-bit full sequence

### LINKED ISSUE #6372: Losing focus when moving workspace to another screen

javgh:
This PR is a follow-up to the discussion in #6372 .

I'm not too familiar with the internals of X11, but it seems that sequence numbers are sometimes of type `uint16_t` and sometimes of type `unsigned int` or `uint32_t`. In some places XCB provides the field `full_sequence` (for example as part of `xcb_generic_event_t`), but other structs only have the 16-bit sequence. Then there is also `xcb_void_cookie_t` which has `unsigned int` for `sequence`.

The current code for tracking ignored events mixes these different data types. This works fine, as long as the sequence number doesn't overflow a `uint16_t`. Afterwards, the tracking of ignored events will fail. This appears to be the reason for #6372 and probably also #6364 . There is also the open PR https://github.com/i3/i3/pull/6559 which notes: "[...] add the sequence to the ignore list, but under some (unknown) circumstances, that does not seem to work reliably."

This problem already existed in 4.23 (and before), but only became visible in 4.24, because it seems that is was masked by using `XCB_GRAB_MODE_SYNC` when executing key bindings. This seems to mask the bug, but sending a command via `i3-msg` also triggers it in 4.23 . Commit https://github.com/i3/i3/commit/b42dc21068196dd60a0806575f0f8bb01610fd0e changed `XCB_GRAB_MODE_SYNC` to `XCB_GRAB_MODE_ASYNC` and surfaced the bug. Additionally, the commit https://github.com/i3/i3/commit/cfa4cf16 increased the number of messages being sent and sequence numbers being used, which triggers this bug more quickly than before.

The first commit in this PR introduces a test for #6372 . I found a more performant way of overflowing the uint16_t sequence number (using sync_with_i3 to create fairly cheap messages). The test runs in 1.3 seconds on my system.

The second commit fixes the problem. It defines a new data type `ignore_event_sequence_t` and a macro `ENSURE_FULL_SEQUENCE` which is then used at every call site for `add_ignore_event` and `event_is_ignored`. This way the compiler checks that a full sequence number is used everywhere. Perhaps this is over-engineered? Please let me know - I don't work much in C.

This change makes the test pass and also fixes #6372 for me in my manual testing.

Looking forward to feedback!

Disclosure of AI assistance: I used Claude Fable 5 and GPT-5.6 Sol as sparring partners, but everything here is hand-written.

### DIFF
--- a/src/handlers.c
+++ b/src/handlers.c
@@ -124,7 +124,7 @@ static void check_crossing_screen_boundary(uint32_t x, uint32_t y) {
  * When the user moves the mouse pointer onto a window, this callback gets called.
  *
  */
-static void handle_enter_notify(xcb_enter_notify_event_t *event) {
+static void handle_enter_notify(xcb_enter_notify_event_t *event, uint32_t full_sequence) {
     Con *con;
 
     last_timestamp = event->time;
@@ -138,7 +138,7 @@ static void handle_enter_notify(xcb_enter_notify_event_t *event) {
     }
     /* Some events are not interesting, because they were not generated
      * actively by the user, but by reconfiguration of windows */
-    if (event_is_ignored(event->sequence, XCB_ENTER_NOTIFY)) {
+    if (event_is_ignored(full_sequence, XCB_ENTER_NOTIFY)) {
         DLOG("Event ignored\n");
         return;
     }
@@ -1514,7 +1514,7 @@ void handle_event(int type, xcb_generic_event_t *event) {
 
         /* Enter window = user moved their mouse over the window */
         case XCB_ENTER_NOTIFY:
-            handle_enter_notify((xcb_enter_notify_event_t *)event);
+            handle_enter_notify((xcb_enter_notify_event_t *)event, event->full_sequence);
             break;
 
         /* Client message are sent to the root window. The only interesting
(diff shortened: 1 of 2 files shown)


javgh:
Wrong branch, sorry

PR #6654, opened 2026-09-20
libi3, i3bar: fix CI on main by keeping const pointers from strchr/strstr

alisa-ulan:
I noticed the CI failed, and wondered if I caused it long enough to fix it. sid moved to glibc 2.43, whose string.h makes strchr/strstr const-correct in C, so with -Werror:
* `libi3/resolve_tilde.c`: assignment discards `const` qualifier
* `i3bar/src/ipc.c`: initialization discards `const` qualifier

This declares the two result pointers const. With these fixes it WOMM in the project's dockerfile.


orestisfl (approved):
Thanks!

